PR #2, opened 2019-03-13
Added Readonly and Partial

marco-fp:
:)


sindresorhus:
`Readonly` is already a built-in. What's the use-case for `Partial`?

marco-fp:
Just found out, both are built-in actually. Didn't know :sweat_smile: 

A use case for Partial would be https://netbasal.com/getting-to-know-the-partial-type-in-typescript-ecfcfbc87cb6

PR #5, opened 2019-03-13
Add `PackageJson` and `LiteralUnion` types

BendingBender:
Will do the eslint types in a later PR.


npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+export type BundledDependency = string[];
+
+/**
+Dependencies of the package. The version range is a string which has one or more space-separated descriptors. Dependencies can also be identified with a tarball or Git URL.
+*/
+export interface Dependency {
+	[packageName: string]: string;
+}
+
+export type PackageJson = {
```
sindresorhus: Wouldn't it be better if all the other types/interfaces were nested under this one? So `Dependency` was referred to as `PackageJson.Dependency`. I think that would be a better namespace than putting everything under `Npm`.

npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+
+	/**
+	Location for test files.
+	*/
+	test?: string;
+
+	[directoryType: string]: unknown;
+}
+
+export interface ScriptsDefinition {
```
sindresorhus: Why the `Definition` postfix?

How about this?

```suggestion
export interface Scripts {
```

npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+	Run with the `npm restart` command. Note: `npm restart` will run the `stop` and `start` scripts if no `restart` script is provided.
+	*/
+	restart?: string;
+
+	/**
+	Run with the `npm restart` command, after `restart`. Note: `npm restart` will run the `stop` and `start` scripts if no `restart` script is provided.
+	*/
+	postrestart?: string;
+
+	[scriptName: string]: string | undefined;
```
sindresorhus: Why allow `undefined`?

npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+	The files included in the package.
+	*/
+	files?: string[];
+
+	/**
+	The module ID that is the primary entry point to the program.
+	*/
+	main?: string;
+
+	bin?:
```
sindresorhus: Needs a doc comment description.

npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+
+	/**
+	Location of the bundled TypeScript declaration file.
+	*/
+	types?: string;
+
+	/**
+	Location of the bundled TypeScript declaration file. Synonymous with `types`.
+	*/
+	typings?: string;
```
sindresorhus: These are not official fields. What do you think about putting them in https://github.com/sindresorhus/type-fest/pull/5/files#diff-6c293fd7802982e4829c46838e706800R438 or a separate type for clarity?

npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+	};
+
+	dist?: {
+		shasum?: string;
+		tarball?: string;
+		integrity?: string;
+		[key: string]: unknown;
+	};
+
+	readme?: string;
```
sindresorhus: Needs a doc comment description.

npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+	private?: boolean;
+
+	/**
+	 * A set of config values that will be used at publish-time. It's especially handy to set the tag, registry or access, to ensure that a given package is not tagged with 'latest', published to the global public registry or that a scoped module is private by default.
+	 */
+	publishConfig?: {
+		[config: string]: unknown;
+	};
+
+	dist?: {
```
sindresorhus: Needs a doc comment description.

npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+	os?: LiteralUnion<
+		'darwin' | 'linux' | 'win32' | '!darwin' | '!linux' | '!win32',
+		string
+	>[];
+
+	/**
+	CPU architectures the module runs on.
+	*/
+	cpu?: LiteralUnion<
+		'x64' | 'ia32' | 'arm' | 'mips' | '!x64' | '!ia32' | '!arm' | '!mips',
```
sindresorhus: > The current possible values are: 'arm', 'arm64', 'ia32', 'mips', 'mipsel', 'ppc', 'ppc64', 's390', 's390x', 'x32', and 'x64'

- https://nodejs.org/api/process.html#process_process_arch

npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+	/**
+	 * @deprecated
+	 */
+	engineStrict?: boolean;
+
+	/**
+	Operating systems the module runs on.
+	*/
+	os?: LiteralUnion<
+		'darwin' | 'linux' | 'win32' | '!darwin' | '!linux' | '!win32',
```
sindresorhus: There are more: https://nodejs.org/api/process.html#process_process_platform

npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+	Package names that are bundled when the package is published.
+	*/
+	bundledDependencies?: string[];
+
+	/**
+	Alias of `bundledDependencies`.
+	*/
+	bundleDependencies?: string[];
+
+	resolutions?: Dependency;
```
sindresorhus: Needs a doc comment.

npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+	*/
+	peerDependencies?: Dependency;
+
+	/**
+	Package names that are bundled when the package is published.
+	*/
+	bundledDependencies?: string[];
+
+	/**
+	Alias of `bundledDependencies`.
```
sindresorhus: Should we use `@alias`? http://usejsdoc.org/tags-alias.html

npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+		[binary: string]: string;
+	};
+
+	/**
+	Location of the bundled TypeScript declaration file.
+	*/
+	types?: string;
+
+	/**
+	Location of the bundled TypeScript declaration file. Alias of `types`.
```
sindresorhus: Should we use `@alias`? http://usejsdoc.org/tags-alias.html

npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+	*/
+	typings?: string;
+
+	/**
+	Filenames to put in place for the `man` program to find.
+	*/
+	man?: string | string[];
+
+	directories?: DirectoryLocations;
```
sindresorhus: Needs a doc comment.

npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+	*/
+	bundleDependencies?: string[];
+
+	resolutions?: Dependency;
+
+	/**
+	Engines that this package runs on.
+	*/
+	engines?: {
+		[engineName: string]: string;
```
sindresorhus: This should at least define `node` and `npm` as engine. https://github.com/chalk/chalk/blob/cd5de7a2f6a7a04fbad4c75c0364a3efc6048ade/package.json#L8 In addition to the above.

https://docs.npmjs.com/files/package.json#engines

npm.d.ts
sindresorhus: Can you also add the `browser` field there? https://github.com/defunctzombie/package-browser-field-spec

npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+	};
+
+	/**
+	 * @deprecated
```
sindresorhus: Don't use the `*` syntax.

npm.d.ts
```diff
@@ -0,0 +1,439 @@
...
+		browser?: string;
+		[moduleName: string]: string | undefined;
+	};
+
+	/**
+	JSPM configuration.
+	*/
+	jspm?: PackageJson;
+} & {
+	[k: string]: unknown;
```
sindresorhus:
```suggestion
	[key: string]: unknown;
```

sindresorhus: Why is this in a separate intersection type?

sindresorhus:
Can you move `npm.ts` to a `source` directory? (Don't forget to add `source` to `files` in package.json).

npm.d.ts
BendingBender: I did the refactoring locally. The problem is that VS Code is no longer able to import the types via code action. So it's kind of a degraded experience.

VS Code uses TypeScript server under the hood, like most other IDEs to provide TS support.

npm.d.ts
BendingBender: In `strict` mode, the type of the index signature has to match with the other keys defined in the type. This is the reason I've used the
```ts
type Foo = {
  key?: string;
} & {
  [key: string]: string;
}
```
below.

npm.d.ts
BendingBender: Never used it before. If I understand the description correctly it seems to have a different purpose - it renames things in the docs. So instead of using the function/property name from the sources it uses the provided name.

IMHO, it doesn't look like a good fit here.

npm.d.ts
BendingBender: See https://github.com/sindresorhus/type-fest/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc#discussion_r265348876

source/npm.d.ts
```diff
@@ -0,0 +1,493 @@
...
+	};
+
+	/**
+	A hint to JavaScript bundlers or component tools when packaging modules for client side use.
+	 */
+	browser?:
+	| string
+	| {
+		[moduleName: string]: string | false;
+	};
```
sindresorhus: `module`, `esnext`, and `browser` are also not official fields and should be in the intersection.

sindresorhus:
I think the filenames for source/test should be `npm` => `package-json`

sindresorhus:
Can you add the types to the readme?

BendingBender:
Which ones? Every single one or only the main `PackageJson`?

sindresorhus:
Only `PackageJson` and `LiteralUnion`

sindresorhus:
Thanks :)

sindresorhus:
@BendingBender Was wondering if you would like to join this project as a maintainer? No commitment, just want to show your name in the readme since you have contributed a lot of code here and you might have opinions about future type additions.

BendingBender:
@sindresorhus It would be a pleasure for me!

PR #1, opened 2019-03-13
Add `RequireAtLeastOne` type

CvX:
(Extracted from https://github.com/sindresorhus/electron-util/pull/11)

I'm wondering if it would be better to change its name to just "AtLeastOne"… 🤔

PR #6, opened 2019-03-13
Add `MergeExclusive` type

BendingBender:


sindresorhus:
Can you use a more readable name than `XOR`? (Not everyone knows what that means)

sindresorhus:
This type looks very useful 👍

test/xor.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const exclusiveVariation1: Options = {exclusive1: true};
+const exclusiveVariation2: Options = {exclusive2: 1};
+
+expectType<{option?: string; exclusive1: boolean; exclusive2?: string}>(
+	exclusiveVariation1
+);
+expectType<{option?: string; exclusive1?: string; exclusive2: number}>(
+	exclusiveVariation2
+);
```
sindresorhus: Can you add some commented-out negative tests.

So we can easily use those when https://github.com/SamVerschueren/tsd-check/issues/2 is fixed.

Example: https://github.com/sindresorhus/type-fest/pull/1/files#diff-fd2873d8135a3976c91e67f021a5cd40R21

janriemer:
> Can you use a more readable name than XOR? (Not everyone knows what that means)

Maybe `MergeExclusive` or `CombineExclusive` is better!? I would prefer `MergeExclusive`, because we already have a `Merge` type.

EdwardDrapkin:
if you implement `Diff` instead of `Without` (return T[key] instead of never in the map), you should be able to make this type `Diff<A, B> & Diff<B,A>` which is much simpler and Diff's a more useful helper than Without is.

sindresorhus:
I just added some contribution guidelines. Would you be able to review them? I'm also interested in feedback if anything there is unclear or could be improved.

BendingBender:
@EdwardDrapkin Do you mean like this:
```ts
type Diff<FirstType, SecondType> = {
	[KeyType in Exclude<keyof FirstType, keyof SecondType>]: FirstType[KeyType]
};
type MergeExclusive<FirstType, SecondType> = Diff<FirstType, SecondType> & Diff<SecondType, FirstType>;
```

This, sadly doesn't work, it basically produces `FirstType & SecondType`, there is nothing exclusive about this type.

sindresorhus:
@BendingBender Can you fix the merge conflict?

test/merge-exclusive.ts
```diff
@@ -0,0 +1,29 @@
...
+
+expectType<{option?: string; exclusive1: boolean; exclusive2?: string}>(
+	exclusiveVariation1
+);
+expectType<{option?: string; exclusive1?: string; exclusive2: number}>(
+	exclusiveVariation2
+);
+
+// TODO: Negative tests, waiting for https://github.com/SamVerschueren/tsd-check/issues/2
+// expectError<Options>({exclusive1: true, exclusive2: 1});
```
SamVerschueren: Upgrade to `tsd-check@5.0.0` and this should work 👌 

```ts
expectError<Options>({exclusive1: true, exclusive2: 1});
```

BendingBender:
Waiting for https://github.com/SamVerschueren/tsd-check/pull/16

BendingBender:
@sindresorhus I think that something is off with the way travis integration is configured. Currently, two different travis runs are spawned up on every commit.

sindresorhus:
> Currently, two different travis runs are spawned up on every commit.

For some reason, there was a webhook to the old `.org` version of Travis too.

sindresorhus:
Thanks :)

Issue #3, opened 2019-03-13
Similar project


mohsen1:
FYI https://github.com/piotrwitek/utility-types is very similar to this project.

sindresorhus:
There are many similar projects, but they all have a lot of stuff that is theoretically cool, but that at least I, would never use in practice. I'm not trying for this to be the "Lodash of TS types". I just want to consolidate the most common types I need.

inad9300:
A few others...
- https://github.com/andnp/SimplyTyped
- https://github.com/gcanti/typelevel-ts
- https://github.com/pelotom/type-zoo
- https://github.com/tycho01/typical

If what makes yours different is that it suits *you* specifically, you should clarify that in the README file so that people don't get confused by the project's goals.

Offirmo:
Also: https://github.com/krzkaczor/ts-essentials

PR #7, opened 2019-03-14
Add `Diff` and `Spread` types

EdwardDrapkin:


sindresorhus:
I just added some contribution guidelines. Would you be able to review them? I'm also interested in feedback if anything there is unclear or could be improved.

index.d.ts
```diff
@@ -84,3 +84,23 @@ const ab: Merge<Foo, Bar> = {a: 1, b: 2};
 ```
 */
 export type Merge<FirstType, SecondType> = Omit<FirstType, Extract<keyof FirstType, keyof SecondType>> & SecondType;
+		  
+/**
+ * Diffs two objects.
```
sindresorhus: Can you add this to the readme too?

index.d.ts
```diff
@@ -84,3 +84,23 @@ const ab: Merge<Foo, Bar> = {a: 1, b: 2};
...
+ *
+ * Given objects with types T and V, returns an object that has all the keys in T that do not also exist in V.
+ *
+ * @example `type Safe = Diff<AllProperties, UnsafeProperties>`
+ */
+export type Diff<T extends {}, V extends {}> = {
```
sindresorhus: Can you use more descriptive type names than `T` and `V`? See the other types for inspiration.

index.d.ts
```diff
@@ -84,3 +84,23 @@ const ab: Merge<Foo, Bar> = {a: 1, b: 2};
 ```
 */
 export type Merge<FirstType, SecondType> = Omit<FirstType, Extract<keyof FirstType, keyof SecondType>> & SecondType;
+		  
+/**
+ * Diffs two objects.
+ *
+ * Given objects with types T and V, returns an object that has all the keys in T that do not also exist in V.
+ *
+ * @example `type Safe = Diff<AllProperties, UnsafeProperties>`
```
sindresorhus: Can you include a more comprehensive example? Should also include the import statement.

index.d.ts
```diff
@@ -84,3 +84,23 @@ const ab: Merge<Foo, Bar> = {a: 1, b: 2};
...
+ */
+export type Diff<T extends {}, V extends {}> = {
+    [P in Exclude<keyof T, keyof V>]: T[P];
+};
+
+		  
+/**
+ * Returns a type modeling the result of spreading two objects together.
+ *
+ * Given objects with types T and V, returns an object that has a type that represents {...T, ...V}
```
sindresorhus: How is this different from `Merge`?

index.d.ts
```diff
@@ -84,3 +84,23 @@ const ab: Merge<Foo, Bar> = {a: 1, b: 2};
...
+ * @example `const a: Spread<X, Y> = { ...x, ...y }`
```
sindresorhus: Doesn't TS already infer the new type correctly from object-spread? Maybe I'm missing the use-case for this type.

sindresorhus:
Ping :)

sindresorhus:
@CvX @BendingBender Do you think these types are something we should add?

CvX:
I prefer to extract types from existing code. I find it difficult to judge if given type will find a use in the future. 😅 Some real-world examples would go a long way.

BendingBender:
@sindresorhus I'm not sure about this PR. Like @CvX I have difficulties imagining using types like this. They are clever solutons but usually a simpler "dumb" solution is easier to understand and is at the same time obvious for every dev who reads it. I'm sure that these types can be useful, especially in the context of ambient declarations but I have yet to encounter a use case for them.

As you've already pointed out, this library is mostly for types that are useful for your projects (https://github.com/sindresorhus/type-fest/issues/3#issuecomment-472474036). So what's the point adding types that you can't even imagine to use?

alixaxel:
I think I have a use case for where this would be helpful. In a ORM I built, I have a few decorators that can potentially be attached for a model, and will automatically populate `id` + `createdAt` + `updatedAt`.

However, in some models, it doesn't make sense to have either one of those:

```typescript
Omit<T, keyof (Identified & Timestamped)>
&
PartialMaybe<T, Identified>
&
PartialMaybe<T, Timestamped>)
```

For context:

```typescript
type PartialMaybe<T, O> = T extends O ? Partial<O> : {};

interface Identified {
  id: string;
}

interface Timestamped {
  createdAt: firestore.Timestamp;
  updatedAt?: firestore.Timestamp;
}
```

So in practice this will conditionally (if they exist) make `Identified` and `Timestamped` types optional.

Richienb:
@sindresorhus My use case for `Spread` is when I'm trying to document return types without creating new interfaces.

```ts
interface returnTypes {
	somePropertyToExtend: string
}

function someFunction(): Spread<returnTypes, {
	somethingElseThatIsNeeded: string
}>

function anotherFunction(): Spread<returnTypes, {
	anotherThingThatIsNeeded: string
}>
```

sindresorhus:
It would be better to open a new issue to discuss this.

PR #12, opened 2019-03-21
Add `@sindresorhus/tsconfig` to devDependencies

CvX:
…since we already “using” it in tsconfig.json. 😉

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
 		"json"
 	],
 	"devDependencies": {
+		"@sindresorhus/tsconfig": "^0.2.1",
 		"@typescript-eslint/eslint-plugin": "^1.4.2",
 		"eslint-config-xo-typescript": "^0.8.0",
 		"tsd": "^0.7.0",


sindresorhus:
Weird that there was no error about it being missing...

CvX:
`tsc` does complain about it:

```
tsconfig.json:2:13 - message TS6096: File '@sindresorhus/tsconfig' does not exist.

2  "extends": "@sindresorhus/tsconfig"
              ~~~~~~~~~~~~~~~~~~~~~~~~
```

…but we don't use `tsc` because there isn't really any code to compile in type-fest and the code that could be compiled is invalid by design (the negative tests 😉).

sindresorhus:
@CvX I meant `@typescript-eslint`. It uses the tsconfig for improved type-checking. That's why we have a `tsconfig.json` file. It should have complained about it, but didn't.

CvX:
I guess it's because it isn't an error or even a warning, just a "message", so the tsconfig parser doesn't throw any exceptions, and eslint continues as if the config is empty. I'll look if there's any way for @typescript-eslint to catch those.

CvX:
@sindresorhus https://github.com/typescript-eslint/typescript-eslint/compare/master...CvX:tsconfig-errors

The output is far from ideal but it's a start: 😉
wip eslint output

sindresorhus:
@CvX Nice! It's a very good start. Anything is better than silent failure.

PR #21, opened 2019-04-06
Add global definition for Symbol.observable

### LINKED ISSUE #19: Property 'observable' does not exist on type 'SymbolConstructor'
Getting the following error right after adding type-fest package

image

BendingBender:
This prevents errors for users who don't use node types in their projects.

Fixes #19.


BendingBender:
@tonyxiao Would be great if you could try out whether this fix works for you.

tonyxiao:
Tested working.

PR #24, opened 2019-04-08
Add `RequireOnlyOne` type

izayl:

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -41,7 +41,7 @@ export type JsonObject = {[key: string]: JsonValue};
 /**
 Matches a JSON array.
 */
-export interface JsonArray extends Array<JsonValue> {} // eslint-disable-line @typescript-eslint/no-empty-interface
+export interface JsonArray extends Array<JsonValue> {}
 
 /**
 Matches any valid JSON value.
--- a/index.d.ts
+++ b/index.d.ts
@@ -225,8 +225,8 @@ export type RequireOnlyOne<ObjectType, KeysType extends keyof ObjectType = keyof
 		[Key in KeysType]: (
 			// …by picking that Key's type and making it required
 			Required<Pick<ObjectType, Key>> &
-			// …and adding the other keys in KeysType as optional and of type `undefined`
-			Partial<Record<Exclude<KeysType, Key>, undefined>>
+			// …and adding the other keys in KeysType as optional and of type `never`
+			Partial<Record<Exclude<KeysType, Key>, never>>
 		)
 	}[KeysType]
 	// …then, make intersection types by adding the remaining properties to each mapped type.
--- a/index.d.ts
+++ b/index.d.ts
@@ -220,10 +220,7 @@ const responder: RequireOnlyOne<Responder, 'text' | 'json'> = {
 ```
 */
 export type RequireOnlyOne<ObjectType, KeysType extends keyof ObjectType = keyof ObjectType> =
-	Pick<
-		ObjectType,
-		Exclude<keyof ObjectType, KeysType>
-	> &
+	Omit<ObjectType, KeysType> &
 	{
 		[Key in KeysType]: Required<Pick<ObjectType, Key>> &
 		Partial<Record<Exclude<KeysType, Key>, undefined>>
--- a/index.d.ts
+++ b/index.d.ts
@@ -223,5 +223,5 @@ export type RequireOnlyOne<ObjectType, KeysType extends keyof ObjectType = keyof
 	Omit<ObjectType, KeysType> &
 	{
 		[Key in KeysType]: Required<Pick<ObjectType, Key>> &
-		Partial<Record<Exclude<KeysType, Key>, undefined>>
+		Partial<Record<Exclude<KeysType, Key>, never>>
 	}[KeysType];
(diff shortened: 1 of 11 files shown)


CvX:
I've just noticed you've changed `never` to `undefined` before. What was the issue with `never`?

izayl:
@CvX 
there is a test to prove that `never` is not suitable for some case.
15dc6dd18a4abcd6f378b6efcf344323eab042ee

CvX:
Hmm… With that change though, it also incorrectly accepts this:

```ts
const system = {macos: 'hey', x: 1} || {linux: 'sup', y: 'wat'};
test({default: 'hello', ...system});
```

izayl:
> Hmm… With that change though, it also incorrectly accepts this:
> 
> ```ts
> const system = {macos: 'hey', x: 1} || {linux: 'sup', y: 'wat'};
> test({default: 'hello', ...system});
> ```

```ts
type SystemMessages = {
	default: string;

	macos?: string;
	linux: string;
	windows: string;

	optional?: string;
};
const test = (_: ValidMessages): void => {};
```

@CvX  Because the property `x` and `y` not in the type `ValidMessages`.

CvX:
That's my point. 🙂 With the current `RequireOnlyOne` implementation and the `…({} || {})` construct you can pass pretty much anything and it doesn't raise an error.

izayl:
you are right,  current implementation `RequireOnlyOne` not support this Object deconstruction scene.

CvX:
> my editor highlight some error confused me, it works after restart my editor

Same here 😅 Turns out there isn't really any difference between `never` and `undefined` in this case?

See the commented-out test I've just pushed. It doesn't raise an error… 🤔

izayl:
> > my editor highlight some error confused me, it works after restart my editor
> 
> Same here 😅 Turns out there isn't really any difference between `never` and `undefined` in this case?
> 
> See the commented-out test I've just pushed. It doesn't raise an error… 🤔

I think it's a bug cause by ts type checker.
The type `{ x?: never }` says either the property does not exist, or it exists with type never. It's mean that x does not exist, though the type should really be isomorphic to never.

sindresorhus:
> I think it's a bug cause by ts type checker.

Can you look for an issue about this in the TS issue tracker and if not, open an issue there?

index.d.ts
```diff
@@ -198,3 +198,36 @@ export type RequireAtLeastOne<ObjectType, KeysType extends keyof ObjectType = ke
 	}[KeysType]
 	// …then, make intersection types by adding the remaining properties to each mapped type.
 	& Omit<ObjectType, KeysType>;
+
+/*
+Create a type that requires at only one of the given properties. The remaining properties are kept as is.
+
```
sindresorhus: Can you mention in text some real-world use-cases? 

> Write about some real-world use-cases where it can be useful. (It can be hard sometimes for users to see where they would use something) - https://github.com/sindresorhus/type-fest/blob/master/.github/contributing.md

index.d.ts
```diff
@@ -198,3 +198,36 @@ export type RequireAtLeastOne<ObjectType, KeysType extends keyof ObjectType = ke
 	}[KeysType]
 	// …then, make intersection types by adding the remaining properties to each mapped type.
 	& Omit<ObjectType, KeysType>;
+
+/*
+Create a type that requires at only one of the given properties. The remaining properties are kept as is.
```
sindresorhus:
```suggestion
Create a type that requires only one of the given properties. The remaining properties are kept as is.
```

index.d.ts
```diff
@@ -198,3 +198,36 @@ export type RequireAtLeastOne<ObjectType, KeysType extends keyof ObjectType = ke
...
+	secure: boolean;
+};
+
+const responder: RequireOnlyOne<Responder, 'text' | 'json'> = {
+	json: () => '{"message": "ok"}',
+	secure: true
+};
+```
+*/
+export type RequireOnlyOne<ObjectType, KeysType extends keyof ObjectType = keyof ObjectType> =
```
sindresorhus: "Require only one" makes me think that it requires only one, but doesn't make it clear that it disallows more than one. Maybe we should simply name it `RequireOne`?

sindresorhus:
Can you fix the merge conflict? (Note that I have moved types to separate files in `master`).

sindresorhus:
@izayl Ping :)

WORMSS:
I just wanted to say that I simplified this concept a little so I could use it to build up a type rather than modify an existing type.

```
type OnlyOne<T> = {
    [K in keyof T]: (
        Required<Pick<T, K>> & Partial<Record<Exclude<keyof T, K>, never>>
    )
}[keyof T];
```
Tested with
```
type Test = OnlyOne<{
    a: boolean;
    b: boolean;
}> & Partial<OnlyOne<{
    c: boolean;
    d: boolean;
}>>;

const t1: Test = {
    a: true,
};
const t2: Test = {
    b: true,
};
const t3: Test = { // Should fail, a or b must be supplied
    
}
const t4: Test = { // Should fail a and b cannot be supplied together
    a: true,
    b: true,
}
const t5: Test = {
    a: true,
    c: true,
}
const t6: Test = { // Should fail c and d cannot be supplied together
    a: true,
    c: true,
    d: true,
}
const t7: Test = { // Should fail, as even though c is supplied, you still need a or b
    c: true,
}
```

I hope this helps anyone else in the future.

sindresorhus:
Closing for lack of response.

https://github.com/sindresorhus/type-fest/issues/54

PR #25, opened 2019-04-08
Add `bigint` primitive

xiaoxiangmoe:

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -12,7 +12,8 @@ export type Primitive =
 	| string
 	| number
 	| boolean
-	| symbol;
+	| symbol
+	| bigint;
 
 /**
 Matches a [`class` constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes).
@@ -31,7 +32,9 @@ export type TypedArray =
 	| Int32Array
 	| Uint32Array
 	| Float32Array
-	| Float64Array;
+	| Float64Array
+	| BigInt64Array
+	| BigUint64Array;
 
 /**
 Matches a JSON object.


sindresorhus:
While you're at it, can you add `BigInt64Array` and `BigUInt64Array` to `TypedArray` too?

BendingBender:
We will need to explicitly reference the `esnext` lib for this to work correctly for all users. And the compiler target has to be set to `esnext`. I don't think that we should release it like this, this will break most users' setups.

We could put it in a different definition file that has to be included separately to fix this issue.

sindresorhus:
> We will need to explicitly reference the esnext lib for this to work correctly for all users. 

For `BigInt64Array` and `BigUInt64Array`, yes, but for `bigint`, it should just work.

> And the compiler target has to be set to esnext.

Are you sure about that? I'm able to compile TS using `bigint` with any `target` setting.

BendingBender:
I didn' try it out, I read it up [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-2.html#caveats-1).

sindresorhus:
That is about the BigInt polyfilling though. We're only using the literal *type* here.

sindresorhus:
@BendingBender If we add `/// <reference lib="esnext"/>` to the top of the file, does it force the user to manually add `esnext` to their tsconfig or will it *just work*? Any downsides?

BendingBender:
I've played a little with `bigint`s and I have no idea what's happening there. It looks like TS has no problems with them even without specifying the `esnext` lib or `esnext` as target. The only thing that seems not to work without `esnext` as target is the literal notation (`1n`). So as far as I see it, it should just work for everyone without further tweaks from our side.

Would be great if someone could verify this with their setup @xiaoxiangmoe?

SamVerschueren:
We will have to add `/// <reference lib="esnext"/>` because of this. It fails if we don't set our target to `esnext`. Will add a quick PR.

Screen Shot 2019-05-15 at 11 16 49

PR #26, opened 2019-04-11
Add `Mutable` type

### LINKED ISSUE #17: Add `Mutable` type
The inverse of the built-in [`Readonly`](https://github.com/Microsoft/TypeScript/blob/2961bc3fc0ea1117d4e53bc8e97fa76119bc33e3/src/lib/es5.d.ts#L1415-L1420) type.

It can be useful to temporarily use a readonly value or a copy of the value as mutable.

Use-case: https://github.com/sindresorhus/pageres/blob/4a5d05fca19a5fbd2f53842cbf3eb7b1b63bddd2/source/index.ts#L72

Thoughts?

NotWoods:
Fixes #17

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -208,7 +208,7 @@ export type RequireAtLeastOne<ObjectType, KeysType extends keyof ObjectType = ke
 /**
 Convert an object with `readonly` properties into a mutable object. Inverse of `Readonly<T>`.
 
-This can be used to [store and mutate options within a class](https://github.com/sindresorhus/pageres/blob/4a5d05fca19a5fbd2f53842cbf3eb7b1b63bddd2/source/index.ts#L72), [edit readonly objects within tests](https://stackoverflow.com/questions/50703834), and [construct a `readonly` object within a function](https://github.com/Microsoft/TypeScript/issues/24509).
+This can be used to [store and mutate options within a class](https://github.com/sindresorhus/pageres/blob/4a5d05fca19a5fbd2f53842cbf3eb7b1b63bddd2/source/index.ts#L72), [edit `readonly` objects within tests](https://stackoverflow.com/questions/50703834), and [construct a `readonly` object within a function](https://github.com/Microsoft/TypeScript/issues/24509).
 
 @example
 ```
--- a/index.d.ts
+++ b/index.d.ts
@@ -174,7 +174,7 @@ export type LiteralUnion<
 	BaseType extends Primitive
 > = LiteralType | (BaseType & {_?: never});
 
-/*
+/**
 Create a type that requires at least one of the given properties. The remaining properties are kept as is.
 
 @example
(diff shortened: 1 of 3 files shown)


test-d/mutable.ts
```diff
@@ -0,0 +1,11 @@
+import {expectType} from 'tsd';
+import {Mutable} from '..';
+
+type Foo = {
+	readonly a: number;
+	readonly b: string;
+};
+
+const ab: Mutable<Foo> = {a: 1, b: '2'};
+expectType<{a: number; b: string}>(ab);
```
BendingBender: Sadly, immutable object types are assignable to mutable ones. So this test is basically a no-op, it will accept mutable and immutable objects without erroring out.

You shoud change this to:
```suggestion
ab.a = 2;
```

test-d/mutable.ts
```diff
@@ -0,0 +1,11 @@
...
+expectType<Mutable<Readonly<Foo>>>(ab);
```
BendingBender: Same as above, change this to:
```suggestion
const ab2 = Mutable<Readonly<Foo>> = ab;
ab2.a = 2;
```

index.d.ts
```diff
@@ -204,3 +204,23 @@ export type RequireAtLeastOne<ObjectType, KeysType extends keyof ObjectType = ke
 	}[KeysType]
 	// …then, make intersection types by adding the remaining properties to each mapped type.
 	& Omit<ObjectType, KeysType>;
+
+/*
+Converts a readonly object into a mutable object. Inverse of Readonly<T>.
```
sindresorhus:
```suggestion
Convert a readonly object into a mutable object. Inverse of `Readonly<T>`.
```

index.d.ts
```diff
@@ -204,3 +204,23 @@ export type RequireAtLeastOne<ObjectType, KeysType extends keyof ObjectType = ke
...
+
```
sindresorhus: Would be good to include some text on real-world use-cases of this type, with links if possible.

index.d.ts
```diff
@@ -204,3 +204,23 @@ export type RequireAtLeastOne<ObjectType, KeysType extends keyof ObjectType = ke
...
+	readonly a: number;
+	readonly b: string;
+};
+
+const mutableFoo: Mutable<Foo> = {a: 1, b: '2'};
+mutableFoo.a = 3;
+```
+*/
+export type Mutable<ObjectType> = {
+	-readonly [KeyType in keyof ObjectType]: ObjectType[KeyType];
```
sindresorhus: Would be useful for learning purposes for users if there was an inline explanation of how it works, like https://github.com/sindresorhus/type-fest/blob/d84eedf0178aa98056cc03f7851dfae39f7fc90c/index.d.ts#L197-L206

index.d.ts
sindresorhus: I don't think we should say `readonly` as that might be confused with the `readonly` keyword, which cannot be used on arbitrary objects. Instead we should maybe say that the properties/keys are readonly. Thoughts?

NotWoods: Agreed, I'll update the comment

sindresorhus:
Looks good :)

PR #29, opened 2019-04-21
Fix broken imports

BendingBender:
Fixes broken imports. I don't have a clue why `tsd` doesn't catch them.

### DIFF
--- a/source/package-json.d.ts
+++ b/source/package-json.d.ts
@@ -307,6 +307,7 @@ export type PackageJson = {
 	/**
 	The licenses for the package.
 	*/
+	// eslint-disable-next-line @typescript-eslint/array-type
 	licenses?: Array<{
 		type?: string;
 		url?: string;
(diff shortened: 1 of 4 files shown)


source/package-json.d.ts
```diff
@@ -307,6 +307,7 @@ export type PackageJson = {
 	/**
 	The licenses for the package.
 	*/
+	// eslint-disable-next-line @typescript-eslint/array-type
```
sindresorhus: These are not needed. I think you have an outdated `node_modules` folder or outdated cache.

https://travis-ci.com/sindresorhus/type-fest/jobs/194522350

sindresorhus:
https://github.com/SamVerschueren/tsd/issues/24

source/package-json.d.ts
BendingBender: Wasn't aware of it, thanks!

SamVerschueren:
Fixed https://github.com/SamVerschueren/tsd/issues/24#issuecomment-508945163

PR #34, opened 2019-05-06
Add `ReadonlyDeep` type

BendingBender:
This is useful when a deeply nested structure needs to be exposed as completely immutable, like for example an imported JSON module.


source/readonly-deep.d.ts
```diff
@@ -0,0 +1,57 @@
+import {Primitive} from './basic';
+
+/**
+Convert `object`s, `Map`s, `Set`s, and `Array`s into immutable stuctures recursively. Converts all properties/contents into immutable structures recusively.
```
sindresorhus:
```suggestion
Convert `object`s, `Map`s, `Set`s, and `Array`s and all of their properties/elements into immutable structures recursively.
```

?

source/readonly-deep.d.ts
```diff
@@ -0,0 +1,57 @@
...
+export default data;
+
+// test.ts
+import data from './main';
+
+data.foo.push('bar');
+//=> error TS2339: Property 'push' does not exist on type 'readonly string[]'
+```
+*/
+export type ReadonlyDeep<T> = T extends Primitive | ((...arguments: any[]) => unknown)
```
sindresorhus: Should we maybe move `((...arguments: any[]) => unknown)` be a `AnyFunction` type?

source/readonly-deep.d.ts
```diff
@@ -0,0 +1,57 @@
...
+	? T
+	: T extends ReadonlyMap<infer KeyType, infer ValueType>
+	? ReadonlyMapDeep<KeyType, ValueType>
+	: T extends ReadonlySet<infer ItemType>
+	? ReadonlySetDeep<ItemType>
+	: T extends object
+	? ReadonlyObjectDeep<T>
```
sindresorhus: Would there be any benefits in specifically targeting array here? Maybe improved debugging. Or would the result be exactly the same as now?

readme.md
```diff
@@ -67,7 +67,8 @@ Click the type names for complete docs.
 - [`Merge`](source/merge.d.ts) - Merge two types into a new type. Keys of the second type overrides keys of the first type.
 - [`MergeExclusive`](source/merge-exclusive.d.ts) - Create a type that has mutually exclusive properties.
 - [`RequireAtLeastOne`](source/require-at-least-one.d.ts) - Create a type that requires at least one of the given properties.
-- [`LiteralUnion`](source/literal-union.d.ts) - Allows creating a union type by combining primitive types and literal types without sacrificing auto-completion in IDEs for the literal type part of the union. Workaround for [Microsoft/TypeScript#29729](https://github.com/Microsoft/TypeScript/issues/29729).
+- [`ReadonlyDeep`](source/readonly-deep.d.ts) - Make a nested `object`/`Map`/`Set`/`Array` deeply immutable.
```
sindresorhus: This makes it sound like it modifies the existing type. I assume it doesn't do that?

Instead, maybe it should read like this:

```suggestion
- [`ReadonlyDeep`](source/readonly-deep.d.ts) - Create a deeply immutable version of a `object`/`Map`/`Set`/`Array` type.
```

?

.travis.yml
```diff
@@ -3,4 +3,3 @@ node_js:
   - '12'
   - '10'
   - '8'
-  - '6'
```
sindresorhus: This is unrelated to this PR and should be done afterwards. Should also be done in package.json.

test-d/readonly-deep.ts
```diff
@@ -0,0 +1,44 @@
...
+	set: new Set<string>(),
+	array: ['foo'],
+	tuple: ['foo'] as ['foo'],
+	readonlyMap: new Map<string, string>() as ReadonlyMap<string, string>,
+	readonlySet: new Set<string>() as ReadonlySet<string>,
+	readonlyArray: ['foo'] as ReadonlyArray<string>,
+	readonlyTuple: ['foo'] as const
+};
+
+const readonlyData: ReadonlyDeep<typeof data> = data;
```
sindresorhus: Any reason of doing this over `const readonlyData = data as ReadonlyDeep<typeof data>;`? Or just a preference?

source/readonly-deep.d.ts
```diff
@@ -0,0 +1,57 @@
+import {Primitive} from './basic';
+
+/**
+Convert `object`s, `Map`s, `Set`s, and `Array`s into immutable stuctures recursively. Converts all properties/contents into immutable structures recusively.
+
+This is useful when a deeply nested structure needs to be exposed as completely immutable, for example, an imported JSON module.
```
sindresorhus:
```suggestion
This is useful when a deeply nested structure needs to be exposed as completely immutable, for example, an imported JSON module or when receiving an API response that is passed around.
```

sindresorhus:
👍 This type is very useful. I've been meaning to open an issue about adding it.

source/readonly-deep.d.ts
```diff
@@ -0,0 +1,57 @@
...
+
```
sindresorhus: Maybe add a note to encourage people to upvote https://github.com/microsoft/TypeScript/issues/13923 if they want to see this type as a built-in?

sindresorhus:
Can you ensure we handle all the cases in https://github.com/krzkaczor/ts-essentials/blob/ddc4e4653b3de716b68d02c961c85d0a703a96fa/test/index.ts#L7-L56 ? Also: https://github.com/krzkaczor/ts-essentials/issues/40

test-d/readonly-deep.ts
```diff
@@ -0,0 +1,44 @@
...
+	null: null,
+	undefined: undefined, // eslint-disable-line object-shorthand
+	map: new Map<string, string>(),
+	set: new Set<string>(),
+	array: ['foo'],
+	tuple: ['foo'] as ['foo'],
+	readonlyMap: new Map<string, string>() as ReadonlyMap<string, string>,
+	readonlySet: new Set<string>() as ReadonlySet<string>,
+	readonlyArray: ['foo'] as ReadonlyArray<string>,
+	readonlyTuple: ['foo'] as const
```
sindresorhus: Can/should we handle values in a Promise?

See: https://gist.github.com/masterkidan/7322752f569b1bba53e0426266768623

ThomasdenH:
Maybe this can be named `DeepReadonly` instead of `ReadonlyDeep`? The former is a commonly used name and would therefore be easier to discover.

source/readonly-deep.d.ts
BendingBender: Not sure. If we do this, why not simply using `Function`?

source/readonly-deep.d.ts
BendingBender: Nope, TS 3.4 handles Arrays correctly with the `ReadonlyObjectDeep`. I've tried it out, it is actually harmful adding special casing for Arrays, it becomes very complicated to handle all variants of `readonly`, `Array`, and `[..., ...]` (tuples) correctly. The way it is right now works like a charm, all the special cases are green in the tests.

test-d/readonly-deep.ts
BendingBender: I suppose I was very uninspired when I wrote this. Do you want me to change this? There is no technical difference between the two.

test-d/readonly-deep.ts
BendingBender: I don't think that we should handle promises here.

BendingBender:
> Can you ensure we handle all the cases in [krzkaczor/ts-essentials:test/index.ts@`ddc4e46`#L7-L56](https://github.com/krzkaczor/ts-essentials/blob/ddc4e4653b3de716b68d02c961c85d0a703a96fa/test/index.ts#L7-L56) ? Also: [krzkaczor/ts-essentials#40](https://github.com/krzkaczor/ts-essentials/issues/40)

The test cases already verify that these edge cases work correctly.

BendingBender:
> Maybe this can be named `DeepReadonly` instead of `ReadonlyDeep`? The former is a commonly used name and would therefore be easier to discover.

@sindresorhus Your opinion?

sindresorhus:
I prefer `ReadonlyDeep`, and I would argue it's actually more discoverable as it will be sorted with `Readonly` alphabetically. Lodash also puts `deep` at the end in their methods.

test-d/readonly-deep.ts
sindresorhus: Alright

source/readonly-deep.d.ts
sindresorhus: Good point. Never mind.

test-d/readonly-deep.ts
sindresorhus: Nah, was just curious.

jpike88:
Does it handle this case?

`delete ReadonlyObjectDeep[key];`

source/readonly-deep.d.ts
```diff
@@ -0,0 +1,57 @@
...
+*/
+export type ReadonlyDeep<T> = T extends Primitive | ((...arguments: any[]) => unknown)
+	? T
+	: T extends ReadonlyMap<infer KeyType, infer ValueType>
+	? ReadonlyMapDeep<KeyType, ValueType>
+	: T extends ReadonlySet<infer ItemType>
+	? ReadonlySetDeep<ItemType>
+	: T extends object
+	? ReadonlyObjectDeep<T>
+	: unknown;
```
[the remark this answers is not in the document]
sindresorhus: // @BendingBender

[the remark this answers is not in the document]
carpben: Checking for primitives at the beginning is necessary for performance.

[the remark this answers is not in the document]
sindresorhus: @carpben Can you move this to a new issue? I don't really know the answer and that way other people could chime in.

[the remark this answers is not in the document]
carpben: @sindresorhus , I opened https://github.com/sindresorhus/type-fest/issues/116. Sincerely sorry for the long delay!

PR #41, opened 2019-06-13
Fix logo

Knutakir:
The logo seems to be acting weird for me in Firefox:
image

This can be fixed by changing a single number: `66964324390299` => `0`. This is found in: `<path fill="url(#y)" d= ... />`


CvX:
@sindresorhus the build fails because `@typescript-eslint/eslint-plugin` isn't in the type-fest's dependencies. 🤔 How did it work before, what has changed in the dependencies?

sindresorhus:
Weird, they seem to have changed to require `@typescript-eslint/parser` being installed manually.

PR #43, opened 2019-07-01
Add `Stringified` type

youssefsharief:
I believe this type can have a lot of uses.


sindresorhus:
What uses? Preferably link to some real code using it.

youssefsharief:
I used before to stringify my interface so that I am able to use the object as a form model. Let us say I have a car with the speed prop. This prop is a number in the domain model, but when I want to use this in a form, I need to stringify it.

source/stringified.d.ts
```diff
@@ -0,0 +1,20 @@
...
+@example
+```
+import {Stringified} from 'type-fest';
+
+type Foo = {
+	a: number;
+	b: string;
+};
+
+type FooWithoutA = Stringified<Foo, 'a'>;
```
bfred-it: `Stringified` only has one generic argument: `ObjectType`

What does `'a'` map to?

youssefsharief: Yes, I was wrong on that. I fixed it

sindresorhus:
Makes sense. Can you please follow each step in https://github.com/sindresorhus/type-fest/blob/master/.github/contributing.md

readme.md
```diff
@@ -71,6 +71,7 @@ Click the type names for complete docs.
 - [`RequireAtLeastOne`](source/require-at-least-one.d.ts) - Create a type that requires at least one of the given properties.
 - [`ReadonlyDeep`](source/readonly-deep.d.ts) - Create a deeply immutable version of a `object`/`Map`/`Set`/`Array` type.
 - [`LiteralUnion`](source/literal-union.d.ts) - Create a union type by combining primitive types and literal types without sacrificing auto-completion in IDEs for the literal type part of the union. Workaround for [Microsoft/TypeScript#29729](https://github.com/Microsoft/TypeScript/issues/29729).
+- [`Stringified`](source/stringified.d.ts) - Create a stringified type of an object
```
sindresorhus: Move it to the bottom (same in the other places)

readme.md
```diff
@@ -71,6 +71,7 @@ Click the type names for complete docs.
 - [`RequireAtLeastOne`](source/require-at-least-one.d.ts) - Create a type that requires at least one of the given properties.
 - [`ReadonlyDeep`](source/readonly-deep.d.ts) - Create a deeply immutable version of a `object`/`Map`/`Set`/`Array` type.
 - [`LiteralUnion`](source/literal-union.d.ts) - Create a union type by combining primitive types and literal types without sacrificing auto-completion in IDEs for the literal type part of the union. Workaround for [Microsoft/TypeScript#29729](https://github.com/Microsoft/TypeScript/issues/29729).
+- [`Stringified`](source/stringified.d.ts) - Create a stringified type of an object.
```
sindresorhus: The description is not clear enough. It doesn't stringify the object itself.

source/stringified.d.ts
```diff
@@ -0,0 +1,20 @@
...
+type Foo = {
+	a: number;
+	b: string;
+};
+
+type StringifiedFoo = Stringified<Foo>;
+//=> {
+	a: string;
+	b: string;
+};
```
sindresorhus: Use multiline comment.

readme.md
sindresorhus: This was not resolved. It's still not clear enough what it does.

source/stringified.d.ts
```diff
@@ -0,0 +1,20 @@
+/**
+Create a stringified type of an object.
```
sindresorhus: Please spend some more time on the description.

Also per PR guidelines, you need to include some use-cases in text form.

source/stringified.d.ts
sindresorhus: Don't resolve things that are not resolved.

sindresorhus:
`Stringified` makes me think it converts the whole object to a string or something. Maybe we should be explicit and call it `StringifiedKeys`? Any other naming ideas?

sindresorhus:
@youssefsharief Friendly bump :)

sindresorhus:
Closing for lack of response: https://github.com/sindresorhus/type-fest/issues/87

Issue #27, opened 2019-04-16
Rename `Omit`


sindresorhus:
We managed to convince the TS team to add `Omit` as a built-in type, unfortunately, they added a loose version, not what we suggested, and [they insist on staying with it](https://github.com/Microsoft/TypeScript/issues/30825).

The great thing about user-land code is that we can fix their dumb decisions.

However, that does mean our `Omit` type will conflict with the native one... But I guess that gives us the opportunity to come up with a better name for it.

I never really found `Omit` that clear. How about `ExcludeProperty` or `ExcludeKey`? I prefer the former. Open to other suggestions.

BendingBender:
I find `ExcludeProperty` very appealing.

CvX:
The another option is `StrictOmit`, to highlight both similarity to the built-in and its main difference. But I'm ok with any of suggested names.

sindresorhus:
@CvX I don't like having to prefix types with `Strict`. They should IMHO be strict by default. In this case, it's not up to us, so I would rather have a different name.

sindresorhus:
Alright. We'll go with `ExcludeProperty`. But let's defer doing anything until it's 100% certain that TS 3.5 will ship with a loose `Omit` type.

sindresorhus:
Relevant discussion: https://github.com/microsoft/TypeScript/pull/31205

sindresorhus:
Since we're renaming this anyway, maybe we should name it `ExcludeProperties` and add support for excluding multiple properties?

BendingBender:
As far as I understand the type, it will already exclude multiple properties when used with a union type:
```
Omit<SomeType, 'foo' | 'bar'>
```
will omit properties `foo` and `bar`.

andyfleming:
While I wish they would have stayed consistent with what the community already understands as the behavior of `Omit`, I do actually like having an omit-key-if-exists AND a omit-key form of the type helper.

andyfleming:
I think I like `ExcludeKey` over `ExcludeProperty` since it matches the language they use with syntax types like `keyof`.

andyfleming:
I'm also not convinced we need a second word in the type name.

Something like `Exclude`, `Remove`, or `Without` could be viable.

sindresorhus:
> As far as I understand the type, it will already exclude multiple properties when used with a union type:

I did not realize that. We should definitely document that better.

sindresorhus:
[...]

PR #44, opened 2019-07-03
Rename `Omit` to `Except`

kainiedziela:
Discussion: #27 

TypeScript incorporated the `Omit` type without making it strict. This pull request changes the type-fest-`Omit` naming to `Except`, that allows for the usage of the native TypeScript-`Omit` type, while maintaining the stricter usage of the type-fest solution.


readme.md
```diff
@@ -64,7 +64,7 @@ Click the type names for complete docs.
 
 ### Utilities
 
-- [`Omit`](source/omit.d.ts) - Create a type from an object type without certain keys.
+- [`Except`](source/except.d.ts) - Create a type from an object type without certain keys.
```
sindresorhus: I think you should mention that this is a stricter version than the built-in `Omit`.

source/except.d.ts
```diff
@@ -0,0 +1,17 @@
+/**
+Create a type from an object type without certain keys.
+
```
sindresorhus: Same as https://github.com/sindresorhus/type-fest/pull/44/files#r300570410 here, but also include how it differs.

sindresorhus: Also a note to encourage people to vote and comment on https://github.com/microsoft/TypeScript/issues/30825 if they want this as a built-in helper.

source/except.d.ts
```diff
@@ -0,0 +1,17 @@
...
+@example
+```
+import {Except} from 'type-fest';
+
+type Foo = {
+	a: number;
+	b: string;
+};
+
+type FooWithoutA = Except<Foo, 'a'>;
```
sindresorhus: Can you make the example exclude two props to make it clear that it works with multiple props. That was not clear to me (https://github.com/sindresorhus/type-fest/issues/27#issuecomment-490267194).

sindresorhus:
Thanks :)

PR #45, opened 2019-07-03
Add `Promisable` type

### LINKED ISSUE #40: Proposal: Promisable
Some times we are just awaiting a function, so we don't really care if the returned value is actually a promise or not. With `Promisable`:

```ts
type Promisable<T> = T | Promise<T>;
```

One could write this:

```ts
type Foo = () => Promisable<string | number | null>
```

Rather than:

```ts
type Foo = () => string | number | null | Promise<string> | Promise<number> | Promise<null>
```

It helps write a bit cleaner code. I'm not sure if it's useful enough to be included here, but I thought I'd mention it.

kainiedziela:
Fixes #40 
TypeScript proposal: [#31394](https://github.com/microsoft/TypeScript/issues/31394)

A type for values that can either be a promise or it's return value.

```TypeScript
type Promisable<T> = T | PromiseLike<T>;
```

**Example**

```TypeScript
async function logger(getLogEntry: () => Awaitable<string>): Promise<void> {
    const entry = await getLogEntry();
    console.log(entry);
}

logger(() => 'foo');
logger(() => Promise.resolve('bar'));
```

**Use cases**

* A function accepts a callback that may either return a value synchronously or may return a promised value.
* Specifically this type could be the return type of `Promise.then()`, `Promise.catch()` and `Promise.finally()` callbacks.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -71,7 +71,7 @@ Click the type names for complete docs.
 - [`RequireAtLeastOne`](source/require-at-least-one.d.ts) - Create a type that requires at least one of the given properties.
 - [`ReadonlyDeep`](source/readonly-deep.d.ts) - Create a deeply immutable version of a `object`/`Map`/`Set`/`Array` type.
 - [`LiteralUnion`](source/literal-union.d.ts) - Create a union type by combining primitive types and literal types without sacrificing auto-completion in IDEs for the literal type part of the union. Workaround for [Microsoft/TypeScript#29729](https://github.com/Microsoft/TypeScript/issues/29729).
-- [`Promisable`](source/promisable.d.ts) - Create a type for values that can either be a promise or it's return value, and as such both accepts synchronous values and promise-likes of those values.
+- [`Promisable`](source/promisable.d.ts) - Create a type that represents either the value or the value wrapped in `PromiseLike`.
 
 ### Miscellaneous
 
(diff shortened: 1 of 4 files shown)


readme.md
```diff
@@ -71,6 +71,7 @@ Click the type names for complete docs.
 - [`RequireAtLeastOne`](source/require-at-least-one.d.ts) - Create a type that requires at least one of the given properties.
 - [`ReadonlyDeep`](source/readonly-deep.d.ts) - Create a deeply immutable version of a `object`/`Map`/`Set`/`Array` type.
 - [`LiteralUnion`](source/literal-union.d.ts) - Create a union type by combining primitive types and literal types without sacrificing auto-completion in IDEs for the literal type part of the union. Workaround for [Microsoft/TypeScript#29729](https://github.com/Microsoft/TypeScript/issues/29729).
+- [`Promisable`](source/promisable.d.ts) - Create a type for values that can either be a promise or it's return value, and as such both accepts synchronous values and promise-likes of those values.
```
sindresorhus: I still don't think this is clear enough. Couldn't it simply be:

> Create a type that represents either the value or the value wrapped in `Promise`.

kainiedziela: I think that would indeed be more readable.

sindresorhus: 👍

PR #47, opened 2019-07-04
Add `PartialDeep` type

### LINKED ISSUE #18: Proposal: DeepPartial
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/type-fest/issues/18)

```ts
interface Child {
  id: string;
  name: string;
  state: string;
  mother: {
    name: string;
    some: string;
  }
}

function fn(child: Partial<Child>) { }

fn({ name: '' }); // OK
fn({ mother: { name: '' }}); // NO, Property 'some' is missing
```

But.

```ts
type DeepPartial<T> = { [P in keyof T]?: DeepPartial<T[P]> };

function fn(child: DeepPartial<Child>) { }

fn({ name: '' }); // OK
fn({ mother: { name: '' }}); // OK
```

How about this?

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $60.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($60.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/type-fest/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

kainiedziela:
Fixes #18 

Create a type from another type with all properties set to optional and the nested properties of all properties set to optional, as opposed to the [`Partial<T>`](https://github.com/Microsoft/TypeScript/blob/2961bc3fc0ea1117d4e53bc8e97fa76119bc33e3/src/lib/es5.d.ts#L1401-L1406) type which constructs a type with properties of T that have a depth of 1 set to optional.

**Example**

```TypeScript
import {PartialDeep} from 'type-fest';

interface Baz {
    qux: string;
    corge: {
        waldo: number[];
        fred: string;
    };
}

function baz(bar: Partial<Baz>) { }
function foo(bar: PartialDeep<Baz>) { }

baz(({ corge: { waldo: [ 1, 2 ] } }); // Property ‘fred’ is missing in type ‘{waldo: number[]; fred: string;}’
foo({ corge: { waldo: [ 1, 2 ] } }); // Passes
```

**Use cases**

* Merging a default settings/config object with another object, the second object would be a deep partial of the default object.
* Mocking and testing complex entities, where populating an entire object with it's properties would be redundant in terms of the mock or test.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -69,7 +69,7 @@ Click the type names for complete docs.
 - [`Merge`](source/merge.d.ts) - Merge two types into a new type. Keys of the second type overrides keys of the first type.
 - [`MergeExclusive`](source/merge-exclusive.d.ts) - Create a type that has mutually exclusive properties.
 - [`RequireAtLeastOne`](source/require-at-least-one.d.ts) - Create a type that requires at least one of the given properties.
-- [`PartialDeep`](source/partial-deep.d.ts) - Create a deeply optional version of another type, as opposed to the [`Partial<T>`](https://github.com/Microsoft/TypeScript/blob/2961bc3fc0ea1117d4e53bc8e97fa76119bc33e3/src/lib/es5.d.ts#L1401-L1406) type which creates a type with properties of T that have a depth of 1 set to optional.
+- [`PartialDeep`](source/partial-deep.d.ts) - Create a deeply optional version of another type, as opposed to the [`Partial<T>`](https://github.com/Microsoft/TypeScript/blob/2961bc3fc0ea1117d4e53bc8e97fa76119bc33e3/src/lib/es5.d.ts#L1401-L1406) type which creates a type with all properties of `T` set to optional, but only one level deep.
 - [`ReadonlyDeep`](source/readonly-deep.d.ts) - Create a deeply immutable version of a `object`/`Map`/`Set`/`Array` type.
 - [`LiteralUnion`](source/literal-union.d.ts) - Create a union type by combining primitive types and literal types without sacrificing auto-completion in IDEs for the literal type part of the union. Workaround for [Microsoft/TypeScript#29729](https://github.com/Microsoft/TypeScript/issues/29729).
 
--- a/readme.md
+++ b/readme.md
@@ -69,8 +69,8 @@ Click the type names for complete docs.
 - [`Merge`](source/merge.d.ts) - Merge two types into a new type. Keys of the second type overrides keys of the first type.
 - [`MergeExclusive`](source/merge-exclusive.d.ts) - Create a type that has mutually exclusive properties.
 - [`RequireAtLeastOne`](source/require-at-least-one.d.ts) - Create a type that requires at least one of the given properties.
-- [`PartialDeep`](source/partial-deep.d.ts) - Create a deeply optional version of another type, as opposed to the [`Partial<T>`](https://github.com/Microsoft/TypeScript/blob/2961bc3fc0ea1117d4e53bc8e97fa76119bc33e3/src/lib/es5.d.ts#L1401-L1406) type which creates a type with all properties of `T` set to optional, but only one level deep.
-- [`ReadonlyDeep`](source/readonly-deep.d.ts) - Create a deeply immutable version of a `object`/`Map`/`Set`/`Array` type.
+- [`PartialDeep`](source/partial-deep.d.ts) - Create a deeply optional version of another type. Use [`Partial<T>`](https://github.com/Microsoft/TypeScript/blob/2961bc3fc0ea1117d4e53bc8e97fa76119bc33e3/src/lib/es5.d.ts#L1401-L1406) if you only need one level deep.
+- [`ReadonlyDeep`](source/readonly-deep.d.ts) - Create a deeply immutable version of an `object`/`Map`/`Set`/`Array` type. Use [`Readonly<T>`](https://github.com/Microsoft/TypeScript/blob/2961bc3fc0ea1117d4e53bc8e97fa76119bc33e3/src/lib/es5.d.ts#L1415-L1420) if you only need one level deep.
 - [`LiteralUnion`](source/literal-union.d.ts) - Create a union type by combining primitive types and literal types without sacrificing auto-completion in IDEs for the literal type part of the union. Workaround for [Microsoft/TypeScript#29729](https://github.com/Microsoft/TypeScript/issues/29729).
 
 ### Miscellaneous
--- a/readme.md
+++ b/readme.md
@@ -69,7 +69,7 @@ Click the type names for complete docs.
 - [`Merge`](source/merge.d.ts) - Merge two types into a new type. Keys of the second type overrides keys of the first type.
 - [`MergeExclusive`](source/merge-exclusive.d.ts) - Create a type that has mutually exclusive properties.
 - [`RequireAtLeastOne`](source/require-at-least-one.d.ts) - Create a type that requires at least one of the given properties.
-- [`PartialDeep`](source/partial-deep.d.ts) - Create a type from another type with all properties set to optional and the nested properties of all properties set to optional, as opposed to the [`Partial<T>`](https://github.com/Microsoft/TypeScript/blob/2961bc3fc0ea1117d4e53bc8e97fa76119bc33e3/src/lib/es5.d.ts#L1401-L1406) type which constructs a type with properties of T that have a depth of 1 set to optional.
+- [`PartialDeep`](source/partial-deep.d.ts) - Create a deeply optional version of another type, as opposed to the [`Partial<T>`](https://github.com/Microsoft/TypeScript/blob/2961bc3fc0ea1117d4e53bc8e97fa76119bc33e3/src/lib/es5.d.ts#L1401-L1406) type which creates a type with properties of T that have a depth of 1 set to optional.
 - [`ReadonlyDeep`](source/readonly-deep.d.ts) - Create a deeply immutable version of a `object`/`Map`/`Set`/`Array` type.
 - [`LiteralUnion`](source/literal-union.d.ts) - Create a union type by combining primitive types and literal types without sacrificing auto-completion in IDEs for the literal type part of the union. Workaround for [Microsoft/TypeScript#29729](https://github.com/Microsoft/TypeScript/issues/29729).
 
--- a/readme.md
+++ b/readme.md
@@ -69,6 +69,7 @@ Click the type names for complete docs.
 - [`Merge`](source/merge.d.ts) - Merge two types into a new type. Keys of the second type overrides keys of the first type.
 - [`MergeExclusive`](source/merge-exclusive.d.ts) - Create a type that has mutually exclusive properties.
 - [`RequireAtLeastOne`](source/require-at-least-one.d.ts) - Create a type that requires at least one of the given properties.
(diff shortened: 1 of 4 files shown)


source/partial-deep.d.ts
```diff
@@ -0,0 +1,27 @@
...
+foo({ corge: { waldo: [ 1, 2 ] } });
+```
+*/
+export type PartialDeep<T> = {
+    [KeyObject in keyof T]?: T[KeyObject] extends Array<infer InferredType>
+        ? Array<PartialDeep<InferredType>>
+        : T[KeyObject] extends ReadonlyArray<infer InferredType>
+            ? ReadonlyArray<PartialDeep<InferredType>>
+            : PartialDeep<T[KeyObject]>;
+};
```
sindresorhus: Does this also handle `Map` and `Set`? Would be good to have tests for those cases. See how it's handled in the other deep type: https://github.com/sindresorhus/type-fest/blob/5809b499720eb30f98967aedc8b9403a1427b26e/source/readonly-deep.d.ts#L32-L40

source/partial-deep.d.ts
```diff
@@ -0,0 +1,27 @@
+/**
+Create a type from another type with all properties set to optional and the nested properties of all properties set to optional
+
```
sindresorhus: Can you include the use-cases mentioned in the PR description here too.

sindresorhus:
Not sure if it applies here, but note https://github.com/sindresorhus/type-fest/pull/34#discussion_r281852687

sindresorhus:
This needs more tests, see: https://github.com/sindresorhus/type-fest/blob/master/test-d/readonly-deep.ts

kainiedziela:
@sindresorhus mind giving this another look?

test-d/partial-deep.ts
```diff
@@ -0,0 +1,32 @@
...
+        object: {key: 'value'},
+        string: 'waldo',
+        number: 1,
+        boolean: false,
+        symbol: Symbol('test'),
+        null: null,
+        undefined: undefined, // eslint-disable-line object-shorthand
+        map: new Map<string, string>(),
+        set: new Set<string>(),
+        array: ['foo']
```
sindresorhus: Use tab-indentation.

source/partial-deep.d.ts
```diff
@@ -0,0 +1,46 @@
...
+    [KeyType in keyof T]?:
+        T[KeyType] extends Map<infer MapKeyType, infer MapValueType>
+        ? PartialMap<MapKeyType, MapValueType>
+        : T[KeyType] extends Set<infer SetType>
+        ? PartialSet<SetType>
+        : T[KeyType] extends Array<infer ArrayType>
+        ? Array<PartialDeep<ArrayType>>
+        : T[KeyType] extends ReadonlyArray<infer ReadonlyArrayType>
+        ? ReadonlyArray<PartialDeep<ReadonlyArrayType>>
+        : PartialDeep<T[KeyType]>;
```
sindresorhus: Use tab-indentation.

source/partial-deep.d.ts
```diff
@@ -0,0 +1,46 @@
...
+    corge: {
+        waldo: number[];
+        fred: string;
+    };
+}
+
+function foo(bar: PartialDeep<Baz>) { }
+
+foo({ corge: { waldo: [ 1, 2 ] } });
+```
```
sindresorhus: This is not really a realistic example. You also don't show the actual usage of `Baz`, only the partial version of it. The point is to show how a type can be useful as both in its original form and as a `PartialDeep` form, in a realistic scenario.

readme.md
```diff
@@ -69,6 +69,7 @@ Click the type names for complete docs.
 - [`Merge`](source/merge.d.ts) - Merge two types into a new type. Keys of the second type overrides keys of the first type.
 - [`MergeExclusive`](source/merge-exclusive.d.ts) - Create a type that has mutually exclusive properties.
 - [`RequireAtLeastOne`](source/require-at-least-one.d.ts) - Create a type that requires at least one of the given properties.
+- [`PartialDeep`](source/partial-deep.d.ts) - Create a type from another type with all properties set to optional and the nested properties of all properties set to optional, as opposed to the [`Partial<T>`](https://github.com/Microsoft/TypeScript/blob/2961bc3fc0ea1117d4e53bc8e97fa76119bc33e3/src/lib/es5.d.ts#L1401-L1406) type which constructs a type with properties of T that have a depth of 1 set to optional.
```
sindresorhus: Would be good to write this more succinctly.

source/partial-deep.d.ts
kainiedziela: Good point, handled in commit 92f9595

test-d/partial-deep.ts
kainiedziela: Fixed in c48b576

source/partial-deep.d.ts
kainiedziela: Fixed in c48b576

source/partial-deep.d.ts
kainiedziela: I provided a more realistic usage of the type where both a type and it's PartialDeep version are used in c48b576. I think I could use some further feedback here.

readme.md
kainiedziela: Fixed in c48b576

readme.md
```diff
@@ -69,6 +69,7 @@ Click the type names for complete docs.
 - [`Merge`](source/merge.d.ts) - Merge two types into a new type. Keys of the second type overrides keys of the first type.
 - [`MergeExclusive`](source/merge-exclusive.d.ts) - Create a type that has mutually exclusive properties.
 - [`RequireAtLeastOne`](source/require-at-least-one.d.ts) - Create a type that requires at least one of the given properties.
+- [`PartialDeep`](source/partial-deep.d.ts) - Create a deeply optional version of another type, as opposed to the [`Partial<T>`](https://github.com/Microsoft/TypeScript/blob/2961bc3fc0ea1117d4e53bc8e97fa76119bc33e3/src/lib/es5.d.ts#L1401-L1406) type which creates a type with properties of T that have a depth of 1 set to optional.
```
BendingBender:
```suggestion
- [`PartialDeep`](source/partial-deep.d.ts) - Create a deeply optional version of another type, as opposed to the [`Partial<T>`](https://github.com/Microsoft/TypeScript/blob/2961bc3fc0ea1117d4e53bc8e97fa76119bc33e3/src/lib/es5.d.ts#L1401-L1406) type which creates a type with all properties of `T` set to optional, but only one level deep.
```

source/partial-deep.d.ts
```diff
@@ -0,0 +1,50 @@
+/**
+Create a type from another type with all properties set to optional and the nested properties of all properties set to optional.
```
BendingBender:
```suggestion
Create a type from another type with all properties and nested properties set to optional.
```

source/partial-deep.d.ts
```diff
@@ -0,0 +1,50 @@
...
+
+Use cases:
+- Merging a default settings/config object with another object, the second object would be a deep partial of the default object.
+- Mocking and testing complex entities, where populating an entire object with it's properties would be redundant in terms of the mock or test.
```
BendingBender:
```suggestion
- Mocking and testing complex entities, where populating an entire object with its properties would be redundant in terms of the mock or test.
```

source/partial-deep.d.ts
BendingBender: Why didn't you just do the same as `ReadonlyDeep`? Why don't you handle primitive types and `Function`s, just like `ReadonlyDeep` (and all the other cases)?

We had a few iterations there, it's quite a solid definition.

test-d/partial-deep.ts
```diff
@@ -0,0 +1,32 @@
...
+		number: 1,
+		boolean: false,
+		symbol: Symbol('test'),
+		null: null,
+		undefined: undefined, // eslint-disable-line object-shorthand
+		map: new Map<string, string>(),
+		set: new Set<string>(),
+		array: ['foo']
+	}
+};
```
BendingBender: Please extend the tests here analogous to the tests for `ReadonlyDeep`.

source/partial-deep.d.ts
kainiedziela: It does handle primitive types as showcased by tests. Regarding functions do you mean that it should convert

```TS
foo: { bar: { baz: function fn(_: string): boolean {} } }
```

to 

```TS
foo: { bar?: { baz?: function fn(_: string): boolean {} }
```

because it already does that, or do you mean that it should convert to

```TS
foo: { bar?: { baz?: function fn(_: string | undefined): boolean | null {} }
```

and if so then this type would indeed need more work. Probably would be wise to then also include Promisables if ReturnTypes are being taken into account.

BendingBender: > It does handle primitive types as showcased by tests.

Yeah, I know it works with primitives the way it's defined right now, it's just a performance tweak because primitives are so common.

> to
> ```
> foo: { bar?: { baz?: function fn(_: string): boolean {} }
> ```
> because it already does that

Are you sure about this? There is no special handling for `Function`s. There is also no test for `Function`s. And it doesn't really work with `Function`s at all: https://www.typescriptlang.org/play/#code/KYDwDg9gTgLgBDAnmYcAKBDWBLDAbAEWGDAB4AVAPjgF44BvAKAEgBtAaWEXOVWwDs4Aay4QAZnHIBdAPwAuRnCXKl5Dlx4opcUDGD8AJgGc4AWQxkBY4FDMXO3XgBo4Vm3bAA1fAFdgm4EoWZhl0LBhcPHMyaIcAl2jvPD8AoOZmOUl1Ry0dED1DEwBlYBhSN1sSmFTg0MwcfCrSKpr0zLU43m1dfWM4AEEoKAxEcv5rW0HhnMDagaGR0nqI-CISUimR1LSMrM7cnsK4ACVgDAMIfjxETdGKk7OLq5uFmZ3Q0-PL69ul8Mi1mRPk8fq9tsFMssAcQyB0NF1KABuRgAX2RAj0UDEGAAxqgofhoqR9sAXEkUrxqIc+kSCYQYcT4ShKC46YDSOT-JTqPQUYwMTZsXiwg08E0qHkCn0mmyGVQeXzGIwAG5YOAYWgMMQQCBwTIACgAlLQFWilarbAAjPUilb09ZIFDidXUOgajAmWUO3jOjBIxiWgCEADptRAjcig6GdSGcfg8Pr+D48HhDYigA

BendingBender: > because it already does that, or do you mean that it should convert to
> ```
> foo: { bar?: { baz?: function fn(_: string | undefined): boolean | null {}
> ```

No, I wouldn't do this.

kainiedziela: Hey, indeed it handled functions wrong and as I was adding more and more cases I realized that I should just move the definition to be in line with `ReadonlyDeep` as you suggested and that caused all the tests to pass, except now the negative test that I commented out doesn't raise an error.

Any ideas why that may be? The type is different from simply `Partial<T>`, but the type checker doesn't seem to think so.

test-d/partial-deep.ts
kainiedziela: Did that, but please take a look at [my comment](https://github.com/sindresorhus/type-fest/pull/47#discussion_r312698373) at the other conversation.

source/partial-deep.d.ts
kainiedziela: Okay, @BendingBender I seem to have fixed some issues that arose during rewriting the type to be more in line with `ReadonlyDeep`. Now all tests pass and the test suite covers all the same types that `ReadonlyDeep`s does.

I'd welcome some more feedback.

kainiedziela:
@sindresorhus @BendingBender this requires another look

PR #48, opened 2019-07-11
Add `SetRequired` and `SetOptional` types

crhistianramirez:
~Now that I'm looking at it maybe SomeRequired and SomeOptional would be better names? Couples the two, as it should and expresses intent pretty well.~

Edit: changed to SetRequired and SetOptional per feedback from sindresorhus

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -12,6 +12,7 @@ export {ReadonlyDeep} from './source/readonly-deep';
 export {LiteralUnion} from './source/literal-union';
 export {Optionalize} from './source/optionalize';
 export {Promisable} from './source/promisable';
+export {Opaque} from './source/opaque';
 
 // Miscellaneous
 export {PackageJson} from './source/package-json';
--- a/index.d.ts
+++ b/index.d.ts
@@ -10,6 +10,7 @@ export {Require} from './source/require';
 export {RequireAtLeastOne} from './source/require-at-least-one';
 export {ReadonlyDeep} from './source/readonly-deep';
 export {LiteralUnion} from './source/literal-union';
+export {Optionalize} from './source/optionalize';
 export {Promisable} from './source/promisable';
 
 // Miscellaneous
--- a/index.d.ts
+++ b/index.d.ts
@@ -6,6 +6,7 @@ export {Except} from './source/except';
 export {Mutable} from './source/mutable';
 export {Merge} from './source/merge';
 export {MergeExclusive} from './source/merge-exclusive';
+export {Require} from './source/require';
 export {RequireAtLeastOne} from './source/require-at-least-one';
 export {ReadonlyDeep} from './source/readonly-deep';
 export {LiteralUnion} from './source/literal-union';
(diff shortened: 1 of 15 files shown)


crhistianramirez:
Now that I'm looking at it maybe `SomeRequired` and `SomeOptional` would be a better name?

sindresorhus:
> maybe SomeRequired and SomeOptional would be better names?

👍 

It's really too bad that the built-in `Required` type is not named `RequiredAll` or `AllRequired`...

source/optionalize.d.ts
```diff
@@ -0,0 +1,30 @@
+/**
+The sister of `Require` type. Allows you to define one or more keys for a type that should be optional.
```
sindresorhus: The sentences here should be in reverse order.

source/optionalize.d.ts
```diff
@@ -0,0 +1,30 @@
...
+
+Use-case: You want to define a single model where the only thing that changes is
+whether or not some of the properties are optional
```
sindresorhus: Don't hard-wrap

source/optionalize.d.ts
```diff
@@ -0,0 +1,30 @@
...
+  a: number;
+  b?: string;
+  c: boolean;
+}
+
+type SomeOptional = Optionalize<Foo, 'b' | 'c'>;
+// type SomeOptional = {
+//   a: number;
+//   b?: string; // was already optional, still is
+//   c?: boolean; // is now optional
```
sindresorhus: Use tab-indentation

source/optionalize.d.ts
```diff
@@ -0,0 +1,30 @@
...
+// }
+```
+*/
+
+export type Optionalize<BaseType, Keys extends keyof BaseType = keyof BaseType> =
+Pick<BaseType, Exclude<keyof BaseType, Keys>> &
+  Partial<Pick<BaseType, Keys>> extends infer InferredType
+  ? { [Property in keyof InferredType]: InferredType[Property] }
+  : never;
```
sindresorhus: > At the bottom, explain how the type works. Some types can be quite advanced and hard to understand. We can use this opportunity to teach users. - https://github.com/sindresorhus/type-fest/blob/master/.github/contributing.md

sindresorhus:
You need to add the types to the readme.

index.d.ts
```diff
@@ -6,9 +6,11 @@ export {Except} from './source/except';
 export {Mutable} from './source/mutable';
 export {Merge} from './source/merge';
 export {MergeExclusive} from './source/merge-exclusive';
+export {Require} from './source/require';
```
sindresorhus: This should be moved to line 12.

source/optionalize.d.ts
sindresorhus: > Allows you to define one or more keys for a type that should be optional.

This is not entirely correct as it creates a new type. It doesn't modify the existing type.

crhistianramirez:
@sindresorhus thank you for the feedback I went ahead and resolved each line item. I'd appreciate additional feeedback you might have :)

Also did you prefer that I squash all the commits to clean it up? Wasn't sure, the readme didn't mention it.

sindresorhus:
Looks good now. Thanks :)

PR #55, opened 2019-07-28
Add `RequireExactlyOne` type

### LINKED ISSUE #54: `RequireOnlyOne` type
See https://github.com/sindresorhus/type-fest/pull/24 and its feedback if you want to tackle this.

kainiedziela:
Fixes #54 

Create a type that requires exactly one of the given properties and disallows more. The remaining properties are kept as is.

**Use cases**
- Creating interfaces for components that only need one of the properties to display properly.
- Declaring generic properties in a single place for a single use case that get narrowed down via `RequireExactlyOne`.

The caveat with `RequireExactlyOne` is that TypeScript doesn't always know at compile time every property that will exist at runtime. Therefore `RequireExactlyOne` can't do anything to prevent extra properties it doesn't know about.

**Example**

```TS
import {RequireExactlyOne} from 'type-fest';

type Responder = {
	text: () => string;
	json: () => string;
	secure: boolean;
};

const responder: RequireExactlyOne<Responder, 'text' | 'json'> = {
	json: () => '{"message": "ok"}',
	secure: true
};
```


test-d/require-one.ts
```diff
@@ -0,0 +1,24 @@
+import {expectType, expectError} from 'tsd';
+import {RequireOne} from '..';
+
+type SystemMessages = {
+    default: string;
```
SamVerschueren: tab

kainiedziela: Changes spaces to tabs

test-d/require-one.ts
```diff
@@ -0,0 +1,24 @@
...
+};
+
+type ValidMessages = RequireOne<SystemMessages, 'macos' | 'linux'>;
+const test = (_: ValidMessages): void => {};
+
+test({macos: 'hey', default: 'hello'});
+test({linux: 'sup', optional: 'howdy', default: 'hello'});
+
+expectError(test({}));
+expectError(test({macos: 'hey', linux: 'sup', default: 'hello'}));
```
BendingBender: This is quite unexpected. So if I understand it correctly, the properties `macos` and `linux` are mutually exclusive in this type? Is this intended? Intuitively, I would expect that one of the both is required but that it's OK if I'd provide both.

Any ideas?

source/require-one.d.ts
```diff
@@ -0,0 +1,30 @@
+/**
+Create a type that requires only one of the given properties, but not more. The remaining properties are kept as is.
```
BendingBender:
```suggestion
Create a type that requires exactly one of the given properties, but not more. The remaining properties are kept as is.
```

test-d/require-one.ts
kainiedziela: Yup, it is intended, this is how it differs from `RequireAtLeastOne`. Still not sure about the naming, it was somewhat discussed in the previous stale pull request, but I don't think any consensus was found.

BendingBender: What do you think, @sindresorhus?

sindresorhus: I agree, `RequireOne` is not clear enough about this. `RequireExactlyOne` sounds good.

readme.md
```diff
@@ -69,6 +69,7 @@ Click the type names for complete docs.
 - [`Merge`](source/merge.d.ts) - Merge two types into a new type. Keys of the second type overrides keys of the first type.
 - [`MergeExclusive`](source/merge-exclusive.d.ts) - Create a type that has mutually exclusive properties.
 - [`RequireAtLeastOne`](source/require-at-least-one.d.ts) - Create a type that requires at least one of the given properties.
+- [`RequireOne`](source/require-one.d.ts) - Create a type that requires a single property of the given properties, but not more.
```
sindresorhus: The description should be clearer about it only allowing one of the properties. Regarding, https://github.com/sindresorhus/type-fest/pull/55/files#r310352159

source/require-one.d.ts
```diff
@@ -0,0 +1,30 @@
+/**
+Create a type that requires exactly one of the given properties, but not more. The remaining properties are kept as is.
+
+Use cases:
+- Creating interfaces for components that only need one of the properties to display properly.
+- Declaring generic properties in a single place for a single use case that get narrowed down via `RequireOne`.
+
+The caveat with `RequireOne` is that TypeScript doesn't always know at compile time every property that will exist at runtime. Therefore `RequireOne` can't do anything to prevent extra properties it doesn't know about.
+
```
sindresorhus: Has there been any issues on the TypeScript issue tracker about a type like this? If so, would be good to include references.

kainiedziela: I agree, couldn't find the myself though.

BendingBender:
LGTM

kainiedziela:
@sindresorhus @BendingBender guys?

SamVerschueren:
Nice! Thanks for working on this!

PR #60, opened 2019-09-21
Allow objects with non-existing keys in PartialDeep

merlinnot:
This fixes the following issue (does not compile with latest TypeScript):

```ts
interface Interface {
  a: 'a',
}

type PartialInterface = PartialDeep<Interface>;

const value: PartialInterface = {};
```


sindresorhus:
@kainiedziela Can you review?

kainiedziela:
I'd say this warrants an extra test case to cover empty objects then.

test-d/partial-deep.ts
```diff
@@ -39,8 +39,8 @@ expectType<undefined>(partialDeepFoo.bar!.undefined);
 expectType<Map<string | undefined, string | undefined> | undefined>(partialDeepFoo.bar!.map);
 expectType<Set<string | undefined> | undefined>(partialDeepFoo.bar!.set);
 expectType<Array<string | undefined> | undefined>(partialDeepFoo.bar!.array);
-expectType<['foo' | undefined] | undefined>(partialDeepFoo.bar!.tuple);
+expectType<[('foo')?] | undefined>(partialDeepFoo.bar!.tuple);
```
kainiedziela: Any particular reason for parentheses here? You don't add them in the `readonly tuple` test (line 46).

merlinnot:
I removed extra parentheses, I'm used to Prettier doing it for me, had to disable it for this repo and didn't notice it.

As for the test case, maybe my example was a little too minimal. Here's another one:

```ts
interface Interface {
  a: 'a',
  b: 'b',
}

type PartialInterface = PartialDeep<Interface>;

// Does not compile.
const value: PartialInterface = { a: 'a' };
```

merlinnot:
I'm not really used to test tools that you guys use, I checked the repositories and it seems it's deprecated, so I wasn't exactly sure what's your approach to this.

I'd appreciate any guidance on what tests would you expect, I'll try to add these.

sindresorhus:
We use https://github.com/SamVerschueren/tsd. It's not deprecated.

merlinnot:
Oh, I thought that's the one: https://github.com/DefinitelyTyped/tsd. First result in Google :)

Thanks for the link, I'll try to add some tests tomorrow.

kainiedziela:
I'd say you can just follow the pattern already established in the present type-fest test cases. The library used is pretty intuitive.

merlinnot:
@kainiedziela I sent you an invite. I was playing with it a little but the only test case I could come up with is to basically paste the code above to the test file, which I guess is not what's expected.

I also changed the existing tests to make them pass, so it's already a regression test of sorts.

kainiedziela:
No, I think that would be quite intuitive. I pushed the test cases that should cover our needs. I also removed the version change as I believe that should be handled by package maintainers.

@sindresorhus this looks good to go.

sindresorhus:
Thanks for fixing :)

Issue #64, opened 2019-10-08
JsonObject doesn't allow optionals


SamVerschueren:
I want to type something as `JsonObject`, because the interface defines a request from an API.

So then I use `extends JsonObject`, but my API has some optionals.

```ts
interface Unicorn extends JsonObject {
	unicorn: string;
	rainbow?: string;
}
```

The compiler will fail with 

> Property 'unicorn' of type 'string | undefined' is not assignable to string index type 'JsonValue'

The reasoning behind this is that `undefined` is not serializable and doesn't exist in JSON. But how would someone solve this issue? Or should we change something in the JSON definitions of `type-fest`?

A possible solution would be to allow `undefined` in `JsonObject`.

```ts
export type JsonObject = {[key: string]: JsonValue | undefined};
```

This allows me to add optionals to something which is a `JsonObject`.

```ts
interface Unicorn extends JsonObject {
	unicorn: string;
	rainbow?: string;
}
```

To me, this kinda makes sense because `JSON.stringify({unicorn: '🦄', rainbow: undefined})` results in `{"unicorn":"🦄"}`, thus allowing `undefined` in JSON like objects.

SamVerschueren:
We could also add `undefined` to `JsonValue`. But wasn't sure about that approach.

yxliang01:
@SamVerschueren I think it's weird to add `undefined` to `JsonValue` since in JSON, we only have `null` but no `undefined`. From my understanding, `JsonObject` is good for casting `JSON.parse` return value but not input to `JSON.stringify` because a non-JSON directly translatable object can be stringified well. E.g. a custom object implementing their own `toJSON()` method.

SamVerschueren:
Adding `undefined` to `JsonValue` isn't my preferred approach either. But currently, the use case explained in the docs of `JsonObject` itself, isn't very usuable. The `extends JsonObject` doesn't allow optional properties which enforces you to do something like `extends Partial<JsonObject>` which is meh... That's why I think adding `| undefined` to the `JsonObject` definition could make sense.

yxliang01:
Then, how about simply make all the keys optional? @SamVerschueren it's pretty clear in this way I think.

SamVerschueren:
By doing `export type JsonObject = {[key: string]: JsonValue | undefined};` the keys become optional. You can't do `[key: string]?: JsonValue` if I'm not mistaken.

yxliang01:
[...]

PR #65, opened 2019-10-08
Allow all keys to be optional in the `JsonObject` type

yxliang01:
By right, when an object has certain keys optional, the object is still a valid `JsonObject`. So, it makes more sense to allow all keys to be optional. Indeed, an `JsonObject` doesn't always have values for all string indexes.

This resolves https://github.com/sindresorhus/type-fest/issues/64 .

### DIFF
--- a/source/basic.d.ts
+++ b/source/basic.d.ts
@@ -40,7 +40,7 @@ Matches a JSON object.
 
 This type can be useful to enforce some input to be JSON-compatible or as a super-type to be extended from. Don't use this as a direct return type as the user would have to double-cast it: `jsonObject as unknown as CustomResponse`. Instead, you could extend your CustomResponse type from it to ensure your type only uses JSON-compatible types: `interface CustomResponse extends JsonObject { … }`.
 */
-export type JsonObject = {[key: string]: JsonValue};
+export type JsonObject = {[key: string]?: JsonValue};
 
 /**
 Matches a JSON array.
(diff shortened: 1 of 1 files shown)


sindresorhus:
Can you add some type tests?

yxliang01:
@sindresorhus Yup I can. But, before that, I need to know whether these changes look good :) Otherwise, the tests will be useless :P

sindresorhus:
@SamVerschueren Is this what you had in mind?

SamVerschueren:
Just wondering, but `key in string` doesn't look like the same as a key which is of type string.

sindresorhus:
@yxliang01 Why did you change it to `Key in string`?

yxliang01:
@sindresorhus It was simply that `[k: string]?: value` is not valid for its syntax. This can actually be written as `[k: string]: value | undefined`, just that I am not aware of semantic difference between `[k: string]: value | undefined` and `[k in string]?: value` for the case of accepting any key of type `string`.

sindresorhus:
I'm not aware of the difference either. One problem with either of these solutions is that it will allow:

```ts
interface Unicorn extends JsonObject {
	rainbow: undefined;
}
```

yxliang01:
So, the intended behavior is that all keys can be optional while it is a must that we can't make the value `undefined` assignable to all defined keys. Although in the underlying JavaScript compiled code, these two has only very subtle semantic difference, the current unintended semantic will enlarge the room for errors.

The technical difficulty to achieve the intended semantic is that the TypeScript type system doesn't differentiate the following cases:
1. Key `k` is defined with value `undefined`
2. Key `k` is undefined
Thus, it might be impossible to achieve our goal with current TypeScript type system?

sindresorhus:
@resynth1943 Any thoughts on this?

resynth1943:
I'll take a look and see what I think.

papb:
Related: https://github.com/Microsoft/TypeScript/issues/13195

resynth1943:
Apologies for the late reply.

I don't really see any major problems with this regarding serialized and de-serialized JSON types, as keys could be optional in that context. LGTM.

resynth1943:
@yxliang01 Could you please follow the instructions in this document?

yxliang01:
Finally merged after created for 4 months :tada:

resynth1943:
Yay!

PR #71, opened 2020-01-19
Make the `Opaque` type stricter

### LINKED ISSUE #70: Opaque types are interchangeable
Do you have any examples for `Opaque`? I thought `Opaque` types weren’t supposed to be interchangeable.

```typescript
import { Opaque } from 'type-fest';

type AccountNumber = Opaque<number>;
type AccountBalance = Opaque<number>;

function getBalance(accountNumber: AccountNumber): AccountBalance {
  return 101 as AccountBalance;
}

// pass an AccountBalance instead of an AccountNumber
const ab = 42 as AccountBalance;
console.log(getBalance(ab)); 
// expected: should not compile
// actual: gets balance for account 42
```

resynth1943:
This commit adds a Token parameter to the Opaque types, allowing them to be
fundamentally unique. This means an oaque type cannot be cast to a separate
opaque type. The previous (flawed) design allowed this. This does not.

Closes #70.


resynth1943:
Should I `eslint-ignore` that error? Also, I'm seeing linting errors for other tests. Not sure what's going on there.

source/opaque.d.ts
```diff
@@ -13,9 +13,28 @@ There have been several discussions about adding this feature to TypeScript via
 ```
 import {Opaque} from 'type-fest';
 
-type AccountNumber = Opaque<number>;
-type AccountBalance = Opaque<number>;
+type AccountNumber = Opaque<number, 'AccountNumber'>;
+type AccountBalance = Opaque<number, 'AccountBalance'>;
+
+// The Token parameter allows the compiler to differentiate between types, whereas "unknown" will not. For example, consider the following structures:
+type ThingOnw = Opaque<string>;
```
papb: typo `ThingOnw` -> `ThingOne`

source/opaque.d.ts
```diff
@@ -13,9 +13,28 @@ There have been several discussions about adding this feature to TypeScript via
...
+type ThingTwo = Opaque<string>;
+
+// To the compiler, these types are allowed to be cast to each other as they have the same underlying type. They are both `string & { __opaque__: unknown }`.
+// To avoid this behaviour, you would instead pass the "Token" parameter, like so.
+type NewThingOnw = Opaque<string, 'ThingOne'>;
```
papb: typo `NewThingOnw` -> `NewThingOne`

resynth1943: Goddamnit, I'm making so many spelling errors tonight across the board. :stuck_out_tongue:

resynth1943:
The thing is, I believe my test resembles the first post of the issue well. It shows opaque types are not interchangeable, which is the desired effect.

sindresorhus:
Because TypeScript supports recursive types, the `Token` could be the parent type itself. This could create even stronger unique types. Would be nice with an example in the docs and a test for this. For example:

```ts
type Person = {
	id: Opaque<number, Person>;
	name: string;
};
```

resynth1943:
@sindresorhus Yes! 😛 I'll add something like that into the TSDoc block.

Thanks.

resynth1943:
Damn. We gotta get this merged.

resynth1943:
Also, I'd like to give apologies for my absence. I've currently been doing my mock examinations in school. Anyway, that's over now, and I'm back. :stuck_out_tongue:

sindresorhus:
Thanks :)

PR #73, opened 2020-01-24
Add `TsConfigJson` type

ifiokjr:
## Description

This type provides the interface for a `tsconfig.json` file. It is compatible with the latest version of TypeScript at the moment (3.7).

I was surprised to find that this isn't exported by default from TypeScript. I'm already using it for a project I'm working on where I need to manually process `tsconfig` files.

Hopefully it's of use to anyone who wants to avoid rolling their own version.


sindresorhus:
Great!

PR #42, opened 2019-06-27
Add `Unpromise` type

streamich:
Adds `Unpromise` type.

```ts
type Str = Unpromise<Promise<string>>;
const str: Str = 'string...';
```

PR #75, opened 2020-01-31
Add `PromiseValue` type

resynth1943:
Boom.

Closes #42.

### DIFF
--- a/source/promise-value.d.ts
+++ b/source/promise-value.d.ts
@@ -1,0 +1,24 @@
+/**
+ * Returns the type that is wrapped inside a `Promise` type.
+ * If the type is not a `Promise`, the type itself is returned.
+ *
+ * @example
+ * import { asyncFunction } from 'api';
+ * import { PromiseValue } from 'type-fest';
+ *
+ * let data: PromiseValue<ReturnType<typeof asyncFunction>>;
+ *
+ * async function setValue () {
+ *   data = await asyncFunction();
+ * }
+ *
+ * setValue();
+ *
+ * // Here's an example of using this with non-Promise types.
+ * function getNumber () {
+ *  return 2;
+ * }
+ *
+ * let number: PromiseValue<ReturnType<typeof getNumber>> = getNumber();
+ */
+export type PromiseValue<PromiseType, Otherwise = PromiseType> = PromiseType extends Promise<infer Value> ? Value : Otherwise;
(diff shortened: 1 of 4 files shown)


source/promise-value.d.ts
```diff
@@ -0,0 +1,24 @@
...
+ * async function setValue () {
+ *   data = await asyncFunction();
+ * }
+ *
+ * setValue();
+ *
+ * // Here's an example of using this with non-Promise types.
+ * function getNumber () {
+ *  return 2;
+ * }
```
fregante: I’m not sure if other types follow this, but, could you lint this piece of code with XO? The style doesn’t pass it

resynth1943:
@fregante How about now?

fregante:
Looks better. How about we drop the asyncFunction mention though? This would let the user focus on this type and nothing else.

resynth1943:
Good idea. One sec.

fregante:
I mean the whole thing, the example should not include `ReturnType<typeof asyncFunction>`. As discussed earlier, it seems that every example of PromiseValue is exactly that `PromiseValue<ReturnType<typeof asyncFunction>>`, suggesting that this type is useless on its own (as opposed to just having `AsyncReturnType<typeof asyncFunction>>`). Is it the case?

resynth1943:
Thanks for your comment, @fregante. I'll say thanks for your contribution, however I'd still like to see this Promise-unpacking type implemented as a separate module to the variant that unpacks the return value from functions. Moreover, I still believe this type has uses on its own.

fregante:
> Moreover, I still believe this type has uses on its own.

Right, I’m not disagreeing, I’m just asking to show a real use in the examples as well.

source/promise-value.d.ts
```diff
@@ -4,21 +4,15 @@ If the type is not a `Promise`, the type itself is returned.
...
-
-// Here's an example of using this with non-Promise types.
-function getNumber(): number {
- return 2;
-}
-
-const number: PromiseValue<ReturnType<typeof getNumber>> = getNumber();
+// Here's an example that shows how this type reacts to non-Promise types.
+type SyncData = PromiseValue<string>;
+let syncData: SyncData = getSyncData();
```
resynth1943: As you can see, this example does not mention unpacking the return type of a function.

fregante: Yeah that works, thanks for making the change.

sindresorhus:
Looks good 👍

resynth1943:
Thanking you. How about we go for `AsyncReturnType` on top of this? ♥️

sindresorhus:
👍

resynth1943:
Let's do it.

PR #77, opened 2020-02-04
Add `workspaces` to `PackageJson` type

ifiokjr:
### Description

Yarn workspaces are becoming a more prominent feature of the `package.json` file.

### DIFF
--- a/source/package-json.d.ts
+++ b/source/package-json.d.ts
@@ -265,37 +265,40 @@ declare namespace PackageJson {
 		An array of workspace pattern strings which contain the workspace packages.
 		*/
 		packages?: WorkspacePattern[];
-		/**
-		`nohoist` is designed to solve the problem of packages which break when their node_modules are moved to the root workspace directory - a process known as hoisting. For these packages both within your workspace and also some that have been installed via node_modules it is important to have a mechanism for preventing the default Yarn workspace behavior. By adding workspace pattern strings here, Yarn will resume non-workspace behavior for any package which matches the patterns defined.
 
-		More information can be found via this blog posthttps://classic.yarnpkg.com/blog/2018/02/15/nohoist/
+		/**
+		Designed to solve the problem of packages which break when their `node_modules` are moved to the root workspace directory - a process known as hoisting. For these packages, both within your workspace, and also some that have been installed via `node_modules`, it is important to have a mechanism for preventing the default Yarn workspace behavior. By adding workspace pattern strings here, Yarn will resume non-workspace behavior for any package which matches the defined patterns.
+
+		[Read more](https://classic.yarnpkg.com/blog/2018/02/15/nohoist/)
 		*/
 		nohoist?: WorkspacePattern[];
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


source/package-json.d.ts
```diff
@@ -257,8 +257,26 @@ declare namespace PackageJson {
...
+		packages: string[];
+		/**
+		A list of glob patterns for packages that should not be hoisted.
+		*/
+		nohoist: string[];
+	}
+
 	export interface YarnConfiguration {
 		/**
+    Used to configure yarn (and other) workspaces.
```
sindresorhus: Use tab-indentation.

Would also be useful to link to the Yarn docs about this property.

source/package-json.d.ts
```diff
@@ -257,8 +257,26 @@ declare namespace PackageJson {
...
+
+    Please note private should be set to true to use workspaces
+    */
+    workspaces?: string[] | Workspaces;
```
sindresorhus: What is `string[]`? I think the docs could be clearer on what it is.

source/package-json.d.ts
```diff
@@ -257,8 +257,26 @@ declare namespace PackageJson {
...
+		A list of glob patterns for packages that should not be hoisted.
+		*/
+		nohoist: string[];
+	}
+
 	export interface YarnConfiguration {
 		/**
+    Used to configure yarn (and other) workspaces.
+
+    Please note private should be set to true to use workspaces
```
sindresorhus:
```suggestion
    Please note private should be set to true to use workspaces.
```

And what is "private". Would be good to make it clear you're talking about the top-level private property.

source/package-json.d.ts
sindresorhus:
```suggestion
    Used to configure Yarn (and other) workspaces.
```

What "other"?

source/package-json.d.ts
```diff
@@ -257,8 +257,26 @@ declare namespace PackageJson {
 		typings?: string;
 	}
 
+	export interface Workspaces {
+		/**
+		The packages blob to use for workspaces.
```
sindresorhus: Not clear to me what "packages blob" is.

source/package-json.d.ts
```diff
@@ -257,8 +257,26 @@ declare namespace PackageJson {
...
+		*/
+		packages: string[];
+		/**
+		A list of glob patterns for packages that should not be hoisted.
```
sindresorhus: I would elaborate on what "hoisting" mean here.

sindresorhus:
It's a welcome addition, but I think the doc comments could be more verbose.

ifiokjr:
I'll address the requested changes soon.

source/package-json.d.ts
ifiokjr: Done, plus I've added a link to the Yarn website.

source/package-json.d.ts
ifiokjr: I've removed the other.

ifiokjr:
I've pushed the requested changes.

sindresorhus:
👌

PR #80, opened 2020-02-10
fix: remove unknown from `PackageJson`

### LINKED ISSUE #79: `PackageJson` is not a valid `JsonObject`
## Description

The following fails type checking. 

```ts
import {PackageJson, JsonObject} from 'type-fest';

const packageJson: PackageJson = {};

const createPackageJson = () => packageJson;
expectType<JsonObject>(createPackageJson()); // => Error
```

The error produced is the following.

```
Argument of type 'PackageJson' is not assignable to parameter of type 'JsonObject'.
  Index signatures are incompatible.
    Type 'unknown' is not assignable to type 'JsonValue'.
      Type 'unknown' is not assignable to type 'JsonArray'.ts(2345)
```

To fix this we would need to remove `unknown` from the `PackageJson` type and replace it with `JsonValue` since the package.json file only supports valid Json this shouldn't cause any issues.

I'll create a PR shortly.

ifiokjr:
Fixes #79 

Adds a test from the failing scenario.

### DIFF
--- a/source/package-json.d.ts
+++ b/source/package-json.d.ts
@@ -1,4 +1,5 @@
-import {LiteralUnion} from '..';
+import {JsonObject, JsonValue} from './basic';
+import {LiteralUnion} from './literal-union';
 
 declare namespace PackageJson {
 	/**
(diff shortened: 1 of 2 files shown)


ifiokjr:
This was also a tired PR and can be ignored until fixed.

PR #82, opened 2020-02-10
fix: make `TsConfigJson` a valid `JsonObject`

ifiokjr:
Similar to #80 for turning the `TsConfigJson` type into a valid `JsonObject`.


ifiokjr:
This was a tired PR and can be ignored until fixed.

PR #86, opened 2020-02-11
Add `Optional` type

### LINKED ISSUE #31: Add Maybe for generic or undefined
An optional type is really missing in the context of return types,
When a function might a value or not you find yourself piping undefined

```ts
type Maybe<T> = T | undefined;
```

Can be really useful

resynth1943:
Closes #31.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -74,6 +74,7 @@ Click the type names for complete docs.
 - [`Opaque`](source/opaque.d.ts) - Create an [opaque type](https://codemix.com/opaque-types-in-javascript/).
 - [`SetOptional`](source/set-optional.d.ts) - Create a type that makes the given keys optional.
 - [`SetRequired`](source/set-required.d.ts) - Create a type that makes the given keys required.
+- [`PromiseValue`](source/promise-value.d.ts) - Returns the type that is wrapped inside a `Promise` type.
 
 ### Miscellaneous
 
--- a/readme.md
+++ b/readme.md
@@ -63,7 +63,7 @@ Click the type names for complete docs.
 
 - [`Except`](source/except.d.ts) - Create a type from an object type without certain keys. This is a stricter version of [`Omit`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-5.html#the-omit-helper-type).
 - [`Mutable`](source/mutable.d.ts) - Convert an object with `readonly` keys into a mutable object. The inverse of `Readonly<T>`.
-- [`Optional`](source/optional.d.ts) - Makes a type optional. Allows a type to be `undefined`, but not `null`.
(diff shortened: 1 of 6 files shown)


sindresorhus:
Same feedback as your other PR => https://github.com/sindresorhus/type-fest/blob/master/.github/contributing.md

readme.md
```diff
@@ -75,6 +75,7 @@ Click the type names for complete docs.
 - [`SetOptional`](source/set-optional.d.ts) - Create a type that makes the given keys optional.
 - [`SetRequired`](source/set-required.d.ts) - Create a type that makes the given keys required.
 - [`PromiseValue`](source/promise-value.d.ts) - Returns the type that is wrapped inside a `Promise` type.
+- [`Optional`](source/optional.d.ts) - Makes a type optional. Allows a type to be `undefined`, but not `null`.
```
```suggestion
- [`Optional`](source/optional.d.ts) - Makes a type optional, allowing it to be `undefined`, but not `null`.
```

?

sindresorhus: And let's move it below `Mutable`. It's a very useful type.

test-d/optional.ts
```diff
@@ -0,0 +1,5 @@
+import {expectType} from 'tsd';
+import {Optional} from '..';
+
+expectType<Optional<number>>(2);
+expectType<Optional<number>>(undefined);
```
sindresorhus: Some more test would be good.

sindresorhus:
> If there has been any discussion somewhere about this type, include a link to it. For example, a discussion on the TypeScript issue tracker. - https://github.com/sindresorhus/type-fest/blob/master/.github/contributing.md

I know for sure this has been discussed in the TypeScript issue tracker before.

source/optional.d.ts
```diff
@@ -0,0 +1,14 @@
...
+Makes a type optional. Allows a type to be `undefined`, but not `null`.
+
+@example
+```
+import {Optional} from 'type-fest';
+
+type Value = number;
+type OptionalValue = Optional<Value>;
+
+const values: [Value, OptionalValue] = [2, undefined];
```
sindresorhus: > Example code block with a realistic example. - https://github.com/sindresorhus/type-fest/blob/master/.github/contributing.md

This example is not really realistic. A more realistic example would be to use `OptionalValue` as a return value or parameter for a function.

source/optional.d.ts
```diff
@@ -0,0 +1,14 @@
+/**
+Makes a type optional. Allows a type to be `undefined`, but not `null`.
+
```
sindresorhus: > Write about some real-world use-cases where it can be useful. (It can be hard sometimes for users to see where they would use something) - https://github.com/sindresorhus/type-fest/blob/master/.github/contributing.md

PR #89, opened 2020-02-23
Add `UnionToIntersection` type

ifiokjr:
## Description

Creates an intersection type from a union type. This has been useful for me in [library](https://github.com/remirror/remirror/blob/b096ed1dd300d682e28f9a7721ad9debbe1c1eb3/%40remirror/core/src/extension-types.ts#L181-L186) code.

### DIFF
--- a/source/union-to-intersection.d.ts
+++ b/source/union-to-intersection.d.ts
@@ -1,23 +1,22 @@
 /**
-A helpful utility which maps a union type to an intersection type using [distributive conditional types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#distributive-conditional-types).
+Convert a union type to an intersection type using [distributive conditional types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#distributive-conditional-types).
 
-Inspired by a [stack overflow answer](https://stackoverflow.com/a/50375286/2172153).
+Inspired by [this Stack Overflow answer](https://stackoverflow.com/a/50375286/2172153).
 
 @example
-
-```ts
+```
 import {UnionToIntersection} from 'type-fest';
 
 type Union = {the(): void} | {great(arg: string): void} | {escape: boolean};
+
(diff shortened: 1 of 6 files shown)


source/union-to-intersection.d.ts
```diff
@@ -0,0 +1,59 @@
...
+
+type Intersection = UnionToIntersection<Union>;
+//=> {a1(): void; b1(): void; a2(argA: string): void; b2(argB: string): void}
+```
+*/
+export type UnionToIntersection<Union> = (
+  // `extends any` is always going to be the case and is used to convert the
+  // `Union` into a [distributive conditional
+  // type](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#distributive-conditional-types).
+  Union extends any
```
sindresorhus: Can you use `unknown` instead of `any`?

source/union-to-intersection.d.ts
```diff
@@ -0,0 +1,59 @@
...
+    // The union type is used as the only argument to a function since the union
+    // of function arguments is an intersection.
+    ? (distributedUnion: Union) => void
+    // This won't happen.
+    : never
+    // Infer the `Intersection` type since TypeScript represents the positional
+    // arguments of unions of functions as an intersection of the union.
+  ) extends ((mergedIntersection: infer Intersection) => void)
+    ? Intersection
+    : never;
```
sindresorhus: Use tab-indentation. Applies everywhere.

source/union-to-intersection.d.ts
ifiokjr: Who would have thought 😄 

Nice catch!

source/union-to-intersection.d.ts
ifiokjr: Fixed 😄

ifiokjr:
I forced push without first checking if you'd made changes yourself.

ifiokjr:
The missing commits are resolved now.

sindresorhus:
👍

PR #96, opened 2020-04-07
Fix `AsyncReturnType` type

### LINKED ISSUE #93: AsyncReturnType doesn't work
I'm not sure how `AsyncReturnType` is supposed to work. If I try it with a basic `async` function, that takes a string argument & returns it, I get a type error.

> Type '(bar: string) => Promise<string>' does not satisfy the constraint 'AsyncFunction'.
  Types of parameters 'bar' and 'args' are incompatible.
    Type 'unknown' is not assignable to type 'string'

I've set up a [Codesandbox](https://codesandbox.io/s/optimistic-shaw-jjxzi) that demonstrates this. I'm not sure what Typescript version it's using, it's not listed in the dependencies, but I am getting it with TS v3.8.2 in my local project.

I'm not sure if there is a specific reason not to use `any` here, but replacing `...args: unknown[]` with `...args: any[]` in the `AsyncFunction` definition resolves the error.

buschtoens:
Fixes #93.

### DIFF
--- a/source/async-return-type.d.ts
+++ b/source/async-return-type.d.ts
@@ -1,6 +1,6 @@
 import {PromiseValue} from './promise-value';
 
-type AsyncFunction = (...args: unknown[]) => Promise<unknown>;
+type AsyncFunction = (...args: any[]) => Promise<unknown>;
 
 /**
 Unwrap the return type of a function that returns a `Promise`.


fregante:
I'm not able to use this type, I get this error:

```
TS2344: Type '(query: string, options?: GHRestApiOptions) => Promise<RestResponse>' does not satisfy the constraint 'AsyncFunction'.
  Types of parameters 'query' and 'args' are incompatible.
    Type 'unknown' is not assignable to type 'string'.
```

fregante:
Repro:
```ts
const ident = async (query: string): Promise<string> => query;
type ShouldBeAString = AsyncReturnType<typeof ident>;
```

```log
Type '(query: string) => Promise<string>' does not satisfy the constraint 'AsyncFunction'.
  Types of parameters 'query' and 'args' are incompatible.
    Type 'unknown' is not assignable to type 'string'.ts(2344)
```

PR #97, opened 2020-04-10
Make `LiteralUnion` type  work for more types

kripod:
While [experimenting](https://github.com/kripod/glaze/blob/57b85375da31008cc2736d1c4c3c7db4e877a050/packages/glaze/src/useStyling.ts#L32-L46) with automatic prop completion for an upcoming CSS-in-JS library, I wanted to use `LiteralUnion` as follows:

```ts
import { CSSProperties } from 'react';

enum Duration {
  SHORT = 1, // Gets resolved to a string somewhere else in the code
  LONG,
}

type AnimationDurationValue = LiteralUnion<
  Duration,
  CSSProperties['animationDuration']
>;
```

`LiteralUnion` works perfectly when `LiteralType` doesn't extend `BaseType`.

### DIFF
--- a/source/literal-union.d.ts
+++ b/source/literal-union.d.ts
@@ -28,6 +28,6 @@ const pet: Pet2 = '';
 ```
  */
 export type LiteralUnion<
-	LiteralType extends BaseType,
+	LiteralType,
 	BaseType extends Primitive
 > = LiteralType | (BaseType & {_?: never});


sindresorhus:
// @BendingBender

kripod:
I would like to ask for a status update on this. Thank you for reviewing in advance!

sindresorhus:
I haven't had a chance to look into this yet. I'm thinking there must be a reason we did it like this in the first place. Maybe some more info in https://github.com/Microsoft/TypeScript/issues/29729.

Can you add a test that shows the use-case you need it for when it's not extending basetype?

kripod:
Unfortunately, I don't have the capacity to set up a test right now, sorry.

forivall:
TypeScript 4.0 no longer needs LiteralUnion (it preserves the literals that are union'd with their base type), so this can probably be closed (and LiteralUnion can eventually be removed)

papb:
@forivall I doubt that, since https://github.com/microsoft/TypeScript/issues/29729 is still open...

PR #100, opened 2020-04-20
Update unicorn link

dcecile:


sindresorhus:
Why?

dcecile:
🦄 As one unicorn fan to another, can you change the link to something that doesn't include violence against unicorns? 🙏 

(It's at 1:41 if you haven't skimmed through the video.)

PR #106, opened 2020-05-12
Add `FixedLengthArray` type

### LINKED ISSUE #102: Proposal: FixedLengthArray
```ts
const myArray: FixedLengthArray<number, 3> = [1, 2, 3]

const [a, b, c] = myArray // Ok
const [a, b, c, d] = myArray // Invalid index error
```

See https://stackoverflow.com/a/59906630/8384910

cc @sindresorhus

kainiedziela:
Fixes #102 

Create a type that represents an array of provided types and of a provided fixed length. The arrays length and Array prototype mutations that manipulate its length are excluded in the resulting type.

**Use-cases:**

- Declaring fixed-length tuples or arrays with a large number of items
- Creating a Range union (e.g. `0 | 1 | 2 | 3 | 4` from the keys of such a type) without having to resort to recursive types
- Creating an array of coordinates with a static length, of for example 3 for a 3D vector

**Example:**

```TS
import {FixedLengthArray} from 'type-fest';

type FencingTeam = FixedLengthArray<string, 3>;

const guestFencingTeam: FencingTeam = ['Josh', 'Michael', 'Robert'];
const homeFencingTeam: FencingTeam = ['George', 'John'];
//=> error TS2322: Type string[] is not assignable to type 'FencingTeam'
guestFencingTeam.push('Sam');
//=> error TS2339: Property 'push' does not exist on type 'FencingTeam'
```

**Caveats:**

I'm not *quite* sure how to test for the absence of Array.prototype methods that manipulate the arrays length without importing a testing framework. Would love to hear a suggestion if this test case is deemed necessary.

### DIFF
--- a/source/fixed-length-array.d.ts
+++ b/source/fixed-length-array.d.ts
@@ -20,7 +20,7 @@ export type FixedLengthArray<T, Length extends number, ArrayPrototype = [T, ...T
 	ArrayPrototype,
 	Exclude<keyof ArrayPrototype, ArrayLengthMutationKeys>
 > & {
-	[I: number]: T;
+	[index: number]: T;
 	[Symbol.iterator]: () => IterableIterator<T>;
 	readonly length: Length;
 };
--- a/source/fixed-length-array.d.ts
+++ b/source/fixed-length-array.d.ts
@@ -4,7 +4,7 @@ Methods to exclude.
 type ArrayLengthMutationKeys = 'splice' | 'push' | 'pop' | 'shift' | 'unshift';
 
 /**
-Create a type that represents an array of the given type and length. The arrays length and the `Array` prototype methods that manipulate its length are excluded in the resulting type.
+Create a type that represents an array of the given type and length. The array's length and the `Array` prototype methods that manipulate its length are excluded in the resulting type.
 
 Please participate in [this issue](https://github.com/microsoft/TypeScript/issues/26223) if you want to have a similiar type built into TypeScript.
 
--- a/source/fixed-length-array.d.ts
+++ b/source/fixed-length-array.d.ts
@@ -1,6 +1,13 @@
 /**
 Create a type that represents an array of a provided type and of a provided fixed length. The arrays length and Array prototype mutations that manipulate its length are excluded in the resulting type.
 
+Please participate in [this issue](https://github.com/microsoft/TypeScript/issues/26223) if you want to have a similiar type built-in in TypeScript.
+
+Use-cases:
+- Declaring fixed-length tuples or arrays with a large number of items
+- Creating a Range union (e.g. `0 | 1 | 2 | 3 | 4` from the keys of such a type) without having to resort to recursive types
+- Creating an array of coordinates with a static length, of for example 3 for a 3D vector
+
 @example
 ```
 import {FixedLengthArray} from 'type-fest';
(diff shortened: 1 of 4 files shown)


papb:
> I'm not _quite_ sure how to test for the absence of Array.prototype methods that manipulate the arrays length without importing a testing framework. Would love to hear a suggestion if this test case is deemed necessary.

I think you can do something like:

```ts
import {expectError} from 'tsd';

expectError(guestFencingTeam.push('Foo'));
```

This is how I *think* `expectError` works and what it looks like from some previous experiments I did with tsd, however [I am not sure](https://github.com/SamVerschueren/tsd/issues/55#issuecomment-616235426).

kainiedziela:
Yeah, I tried that before making a PR and unfortunately it didn't work. The `guestFencingTeam has no property 'push'` error still fires and `expectError` also throws an error stating that no error is present in the passed operation.

papb:
Yeah, I guess the documentation really needs update first then... Looks like `expectError` only works with an immediate function call inside the parentheses, and probably just checks for errors in the argument assignment. Very specific for something called `expectError` if you ask me...

source/fixed-length-array.d.ts
```diff
@@ -0,0 +1,29 @@
...
+//=> error TS2322: Type string[] is not assignable to type 'FencingTeam'
+guestFencingTeam.push('Sam');
+//=> error TS2339: Property 'push' does not exist on type 'FencingTeam'
+```
+*/
+export type FixedLengthArray<T, Length extends number, ArrayPrototype = [T, ...T[]]> = Pick<
+	ArrayPrototype,
+	Exclude<keyof ArrayPrototype, ArrayLengthMutationKeys>
+> & {
+	[I: number]: T;
```
sindresorhus, line 21:
```suggestion
	[index: number]: T;
```

source/fixed-length-array.d.ts
```diff
@@ -0,0 +1,29 @@
...
+type FencingTeam = FixedLengthArray<string, 3>;
+
+const guestFencingTeam: FencingTeam = ['Josh', 'Michael', 'Robert'];
+const homeFencingTeam: FencingTeam = ['George', 'John'];
+//=> error TS2322: Type string[] is not assignable to type 'FencingTeam'
+guestFencingTeam.push('Sam');
+//=> error TS2339: Property 'push' does not exist on type 'FencingTeam'
+```
+*/
+export type FixedLengthArray<T, Length extends number, ArrayPrototype = [T, ...T[]]> = Pick<
```
sindresorhus, line 17: `T` could be more descriptive. Maybe `Element`.

source/fixed-length-array.d.ts
```diff
@@ -0,0 +1,29 @@
+/**
+Create a type that represents an array of provided types and of a provided fixed length. The arrays length and Array prototype mutations that manipulate its length are excluded in the resulting type.
```
sindresorhus, line 2: This makes it sound like the array supports elements of various types, which is not the case.

source/fixed-length-array.d.ts
```diff
@@ -0,0 +1,29 @@
...
+
```
sindresorhus, line 3: > Write about some real-world use-cases where it can be useful. (It can be hard sometimes for users to see where they would use something) - https://github.com/sindresorhus/type-fest/blob/master/.github/contributing.md

---

> If there has been any discussion somewhere about this type, include a link to it. For example, a discussion on the TypeScript issue tracker.

source/fixed-length-array.d.ts
kainiedziela: Check out the current description. Should be clearer.

sindresorhus:
Looks good. Thanks 👍

PR #109, opened 2020-05-25
Add support for ESM exports in the `PackageJson` type

kripod:
Please see the [ECMAScript Modules docs of Node.js](https://nodejs.org/api/esm.html) for details about the implementation.

### DIFF
--- a/source/package-json.d.ts
+++ b/source/package-json.d.ts
@@ -231,8 +231,8 @@ declare namespace PackageJson {
 	*/
 	export type Exports =
 	| string
-	| { [key in ExportCondition]: Exports }
-	| { [key: string]: Exports };
+	| {[key in ExportCondition]: Exports}
+	| {[key: string]: Exports};
 
 	export interface NonStandardEntryPoints {
 		/**
--- a/source/package-json.d.ts
+++ b/source/package-json.d.ts
@@ -417,6 +417,8 @@ export type PackageJson = {
 
 	/**
 	Standard entry points of the package, with enhanced support for ECMAScript Modules.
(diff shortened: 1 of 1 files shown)


source/package-json.d.ts
```diff
@@ -380,11 +403,23 @@ export type PackageJson = {
...
+
 	/**
 	The module ID that is the primary entry point to the program.
 	*/
 	main?: string;
 
+	/**
+	Standard entry points of the package, with enhanced support for ECMAScript Modules.
+	*/
+	exports?: PackageJson.Exports;
```
sindresorhus, line 421: Can you link to its docs here?

kripod: Sure, thanks for the fast review!

source/package-json.d.ts
```diff
@@ -211,6 +211,29 @@ declare namespace PackageJson {
...
+	Conditions which provide a way to resolve a package entry point based on the environment.
+	*/
+	export type ExportCondition = LiteralUnion<
+		| 'import'
+		| 'require'
+		| 'node'
+		| 'deno'
+		| 'browser'
+		| 'electron'
+		| 'react-native'
```
privatenumber, line 224: @kripod

Can you reference where the values `browser`, `electron`, `react-native` are from? AFAIK they're unofficial and these values depend on specific bundlers (eg. Webpack)

kripod: Thank you for asking. Unfortunately, I’m not sure anymore. It could possibly be from (an older version of) the Node.js docs.

PR #112, opened 2020-06-11
Make the `PromiseValue` type handle nested promises

onichandame:
As per [this discussion](https://stackoverflow.com/questions/48011353/how-to-unwrap-type-of-a-promise?rq=1), a more realistic use case of unwrapping a type from Promise is when the Promise is resolved to another Promise.

For example, `await Promise.resolve(Promise.resolve(2))` is resolved as `number`, not `Promise<number>`. The current type `PromiseValue` however, does not reflect this behaviour.

Thanks to [Gerrit0](https://github.com/gerrit0) who shared the trick, I stole his brilliant idea and brought to this package.


sindresorhus:
Shouldn't `PromiseValue` just do this by default? Instead of a separate type.

onichandame:
- reason 1: I thought it would be clearer for the reviewer if changes are isolated in new files

- reason 2: Let me quote from the creator of this type, [Gerrit0 said](https://stackoverflow.com/questions/48011353/how-to-unwrap-type-of-a-promise?rq=1),

  ```text
  You can work around this issue by using a recursive type alias, but be aware that this is officially not supported and may break in future versions.
  ```

  Therefore I think it would be better to keep a fallback type `PromiseValue` so that when this type is broken someday, the maintainer can simply delete this type or alias it to `PromiseValue` to workaround the possible fatal bugs.

- reason 3: There may be some use cases where people do not want to recursively unwrap the promise(although I cannot think of any)

So I decided to post the changes first and see how people think about this.

sindresorhus:
Are you aware that they plan to solve this problem in TypeScript itself? https://github.com/microsoft/TypeScript/issues/27711#issuecomment-604142887

sindresorhus:
I doubt it will break in the future, but if so, we could probably find a workaround.

onichandame:
I did look at that issue. The problem is that the official support won't be available before TS 4.0. Before TS people have solved the problem, it is necessary to have this type as a temporary workaround.

I do wish to alter the behaviour of `PromiseValue` so that it unwraps recursively. So if you agree too, I will make another commit that implements it.

source/promise-value.d.ts
```diff
@@ -15,6 +15,12 @@ let data: Data = await asyncData;
...
 let syncData: SyncData = getSyncData();
+
+// Here's an example that shows how this type reacts to recursive Promise types.
+type RecursiveAsyncData = Promise<Promise<string> >;
+let recursiveAsyncData: PromiseValue<RecursiveAsyncData> = Promise.resolve(Promise.resolve('ABC'));
 ```
 */
-export type PromiseValue<PromiseType, Otherwise = PromiseType> = PromiseType extends Promise<infer Value> ? Value : Otherwise;
+export type PromiseValue<PromiseType, Otherwise = PromiseType> = PromiseType extends Promise<infer Value>
+	? {0: PromiseValue<Value>; 1: Value}[PromiseType extends Promise<any> ? 0 : 1]
```
sindresorhus, line 25:
```suggestion
	? {0: PromiseValue<Value>; 1: Value}[PromiseType extends Promise<unknown> ? 0 : 1]
```

source/promise-value.d.ts
```diff
@@ -15,6 +15,12 @@ let data: Data = await asyncData;
 // Here's an example that shows how this type reacts to non-Promise types.
```
sindresorhus, line 15: Can you mention "nested" support on line 2?

onichandame: please see the latest change. It is on line 3.

source/promise-value.d.ts
onichandame: done

PR #127, opened 2020-08-20
Add `SetReturnType` and `Asyncify` types

### LINKED ISSUE #88: Add `WithReturnType` type?
```ts
type WithReturnType<
  TFn extends (...args: any) => any,
  TR
> = TFn extends (...args: infer TArgs) => any ? (...args: TArgs) => TR : any;
```

Creates a new function type that replaces the given function type's return type with the given type.

Use case: functions that return wrapped callbacks, i.e. any generic function in the general form of

```ts
(cb: (...args: A) => any) => (...args: A) => B
```

e.g. here's a React hook:

```ts
function useAsync<TFn extends (...args: any) => Promise<any>>(
  asyncFn: TFn
): [
  { loading: boolean; result?: AsyncReturnType<TFn> },
  WithReturnType<TFn, Promise<void>>
] {
  const [state, setState] = React.useState({
    loading: false,
    result: undefined as AsyncReturnType<TFn> | undefined
  });
  return [state, (async (...args: any) => {
    setState({ ...state, loading: true });
    try {
      const result = await asyncFn(...args);
      setState({ ...state, loading: false, result });
    } catch (e) {
      setState({ ...state, loading: false });
    }
  }) as WithReturnType<TFn, Promise<void>>];
}
```

I'm not sure if this use case is widespread enough but the solution is non-obvious if you're not familiar with conditional types (it's derived from how `ReturnType` is defined). I can create a PR if there's interest in adding this.

### LINKED ISSUE #125: Proposal: `ReplaceReturnType`
A type where you can change the return type of a function type. This makes it possible to reuse a function type definition where you need slightly different return values. For example, a Node.js package that has both an async and sync API, and the only difference is that the async API returns the value wrapped in a `Promise` type.

Implementation and example use-case: https://github.com/sindresorhus/make-synchronous/blob/6796cca6109a5e5068c5f1cadc99ef7e0d9b8e6a/index.d.ts

Thoughts?

### LINKED ISSUE #126: Proposal: `Asyncify`
Many Node.js packages ([example](https://github.com/sindresorhus/make-dir)) expose both an async and sync API. Usually, the only difference type-wise is that the async versions' return value is wrapped in a `Promise`. It would be useful if the type definition could be reused, so you could just take the definition for the sync version and wrap it in `type FooAsync = Asyncify<FooSync>`.

papb:
This PR adds two types:

* `SetReturnType`
* `Asyncify`

I decided to add the two in a single PR because the second depends directly on the first. I will be happy to separate into two PRs upon request (but the second would have to wait for the first to be merged first).

I can also change the name of the first to `WithReturnType` or `ReplaceReturnType` upon request; I don't have a *strong* preference but I slightly prefer `SetReturnType`.

Closes #88
Closes #125
Closes #126

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -86,7 +86,7 @@ Click the type names for complete docs.
 - [`IterableElement`](source/iterable-element.d.ts) - Get the element type of an `Iterable`/`AsyncIterable`. For example, an array or a generator.
 - [`Entry`](source/entry.d.ts) - Create a type that represents the type of an entry of a collection.
 - [`Entries`](source/entries.d.ts) - Create a type that represents the type of the entries of a collection.
-- [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters of a given function type.
+- [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`Asyncify`](source/asyncify.d.ts) - Create an async version of the given function type.
 
 ### Miscellaneous
--- a/readme.md
+++ b/readme.md
@@ -83,6 +83,8 @@ Click the type names for complete docs.
 - [`UnionToIntersection`](source/union-to-intersection.d.ts) - Convert a union type to an intersection type.
 - [`Stringified`](source/stringified.d.ts) - Create a type with the keys of the given type changed to `string` type.
 - [`FixedLengthArray`](source/fixed-length-array.d.ts) - Create a type that represents an array of the given type and length.
+- [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters of a given function type.
+- [`Asyncify`](source/asyncify.d.ts) - Create an async version of the given function type.
 
 ### Miscellaneous
 
--- a/source/asyncify.d.ts
+++ b/source/asyncify.d.ts
@@ -2,8 +2,9 @@ import {PromiseValue} from './promise-value';
 import {SetReturnType} from './set-return-type';
 
 /**
-Create an async version of the given function type.
+Create an async version of the given function type, by boxing the return type in `Promise`, while keeping the same parameter types.
 
+Use-case: you have two functions, one synchronous and one asynchronous, that do the same thing. Instead of having to duplicate the type specification, you can use `Asyncify` to reuse the synchronous type.
 @example
 ```
 import {Asyncify} from 'type-fest';
--- a/source/asyncify.d.ts
+++ b/source/asyncify.d.ts
@@ -27,4 +27,4 @@ const getFooAsync: AsyncifiedFooGetter = (someArg) => {
 }
 ```
 */
-export type Asyncify<Fn extends (...args: any[]) => unknown> = SetReturnType<Fn, Promise<PromiseValue<ReturnType<Fn>>>>;
+export type Asyncify<Fn extends (...args: any[]) => any> = SetReturnType<Fn, Promise<PromiseValue<ReturnType<Fn>>>>;
--- a/source/asyncify.d.ts
+++ b/source/asyncify.d.ts
@@ -8,13 +8,15 @@ Create an async version of the given function type.
 ```
 import {Asyncify} from 'type-fest';
 
+// Synchronous function.
 function getFooSync(someArg: SomeType): Foo {
-	// ...
+	// …
 }
 
 type AsyncifiedFooGetter = Asyncify<typeof getFooSync>;
-// type AsyncifiedFooGetter = (someArg: SomeType) => Promise<Foo>;
+//=> type AsyncifiedFooGetter = (someArg: SomeType) => Promise<Foo>;
 
+// Same as `getFooSync` but asynchronous.
 const getFooAsync: AsyncifiedFooGetter = (someArg) => {
 	// TypeScript now knows that `someArg` is `SomeType` automatically.
 	// It also knows that this function must return `Promise<Foo>`.
--- a/source/asyncify.d.ts
+++ b/source/asyncify.d.ts
@@ -2,9 +2,10 @@ import {PromiseValue} from './promise-value';
 import {SetReturnType} from './set-return-type';
 
 /**
-Create an async version of the given function type, by boxing the return type in `Promise`, while keeping the same parameter types.
+Create an async version of the given function type, by boxing the return type in `Promise` while keeping the same parameter types.
 
-Use-case: you have two functions, one synchronous and one asynchronous, that do the same thing. Instead of having to duplicate the type specification, you can use `Asyncify` to reuse the synchronous type.
+Use-case: You have two functions, one synchronous and one asynchronous that do the same thing. Instead of having to duplicate the type definition, you can use `Asyncify` to reuse the synchronous type.
+
 @example
 ```
 import {Asyncify} from 'type-fest';
@@ -21,9 +22,9 @@ type AsyncifiedFooGetter = Asyncify<typeof getFooSync>;
 const getFooAsync: AsyncifiedFooGetter = (someArg) => {
 	// TypeScript now knows that `someArg` is `SomeType` automatically.
 	// It also knows that this function must return `Promise<Foo>`.
-	// If you have `@typescript-eslint/promise-function-async` linter rule enabled, it will even report that "Functions that return promises must be async."
+	// If you have `@typescript-eslint/promise-function-async` linter rule enabled, it will even report that "Functions that return promises must be async.".
 
-	// ...
+	// …
 }
 ```
 */
--- a/source/asyncify.d.ts
+++ b/source/asyncify.d.ts
@@ -24,4 +24,4 @@ const getFooAsync: AsyncifiedFooGetter = (someArg) => {
 }
 ```
 */
-export type Asyncify<Fn extends (...args: any[]) => any> = SetReturnType<Fn, Promise<PromiseValue<ReturnType<Fn>>>>;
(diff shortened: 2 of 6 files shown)


source/set-return-type.d.ts
```diff
@@ -0,0 +1,29 @@
...
+import {SetReturnType} from 'type-fest';
+
+type MyFunction = (foo: SomeType, bar: unknown) => SomeOtherType;
+
+type MyWrappedFunction = SetReturnType<MyFunction, boolean>;
+// type MyWrappedFunction = (foo: SomeType, bar: unknown) => boolean;
+```
+*/
+export type SetReturnType<Fn extends (...args: any[]) => any, TypeToReturn> =
+	// Just using `Parameters<Fn>` isn't ideal because it doesn't handle the `this` fake parameter
```
sindresorhus, line 21: Is this intended TS behavior or a bug? If it's a bug, it would be good to open an issue on TS.

source/asyncify.d.ts
```diff
@@ -0,0 +1,27 @@
...
+const getFooAsync: AsyncifiedFooGetter = (someArg) => {
+	// TypeScript now knows that `someArg` is `SomeType` automatically.
+	// It also knows that this function must return `Promise<Foo>`.
+	// If you have `@typescript-eslint/promise-function-async` linter rule enabled, it will even report that "Functions that return promises must be async."
+
+	// ...
+}
+```
+*/
+export type Asyncify<Fn extends (...args: any[]) => any> = SetReturnType<Fn, Promise<PromiseValue<ReturnType<Fn>>>>;
```
sindresorhus, line 27:
```suggestion
export type Asyncify<Fn extends (...args: any[]) => unknown> = SetReturnType<Fn, Promise<PromiseValue<ReturnType<Fn>>>>;
```

Can you use `unknown` here?

source/asyncify.d.ts
```diff
@@ -0,0 +1,27 @@
+import {PromiseValue} from './promise-value';
+import {SetReturnType} from './set-return-type';
+
+/**
+Create an async version of the given function type.
+
+@example
+```ts
```
sindresorhus, line 8:
```suggestion
```
```

File: source/set-return-type.d.ts [hunk 4]
```diff
@@ -0,0 +1,29 @@
+type IsAny<T> = 0 extends (1 & T) ? true : false; // https://stackoverflow.com/a/49928360/3406963
+type IsNever<T> = [T] extends [never] ? true : false;
+type IsUnknown<T> = IsNever<T> extends false ? T extends unknown ? unknown extends T ? IsAny<T> extends false ? true : false : false : false : false;
```
sindresorhus, line 3: Do you think it's worth exposing any of these types? (Could be done later on if decided on. Just curious about your opinion).

sindresorhus (changes requested):

sindresorhus:
I'm ok with `SetReturnType`.

source/set-return-type.d.ts
papb: That is a good question, I don't know the answer :sweat_smile:

I have never really used the `this` fake parameter explicitly... I recalled its existence while writing the PR, tested for it and noticed `Parameters<Fn>` doesn't catch it... Didn't even think it might be a bug :sweat_smile:

source/asyncify.d.ts
papb: Yes! To the best of my knowledge, yes. Anything that extends `() => any` also extends `() => unknown` and vice versa, since:

* `any` is assignable to `unknown`, and
* `unknown` is also assignable to `any`

source/asyncify.d.ts
papb: You probably mean to delete just the `ts`, not the whole thing, right?

source/set-return-type.d.ts
papb: I would really like that, personally! Whenever I want to use any of them (mostly for testing purposes), I have to look them up around the internet, they're not so easy to memorize. If they were here I'd just quickly import instead :grimacing: I use them when trying to build some complex type every now and then. I think I used `IsAny` somewhere else in a PR to another repository of yours...

It may be a little hard to give examples of real use cases though... It might be fun to try to find good examples... Maybe just saying that they can be used to work around edge cases when creating more complex types?

sindresorhus:
@papb Friendly bump :)

papb:
I will come back to this soon, sorry for the delay :)

source/set-return-type.d.ts
```diff
@@ -0,0 +1,29 @@
...
+
+type MyFunction = (foo: SomeType, bar: unknown) => SomeOtherType;
+
+type MyWrappedFunction = SetReturnType<MyFunction, boolean>;
+// type MyWrappedFunction = (foo: SomeType, bar: unknown) => boolean;
+```
+*/
+export type SetReturnType<Fn extends (...args: any[]) => any, TypeToReturn> =
+	// Just using `Parameters<Fn>` isn't ideal because it doesn't handle the `this` fake parameter
+	Fn extends (this: infer ThisArg, ...args: infer Arguments) => any ? (
```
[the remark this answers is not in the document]
papb: To my surprise, no. I have just found out that `(a: never) => any` does not extend `(...args: any[]) => unknown`, only `(...args: any[]) => any`. I have absolutely no idea why though.

source/set-return-type.d.ts
papb: Done! https://github.com/microsoft/TypeScript/issues/41072 and https://github.com/microsoft/TypeScript/issues/41073

readme.md
```diff
@@ -83,6 +83,8 @@ Click the type names for complete docs.
 - [`UnionToIntersection`](source/union-to-intersection.d.ts) - Convert a union type to an intersection type.
 - [`Stringified`](source/stringified.d.ts) - Create a type with the keys of the given type changed to `string` type.
 - [`FixedLengthArray`](source/fixed-length-array.d.ts) - Create a type that represents an array of the given type and length.
+- [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters of a given function type.
```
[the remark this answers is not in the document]
papb:
```suggestion
- [`SetReturnType`](source/set-return-type.d.ts) - Create a function type by modifying the return type of an existing function type.
```

What do you think?

source/asyncify.d.ts
```diff
@@ -0,0 +1,27 @@
+import {PromiseValue} from './promise-value';
+import {SetReturnType} from './set-return-type';
+
+/**
+Create an async version of the given function type.
```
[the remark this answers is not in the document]
papb:
```suggestion
Create an async version of the given function type, by boxing the return type in `Promise`, while keeping the same parameter types.
```

source/set-return-type.d.ts
```diff
@@ -0,0 +1,29 @@
+type IsAny<T> = 0 extends (1 & T) ? true : false; // https://stackoverflow.com/a/49928360/3406963
+type IsNever<T> = [T] extends [never] ? true : false;
+type IsUnknown<T> = IsNever<T> extends false ? T extends unknown ? unknown extends T ? IsAny<T> extends false ? true : false : false : false : false;
+
+/**
+Get a function type based on another, changing the return type to one of your choice.
```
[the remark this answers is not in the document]
papb:
```suggestion
Create a function type by modifying the return type of the given function type.
```

source/set-return-type.d.ts
```diff
@@ -0,0 +1,29 @@
...
+
+Use-case: You want to define a wrapped function that returns something different while receiving the same parameters.
```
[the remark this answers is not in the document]
papb:
```suggestion
Use-case: You want to define a wrapped function that returns something different while receiving the same parameters. For example, you might want to wrap a function that can throw an error into one that will return `undefined` instead.
```

source/set-return-type.d.ts
```diff
@@ -0,0 +1,29 @@
...
+
+@example
+```
+import {SetReturnType} from 'type-fest';
+
+type MyFunction = (foo: SomeType, bar: unknown) => SomeOtherType;
+
+type MyWrappedFunction = SetReturnType<MyFunction, boolean>;
+//=> type MyWrappedFunction = (foo: SomeType, bar: unknown) => boolean;
```
papb, line 17:
```suggestion
type MyFunctionThatCanThrow = (foo: SomeType, bar: unknown) => SomeOtherType;

type MyWrappedFunction = SetReturnType<MyFunction, SomeOtherType | undefined>;
//=> type MyWrappedFunction = (foo: SomeType, bar: unknown) => SomeOtherType | undefined;
```

source/asyncify.d.ts
```diff
@@ -0,0 +1,29 @@
...
+
```
papb, line 6:
```suggestion

Use-case: you have two functions, one synchronous and one asynchronous, that do the same thing. Instead of having to duplicate the type specification, you can use `Asyncify` to reuse the synchronous type.
```

source/asyncify.d.ts
```diff
@@ -0,0 +1,27 @@
...
```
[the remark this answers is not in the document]
papb: Done in 4868235abd05ff476fe3026851df3f8ecb181498

papb:
> 
> 
> > If there has been any discussion somewhere about this type, include a link to it. For example, a discussion on the TypeScript issue tracker. - https://github.com/sindresorhus/type-fest/blob/master/.github/contributing.md

Sorry, I am sure this is a common request but I could not find any particular discussion that I think is worth linking, do you want to suggest one?

sindresorhus:
Nice work 👍🏻

Issue #134, opened 2020-10-12
Template literal type ideas


sindresorhus:
TypeScript 4.1 will bring a lot of cool abilities: https://devblogs.microsoft.com/typescript/announcing-typescript-4-1-beta/

Let's brainstorm some ideas.

---

[...]

PR #138, opened 2020-11-04
Add `CamelCase`, `KebabCase`, `SnakeCase`, `PascalCase`, `DelimitorCase` types

voxpelli:
Putting this here to start a conversation around #134 

I'm currently running in to some problems with xo tests locally (both #134 and `Parsing error: Cannot read property map of undefined`). I don't think it depends on TypeScript 4.1, but can't say for sure.

**Update 2020-11-16:** Added `SnakeCase<value>` and generic `DelimitorCase<value, delimitor>` as well


test-d/kebab-case.ts
```diff
@@ -0,0 +1,34 @@
...
+
+// Verifying example
+type KebabCasedProps<T> = {
+    [K in keyof T as KebabCase<K>]: T[K]
+};
+
+interface CliOptions {
+		dryRun: boolean;
+		includeFile: string;
+    foo: number;
```
papb, line 27: Indentation is odd here and in other places as well. You should only use tabs.

Hint: if you're using VSCode, install the editorconfig extension and run `npm i -g editorconfig` once and this will become much easier.

voxpelli: @papb Thanks for pointing that out 👍 I do have the editorconfig extension installed, however this code was pasted from a place where it had space indention and I unfortunately hadn't turned on whitespace visualization in VS Code for some reason.

Fixed and pushed 👍

voxpelli:
Added `SnakeCase<value>` and generic `DelimitorCase<value, delimitor>` as well

sindresorhus:
Thanks for working on this 🙌🏻 These types look really useful.

Can you add the types to the readme? And ensure you adhere to https://github.com/sindresorhus/type-fest/blob/master/.github/contributing.md

test-d/delimitor-case.ts
```diff
@@ -0,0 +1,104 @@
...
+const kebabFromSpace: KebabCase<'foo bar'> = 'foo-bar';
+expectType<'foo-bar'>(kebabFromSpace);
+
+const kebabFromSnake: KebabCase<'foo_bar'> = 'foo-bar';
+expectType<'foo-bar'>(kebabFromSnake);
+
+const noKebabFromMono: KebabCase<'foobar'> = 'foobar';
+expectType<'foobar'>(noKebabFromMono);
+
+// SnakeCase
```
sindresorhus, line 53: This should be in a separate test file.

index.d.ts
```diff
@@ -10,6 +10,7 @@ export {RequireAtLeastOne} from './source/require-at-least-one';
 export {RequireExactlyOne} from './source/require-exactly-one';
 export {PartialDeep} from './source/partial-deep';
 export {ReadonlyDeep} from './source/readonly-deep';
+export {DelimitorCase, KebabCase, SnakeCase} from './source/delimitor-case';
```
sindresorhus, line 13:
```suggestion
export {DelimiterCase, KebabCase, SnakeCase} from './source/delimitor-case';
```

voxpelli:
I'll look into this once I have completed my deadline today

voxpelli:
> And ensure you adhere to https://github.com/sindresorhus/type-fest/blob/master/.github/contributing.md

Do you want DelimiterCase, KebabCase and SnakeCase in one file each?

sindresorhus:
> Do you want DelimiterCase, KebabCase and SnakeCase in one file each?

Yes

sindresorhus:
Would you be interested in including a `CamelCase` type here too?

voxpelli:
Yes, I have one working in a project of mine, I’ll add it here

mån 23 nov. 2020 kl. 08:02 skrev Sindre Sorhus <notifications@github.com>:

voxpelli:
> > Do you want DelimiterCase, KebabCase and SnakeCase in one file each?
> 
> Yes. If they have any utility types they share, you could move those to a `utilities.d.ts` file.

The `DelimiterCase` is pretty much the utility type, maybe we don't want to export that one?

Anyhow: I have separated them all into separate files, added them to the README + added a new CamelCase implementation as well (ping @knpwrs)

The tests probably fails because `tsd` hasn't been updated to include TS 4.1?

Also: As these are exported in `index.d.ts`, `type-fest` at large will probably only work with TS 4.1 going forward if this is merged. One could probably work around that if it is desired.

test-d/camel-case.ts
```diff
@@ -0,0 +1,47 @@
...
+expectType<'fooBar'>(camelFromSnake);
+
+const noDelimiterFromMono: CamelCase<'foobar'> = 'foobar';
+expectType<'foobar'>(noDelimiterFromMono);
+
+const camelFromMixed: CamelCase<'foo-bar_abc xyzBarFoo'> = 'fooBarAbcXyzBarFoo';
+expectType<'fooBarAbcXyzBarFoo'>(camelFromMixed);
+
+const camelFromVendorPrefixedCssProperty: CamelCase<'-webkit-animation'> = 'WebkitAnimation';
+expectType<'WebkitAnimation'>(camelFromVendorPrefixedCssProperty);
```
knpwrs, line 24: Lodash's `camelCase` function returns `webkitAnimation` (lowercase `w`). I don't know what other libraries we should test but if we're preferring one library over another I think we should document which behavior is picked.

If libraries all consistently return `webkitAnimation` then this should be simple enough to solve with `Uncapitalize` or some otherwise minor reworking.

voxpelli: Right, I can agree with that, would require the adding of a `trim<>`, but that's doable 👍

voxpelli: @knpwrs Fixed it 👍

sindresorhus:
> Also: As these are exported in index.d.ts, type-fest at large will probably only work with TS 4.1 going forward if this is merged. One could probably work around that if it is desired.

I think we'll have to do that. For example, I plan to use the `CamelCase` type in https://github.com/sindresorhus/camelcase-keys, and I don't think I can require TS 4.1 there already. Could you just do a ts41.d.ts where you import index.d.ts and add the new exports instead and then use `typesVersions` in package.json?

sindresorhus:
Would you be able to also add `PascalCase`? I think we're then pretty well covered.

voxpelli:
> Would you be able to also add `PascalCase`? I think we're then pretty well covered.

Sure! And force the first char to be uppercase there while forcing the first one to be lowercase for CamelCase

voxpelli:
> Could you just do a ts41.d.ts where you import index.d.ts and add the new exports instead and then use `typesVersions` in package.json?

I like that approach 👍

Ideally we should get tests running on both TS 4.1 and an older/the oldest supported TS version to ensure that it works as expected and that no unintentional 4.1 dependencies gets introduced down the line

sindresorhus:
> And force the first char to be uppercase there while forcing the first one to be lowercase for CamelCase

Yes

readme.md
```diff
@@ -89,6 +89,15 @@ Click the type names for complete docs.
 - [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`Asyncify`](source/asyncify.d.ts) - Create an async version of the given function type.
 
+### Template literal types
+
+*Note:* These require [TypeScript 4.1 or newer](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#template-literal-types).
+
+- [`CamelCase`](source/camel-case.d.ts) – Converts a typical casing to `camelCasing`.
```
sindresorhus: "typical casing" is bit weird way to say it.

source/snake-case.d.ts
```diff
@@ -0,0 +1,30 @@
+import {DelimiterCase} from './delimiter-case';
+
+/**
+Converts a string literal from any typical non-snake-case casing to a snake-case
```
sindresorhus: Is it clear to the user what "typical non-snake-case casing" means?

source/delimiter-case.d.ts
```diff
@@ -0,0 +1,64 @@
+export type UpperCaseChars = 'A' | 'B' | 'C' | 'D' | 'E' | 'F' | 'G' | 'H' | 'I' | 'J' | 'K' | 'L' | 'M' | 'N' | 'O' | 'P' | 'Q' | 'R' | 'S' | 'T' | 'U' | 'V' | 'X' | 'Y' | 'Z';
```
sindresorhus, line 1:
```suggestion
export type UpperCaseCharacters = 'A' | 'B' | 'C' | 'D' | 'E' | 'F' | 'G' | 'H' | 'I' | 'J' | 'K' | 'L' | 'M' | 'N' | 'O' | 'P' | 'Q' | 'R' | 'S' | 'T' | 'U' | 'V' | 'X' | 'Y' | 'Z';
```

source/camel-case.d.ts
```diff
@@ -0,0 +1,49 @@
...
+type InnerCamelCaseStringArray<Parts extends any[], PreviousPart> =
+	Parts extends [`${infer FirstPart}`, ...infer RemainingParts]
+		? FirstPart extends undefined
+			? ''
+			: FirstPart extends ''
+					? InnerCamelCaseStringArray<RemainingParts, PreviousPart>
+					: `${PreviousPart extends '' ? FirstPart : Capitalize<FirstPart>}${InnerCamelCaseStringArray<RemainingParts, FirstPart>}`
+		: '';
+
+type CamelCaseStringArray<Parts extends string[]> =
```
sindresorhus, line 18: Could you add a short code comment to this and the above ones about what they do?

source/camel-case.d.ts
```diff
@@ -0,0 +1,49 @@
+import {WordSeparators} from './delimiter-case';
```
sindresorhus, line 1: `WordSeparators` should be in `utilities.d.ts`. Types should ideally not import from each other.

source/camel-case.d.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+Converts a string literal from any typical non-camel-case casing to camel-case.
+
+This can be useful when, for example, converting some kebab-cased command-line flags or a snake-cased database result.
+
+@example
+```
+import {CamelCase} from 'type-fest';
+
+type CamelCasedProps<T> = {
```
sindresorhus, line 32: We can do this later, but I think this is an extremely common situation, so I think we should expose this as a type too.

source/camel-case.d.ts
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+Converts a string literal from any typical non-camel-case casing to camel-case.
+
+This can be useful when, for example, converting some kebab-cased command-line flags or a snake-cased database result.
+
+@example
+```
+import {CamelCase} from 'type-fest';
+
```
sindresorhus, line 31: Could you do a super simple example first? A single property or something.

source/delimiter-case.d.ts
```diff
@@ -0,0 +1,64 @@
...
+	Parts extends [`${infer FirstPart}`, ...infer RemainingParts]
+		? `${StringPartToDelimiterCase<FirstPart, UsedWordSeparators, UsedUpperCaseChars, Delimiter>}${StringArrayToDelimiterCase<RemainingParts, UsedWordSeparators, UsedUpperCaseChars, Delimiter>}`
+		: '';
+
+/**
+Converts a string literal from any typical casing to one based on a custom delimiter
+
+This can be useful when eg. converting a camel cased object property to an oddly cased one.
+
+Also see KebabCase and SnakeCase.
```
sindresorhus, line 33: Could we use `@see` here? https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#jsdoc-see-tag

readme.md
```diff
@@ -89,6 +89,15 @@ Click the type names for complete docs.
 - [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`Asyncify`](source/asyncify.d.ts) - Create an async version of the given function type.
 
+### Template literal types
+
+*Note:* These require [TypeScript 4.1 or newer](https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#template-literal-types).
+
+- [`CamelCase`](source/camel-case.d.ts) – Convert a typical casing to camel-case (`fooBar`).
```
sindresorhus, line 96: I think the description here could be clearer. Not clear what "typical casing" means.

sindresorhus (changes requested):

source/snake-case.d.ts
voxpelli: We could simplify  it to:

> Converts a string literal to snake case

Alternative would be to list them all:

> Converts a kebab cased, camel cased and/or pascal cased string literal to snake case

sindresorhus: Let's go with the former

source/camel-case.d.ts
voxpelli: Very much agree that this would be a good next step

source/delimiter-case.d.ts
voxpelli: @sindresorhus I'm a bit unsure how to do that actually, in the reference they have imported the module that they are referring to, but that's not the case here.

I'll just add this for now:

```
@see KebabCase
@see SnakeCase
```

voxpelli:
Fixed backwards compatibility, added PascalCase and  pretty much all other feedback as well

sindresorhus:
Yay! Really awesome work, @voxpelli. 🎉

sindresorhus:
@voxpelli Would you be interested in being added as a maintainer here? No worries if not.

voxpelli:
@sindresorhus Would be happy to be a co-maintainer! Have you written up your preferred flow / your expectations for co-maintenance?

sindresorhus:
Nothing much. I generally trust people I add. Just do PRs for non-trivial things and new types. Expectations are whatever you feel like. It's open-source. We do it for fun. But if you need any ideas, comment on type proposals, review open PRs, etc.

PR #139, opened 2020-11-04
Add GitHub CI setup

voxpelli:
As Travis CI is now severely limited for OSS projects and I would say GitHub Actions are more accessible for GitHub projects.

Based on a modified copy of [the one in Unicorn](https://raw.githubusercontent.com/sindresorhus/eslint-plugin-unicorn/master/.github/workflows/ci.yml), which we back then discussed in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/504.

Added tests for Node `14` and `15` on top of the `10` and `12` that's in the Travis file.

Once https://github.com/actions/setup-node/issues/26 has been solved the version numbers can be replaced with aliases like `lts` instead, until then eg. https://github.com/dcodeIO/setup-node-nvm could be used if wanted.

Runs well: https://github.com/voxpelli/type-fest/actions?query=workflow%3ACI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,28 @@
+name: CI
+
+on:
+  - push
+  - pull_request
+
+jobs:
+  test:
+    name: Node.js ${{ matrix.node_version }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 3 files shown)


sindresorhus:
Can you remove the travis file and replace the badge?

voxpelli:
Done now @sindresorhus 👍

PR #142, opened 2020-11-11
`RequireAtLeastOne`: Make other given keys optional

### LINKED ISSUE #59: `RequireAtLeaseOne` issues
```ts
import { RequireAtLeastOne } from 'type-fest';

interface BaseConditionProps {
  when: string;
  is?: any;
  isNot?: any;
  children: (value: any) => React.ReactNode;
}

type ConditionProps = RequireAtLeastOne<BaseConditionProps, 'isNot' | 'is'>;

const Condition: React.FC<ConditionProps> = ({ when, is, isNot, children }) => (  // ERROR HERE
  <Field name={when} subscription={{ value: true }}>
    {({ input: { value } }) => {
      const show = is ? value === is : value !== isNot;
      return show ? children(value) : null;
    }}
  </Field>
);

```

Typescript complain that 
1. Property 'isNot' does not exist on type 'PropsWithChildren<RequireAtLeastOne<BaseConditionProps, "is" | "isNot">>'.ts(2339) 
2. Property 'is' does not exist on type 'PropsWithChildren<RequireAtLeastOne<BaseConditionProps, "is" | "isNot">>'.ts(2339)

I had no problem if I use another version that I found in the issue.

```ts
import { Except } from 'type-fest';

export type RequireOne<ObjectType, KeysType extends keyof ObjectType = keyof ObjectType> = {
  // For each Key in KeysType make a mapped type
  [Key in KeysType]: (// …by picking that Key's type and making it required
  Required<Pick<ObjectType, Key>> &
    // …and adding the other keys in KeysType as optional and of type `never`
    Partial<Record<Exclude<KeysType, Key>, never>>);
}[KeysType] &
  Except<ObjectType, KeysType>;
```

ulken:
For each mapped type with given required key, make all other given keys optional.

Fixes #59

### DIFF
--- a/source/require-at-least-one.d.ts
+++ b/source/require-at-least-one.d.ts
@@ -20,13 +20,14 @@ const responder: RequireAtLeastOne<Responder, 'text' | 'json'> = {
 };
 ```
 */
-export type RequireAtLeastOne<ObjectType, KeysType extends keyof ObjectType = keyof ObjectType> =
-	{
-		// For each Key in KeysType make a mapped type
-		[Key in KeysType]: (
-			// …by picking that Key's type and making it required
-			Required<Pick<ObjectType, Key>>
-		)
-	}[KeysType]
-	// …then, make intersection types by adding the remaining keys to each mapped type.
-	& Except<ObjectType, KeysType>;
+export type RequireAtLeastOne<
+	ObjectType,
+	KeysType extends keyof ObjectType = keyof ObjectType
+> = {
+	// For each `Key` in `KeysType` make a mapped type:
+	[Key in KeysType]-?: Required<Pick<ObjectType, Key>> & // 1. Make `Key`'s type required
+		// 2. Make all other keys in `KeysType` optional
+		Partial<Pick<ObjectType, Exclude<KeysType, Key>>>;
+}[KeysType] &
+	// 3. Add the remaining keys not in `KeysType`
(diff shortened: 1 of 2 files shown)


sindresorhus:
From the description:

> The remaining keys are kept as is.

That's not really true anymore, right?

ulken:
> That's not really true anymore, right?

I get confusion, but I believe it still holds. It gets a little hairy with the terminology, but I'll try to explain the differences, as I understand it:
- **Remaining keys**: keys not given as parameters, i.e. keys of `ObjectType` not present in `KeysType`. Note: if no keys are given, then `KeysType` includes all keys of `ObjectType`. 
- **Other given keys**: all keys of `KeysType` except the one we're currently mapping over.

**Example**
```ts
type SystemMessages = {
  default: string;
  
  macos?: string;
  linux?: string;
  windows?: string;

  optional?: string;
}

type ValidMessages = RequireAtLeastOne<
  SystemMessages,
  "macos" | "linux" | "windows"
>
```

In this case, `default` and `optional` are _remaining keys_ and they are left intact/as is.

When we're creating the type where "macos" is required, "linux" and "windows" are the _other given keys_.

Does that make sense?

The change this PR introduces is making the _other given keys_ optional.

ulken:
@sindresorhus do you feel me?

sindresorhus:
That makes sense. Thanks for explaining.

PR #143, opened 2020-11-12
Fix `entry.d.ts` missing exports

fjc0k:
https://github.com/sindresorhus/type-fest/blob/e6827d94e3af8b5611cf330263df53f62ce683d6/source/entries.d.ts#L1

`rollup-plugin-dts`:

```bash
Error: 'ArrayEntry' is not exported by ****/node_modules/type-fest/source/entries.d.ts
```

### DIFF
--- a/source/entry.d.ts
+++ b/source/entry.d.ts
@@ -1,10 +1,10 @@
 type MapKey<BaseType> = BaseType extends Map<infer KeyType, unknown> ? KeyType : never;
 type MapValue<BaseType> = BaseType extends Map<unknown, infer ValueType> ? ValueType : never;
 
-type ArrayEntry<BaseType extends readonly unknown[]> = [number, BaseType[number]];
-type MapEntry<BaseType> = [MapKey<BaseType>, MapValue<BaseType>];
-type ObjectEntry<BaseType> = [keyof BaseType, BaseType[keyof BaseType]];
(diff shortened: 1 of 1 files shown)


sindresorhus:
Those are internal types. I don't see why they need to be exported. Sounds like a bug in that plugin.

fjc0k:
@sindresorhus 

eeee, https://api-extractor.com/pages/overview/demo_rollup/:

```bash
Warning: node_modules/type-fest/source/entry.d.ts:55:1 - (ae-forgotten-export) The symbol "MapEntry" needs to be exported by the entry point index.d.ts
Warning: node_modules/type-fest/source/entry.d.ts:55:1 - (ae-forgotten-export) The symbol "SetEntry" needs to be exported by the entry point index.d.ts
Warning: node_modules/type-fest/source/entry.d.ts:55:1 - (ae-forgotten-export) The symbol "ArrayEntry" needs to be exported by the entry point index.d.ts
Warning: node_modules/type-fest/source/entry.d.ts:55:1 - (ae-forgotten-export) The symbol "ObjectEntry" needs to be exported by the entry point index.d.ts
```

papb:
Why do they 'need to be exported'?

sindresorhus:
Ah, because they're used here: https://github.com/sindresorhus/type-fest/blob/7714f57bdb0f9841c525d9b81079066cb9de9091/source/entries.d.ts#L1

papb:
Oh, wow. Indeed. Amazing.

PR #153, opened 2020-11-25
Add `Get` type

### LINKED ISSUE #147: Proposal: Key path type
For example, to correctly get the type of a nested property using a string key path: `get(object, 'foo.bar')`, like https://github.com/sindresorhus/dot-prop

I found this playground:
https://www.typescriptlang.org/play?ts=4.1.0-pr-40336-88#code/FAFwngDgpgBACgQxACwJIFsIBsA8AVAGhgGkowYoAPEKAOwBMBnGAazIHsAzGPAPhgC8wGCTIVqdJjEYgATgEtaAc2EwA-DwDapMAF1xNBswBKUAMbtZ9HDIXKiCWmF6qRGgD4wABgBIA3joAvgB0-ogoGNj42mS6RACilGZYAK70UDhsYFxaOnGsHNyOYJq6vPwAZNJyikqBXq4iMJ6+AWQh-onJaRlZOXgxekR9RU6lldV2dQ1NMABcMLRQAG5QsqoLS6uyANzAoJCw4WiYWABM+PwC8EgnUYQF2dx8zY-9ewfQNyiXgt9351+VEMUlstVeIx46n+kUBLwWkLwH3AX2OADUEKkMg84AZJMxjpcrqpccD8d5-IpOGtRGAOn4qTTTDJ6qoNDo8UY3s9GhpmSBOVJCQM8i5Zm5-hisdE8kR+WLxZsVmtGkrtht4ILmIjeVo4LpVYtlbt9ulkghZLBOClaGYQPJ2LQYEooCB8ERSRIucLygAKdgAIwAVgsHhBbgs4ABKSO3KUpbEe3gfCy0GQwQNB8wC65+VSceSyGQAOQQ6CgCwARAAReRQJTsSsEVRYBAlssVmCVgASCAAXk3VAgXQsAMwABmbIggsnYWbtjAWmkafkWHarpgQLHkICbMFTNQDKRAlkXMAAbAAOGCBKdNVe0dddgDCjpkSCge4PCiPJ6LCwARjOG87wNQIYDbfc3xAD4XRAf1g2zIhKxnOds0YYJx2CR9y0rKMdiAA

mmkal:
Gets a deeply-nested property from an object, like lodash's [`get` function](https://lodash.com/docs/#get).

Use-case: retrieve a property from deep inside an API response or other complex object.

I couldn't find anything like this, so thought I'd put up a proposal since I've written shoddy versions of it a few times now.

```typescript
import { Get } from 'type-fest'

interface ApiResponse {
  hits: {
    hits: Array<{
      _id: string
      _source: {
        name: Array<{
          given: string[]
          family: string
        }>
        birthDate: string
      }
    }>
  }
}
type Name = Get<ApiResponse, 'hits.hits[0]._source.name'> // Array<{ given: string[]; family: string }>
```

This works by first splitting the path based on `.` and `[...]` characters into a tuple of string keys.
Then it recursively uses the head key to get the next property of the current object, until there are no keys
left. Number keys extract the item type from arrays, or are converted to strings to extract types from tuples
and dictionaries with number keys.

Todo:

- [x] propagate nullable props
- [x] add tests for nullable props (~likely~ not blocked by #162 but I don't think tests are being run by CI properly right now so maybe)
- [x] make bad paths return `unknown`, not `undefined`
- [x] add test for readonly props/arrays
- [x] rename `Numbers` to `StringDigit` or similar

Follow-ups:
- Extract `KeyPath` type like in #158
- `Set` type - `Set<{ foo: number }, 'bar.baz', string> = { foo: number; bar: { baz: string } }`
- (maybe) support `foo\\.bar.baz` to allow for dots within keys, similar to dot-prop.
---

Fixes #147


source/split.d.ts
```diff
@@ -0,0 +1,8 @@
+/**
+Recursively split a string literal into two parts on the first occurence of the given string, returning an array literal of all the separate parts.
+*/
+export type Split<S extends string, D extends string> =
```
mmkal, line 4: Moved from `camel-case.d.ts`.

test-d/get.ts
```diff
@@ -0,0 +1,50 @@
...
+      };
+    }>;
+  };
+}
+
+expectType<Get<ApiResponse, 'hits.hits[0]._source.name'>>([{given: ['Homer', 'J'], family: 'Simpson'}]);
+expectType<Get<ApiResponse, 'hits.hits.0._source.name'>>([{given: ['Homer', 'J'], family: 'Simpson'}]);
+expectType<Get<ApiResponse, 'hits.hits[12345]._source.name'>>([{given: ['Homer', 'J'], family: 'Simpson'}]);
+
+expectType<Get<ApiResponse, 'hits.someNonsense.notTheRightPath'>>({} as never);
```
mmkal, line 23: Please let me know if there's a better way to ensure `never` from `expectType`, or to do type-to-type (rather than type-to-value) comparisons.

I have a similar library called [expect-type](https://npmjs.com/package/expect-type) which I'd normally use, the syntax in this case would be a bit clearer (I find the difference between actual and expected here confusing) and wouldn't require the dummy variable:

```typescript
expectTypeOf<Get<ApiResponse, 'hit.hits[0]._source.name'>>()
  .toEqualTypeOf<Array<{ given: string[]; family: string }>>()

expectTypeOf<Get<ApiResponse, 'hits.someNonsense.notTheRightPath'>>().toBeNever()
```

ts41/get.d.ts
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Gets a deeply-nested property from an object, like lodash's `get` method.
+ *
+ * Use-case: retrieve a property from deep inside an API response or other complex object.
+ *
+ * @example
+ * import { Get } from 'type-fest'
+ *
+ * interface ApiResponse {
```
mmkal, line 12: This is what [elasticsearch responses](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-request-body.html) look like. Just as an example of the kind of very-nested payloads you can get in the real world. The `_source` document is a [FHIR Patient](http://www.hl7.org/fhir/patient.html) - again, wanted to keep it somewhat realistic.

sindresorhus:
Yes, we definitely want this. Thanks for working on it. We actually have an issue for it: https://github.com/sindresorhus/type-fest/issues/147

Did you see https://www.typescriptlang.org/play?ts=4.1.0-pr-40336-88#code/FAFwngDgpgBACgQxACwJIFsIBsA8AVAGhgGkowYoAPEKAOwBMBnGAazIHsAzGPAPhgC8wGCTIVqdJjEYgATgEtaAc2EwA-DwDapMAF1xNBswBKUAMbtZ9HDIXKiCWmF6qRGgD4wABgBIA3joAvgB0-ogoGNj42mS6RACilGZYAK70UDhsYFxaOnGsHNyOYJq6vPwAZNJyikqBXq4iMJ6+AWQh-onJaRlZOXgxekR9RU6lldV2dQ1NMABcMLRQAG5QsqoLS6uyANzAoJCw4WiYWABM+PwC8EgnUYQF2dx8zY-9ewfQNyiXgt9351+VEMUlstVeIx46n+kUBLwWkLwH3AX2OADUEKkMg84AZJMxjpcrqpccD8d5-IpOGtRGAOn4qTTTDJ6qoNDo8UY3s9GhpmSBOVJCQM8i5Zm5-hisdE8kR+WLxZsVmtGkrtht4ILmIjeVo4LpVYtlbt9ulkghZLBOClaGYQPJ2LQYEooCB8ERSRIucLygAKdgAIwAVgsHhBbgs4ABKSO3KUpbEe3gfCy0GQwQNB8wC65+VSceSyGQAOQQ6CgCwARAAReRQJTsSsEVRYBAlssVmCVgASCAAXk3VAgXQsAMwABmbIggsnYWbtjAWmkafkWHarpgQLHkICbMFTNQDKRAlkXMAAbAAOGCBKdNVe0dddgDCjpkSCge4PCiPJ6LCwARjOG87wNQIYDbfc3xAD4XRAf1g2zIhKxnOds0YYJx2CR9y0rKMdiAA ? Maybe there's some inspiration in there.

ts41/split.d.ts
```diff
@@ -0,0 +1,8 @@
+/**
+Recursively split a string literal into two parts on the first occurence of the given string, returning an array literal of all the separate parts.
+*/
+export type Split<S extends string, D extends string> =
+	string extends S ? string[] :
+	S extends '' ? [] :
+	S extends `${infer T}${D}${infer U}` ? [T, ...Split<U, D>] :
+	[S];
```
sindresorhus, line 8: Can you move this to a `ts41/utilities.d.ts` file? I prefer to only have files for exported types.

ts41/get.d.ts
```diff
@@ -0,0 +1,82 @@
...
+ * type Name = Get<ApiResponse, 'hits.hits[0]._source.name'> // Array<{ given: string[]; family: string }>
+ *
+ * @explanation
+ *
+ * This works by first splitting the path based on `.` and `[...]` characters into a tuple of string keys.
+ * Then it recursively uses the head key to get the next property of the current object, until there are no keys
+ * left. Number keys extract the item type from arrays, or are converted to strings to extract types from tuples
+ * and dictionaries with number keys.
+ */
+export type Get<Object, Path extends string> = GetWithPath<Object, ToPath<Path>>;
```
sindresorhus, line 36: The exported type should be at the bottom.

ts41/get.d.ts
```diff
@@ -0,0 +1,82 @@
+import {Split} from '../source/split';
+import {Integers} from '../source/utilities';
+
+/**
+ * Gets a deeply-nested property from an object, like lodash's `get` method.
+ *
```
sindresorhus, line 6: Don't prefix the doc comments with `*`. See the other type's doc comments.

ts41/get.d.ts
```diff
@@ -0,0 +1,82 @@
...
+ *         }>
+ *         birthDate: string
+ *       }
+ *     }>
+ *   }
+ * }
+ *
+ * type Name = Get<ApiResponse, 'hits.hits[0]._source.name'> // Array<{ given: string[]; family: string }>
+ *
+ * @explanation
```
sindresorhus, line 29: Is this a valid JSDoc tag? I couldn't find it.

ts41/get.d.ts
```diff
@@ -0,0 +1,82 @@
+import {Split} from '../source/split';
+import {Integers} from '../source/utilities';
+
+/**
+ * Gets a deeply-nested property from an object, like lodash's `get` method.
+ *
+ * Use-case: retrieve a property from deep inside an API response or other complex object.
+ *
+ * @example
+ * import { Get } from 'type-fest'
```
sindresorhus, line 10: This should be wrapped in backticks, see the other types as example.

test-d/get.ts
sindresorhus: This is actually a bug with the tool we use: https://github.com/SamVerschueren/tsd/issues/78 Feel free to bring in your package here.

ts41/get.d.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Then it recursively uses the head key to get the next property of the current object, until there are no keys
+ * left. Number keys extract the item type from arrays, or are converted to strings to extract types from tuples
+ * and dictionaries with number keys.
+ */
+export type Get<Object, Path extends string> = GetWithPath<Object, ToPath<Path>>;
+
+/**
+ * Like @see Get but receives an array of strings as a path parameter.
+ */
+type GetWithPath<T, Keys extends string[]> =
```
sindresorhus, line 41:
```suggestion
type GetWithPath<T, Keys extends readonly string[]> =
```

Should this be `readonly` so it also works with readonly arrays?

ts41/get.d.ts
```diff
@@ -0,0 +1,82 @@
...
+type ConsistsOnlyOf<S extends string, C extends string> =
+  S extends ''
+  ? true
+  : S extends `${C}${infer Tail}`
+  ? ConsistsOnlyOf<Tail, C>
+  : false;
+
+type IsInteger<S extends string> = ConsistsOnlyOf<S, Integers>;
+
+type WithStringKeys<T extends Record<string | number, any>> = {
```
sindresorhus, line 64: Can you add a short code comment about the purpose of this one? Just for readability.

ts41/get.d.ts
sindresorhus: Regardless, can you add some tests that uses readonly arrays? Just to ensure it works.

readme.md
```diff
@@ -97,6 +97,7 @@ Click the type names for complete docs.
 - [`PascalCase`](ts41/pascal-case.d.ts) – Converts a string literal to pascal-case (`FooBar`)
 - [`SnakeCase`](ts41/snake-case.d.ts) – Convert a string literal to snake-case (`foo_bar`).
 - [`DelimiterCase`](ts41/delimiter-case.d.ts) – Convert a string literal to a custom string delimiter casing.
+- [`Get`](ts41/get.d.ts) - Gets a deeply-nested property from an object, like lodash's get method.
```
sindresorhus, line 100:
```suggestion
- [`Get`](ts41/get.d.ts) - Get a deeply-nested property from an object using a key path, like Lodash's `.get()` method.
```

?

ts41/get.d.ts
```diff
@@ -0,0 +1,82 @@
...
+  [K in `${Extract<keyof T, string | number>}`]: T[K]
+};
+
+/**
+ * Get a property of an object or array. Works when indexing arrays using number-literal-strings, e.g. `PropertyOf<number[], '0'>  = number`,
+ * and when indexing objects with number keys.
+ * Returns `never` if `Key` is not a property of `Object`,
+ */
+type PropertyOf<Object, Key extends string> =
```
sindresorhus, line 73:
```suggestion
type PropertyOf<ObjectType, Key extends string> =
```

I kept reading it as the built-in `Object` type before seeing the generic here.

sindresorhus (changes requested):

sindresorhus:
Make sure you adhere to https://github.com/sindresorhus/type-fest/blob/master/.github/contributing.md

ts41/get.d.ts
mmkal: No! I was struggling to get formatting to work properly, and didn't mean to leave this in. Removing.

ts41/get.d.ts
mmkal: Have made this readonly but this type isn't exported so can't be directly tested. That's fine though IMO - it also can't be incorrectly used since it's not exported.

ts41/get.d.ts
mmkal: 👍 have removed the prefixes for consistency, but I wonder if they should be added to the other comments (in a follow-up)? VSCode seems to prefer the stars:

Without stars:

image

With stars:

image

(the formatting gets flattened without the stars)

Either way, totally agree this type should be inline with all the other existing ones right now.

mmkal:
@sindresorhus thanks for reviewing, have updated based on comments. One note - in the latest revision I changed the `never`s to `undefined`s since that's the standard behaviour - `never` would imply "throwing"

```js
get({ foo: 'bar' }, 'some.nonsense.string') // returns undefined, rather than throws
```

source/utilities.d.ts
```diff
@@ -1,3 +1,5 @@
 export type UpperCaseCharacters = 'A' | 'B' | 'C' | 'D' | 'E' | 'F' | 'G' | 'H' | 'I' | 'J' | 'K' | 'L' | 'M' | 'N' | 'O' | 'P' | 'Q' | 'R' | 'S' | 'T' | 'U' | 'V' | 'X' | 'Y' | 'Z';
 
 export type WordSeparators = '-' | '_' | ' ';
+
+export type Integers = '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9';
```
sindresorhus, line 5: The name should probably mention that it's strings.

test-d/get.ts
```diff
@@ -0,0 +1,49 @@
+import {Get} from '../ts41/get';
+import {expectTypeOf} from 'expect-type';
+
+interface ApiResponse {
+  hits: {
+    hits: Array<{
+      _id: string;
```
sindresorhus, line 7: Use tab-indentation. Applies elsewhere too.

test-d/get.ts
```diff
@@ -0,0 +1,49 @@
...
+      _source: {
+        name: Array<{
```
sindresorhus, line 9:
```suggestion
        name: ReadonlyArray<{
```

just to ensure it works with readonly arrays too.

ts41/get.d.ts
```diff
@@ -0,0 +1,93 @@
...
+Like @see Get but receives an array of strings as a path parameter.
+*/
+type GetWithPath<T, Keys extends readonly string[]> =
+  Keys extends []
+  ? T
+  : Keys extends [infer Head, ...infer Tail]
+  ? GetWithPath<PropertyOf<T, Extract<Head, string>>, Extract<Tail, string[]>>
+  : never;
+
+type ToPath<S extends string> = Split<FixPathSquareBrackets<S>, '.'>;
```
sindresorhus, line 14: Could use a short code comment on what it does.

ts41/get.d.ts
```diff
@@ -0,0 +1,93 @@
...
+
+type FixPathSquareBrackets<S extends string> =
+  S extends `${infer T}[${infer U}]${infer V}`
+  ? `${T}.${U}${FixPathSquareBrackets<V>}`
+  : S;
+
+type ConsistsOnlyOf<S extends string, C extends string> =
```
sindresorhus, line 21: Could use a short code comment.

ts41/get.d.ts
```diff
@@ -0,0 +1,93 @@
...
+  S extends ''
+  ? true
+  : S extends `${C}${infer Tail}`
+  ? ConsistsOnlyOf<Tail, C>
+  : false;
+
+type IsInteger<S extends string> = ConsistsOnlyOf<S, Integers>;
```
sindresorhus, line 28:
```suggestion
type IsIntegerString<S extends string> = ConsistsOnlyOf<S, Integers>;
```

?

ts41/get.d.ts
```diff
@@ -0,0 +1,93 @@
...
+```
+*/
+type WithStringKeys<T extends Record<string | number, any>> = {
+  [K in `${Extract<keyof T, string | number>}`]: T[K]
+};
+
+/**
+Get a property of an object or array. Works when indexing arrays using number-literal-strings, e.g. `PropertyOf<number[], '0'>  = number`, and when indexing objects with number keys.
+
+Returns `neundefinedver` if `Key` is not a property of `Object`,
```
sindresorhus, line 51: Typo?

And `Object` => `ObjectType`.

ts41/get.d.ts
```diff
@@ -0,0 +1,93 @@
...
+type WithStringsKeys = keyof WithStrings;
+//=> 'foo' | '0'
+```
+*/
+type WithStringKeys<T extends Record<string | number, any>> = {
+  [K in `${Extract<keyof T, string | number>}`]: T[K]
+};
+
+/**
+Get a property of an object or array. Works when indexing arrays using number-literal-strings, for example, `PropertyOf<number[], '0'> = number`, and when indexing objects with number keys.
```
sindresorhus, line 49:
```suggestion
Get a property of an object or array. Works when indexing arrays using number-literal-strings, for example, `PropertyOf<number[], '0'> = number`, and when indexing objects with number keys.
```

Is ` = number` correct there?

sindresorhus (changes requested):

sindresorhus:
Can you add some more tests? And fix the lint errors.

ts41/get.d.ts
```diff
@@ -0,0 +1,93 @@
...
+          given: string[]
+          family: string
+        }>
+        birthDate: string
+      }
+    }>
+  }
+}
+
+type Name = Get<ApiResponse, 'hits.hits[0]._source.name'>;
```
sindresorhus, line 89: Can you show a more real-world example with a function? I don't think anyone would actually use the key path directly in the type. It would come from some function. Like: https://github.com/sindresorhus/type-fest/pull/158/files#diff-c4a707129eadc5349a83e23c9bd5a1fc2f05e0c88b1365e6fb7efe9961ae95ddR13-R21

ts41/get.d.ts
kainiedziela: `NumericalString`?

ts41/utilities.d.ts
```diff
@@ -0,0 +1,8 @@
+/**
```
kainiedziela, line 1: This isn't really a "utility" as they are used in source/utilities.d.ts. It's more of a separate type – and as such deserves a separate file, description, use case and needs to be included in readme.

ts41/utilities.d.ts
```diff
@@ -0,0 +1,8 @@
...
+Recursively split a string literal into two parts on the first occurence of the given string, returning an array literal of all the separate parts.
+*/
+export type Split<S extends string, D extends string> =
```
kainiedziela, line 4: This isn't really a "utility" as they are used in source/utilities.d.ts. It's more of a separate type – and as such deserves a separate file, description, use case and needs to be included in readme.

source/utilities.d.ts
kainiedziela: Same as [here](https://github.com/sindresorhus/type-fest/pull/153/files#r531693170) – I recommend `NumericalString`

kainiedziela (changes requested):

ts41/utilities.d.ts
mmkal: 👍 I'll mimic what you've done in your PR then it won't matter which goes in first (I won't get a chance to push another revision until Monday, most likely)

ts41/get.d.ts
sindresorhus: `1.1` is a number, but it's not allowed, so I think "numerical" is to general.

ts41/utilities.d.ts
sindresorhus: I don't agree with the premise that all individual non-utility types should be exposed. We should expose what's generally useful. In this specific case, I do agree `Split` could be useful for people.

source/utilities.d.ts
mmkal: What about `StringDigit`? Usually literal unions are singular, e.g. `type Direction = 'up' | 'down'`. If we go with that in a follow up I could also change the other type utilities in this file to be singular (reason becomes kinda clear if you try to use them in a "normal" way: `let my letter: UppercaseCharacters = 'F'`)

ts41/get.d.ts
mmkal: Yes I think `= number` is correct. It's roughly modeling normal array usage in typescript:

```typescript
const myArray: number[] = [1, 2, 3]

const myNumber: number = myArray[0]
```

Of course this is `'0'` rather than `0` but the stringified key matches the runtime behaviour of `lodash.get(myNumber, '0')`.

ts41/utilities.d.ts
mmkal: In the end I left it here unexposed, because I don't want to add scope/keep in sync with a different PR, even if the eventual plan is to expose and document it. @kainiedziela maybe whichever of us merges in second can just deal with conflicts, should be v simple.

ts41/get.d.ts
mmkal: I removed this type completely. It was only used once so I replaced its usage with `ConsistsOnlyOf<Key, StringDigit>`.

test-d/get.ts
mmkal: Added tests for `readonly`, `ReadonlyArray`, as well as nullable/undefinedable props below.

mmkal:
I changed the main test to use a lodash/dot-prop-like `get` function rather than the type util directly, since that is likely how it'll most commonly be used. Then added a bunch more tests for edge cases which do use the type util directly.

sindresorhus:
> BTW @sindresorhus, there are a lot of commits in this PR. I think it should be squash-and-merged, but if you prefer not to do that, let me know and I can rebase and clean up on this branch before it goes in.

I always squash on merge. I prefer keeping the commits visible in the PR as it makes it easier to follow changes.

sindresorhus:
> Edit: hadn't seen the issue/playground - will take a look and see if there are any simplifications/improvements to be made.

Did you find any?

ts41/get.d.ts
```diff
@@ -0,0 +1,118 @@
+import {Split} from './utilities';
+import {StringDigit} from '../source/utilities';
+
+/**
+Like @see Get but receives an array of strings as a path parameter.
```
sindresorhus, line 5: You cannot use `@see` like this.

https://tsdoc.org/pages/tags/see/

sindresorhus (changes requested):

ts41/get.d.ts
```diff
@@ -0,0 +1,123 @@
...
+type WithStrings = WithStringKeys<WithNumbers>;
+
+type WithNumbersKeys = keyof WithNumbers;
+//=> 'foo' | 0
+type WithStringsKeys = keyof WithStrings;
+//=> 'foo' | '0'
+```
+*/
+type WithStringKeys<ObjectType extends Record<string | number, any>> = {
+	[Key in `${Extract<keyof ObjectType, string | number>}`]: ObjectType[Key]
```
voxpelli, line 68: How about the following instead? I think it would be simpler to read?

```suggestion
	[Key in keyof ObjectType as `${Key}`]: ObjectType[Key]
```

ts41/get.d.ts
```diff
@@ -0,0 +1,123 @@
...
+*/
+type PropertyOf<ObjectType, Key extends string> =
+	ObjectType extends null | undefined
+	? undefined
+	: Key extends keyof ObjectType
+	? ObjectType[Key]
+	: ObjectType extends ArrayLike<infer Item>
+	? ConsistsOnlyOf<Key, StringDigit> extends true
+	? Item
+	: unknown
```
voxpelli, line 84: We needs some parenthesis here to make things readable.

Only when the rows alternate between `?` and `:` every other row can parenthesis be skipped, as they are then essentially the `else if`:s of ternary operators. But this isn't an `else if`, this is a nested `if`, so we need parenthesis to highlight that and separate it from the larger `else if`-style ternary:

```suggestion
	? (
		ConsistsOnlyOf<Key, StringDigit> extends true
		? Item
		: unknown
	)
```

Also: `unknown` here should be `never`?

ts41/get.d.ts
```diff
@@ -0,0 +1,123 @@
...
+Get a property of an object or array. Works when indexing arrays using number-literal-strings, for example, `PropertyOf<number[], '0'> = number`, and when indexing objects with number keys.
+
+Returns `unknown` if `Key` is not a property of `ObjectType`.
+*/
+type PropertyOf<ObjectType, Key extends string> =
+	ObjectType extends null | undefined
+	? undefined
+	: Key extends keyof ObjectType
+	? ObjectType[Key]
+	: ObjectType extends ArrayLike<infer Item>
```
voxpelli, line 81: If `Key` is eg. `'length'` then `Key extends keyof ObjectType` will be truthy if `ObjectType` is an array.

I think we therefore need to handle the array case first to avoid this?

And in regards to the array case, a couple of things:

First:

`ArrayLike<>` is barely documented at all, neither in this repo nor in the official TypeScript docs, I had to go to the source code of TypeScript itself to find what it does: https://github.com/microsoft/TypeScript/blob/d0c28ab0c2c54c8d4ab2f0c34385aed501ef73c6/src/lib/es5.d.ts#L1431-L1434

Are there any benefits of this over `(infer Item)[]`, `Array<infer Item>` or even `{ [n: number]: infer Item; }`?

Secondly:

I would suggest splitting the logic for arrays out of this + ensure that it works correctly with array literals, like this:

```typescript
// Safe to assume that all array literals have a 0 key?
type isArrayLiteral<ArrayType> = '0' extends keyof ArrayType ? true : false;

type KeyOfArray<ArrayType extends any[], Key extends string> =
  // One should only use numeric types for arrays, so any other keys gets the "never" treatment
  ConsistsOnlyOf<Key, StringDigit> extends false
  ? never
  // If we have found ourselves an array literal, then get value by key
  : isArrayLiteral<ArrayType> extends true
  ? ArrayType[Key]
  // Else infer type
  : ArrayType extends (infer ItemType)[]
  ? ItemType
  // This won't ever happen
  : never
)
```

Using that we could rewrite this part of `PropertyOf` to something like:

```suggestion
	: ObjectType extends any[]
	? KeyOfArray<ObjectType, Key>
	: Key extends keyof ObjectType
	? ObjectType[Key]
```

voxpelli (commented):
Here's some thoughts

ts41/get.d.ts
mmkal: I did something like that at first, but weirdly typescript isn't able to keep track of the fact that `Key` must be a string or number (I'd have thought it would be able to, since `ObjectType extends Record<string | number, any>`:

image

ts41/get.d.ts
```diff
@@ -0,0 +1,123 @@
...
+		}>
+	}
+}
+
+const getName = (apiResponse: ApiResponse) =>
+	get(apiResponse, 'hits.hits[0]._source.name');
+	//=> Array<{given: string[]; family: string}>
+```
+*/
+export type Get<Object, Path extends string> = GetWithPath<Object, ToPath<Path>>;
```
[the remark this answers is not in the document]
mmkal: 👍 I prefer that, will change all the `ObjectType`s to `BaseType`

ts41/get.d.ts
```diff
@@ -0,0 +1,123 @@
...
+};
+
+/**
+Get a property of an object or array. Works when indexing arrays using number-literal-strings, for example, `PropertyOf<number[], '0'> = number`, and when indexing objects with number keys.
+
+Returns `unknown` if `Key` is not a property of `ObjectType`.
+*/
+type PropertyOf<ObjectType, Key extends string> =
+	ObjectType extends null | undefined
+	? undefined
```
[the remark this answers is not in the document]
mmkal: I wrote this as a model for the `get` functions from libraries like lodash, dot-prop and jest's `expect({ foo: { bar: 1 } }).toHaveProperty('foo.bar', 1)`. For that behaviour to be properly mapped to the type system, nullish values need to propagate:

```js
declare const myObj: {
  foo?: {
    bar: { baz: number }
  }
}

get(myObj, 'foo.bar.baz')
```

The return type for the expression above should be `number | undefined`, but if we changed this to `never`, the return type would be `number` (because `number | never` => `number`). Which would be an accurate model for a library which _throws_ if the path can't be fully resolved, but I'm not aware of any that work that way.

ts41/get.d.ts
mmkal: Will improve the formatting. @sindresorhus do you generally use prettier or anything similar? I'd love to have the decision taken out of my hands!

Re `unknown` vs `never` - similar to the other comment - `never` would imply an error being thrown when the key isn't found. All of this assumes that we're trying to modal a deep-get function as accurately as possible. There could definitely be type-land use-cases for `never`s - maybe `PathKey` and `PathValue` from the playground could co-exist with this type?

ts41/get.d.ts
mmkal: I'll try this out - I agree, `ArrayLike` is weird. FWIW I _think_ we'd be ok if the key is `length` as the other check would just get `ObjectType['length']` which would resolve to `number`.

Also I'm not sure it is safe to assume all array/tuple literals have a '0' key. `Parameters<() => {}>` is a not-totally-unrealistic counterexample which has type `[]`.

mmkal: Ok, here are the various tradeoffs to each approach:

`Array<infer Item>` of `(infer Item)[]` won't work without a special case for `ReadonlyArray`. Failing test:
```typescript
type Test = { a: ReadonlyArray<{b: 1 }> }
expectTypeOf<Get<Test, 'a[0].b'>>().toEqualTypeOf<1>()
```

`{ [n: number]: infer Item }` breaks number-indexed types. This test passes when it shouldn't:
```typescript
type Test = { 1: { b: string } }
expectTypeOf<Get<Test, '123456.b'>>().toBeString()
```

The following all work:
1. `ReadonlyArray<infer Item>`
2. `ArrayLike<infer Item>`
3. `{ length: number; [n: number]: infer Item }` (equivalent to ArrayLike)

Given this conversation, I'd like to go for 3 to avoid using a little-documented type like in 2, or relying on `Array` being assignable to `ReadonlyArray` like in 1.

Tests for all these cases already exist in `test-d/get.ts` - feel free to try them out and let me know if you find a better fit. (Note: I had to un-exclude `"test-d/**/*"` from tsconfig to see errors in my IDE).

package.json
```diff
@@ -35,6 +35,7 @@
 	],
 	"devDependencies": {
 		"@sindresorhus/tsconfig": "~0.7.0",
+		"expect-type": "^0.9.0",
```
sindresorhus, line 38: Can you update to the latest version?

ts41/get.d.ts
```diff
@@ -0,0 +1,131 @@
+import {Split} from './utilities';
+import {StringDigit} from '../source/utilities';
+
+/**
+Like @see {@link Get} but receives an array of strings as a path parameter.
```
sindresorhus, line 5: I don't think this is valid `@see` syntax. VS Code renders it raw and https://tsdoc.org/play only renders `Like`.

sindresorhus (changes requested):

ts41/get.d.ts
sindresorhus: > Will improve the formatting. @sindresorhus do you generally use prettier or anything similar? I'd love to have the decision taken out of my hands!

I don't. Prettier is a good idea in theory, but in practice, it makes too many weird decisions, often making code less readable.

ts41/get.d.ts
voxpelli: Yeah, it should be at the beginning of the line: https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#jsdoc-see-tag

package.json
voxpelli: Do we still want `expect-type` in there now that `tsd` is activated again as of #162?

sindresorhus: Yes, it covers some cases `tsd` currently does not handle.

mmkal: Will update the version and obviously no problem if this gets replaced with tsd in future.

sindresorhus:
@mmkal This looks good to me when https://github.com/sindresorhus/type-fest/pull/153#discussion_r546335324 is fixed.

@voxpelli LGTY?

mmkal:
Sounds good - will be able to work on it in a couple of days. Happy new year!

ts41/get.d.ts
```diff
@@ -2,7 +2,7 @@ import {Split} from './utilities';
 import {StringDigit} from '../source/utilities';
 
 /**
-Like @see {@link Get} but receives an array of strings as a path parameter.
+Like @see Get but receives an array of strings as a path parameter.
```
sindresorhus, line 5: Per the linked page, you cannot have `@see` in the middle of a sentence.

sindresorhus, line 5:
```suggestion
Like the `Get` type but receives an array of strings as a path parameter.
```

mmkal: 🤦 sorry.

sindresorhus (approved):

PR #157, opened 2020-11-26
Add `Mutable` type

### LINKED ISSUE #146: [Proposal] Writable and SetWritable
This util would allow removing `readonly` prop on an object

```ts
/**
 * Make all properties in T writable (removes readonly)
 */
type Writable<T> = {
  -readonly [P in keyof T]: T[P];
};
```

Based on SetRequired:

```ts
type SetWritable<BaseType, Keys extends keyof BaseType = keyof BaseType> =
  // Pick just the keys that are not required from the base type.
  Except<BaseType, Keys> &
    // Pick the keys that should be required from the base type and make them required.
    Writable<Pick<BaseType, Keys>> extends infer InferredType // If `InferredType` extends the previous, then for each key, use the inferred type key.
    ? { [KeyType in keyof InferredType]: InferredType[KeyType] }
    : never;
```

Real world example:

For compatibility, Notification.permission should be written (see example https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API#Example) but Notification.permission is readonly.

```ts
const requestNotification = async (): Promise<void> => {
  if (Notification.permission === 'granted') return;

  try {
    const status = await Notification.requestPermission();
    // Whatever the user answers, we make sure Chrome stores the information
    if (!('permission' in Notification)) {
      (Notification as SetWritable<typeof Notification, 'permission'>).permission = status;
    }
  } catch {}
};
```

kainiedziela:
Fixes  #146

### SetMutable

Creates a type that converts the given keys from `readonly` to mutable. The remaining keys are kept as is.

**Use-case:** You want to define a single model where the only thing that changes is whether or not some of the keys are mutable.

```TS
import {SetMutable} from 'type-fest';

type Foo = {
	readonly a: number;
	b: string;
	readonly c: boolean;
}

type SomeMutable = SetMutable<Foo, 'b' | 'c'>;
// type SomeMutable = {
// 	readonly a: number;
// 	b: string; // Was already mutable and still is.
// 	c: boolean; // Is now mutable.
// }
```


source/set-mutable.d.ts
```diff
@@ -0,0 +1,35 @@
...
+
+type SomeMutable = SetMutable<Foo, 'b' | 'c'>;
+// type SomeMutable = {
+// 	readonly a: number;
+// 	b: string; // Was already mutable and still is.
+// 	c: boolean; // Is now mutable.
+// }
+```
+*/
+export type SetMutable<BaseType, Keys extends keyof BaseType = keyof BaseType> =
```
mmkal, line 27: The second typearg has a default of `keyof BaseType`, doesn't this reduce down to the same as `Mutable` if that default is used? Is there a use case for `SetMutable<T>`?

kainiedziela: The difference between `Mutable` and `SetMutable` is that in the latter you can set only some keys as mutable, while leaving the rest unafffected.

kainiedziela: There really isn't, but that's the way the other `Set-` types work, so I just replicated that pattern.

source/set-mutable.d.ts
```diff
@@ -0,0 +1,35 @@
+import {Except} from './except';
+import {Mutable} from './mutable';
+
+/**
+Create a type that converts the given keys from `readonly` to mutable. The remaining keys are kept as is.
+
+Use-case: You want to define a single model where the only thing that changes is whether or not some of the keys are mutable.
+
```
sindresorhus, line 8: Can you add a `@see` tag for the `Mutable` type. And the inverse in the `Mutable` type file?

source/set-mutable.d.ts
sindresorhus: If there really isn't any use-case for that type parameter, we should remove it from those other types too.

source/set-mutable.d.ts
```diff
@@ -0,0 +1,35 @@
+import {Except} from './except';
+import {Mutable} from './mutable';
+
+/**
+Create a type that converts the given keys from `readonly` to mutable. The remaining keys are kept as is.
+
```
sindresorhus, line 6: While it's implied, think we should make it clear here and in `Mutable` that this does not affect the values, like `{foo: readonly string[]}`.

sindresorhus (changes requested):

papb:
Hello, nice work, is there a plan to create `SetReadonly` as well?

source/set-mutable.d.ts
```diff
@@ -0,0 +1,37 @@
...
+Use-case: You want to define a single model where the only thing that changes is whether or not some of the keys are mutable.
+
+@see [`Mutable`](source/mutable.d.ts)
```
sindresorhus, line 9: This gives the user no context on "why" it's `@see`. It should explain how it relates. The other way too.

sindresorhus:
> Hello, nice work, is there a plan to create SetReadonly as well?

Sure. Open an issue or PR about it.

source/set-mutable.d.ts
```diff
@@ -0,0 +1,37 @@
...
+// 	c: boolean; // Is now mutable.
+// }
+```
+*/
+export type SetMutable<BaseType, Keys extends keyof BaseType> =
+	// Pick just the keys that are not mutable from the base type.
+	Except<BaseType, Keys> &
+	// Pick the keys that should be mutable from the base type and make them mutable.
+	Mutable<Pick<BaseType, Keys>> extends
+	// If `InferredType` extends the previous, then for each key, use the inferred type key.
```
voxpelli, line 34: There's really no `If` here, right? You use `infer InferredType`  to assign the combination of `Except<BaseType, Keys>` and `Mutable<Pick<BaseType, Keys>>` to a single merged type? So `InferredType` will always get a value.

```suggestion
	// Assigns the combined values to `InferredType` for easy reference
```

source/set-mutable.d.ts
```diff
@@ -0,0 +1,37 @@
...
+	infer InferredType
+		? {[KeyType in keyof InferredType]: InferredType[KeyType]}
```
voxpelli, line 36: What's the reason for doing this step rather than just doing:

```typescript
export type SetMutable<BaseType, Keys extends keyof BaseType> =
  // Pick just the keys that are not mutable from the base type.
  Except<BaseType, Keys> &
  // Pick the keys that should be mutable from the base type and make them mutable.
  Mutable<Pick<BaseType, Keys>>;
```

I guess the reason is that it makes it easier to read what the type actually is? As eg. VS Code Intellisense shows the actual keys and values rather than `Pick<Foo, "a"> & Mutable<Pick<Foo, "b" | "c">>`?

 don't know if this pattern is really something that's used throughout `type-fest` or whether it should be a pattern that's used throughout `type-fest`. What are your thoughts @sindresorhus?

The purist/perfectionist in me says that this is bad and just complicating the type setup to workaround a developer experience thing that should really be improved in TypeScript itself, although the convenient/pragmatist part of me says: This is great, so much more readable!

If we want to double down on this patter, which we for developer experience and consistency's sake should if it should remain, then creating a simple shared "simplifier" type would be preferable:

```typescript
type Simplify<T> = {[KeyType in keyof T]: T[KeyType]};
```

Using such a `Simplify<>` type, `SetMutable` could – while being functionally identical – be simplified to:

```typescript
export type SetMutable<BaseType, Keys extends keyof BaseType> =
  Simplify<
    // Pick just the keys that are not mutable from the base type.
    Except<BaseType, Keys> &
    // Pick the keys that should be mutable from the base type and make them mutable.
    Mutable<Pick<BaseType, Keys>>
  >;

File: source/set-mutable.d.ts [hunk 7]
```diff
@@ -0,0 +1,37 @@
...
+
+type SomeMutable = SetMutable<Foo, 'b' | 'c'>;
+// type SomeMutable = {
+// 	readonly a: number;
+// 	b: readonly string[]; // Was already mutable and still is, type of the property remains unaffected.
+// 	c: boolean; // Is now mutable.
+// }
+```
+*/
+export type SetMutable<BaseType, Keys extends keyof BaseType> =
```
voxpelli, line 29: I'm kind of wondering why we simply doesn't add `Keys extends keyof BaseType = keyof BaseType` to `Mutable<>` and makes this implementation be the future `Mutable<>` where one can optionally pick a subset of keys to make mutable.

Does it make sense to have two so similar ones? I know @mmkal had this same wondering in https://github.com/sindresorhus/type-fest/pull/157#discussion_r531389560 but it was resolved by making all `Set*` require a keys attribute without any discussion on why this one doesn't simply replace `Mutable<>`, I would like to discuss the latter 🙂

test-d/set-mutable.ts
```diff
@@ -0,0 +1,18 @@
...
+declare const variation2: SetMutable<{readonly a: number; readonly b: string; readonly c: boolean}, 'a' | 'b'>;
+expectType<{a: number; b: string; readonly c: boolean}>(variation2);
+
+// Three required remain required.
+declare const variation3: SetMutable<{a: number; b: string; c: boolean}, 'a' | 'b' | 'c'>;
+expectType<{a: number; b: string; c: boolean}>(variation3);
+
+// Fail if type changes even if optional is right.
+declare const variation4: SetMutable<{readonly a: number; b: string; readonly c: boolean}, 'b' | 'c'>;
+expectError<{readonly a: boolean; b: string; c: boolean}>(variation4);
```
voxpelli, line 18: I think you have a copy and paste mistake with the comments here. They are identical to the ones in https://github.com/sindresorhus/type-fest/blob/master/test-d/set-required.ts and doesn't make much sense in this context, making me somewhat confused at first 😉

source/mutable.d.ts
```diff
@@ -3,17 +3,21 @@ Convert an object with `readonly` keys into a mutable object. Inverse of `Readon
...
+@see [`SetMutable`](source/set-mutable.d.ts) – a type that can selectively set some keys as mutable.
+
 @example
 ```
 import {Mutable} from 'type-fest';
 
 type Foo = {
 	readonly a: number;
-	readonly b: string;
+	readonly b: readonly string[];
```
voxpelli, line 14: I think the addition of `readonly string[];` to the example confuses things.

Can that maybe be in a second, more advanced example or as a caveat in the explanatory text?

As people who skim a test could otherwise interpret that this also changes `readonly` values, as why else would it be mentioned there?

voxpelli (changes requested):
Sorry for not reviewing earlier, I had my notifications for this repo setup in a way that made me miss the request for reviews.

My "request changes" isn't for all of my comments, but mostly for things like the erroneous comments in the test file.

As a new co-maintainer on this repo I'm still figuring out some things myself.

test-d/set-mutable.ts
kainiedziela: Oh yeah, my bad! Nice catch 🥇

source/mutable.d.ts
kainiedziela: This was requested by Sindre, but I think we can explain ourselves in the example itself, highlighting what happens and why is that showcased.

kainiedziela: Huh, actually doesn't line 19 kinda explains the reasoning already?

```ts
mutableFoo.b[0] = '3'; //=> Error: Index signature in type 'readonly string[]' only permits reading.ts(2542).
```

source/set-mutable.d.ts
kainiedziela: I think it's done to be consistent with the other `Set-` types and their non-set siblings.

source/set-mutable.d.ts
kainiedziela: This implementation is really just an extension of the other `Set-` types. If it is decided that it should be changed I believe all of them should be updated.

source/mutable.d.ts
kainiedziela: Added an additional clarifying comment.

source/set-mutable.d.ts
voxpelli: Yeah, I guess @sindresorhus can explain 🙂

source/mutable.d.ts
```diff
@@ -3,17 +3,22 @@ Convert an object with `readonly` keys into a mutable object. Inverse of `Readon
...
 };
 
-const mutableFoo: Mutable<Foo> = {a: 1, b: '2'};
+const mutableFoo: Mutable<Foo> = {a: 1, b: ['2']};
 mutableFoo.a = 3;
+// Type of the property remains unaffected
+mutableFoo.b[0] = '3'; // -> Error: Index signature in type 'readonly string[]' only permits reading.ts(2542).
+mutableFoo.b = ['3'];
```
voxpelli, line 21: Maybe something like this?

```suggestion
mutableFoo.b[0] = 'new value'; // -> Will still fail as the value of property "b" is still a readonly type
mutableFoo.b = ['something']; // -> Will work as the "b" property itself is no longer readonly
```

source/set-mutable.d.ts
voxpelli: Right, another question for @sindresorhus then 👍

voxpelli (commented):
Wow, you're fast, I only had time to review https://github.com/sindresorhus/type-fest/pull/153 and then I'm back to this one 😅

Well, here's some follow up thoughts

source/set-mutable.d.ts
sindresorhus: I think the initial reason was consistency. For example, `SetOptional`. We couldn't extend the built-in `Partial`, right?

Thinking more about it now and seeing how the "duplication" continues, I'm also thinking that we should merge them.

Can you think of any downsides?

source/set-mutable.d.ts
sindresorhus: > If we want to double down on this patter, which we for developer experience and consistency's sake should if it should remain, then creating a simple shared "simplifier" type would be preferable:

That sounds like a good idea. I think it's just the case of no one noticing the repetition/duplication.

kainiedziela: Applied.

source/set-mutable.d.ts
kainiedziela: Honestly the `Set-` types are kinda confusing in naming. There's no way to immediately know how `SetRequired` differs from `Required` until you check the docs.

And since there's no `Mutable<T>` built-in type, merging those two is a no-brainer. We just have to document them properly to make it obvious that there's a `Mutable<T, keys?>` option.

kainiedziela: I merged these two types. Take a gander. We can roll back that commit with necessary.

sindresorhus: Continued in https://github.com/sindresorhus/type-fest/issues/171

test-d/set-mutable.ts
```diff
@@ -0,0 +1,18 @@
+import {expectType, expectError} from 'tsd';
```
sindresorhus, line 1: This file should be merged with the `mutable.ts` test.

readme.md
```diff
@@ -61,7 +61,7 @@ Click the type names for complete docs.
 ### Utilities
 
 - [`Except`](source/except.d.ts) - Create a type from an object type without certain keys. This is a stricter version of [`Omit`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-5.html#the-omit-helper-type).
-- [`Mutable`](source/mutable.d.ts) - Convert an object with `readonly` keys into a mutable object. The inverse of `Readonly<T>`.
+- [`Mutable`](source/mutable.d.ts) - Convert an object with `readonly` keys into a mutable object. The inverse of `Readonly<T>`. Alternatively create a type that converts the given keys from `readonly` to mutable.
```
sindresorhus, line 64: I don't think the added sentence here is clear enough.

source/mutable.d.ts
```diff
@@ -1,22 +1,37 @@
 /**
 Convert an object with `readonly` keys into a mutable object. Inverse of `Readonly<T>`.
+Alternatively create a type that converts the given keys from `readonly` to mutable. The remaining keys are kept as is.
 
 This can be used to [store and mutate options within a class](https://github.com/sindresorhus/pageres/blob/4a5d05fca19a5fbd2f53842cbf3eb7b1b63bddd2/source/index.ts#L72), [edit `readonly` objects within tests](https://stackoverflow.com/questions/50703834), and [construct a `readonly` object within a function](https://github.com/Microsoft/TypeScript/issues/24509).
+Or to define a single model where the only thing that changes is whether or not some of the keys are mutable.
```
sindresorhus, line 6: This should be part of the previous sentence.

source/simplify.d.ts
```diff
@@ -0,0 +1,4 @@
+/**
+Flatten the type output for ease of reference.
```
sindresorhus, line 2: "ease of reference"? I think it could more clearly state that it's to improve the type hints shown in editors.

sindresorhus, line 2: "ease of reference"? I think it could more clearly state that it's to improve the type hints shown in editors.

sindresorhus (changes requested):

test-d/set-mutable.ts
kainiedziela: Yeah, I've done that but forgot to remove `set-mutable.ts`. Applied.

readme.md
```diff
@@ -61,7 +61,7 @@ Click the type names for complete docs.
 ### Utilities
 
 - [`Except`](source/except.d.ts) - Create a type from an object type without certain keys. This is a stricter version of [`Omit`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-5.html#the-omit-helper-type).
-- [`Mutable`](source/mutable.d.ts) - Convert an object with `readonly` keys into a mutable object. The inverse of `Readonly<T>`.
+- [`Mutable`](source/mutable.d.ts) - Convert an object with `readonly` keys into a mutable object. The inverse of `Readonly<T>`. Alternatively creates a type that converts the given keys from `readonly` to mutable and the remaining keys are kept as is, as opposed to converting all keys to mutable.
```
sindresorhus, line 64: This is too long. I think you could just integration a mention of support for only applying to the given keys in the first sentence.

source/mutable.d.ts
```diff
@@ -1,22 +1,36 @@
 /**
 Convert an object with `readonly` keys into a mutable object. Inverse of `Readonly<T>`.
+Alternatively create a type that converts the given keys from `readonly` to mutable. The remaining keys are kept as is.
 
-This can be used to [store and mutate options within a class](https://github.com/sindresorhus/pageres/blob/4a5d05fca19a5fbd2f53842cbf3eb7b1b63bddd2/source/index.ts#L72), [edit `readonly` objects within tests](https://stackoverflow.com/questions/50703834), and [construct a `readonly` object within a function](https://github.com/Microsoft/TypeScript/issues/24509).
+This can be used to [store and mutate options within a class](https://github.com/sindresorhus/pageres/blob/4a5d05fca19a5fbd2f53842cbf3eb7b1b63bddd2/source/index.ts#L72), [edit `readonly` objects within tests](https://stackoverflow.com/questions/50703834), and [construct a `readonly` object within a function](https://github.com/Microsoft/TypeScript/issues/24509) or to define a single model where the only thing that changes is whether or not some of the keys are mutable.
```
sindresorhus, line 5:
```suggestion
This can be used to [store and mutate options within a class](https://github.com/sindresorhus/pageres/blob/4a5d05fca19a5fbd2f53842cbf3eb7b1b63bddd2/source/index.ts#L72), [edit `readonly` objects within tests](https://stackoverflow.com/questions/50703834), [construct a `readonly` object within a function](https://github.com/Microsoft/TypeScript/issues/24509), or to define a single model where the only thing that changes is whether or not some of the keys are mutable.
```

source/simplify.d.ts
```diff
@@ -0,0 +1,4 @@
+/**
+Flattens the type output to improve type hints shown in editors.
```
sindresorhus, line 2:
```suggestion
Flatten the type output to improve type hints shown in editors.
```

readme.md
kainiedziela: Take a look now.

papb:
What do you think of also making `Mutable<readonly [string, number]` return `[string, number]`? I needed this recently and had to write my own.

readme.md
```diff
@@ -61,7 +61,7 @@ Click the type names for complete docs.
 ### Utilities
 
 - [`Except`](source/except.d.ts) - Create a type from an object type without certain keys. This is a stricter version of [`Omit`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-5.html#the-omit-helper-type).
-- [`Mutable`](source/mutable.d.ts) - Convert an object with `readonly` keys into a mutable object. The inverse of `Readonly<T>`. Alternatively creates a type that converts the given keys from `readonly` to mutable and the remaining keys are kept as is, as opposed to converting all keys to mutable.
+- [`Mutable`](source/mutable.d.ts) - Strip `readonly` from all or some of an object's keys. The inverse of `Readonly<T>`.
```
sindresorhus, line 64: This makes it sound like it strips `readonly` from the existing type, but it actually creates a new type with those keys stripped. That's an important distinction.

kainiedziela: Take a gander.

kainiedziela:
> What do you think of also making `Mutable<readonly [string, number]` return `[string, number]`? I needed this recently and had to write my own.

This sounds useful to me. @sindresorhus thoughts?

voxpelli:
I think we should open a separate issue / PR for accepting arrays into Mutable, to keep this PR as focused as possible so we can get it reviewed and merged without too much effort 😉

Sorry for lagging with my re-review, I’m a bit behind at work right now.

sindresorhus:
> I think we should open a separate issue / PR for accepting arrays into Mutable, to keep this PR as focused as possible so we can get it reviewed and merged without too much effort 😉

Agreed: https://github.com/sindresorhus/type-fest/issues/176

sindresorhus (approved):

sindresorhus:
@voxpelli LGTY?

sindresorhus:
Actually, I missed that the tests are failing. You need to fix those.

voxpelli (approved):
LGTM 👍

voxpelli:
@sindresorhus Do you want me to merge?

sindresorhus:
Sure, go ahead.

PR #159, opened 2020-11-26
Add `LastArrayElement`, `Split`,  and `Trim` types

### LINKED ISSUE #148: Proposal: Enhanced `document.querySelector` type
See: https://twitter.com/MikeRyanDev/status/1308472279010025477

Or maybe that's too opinionated for `type-fest`, but at least we could add many of the utility methods used which would make it easier to create such a type. See the below code for the utility types.

Code: https://www.typescriptlang.org/play?ts=4.1.0-dev.20200920#code/C4TwDgpgBAymA2BLYAeGUIA9gQHYBMBnKQ4AJ0VwHMAaKAEQ2zyJPMqoD4oBeWJnAWIADACQBvSgDMIZKABUAvhPrLJuGXICqi4VAD8UANry6AOgtwkqLXXqcAulABcxmA4DcAKFCQFAQwBrCAAZf1IUADVuPkiBFmIjJ0NcCAA3WRcoOKxBViNSCmpk7KMABidXHOYhY0KOcwtpTIAlEvkg0PDUFu5XVIyyb19oeQoAWxCIKVRqvOJ66hjs+Nq9CWa5Ft0DBQmpmZRerMjh8FGJlsQqAAtZ1dZFrl4V3ISoMXVNKG2P3bHEOMrrcen1smc-ADxlEHgt2EsXlCDqgocC7lFOJwIdAYOwwABZAD2+EQUkQshhb1qTzo+NhbCKz1i9M+mygITU+LUbIAgjtDCETtjYHiiSSybJCJSao94UyRRQCcTSeSyGhRcqJWrcYqxSqKTrEErxaqonQAORmc2cC0AYmtFqMDqg5uc1uFHWCYVIPJmsnkhOCuGl8wZDQU9Key0Nxv1ZClnq6ESsyBQUIxdHkmKxPnOUAA4hBgN7gABReAQcZ4YAAOX8VZD7yjiM6Jd9ODIAaDadb3Xb-sDeDNLpdNtH7tzfkLZf8AGMbuXK9W6w3InRBVTZYyki8kssvFBD68ZYknAejxeBVBzxfD1V6QU5Wfb0fDEYLGYQnRpyXF1XcLW9YQFE5QOI414vnex6ho+jKNGYbJtBBkGGNOpZzguFb-oBDYtHQ74WF+BZFr+WHLkBIEVI43A3rerhJMKaFkQBK4QFKcxNnKyxoRhf7kQ2KazBaNDWjmIxQHxAEAEIgKxjbUlxLy0RxtTBCAhJSFAAAS8j4iEknAB0VCsfi-hgMhL6GDpekGUZJlmUYkTPpB96bsQakabAkT5rZ-jGUBpnmbRl5eT5zGGX59lgI5zkvq4BnCqZwDzuhKXhcOG4nhJ4VQAAPlAuAAK7wPALxFSV+4Xip+SxbeV7BUerlZbBHC1SFgr5QZMlyZEoE0ZBJwPjSUAfohbWvlASUpbx6V4eyeXZUu0myRRvVUf1Lnst4k7QAAioVsggC0bHFSikaKXwU0LjNS2oExt2sQmmLbQA9AAVG9B5vRJmD1ggEBfS9Xj4BAs7wP4ZDQFIhW4LOwCIISuBQAAjgdZAgDAEAVnDhJqvI52MpwAAUqOHa48gAJSuPth3HYQp1pjmXizojpBQP4Lyk+jmPY8AuNE+aJJpGYABG-i4KkcjcP4Zizv4JUALR8wrc7w4j5oU1AL0vQr3DWSEPKwzcuMGczrPAFAIuc2jGNY6DfNkALlBgIVwB0ELGtazreu6SEACSuAu2WOX5fr9CIGkpss7gbOztbh08-b-PmrOiBkGDEBGLOmA8AARAAjAArGUucOJ72u66FADCacZ1H5tQPg8fc3bOOO+aIuu3zuC2p3IAK7ghIAO7l972m+1JXeI-XMcW9AfBc7bvPJ4Q9sI0jYDOGS8bAArGlK+cGseF7lf6wAChDflkGZmG3V4QA

kainiedziela:
Fixes  #148

### Split

Represents an array of strings split using a passed-in character or character set.

**Use case:** defining the return type of the `String.prototype.split` method.

```TS
declare function split<S extends string, D extends string>(string: S, separator: D): Split<S, D>;

type Item = 'foo' | 'bar' | 'baz' | 'waldo';
const items = 'foo,bar,baz,waldo';
let array: Item[];

array = split(items, ',');
```

### TakeLast

Extracts the type of the last element of an array.

**Use case:** Defining the return type of functions that extract the last element of an array, for example [lodash.last](https://lodash.com/docs/4.17.15#last).

```TS
declare function lastOf<V extends any[], L extends TakeLast<V>>(array: V): L;
const array = ['foo', 2];
typeof lastOf(array); // -> number;
```

### Trim

Removes spaces from the edges of a string.

```TS
Trim<' foo '> // => 'foo';
```


readme.md
```diff
@@ -87,6 +87,7 @@ Click the type names for complete docs.
 - [`Entries`](source/entries.d.ts) - Create a type that represents the type of the entries of a collection.
 - [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`Asyncify`](source/asyncify.d.ts) - Create an async version of the given function type.
+- [`TakeLast`](source/take-last.d.ts) - Extracts the type of the last element of an array.
```
sindresorhus, line 90: Since it only works on arrays, I think this name is too generic.
```suggestion
- [`LastArrayElement`](source/take-last.d.ts) - Extracts the type of the last element of an array.
```

?

readme.md
```diff
@@ -97,6 +98,9 @@ Click the type names for complete docs.
 - [`PascalCase`](ts41/pascal-case.d.ts) – Converts a string literal to pascal-case (`FooBar`)
 - [`SnakeCase`](ts41/snake-case.d.ts) – Convert a string literal to snake-case (`foo_bar`).
 - [`DelimiterCase`](ts41/delimiter-case.d.ts) – Convert a string literal to a custom string delimiter casing.
+- [`QueryResult`](ts41/query-result.d.ts) - Matches element types returned from a query.
```
sindresorhus, line 101: "query" of what? Not clear to the user.

readme.md
```diff
@@ -97,6 +98,9 @@ Click the type names for complete docs.
...
+- [`Split`](ts41/split.d.ts) - Represents an array of strings split using a passed-in character or character set.
+- [`Trim`](ts41/trim.d.ts) - Removes spaces from the edges of a string.
```
sindresorhus, line 103:
```suggestion
- [`Trim`](ts41/trim.d.ts) - Remove leading and trailing spaces from a string.
```

source/take-last.d.ts
```diff
@@ -0,0 +1,18 @@
...
+typeof lastOf(array); // -> number;
+*/
+export type TakeLast<V> =
+    V extends []
+        ? never
+        : V extends [string]
+            ? V[0]
+            : V extends [string, ...infer R]
+                ? TakeLast<R>
+                : never;
```
sindresorhus, line 18: Use tab-indentation.

source/take-last.d.ts
```diff
@@ -0,0 +1,18 @@
+/**
+Extracts the type of the last element of an array.
+
+Use case: Defining the return type of functions that extract the last element of an array, for example [lodash.last](https://lodash.com/docs/4.17.15#last).
+
+@example
+declare function lastOf<V extends any[], L extends TakeLast<V>>(array: V): L;
+const array = ['foo', 2];
+typeof lastOf(array); // -> number;
```
sindresorhus, line 9: Use backticks to wrap this code block. Should also have an import statement.

ts41/query-result.d.ts
```diff
@@ -0,0 +1,66 @@
...
+@example
+declare function querySelector<T extends string>(query: T): QueryResult<T>;
+
+const anchor = querySelector('div.banner > a.call-to-action') //-> HTMLAnchorElement
+const div = querySelector('input, div') //-> HTMLInputElement | HTMLDivElement
+const svg = querySelector('circle[cx="150"]') //-> SVGCircleElement
+const button = querySelector('button#buy-now') //-> HTMLButtonElement
+const paragraph = querySelector('section p:first-of-type'); //-> HTMLParagraphElement
+*/
+export type QueryResult<T extends string> = MatchEachElement<GetElementNames<T>>;
```
sindresorhus, line 23: Should be named `QueryResult`.

And the export should be at the bottom.

ts41/query-result.d.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/** Internal helper for `QueryResult`. Strips a given modifier from a string. */
```
sindresorhus, line 25: No need to prefix with internal helper. That's clear from the context.

test-d/query-result.ts
```diff
@@ -0,0 +1,10 @@
+import {expectType} from 'tsd';
+import {QueryResult} from '../ts41';
+
+declare function querySelector<T extends string>(query: T): QueryResult<T>;
+
+expectType<HTMLAnchorElement>(querySelector('div.banner > a.call-to-action'));
+expectType<HTMLInputElement | HTMLDivElement>(querySelector('input, div'));
+expectType<SVGCircleElement>(querySelector('circle[cx="150"]'));
+expectType<HTMLButtonElement>(querySelector('button#buy-now'));
+expectType<HTMLParagraphElement>(querySelector('section p:first-of-type'));
```
sindresorhus, line 10: I think this type could use some more tests. It's quite complex.

sindresorhus (changes requested):

sindresorhus:
I struggle to see where I would use `TakeLast` in real-world code though? `lodash.last` would already ship the type itself, so not the best argument. Any other use-cases?

sindresorhus:
// @mmkal In case you would be willing to help review.

ts41/query-result.d.ts
kainiedziela: Applied.

test-d/take-last.ts
```diff
@@ -0,0 +1,8 @@
+import {expectType} from 'tsd';
+import {TakeLast} from '..';
+
+declare function lastOf<V extends any[], L extends TakeLast<V>>(array: V): L;
```
mmkal:
```suggestion
declare function lastOf<V extends any[]>(array: V): TakeLast<V>;
```

ts41/query-result.d.ts
kainiedziela: It is named that though? Moved to the end of the file.

readme.md
kainiedziela: Added some clarity.

readme.md
kainiedziela: Applied.

test-d/last-array-element.ts
```diff
@@ -0,0 +1,8 @@
+import {expectType} from 'tsd';
+import {LastArrayElement} from '../ts41';
+
+declare function lastOf<V extends any[], L extends LastArrayElement<V>>(array: V): L;
```
mmkal, line 4:
```suggestion
declare function lastOf<V extends any[]>(array: V): LastArrayElement<V>;
```

mmkal (commented):
I can see valid userland use cases for `TakeLast`, maybe getting the last argument from a function? e.g. `type MyFuncOptions = TakeLast<Parameters<typeof myFunc>>`

source/last-array-element.d.ts
```diff
@@ -0,0 +1,22 @@
+/**
+Extracts the type of the last element of an array.
+
+Use case: Defining the return type of functions that extract the last element of an array, for example [lodash.last](https://lodash.com/docs/4.17.15#last).
+
+@example
+```
+import {LastArrayElement} from 'type-fest';
+
+declare function lastOf<V extends any[], L extends LastArrayElement<V>>(array: V): L;
```
mmkal, line 10:
```suggestion
declare function lastOf<V extends any[]>(array: V): LastArrayElement<V>;
```

source/last-array-element.d.ts
```diff
@@ -0,0 +1,22 @@
...
+const array = ['foo', 2];
+typeof lastOf(array); // -> number;
+```
+*/
+export type LastArrayElement<V> =
```
mmkal, line 15:
```suggestion
export type LastArrayElement<V extends unknown[]> =
```

source/last-array-element.d.ts
```diff
@@ -0,0 +1,22 @@
...
+	V extends []
+		? never
+		: V extends [string]
```
mmkal, line 18: Do you need this check? Why `[string]` specifically?

Could the whole implementation reduce to something like

```typescript
export type LastArrayElement<V extends unknown[]> =
  V extends [infer _, ...infer Tail] ? LastArrayElement<Tail> : never
```

ts41/query-result.d.ts
```diff
@@ -0,0 +1,70 @@
...
+declare function querySelector<T extends string>(query: T): QueryResult<T>;
+
+const anchor = querySelector('div.banner > a.call-to-action') //-> HTMLAnchorElement
+const div = querySelector('input, div') //-> HTMLInputElement | HTMLDivElement
+const svg = querySelector('circle[cx="150"]') //-> SVGCircleElement
+const button = querySelector('button#buy-now') //-> HTMLButtonElement
+const paragraph = querySelector('section p:first-of-type'); //-> HTMLParagraphElement
+```
+*/
+export type QueryResult<T extends string> = MatchEachElement<GetElementNames<T>>;
```
mmkal, line 70: Not a problem if this is ready to go in but this type is much more complex than the string ones so maybe should go in separately after them to leave time for more testing/feedback?

test-d/trim.ts
```diff
@@ -0,0 +1,22 @@
...
+let rightSpaced: {withSpaces: 'bar '; trimmed: 'bar'};
+let onceSpaced: {withSpaces: ' baz '; trimmed: 'baz'};
+let twiceSpaced: {withSpaces: '  waldo  '; trimmed: 'waldo'};
+
+type TrimmedFromLeft = Trim<typeof leftSpaced.withSpaces>;
+type TrimmedFromRight = Trim<typeof rightSpaced.withSpaces>;
+type TrimmedOnce = Trim<typeof onceSpaced.withSpaces>;
+type TrimmedTwice = Trim<typeof twiceSpaced.withSpaces>;
+
+expectType<TrimmedFromLeft>(leftSpaced.trimmed);
```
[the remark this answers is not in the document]
kainiedziela: Great idea! Applied!

test-d/trim.ts
```diff
@@ -1,22 +1,9 @@
...
-expectType<TrimmedTwice>(twiceSpaced.trimmed);
-
-expectError(expectType<TrimmedFromLeft>(rightSpaced.trimmed));
-expectError(expectType<TrimmedFromRight>(onceSpaced.trimmed));
-expectError(expectType<TrimmedOnce>(twiceSpaced.trimmed));
-expectError(expectType<TrimmedTwice>(leftSpaced.trimmed));
+expectType<'foo'>(trim(' foo'));
+expectType<'bar'>(trim('bar '));
+expectType<'baz'>(trim(' baz '));
+expectType<'waldo'>(trim('  waldo  '));
```
mmkal, line 9: Maybe a test with spaces in the middle?

source/last-array-element.d.ts
kainiedziela: Good catch and smart reimplementation. Applied.

ts41/query-result.d.ts
kainiedziela: I wouldn't mind it, I also don't mind handling the arising merge conflicts if #153 gets merged first though.

test-d/last-array-element.ts
```diff
@@ -0,0 +1,8 @@
+import {expectType} from 'tsd';
+import {LastArrayElement} from '../ts41';
+
+declare function lastOf<V extends any[]>(array: V): LastArrayElement<V>;
+const array = ['foo', 2];
+
+expectType<number>(lastOf(array));
+expectType<string>(lastOf(array.reverse()));
```
mmkal, line 8: I'm a bit surprised these tests are passing? Doesn't `array` have type `(string | number)` rather than `[string, number]`? Plus, does `.reverse()` actually reverse tuples in type-land?

test-d/query-result.ts
```diff
@@ -0,0 +1,17 @@
...
+expectType<HTMLDivElement>(querySelector('.menu #leftSide aside div.menu #leftSide aside div'));
+expectType<HTMLButtonElement | HTMLButtonElement | HTMLParagraphElement>(querySelector('button#buy-now, input.input, section>p'));
+expectType<HTMLVideoElement>(querySelector('.main video#trailer'));
+expectType<HTMLTableRowElement>(querySelector('table.main tr'));
+expectType<HTMLStyleElement>(querySelector('html style'));
+expectType<HTMLSourceElement>(querySelector('.main source'));
+expectType<HTMLScriptElement>(querySelector('head script[src="script-source"]'));
```
mmkal, line 17: Maybe a test for unknown type like `querySelector('.some-class')`?

ts41/query-result.d.ts
```diff
@@ -0,0 +1,70 @@
...
+type StripModifiers<V extends string> = StripModifier<StripModifier<StripModifier<StripModifier<V, '.'>, '#'>, '['>, ':'>;
+
+/** Takes the last element after the given token. */
+type LastArrayElementAfterToken<V extends string, T extends string> = StripModifiers<LastArrayElement<Split<Trim<V>, T>>>;
+
+/** Extracts the last element from a CSS selector string. */
+type GetLastElementName<V extends string> = LastArrayElementAfterToken<LastArrayElementAfterToken<V, ' '>, '>'>;
+
+/** Returns an array of element names with stripped modifiers. */
+type GetEachElementName<V, L extends string[] = []> =
```
mmkal, line 18: This type is a little confusing to me. Again it feels like the `[string]` check isn't necessary since it's covered by `[string, ...infer Tail]` but I could be wrong. Maybe some examples in the jsdoc could help either way. I know not exposed but I suspect this implementation will be used as a reference.

mmkal (commented):
Sorry to keep submitting separate reviews! Won't have much chance to look today after this.

ts41/query-result.d.ts
mmkal: Ok - I'm also fine dealing with them, it shouldn't be too hard.

mmkal, line 70: Q re the implementation - why do we need to match each element? Don't we just care about the last tag?

Also, it might be nice to expose a type for getting the tag name (as well as the tag type), which could be useful for libraries like [mithril](https://npmjs.com/package/mithril) which use the same format for element creation, and could have an internal mapping from tag name -> allowed attributes 

i.e.
```typescript
type B = QueryResultTag<'button.foo'> // 'button'
```

test-d/last-array-element.ts
kainiedziela: It's not a tuple, it's an array. Also the entire point of `LastArrayElement` is to get te actual type of the last element. Yes, the array's type is `(string | number)[]` but the type of `array[array.length - 1]` is what we want to extract here, which in this case is `number`.

mmkal: Right, but isn't `LastArrayElement<(string | number)[]>` equal to `string | number`? This test suggests it's `string` which seems odd to me. Not necessarily something wrong with the type itself, but the test (and maybe even `expectType`)

kainiedziela: Okay, so you inspired to take a more thourough look at this and my VS Code's TS Server seems to think that `lastOf(array)` returns `never`. That would indeed indicate that `expectType` isn't working as intended I think. I need some more time on this type then.

kainiedziela: It looks like `LastArrayElement` works only on specifically typed arrays:

```TS
const array: ['foo', 2, 'bar'] = ['foo', 2, 'bar'];
lastOf(array); // -> is of type 'bar'

const integers: [0, 1, 2] = [0, 1, 2];
lastOf(array); // -> is of type 2

const elements: [HTMLDivElement, HTMLSpanElement] = [querySelector('div'), querySelector('span')];
lastOf(elements); // -> is of type HTMLSpanElement
```

Otherwise it just returns `never`. Not that helpful of a type to be honest, but necessary for `QueryResult`. I think we might reconsider if it should be exposed.

mmkal: Makes sense. I guess this is another instance of the tsd bug mentioned in https://github.com/sindresorhus/type-fest/pull/153#discussion_r530733893

But also the LastArrayElement type could be improved to fix - maybe instead of the last `: never` it should be `: V[number]`. Then this test would have caught it.

I agree that it simplifies things to not expose the type at all, though.

Generally specifically typed arrays are called tuples, and the slightly easier way to declare them is to use `as const`:

```typescript
const array = ['foo', 2] as const
```

But yeah generally people have arrays, not tuples, available. So this test is probably a bit unrealistic.

fregante:
I don’t think QuerySelector needs to be part of type-fest. The original author of those types already published and improved on them: https://github.com/g-plane/typed-query-selector

sindresorhus:
I agree with @fregante. That package does it better than what we're doing here.

I still think the other types added here are useful though.

@kainiedziela Would you be able to remove the `QueryResult` type, but keep the others? And also link to https://github.com/g-plane/typed-query-selector in a `Related` section in the readme.

kainiedziela:
There you go!

readme.md
```diff
@@ -137,7 +137,7 @@ Click the type names for complete docs.
 
 ### Related
 
-- [typed-query-selector](https://github.com/g-plane/typed-query-selector) - Enhances `querySelector` and `querySelectorAll` with a template literal type that matches element types returned from an HTML element query selector.
+- [typed-query-selector](https://github.com/g-plane/typed-query-selector) - Enhances `document.querySelector` and `document.querySelectorAll` with a template literal type that matches element types returned from a HTML element query selector.
```
papb, line 140: I think *[...] an HTML [...]* is the correct form...

https://english.stackexchange.com/questions/1016/do-you-use-a-or-an-before-acronyms-initialisms

Issue #160, opened 2020-11-27
type-fest@0.20.1 typescript@4.1.2 SnakeCase has bug


kahirokunn:
type-fest@0.20.1
typescript@4.1.2

SnakeCase<'helloWorld'> expect hello_world, but got helloWorld. 😢 

```typescript
import type { SnakeCase } from 'type-fest';

const snakeFromCamel: SnakeCase<'helloWorld'> = 'hello_world';
const abcd: SnakeCase<'aBCD'> = 'a_b_c_d';
```

<img width="619" alt="スクリーンショット 2020-11-27 10 58 18" src="https://user-images.githubusercontent.com/22343391/100402387-76714480-309f-11eb-9891-fe39b28114eb.png">

PR #161, opened 2020-11-27
Fix `SnakeCase` type issue

kahirokunn:
I just fixed this bug.
Fixes https://github.com/sindresorhus/type-fest/issues/160


voxpelli:
Yikes, that's awkward, forgetting the one char that my last name starts with 🙈 Thanks for finding and fixing @kahirokunn!

Issue #150, opened 2020-11-25
Proposal: `MergeDeep` type


sindresorhus:
See https://github.com/sindresorhus/type-fest/pull/94 for the initial attempt.

Isabellarossi:
Uppercase
Lowercase
Capitalize
_Uncapitalize****_

sindresorhus:
If anyone wants to work on this, see https://github.com/sindresorhus/type-fest/pull/164 and the feedback given there.

sindresorhus:
We have this now: https://github.com/sindresorhus/type-fest/blob/main/source/merge-deep.d.ts

PR #164, opened 2020-11-30
Add `MergeDeep` type

kainiedziela:
### `MergeDeep`

Merge two types deeply into a new type. Keys of the second type overrides keys of the first type.

**Use-cases:**
- where intersection types would result in a prop having `never` – if two different types for the same property are used, intersection types assume it's type is `never`. `MergeDeep` allows for safe intersection of two types and having the second passed in type have precedence, hence resulting in no unwanted `never` prop types.
- merging complex, multi-level interfaces that share common props across different levels

@example
```ts
import {MergeDeep} from 'type-fest';

type Foo = {
	a: number;
	b: string;
	c: {
		d: number;
		e: string;
	}
};
type Bar = {
	b: number;
	c: {
		e: number;
	}
};

const foobar: MergeDeep<Foo, Bar> = {
	a: 1,
	b: 2,
	c: {
		d: 3,
		e: 4,
	}
};
const intersectFooWithBar: Foo & Bar = {b: 2}; // -> Error: Type 'number' is not assignable to type 'never'.
```


source/merge-deep.d.ts
```diff
@@ -0,0 +1,55 @@
...
+const intersectFooWithBar: Foo & Bar = {b: 2}; // -> Error: Type 'number' is not assignable to type 'never'.
+```
+*/
+export type MergeDeep<FirstType, SecondType> = {
+	[KeyType in keyof (FirstType & SecondType)]:
+		KeyType extends keyof SecondType
+			? KeyType extends keyof FirstType
+				? FirstType[KeyType] extends object
+					? SecondType[KeyType] extends object
+						? FirstType[KeyType] extends Array<infer FirstArrayElement>
```
mmkal, line 44: Do tuples need special handling? Or are they covered by `Array<infer ...>`?

test-d/merge-deep.ts
```diff
@@ -0,0 +1,37 @@
...
+    baz: Array<{
+        yowza: string;
+        george: number;
+    }>;
+    waldo: number;
+    fred: {
+        ['1']: number;
+        ['2']: number[];
+    };
+}>(foobar);
```
mmkal, line 37: Should probably have some tests for arrays and tuples (of different sizes), optional properties, unions, intersections, readonly properties, readonly arrays.

readme.md
```diff
@@ -63,6 +63,7 @@ Click the type names for complete docs.
 - [`Except`](source/except.d.ts) - Create a type from an object type without certain keys. This is a stricter version of [`Omit`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-5.html#the-omit-helper-type).
 - [`Mutable`](source/mutable.d.ts) - Convert an object with `readonly` keys into a mutable object. The inverse of `Readonly<T>`.
 - [`Merge`](source/merge.d.ts) - Merge two types into a new type. Keys of the second type overrides keys of the first type.
+- [`MergeDeep`](source/merge-deep.d.ts) - Merge two types deeply into a new type. Keys of the second type overrides keys of the first type.
```
mmkal, line 66: Should we have a runtime reference implementation we're basing it on? Are we basically trying to type `lodash.merge` or is there another?

source/merge-deep.d.ts
```diff
@@ -0,0 +1,55 @@
...
+
+Use-cases:
+- where intersection types would result in a prop having `never` – if two different types for the same property are used, intersection types assume it's type is `never`. `MergeDeep` allows for safe intersection of two types and having the second passed in type have precedence, hence resulting in no unwanted `never` prop types.
+- merging complex, multi-level interfaces that share common props across different levels
+
+@example
+```
+import {MergeDeep} from 'type-fest';
+
+type Foo = {
```
mmkal, line 12: Might be useful to have a real world scenario/something that shows intuitively how the right-hand side "wins" and why that's probably what we'd want?

sindresorhus:
The docs should be explicit about collection types it handles and how. I see array is supported by `Set` and `Map` are not.

readme.md
kainiedziela: Not sure what you mean here. You'd like to link to `lodash.merge` in the description?

source/merge-deep.d.ts
kainiedziela: Is that really necessary? In all JS syntax right-hand "wins" – see `Object.assign(foo, bar)` or the spread operator `{...foo, ...bar}`. I think it's pretty intuitive for JS/TS devs.

If you have an idea how to succintly explain this I'm not against including it in the docs.

kainiedziela:
@mmkal @sindresorhus I have rewritten `MergeDeep` to handle edge cases and test for them. Take a look.

readme.md
mmkal: Yeah maybe a link would make sense. I've always found it difficult to reason about deep merges, they're complex. e.g. what happens to arrays? What about arrays with complex objects? Having a well established "target" might help make some of those decisions, and give a hint to users how it works.

source/merge-deep.d.ts
mmkal: I think in terms of description what you have now is good. I meant more that the example should have fields like name/address etc rather than a b c

test-d/merge-deep.ts
```diff
@@ -0,0 +1,69 @@
...
+expectAssignable<object>(foobar.waldo.object);
+expectType<string>(foobar.waldo.string);
+expectType<Partial<string>>(foobar.waldo.partialString);
+expectType<number>(foobar.waldo.number);
+expectType<boolean>(foobar.waldo.boolean);
+expectType<symbol>(foobar.waldo.symbol);
+expectType<null>(foobar.waldo.null);
+expectType<undefined>(foobar.waldo.undefined);
+expectAssignable<Map<string, string>>(foobar.waldo.map);
+expectAssignable<Set<string>>(foobar.waldo.set);
```
mmkal, line 62: These tests are great! Do we know what's supposed to happen if you merge two maps e.g.

```typescript
MergeDeep<{ foo: Map<string, { bar: number }> }, { foo: Map<string, { baz: boolean }> }>
```

readme.md
kainiedziela: Oh, gotcha. I basically tried to replicate our other `-Deep` types in how this is reasoned about. Just go as deep as possible and when it longer makes sense to go deeper, apply right-hand side type. Not based on lodash, but that might be a mistake I suppose.

source/merge-deep.d.ts
kainiedziela: Oh, ok. I updated the example with a use case I found on lodash issues.

test-d/merge-deep.ts
kainiedziela: Right-hand side map wins with no recursion. So in your example the resulting type would be:

```ts
{ foo: Map<string, { baz: boolean }> }
```

kainiedziela:
Okay, since my local `npm run test` seemed to report no errors I didn't catch this earlier (and neither did the CI, btw – check [commit ab792c0](https://github.com/sindresorhus/type-fest/pull/164/commits/ab792c098bcd7e5f8cae2e77d778340d40a0c111)) *but* the reimplementation was faulty.

@sindresorhus @mmkal check it again please, the tests are unchanged. I know it seems hacky but that was the single only way of making both arrays and tuples work. Otherwise `[KeyType in keyof (First & Second)]` seems wholly incompatible with those types and with how [mapping over them](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#mapped-types-on-tuples-and-arrays) works.

kainiedziela:
Well, actually, now I have no idea what's going on – nothing is failing locally. Can someone check this for me, please?

test-d/merge-deep.ts
sindresorhus: I think this should be documented.

sindresorhus:
Should this also use the `Simplify` type being added in #157?

readme.md
sindresorhus: I don't think we should mention `lodahs.merge` or base it on it, but in (not here) `merge-deep.d.ts`, we should clearly document how the merging works regarding arrays/maps/sets/etc.

source/merge-deep.d.ts
```diff
@@ -0,0 +1,77 @@
...
+/**
+Types that will cause the right-hand side of the merge operation to always win with no further recursion.
+*/
+type Unmergeable =
+	| Primitive
+	| ((...arguments: any[]) => unknown)
+	| ReadonlyMap<any, any>
+	| ReadonlySet<any>
+	| Map<any, any>
+	| Set<any>;
```
sindresorhus, line 12: Could we use `unknown` here?

source/merge-deep.d.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+Merge two types deeply into a new type. Keys of the second type overrides keys of the first type.
+Recursively merges own and inherited enumerable string keyed properties of source types into the destination type.
+
+Use-cases:
+- Where intersection types would result in a property having `never`. If two different types for the same property are used, intersection types assume its type is `never`. `MergeDeep` allows for safe intersection of two types and having the second passed in type have precedence, hence resulting in no unwanted `never` prop types.
```
sindresorhus, line 19:
```suggestion
- Where intersection types would result in a property having `never`. If two different types for the same property are used, intersection types assume its type is `never`. `MergeDeep` allows for safe intersection of two types and having the second passed in type take precedence, resulting in no unwanted `never` property types.
```

source/merge.d.ts
```diff
@@ -3,6 +3,8 @@ import {Except} from './except';
 /**
 Merge two types into a new type. Keys of the second type overrides keys of the first type.
 
+Used in cases where intersection types would result in a prop having `never` – if two different types for the same property are used, intersection types assume it's type is `never`. `MergeDeep` allows for safe intersection of two types and having the second passed in type have precedence, hence resulting in no unwanted `never` prop types.
```
sindresorhus: I don't think this text makes it immediately clear enough that you're talking about a different type. Use `@see`.

source/merge-deep.d.ts
```diff
@@ -0,0 +1,77 @@
...
+- Merging complex, multi-level interfaces that share common properties across different levels.
+
```
sindresorhus, line 21: Add a `@see` for the `Merge` type.

sindresorhus (changes requested):

sindresorhus:
For https://github.com/sindresorhus/type-fest/pull/164#discussion_r544089215, I meant instead of the `any`'s.

kainiedziela:
> For [#164 (comment)](https://github.com/sindresorhus/type-fest/pull/164#discussion_r544089215), I meant instead of the `any`'s.

Yeah, that makes sense... Why though? Does it make any difference?

sindresorhus:
> Yeah, that makes sense... Why though? Does it make any difference?

That's what I'm asking. If it doesn't change the behavior, I prefer `unknown` as that's what I use anytime I can, so basically for consistency.

sindresorhus:
> Should this also use the Simplify type being added in #157?

^ @kainiedziela

sindresorhus:
> Well, actually, now I have no idea what's going on – nothing is failing locally. Can someone check this for me, please?

Tests are failing locally for me. Try reinstalling `node_modules`.

sindresorhus:
@kainiedziela Would you be able to finish up this PR? 🙏

kainiedziela:
I pushed something that I thought might fix some or all of the compiler issues, but still something's not clicking. Honestly I think I won't be able to finish this. Feel free to close it or defer this PR to someone more capable.

sindresorhus:
https://github.com/sindresorhus/type-fest/issues/150

PR #167, opened 2020-12-10
Add `UnionToTuple` type

joaopaulobdac:
_This one is controversial, but I and a lot of other people have run into this..._

Convert an explicit union type to an tuple.

The elements order of the tuple and of the union will be the same. Because of that, you should not use this if the union is computed instead of manually declared, because the TypeScript compiler does not guarantee that the order of the union elements will always be the same.

Inspired by [this issue in the TypeScript repo](https://github.com/microsoft/TypeScript/issues/13298).

Use-case: You need to do validation but you don't have an enum, only an union type. Similar to the [`Extract` example in the readme](https://typescript-play.js.org/?target=6#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXzSwEdkQBJYACgEoAueVZAWwCMQYBuAKDDwGcM8MgBF4AXngBlAJ6scESgHIRi6ty5ZUGdoihgEABXZ888AN5d48ANoiAuvUat23K6ihMQ9ATE0BzV3goPy8GZjZOLgBfLi4Aejj4AEEICBwAdz54MAALKFQQ+BxEeAAHY1NgKAwoIKy0grr4DByEUpgccpgMaXgAaxBerCzi+B9-ZulygDouFHRsU1z8kKMYE1RhaqgAHkt4AHkWACt4EAAPbVRgLLWNgBp9gGlBs8uQa6yAUUuYPQwdgNpKM7nh7mMML4CgA+R5WABqUAgpDeVxuhxO1he0jsXGh8EoOBO9COx3BQPo2PBADckaR6IjkSA6PBqTgsMBzPsicdrEC7OJWXSQNwYvFEgAVTS9JLXODpeDpKBZFg4GCoWa8VACIJykAKiQWKy2YQOAioYikCg0OEMDyhRSy4DyxS24KhAAMjyi6gS8AAwjh5OD0iBFHAkJoEOksC1mnkMJq8gUQKDNttKPlnfrwYp3J5XfBHXqoKpfYkAOI4ansTxaeDADmoRSCCBYAbxhC6TDx6rwYHIRX5bScjA4bLJwoDmDwDkfbA9JMrVMVdM1TN69LgkTgwgkchUahqIA), but in that case the `changePersonData` function does not validate the `key` parameter at runtime, making it only safe inside TypeScript. Using a tuple and `includes` for validation gives us runtime safety for that kind of function.

```ts
type Union = 'A' | 'B';

const tuple: UnionToTuple<Union> = ['A', 'B'];
```

```ts
// This type may come from a library that you are using
type ActionsInSomeLibrary = 'create' | 'read' | 'update' | 'delete' | 'aggregate';

type AllowedActions = Exclude<ActionsInSomeLibrary, 'aggregate'>;

const actions: UnionToTuple<AllowedActions> = ['create', 'read', 'update', 'delete'];

actions.includes(requestedAction) // Validate
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -80,7 +80,7 @@ Click the type names for complete docs.
 - [`ConditionalPick`](source/conditional-pick.d.ts) - Like `Pick` except it selects properties from a shape where the values extend the given `Condition` type.
 - [`ConditionalExcept`](source/conditional-except.d.ts) - Like `Omit` except it removes properties from a shape where the values extend the given `Condition` type.
 - [`UnionToIntersection`](source/union-to-intersection.d.ts) - Convert a union type to an intersection type.
-- [`UnionToTuple`](source/union-to-tuple.d.ts) - Convert an explicit union type to a tuple.
+- [`UnionToTuple`](source/union-to-tuple.d.ts) - Convert a user defined union type to a tuple type. Order can not be guaranteed on computed union types.
 - [`Stringified`](source/stringified.d.ts) - Create a type with the keys of the given type changed to `string` type.
 - [`FixedLengthArray`](source/fixed-length-array.d.ts) - Create a type that represents an array of the given type and length.
 - [`IterableElement`](source/iterable-element.d.ts) - Get the element type of an `Iterable`/`AsyncIterable`. For example, an array or a generator.
--- a/readme.md
+++ b/readme.md
@@ -80,7 +80,7 @@ Click the type names for complete docs.
 - [`ConditionalPick`](source/conditional-pick.d.ts) - Like `Pick` except it selects properties from a shape where the values extend the given `Condition` type.
 - [`ConditionalExcept`](source/conditional-except.d.ts) - Like `Omit` except it removes properties from a shape where the values extend the given `Condition` type.
 - [`UnionToIntersection`](source/union-to-intersection.d.ts) - Convert a union type to an intersection type.
-- [`UnionToTuple`](source/union-to-tuple.d.ts) - Convert an explicit union type to an tuple.
+- [`UnionToTuple`](source/union-to-tuple.d.ts) - Convert an explicit union type to a tuple.
 - [`Stringified`](source/stringified.d.ts) - Create a type with the keys of the given type changed to `string` type.
 - [`FixedLengthArray`](source/fixed-length-array.d.ts) - Create a type that represents an array of the given type and length.
 - [`IterableElement`](source/iterable-element.d.ts) - Get the element type of an `Iterable`/`AsyncIterable`. For example, an array or a generator.
--- a/readme.md
+++ b/readme.md
@@ -80,6 +80,7 @@ Click the type names for complete docs.
 - [`ConditionalPick`](source/conditional-pick.d.ts) - Like `Pick` except it selects properties from a shape where the values extend the given `Condition` type.
 - [`ConditionalExcept`](source/conditional-except.d.ts) - Like `Omit` except it removes properties from a shape where the values extend the given `Condition` type.
 - [`UnionToIntersection`](source/union-to-intersection.d.ts) - Convert a union type to an intersection type.
+- [`UnionToTuple`](source/union-to-tuple.d.ts) - Convert an explicit union type to an tuple.
 - [`Stringified`](source/stringified.d.ts) - Create a type with the keys of the given type changed to `string` type.
 - [`FixedLengthArray`](source/fixed-length-array.d.ts) - Create a type that represents an array of the given type and length.
 - [`IterableElement`](source/iterable-element.d.ts) - Get the element type of an `Iterable`/`AsyncIterable`. For example, an array or a generator.
--- a/source/union-to-tuple.d.ts
+++ b/source/union-to-tuple.d.ts
@@ -1,13 +1,13 @@
 import {UnionToIntersection} from './union-to-intersection';
 
 /**
-Convert an explicit union type to an tuple.
+Convert an explicit union type to a tuple.
 
-The elements order of the tuple and of the union will be the same. Because of that, you should not use this if the union is computed instead of manually declared, because the TypeScript compiler does not guarantee that the order of the union elements will always be the same.
+The element order of the tuple and of the union will be the same. Because of that, you should not use this if the union is computed instead of manually declared, because the TypeScript compiler does not guarantee that the order of such union elements in that case.
 
 Inspired by [this issue in the TypeScript repo](https://github.com/microsoft/TypeScript/issues/13298).
 
-Use-case: You need to do validation but you don't have an enum, only an union type. Similar to the [`Extract` example in the readme](https://typescript-play.js.org/?target=6#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXzSwEdkQBJYACgEoAueVZAWwCMQYBuAKDDwGcM8MgBF4AXngBlAJ6scESgHIRi6ty5ZUGdoihgEABXZ888AN5d48ANoiAuvUat23K6ihMQ9ATE0BzV3goPy8GZjZOLgBfLi4Aejj4AEEICBwAdz54MAALKFQQ+BxEeAAHY1NgKAwoIKy0grr4DByEUpgccpgMaXgAaxBerCzi+B9-ZulygDouFHRsU1z8kKMYE1RhaqgAHkt4AHkWACt4EAAPbVRgLLWNgBp9gGlBs8uQa6yAUUuYPQwdgNpKM7nh7mMML4CgA+R5WABqUAgpDeVxuhxO1he0jsXGh8EoOBO9COx3BQPo2PBADckaR6IjkSA6PBqTgsMBzPsicdrEC7OJWXSQNwYvFEgAVTS9JLXODpeDpKBZFg4GCoWa8VACIJykAKiQWKy2YQOAioYikCg0OEMDyhRSy4DyxS24KhAAMjyi6gS8AAwjh5OD0iBFHAkJoEOksC1mnkMJq8gUQKDNttKPlnfrwYp3J5XfBHXqoKpfYkAOI4ansTxaeDADmoRSCCBYAbxhC6TDx6rwYHIRX5bScjA4bLJwoDmDwDkfbA9JMrVMVdM1TN69LgkTgwgkchUahqIA), but in that case the `changePersonData` function does not validate the `key` parameter at runtime, making it only safe inside TypeScript. Using a tuple and `includes` for validation gives us runtime safety for that kind of function.
+Use-case: You need to do validation but you don't have an enum, only an union type. Similar to the [`Extract` example in the readme](https://typescript-play.js.org/?target=6#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXzSwEdkQBJYACgEoAueVZAWwCMQYBuAKDDwGcM8MgBF4AXngBlAJ6scESgHIRi6ty5ZUGdoihgEABXZ888AN5d48ANoiAuvUat23K6ihMQ9ATE0BzV3goPy8GZjZOLgBfLi4Aejj4AEEICBwAdz54MAALKFQQ+BxEeAAHY1NgKAwoIKy0grr4DByEUpgccpgMaXgAaxBerCzi+B9-ZulygDouFHRsU1z8kKMYE1RhaqgAHkt4AHkWACt4EAAPbVRgLLWNgBp9gGlBs8uQa6yAUUuYPQwdgNpKM7nh7mMML4CgA+R5WABqUAgpDeVxuhxO1he0jsXGh8EoOBO9COx3BQPo2PBADckaR6IjkSA6PBqTgsMBzPsicdrEC7OJWXSQNwYvFEgAVTS9JLXODpeDpKBZFg4GCoWa8VACIJykAKiQWKy2YQOAioYikCg0OEMDyhRSy4DyxS24KhAAMjyi6gS8AAwjh5OD0iBFHAkJoEOksC1mnkMJq8gUQKDNttKPlnfrwYp3J5XfBHXqoKpfYkAOI4ansTxaeDADmoRSCCBYAbxhC6TDx6rwYHIRX5bScjA4bLJwoDmDwDkfbA9JMrVMVdM1TN69LgkTgwgkchUahqIA), but in that case the `changePersonData` function does not validate the `key` parameter at runtime, making it only safe inside TypeScript. Using a tuple and `Array#includes` for validation gives us runtime safety for that kind of function.
 
 @example
 ```
@@ -18,7 +18,7 @@ const tuple: UnionToTuple<Union> = ['A', 'B'];
 
 @example
 ```
-// This type may come from a library that you are using
+// This type may come from a library that you are using.
 type ActionsInSomeLibrary = 'create' | 'read' | 'update' | 'delete' | 'aggregate';
 
 type AllowedActions = Exclude<ActionsInSomeLibrary, 'aggregate'>;
--- a/source/union-to-tuple.d.ts
+++ b/source/union-to-tuple.d.ts
@@ -3,7 +3,7 @@ import {UnionToIntersection} from './union-to-intersection';
 /**
 Convert an explicit union type to a tuple.
 
-The element order of the tuple and of the union will be the same. Because of that, you should not use this if the union is computed instead of manually declared, because the TypeScript compiler does not guarantee that the order of such union elements in that case.
+The element order of the tuple and of the union will be the same. Because of that, you should not use this if the union is computed instead of manually declared, because the TypeScript compiler does not guarantee the order of such union elements in that case.
 
 Inspired by [this issue in the TypeScript repo](https://github.com/microsoft/TypeScript/issues/13298).
 
--- a/source/union-to-tuple.d.ts
+++ b/source/union-to-tuple.d.ts
@@ -11,6 +11,8 @@ Use-case: You need to do validation but you don't have an enum, only an union ty
 
 @example
 ```
+import {UnionToTuple} from 'type-fest';
+
 type Union = 'A' | 'B';
 
 const tuple: UnionToTuple<Union> = ['A', 'B'];
@@ -18,12 +20,17 @@ const tuple: UnionToTuple<Union> = ['A', 'B'];
 
 @example
 ```
+import {UnionToTuple} from 'type-fest';
+
 // This type may come from a library that you are using.
 type ActionsInSomeLibrary = 'create' | 'read' | 'update' | 'delete' | 'aggregate';
 
 type AllowedActions = Exclude<ActionsInSomeLibrary, 'aggregate'>;
 
 const actions: UnionToTuple<AllowedActions> = ['create', 'read', 'update', 'delete'];
+
+// Represents an parameter of the API.
+const requestedAction = 'create';
 
 actions.includes(requestedAction) // Validate
 ```
--- a/source/union-to-tuple.d.ts
+++ b/source/union-to-tuple.d.ts
@@ -1,0 +1,41 @@
+import {UnionToIntersection} from './union-to-intersection';
+
+/**
+Convert an explicit union type to an tuple.
+
+The elements order of the tuple and of the union will be the same. Because of that, you should not use this if the union is computed instead of manually declared, because the TypeScript compiler does not guarantee that the order of the union elements will always be the same.
+
+Inspired by [this issue in the TypeScript repo](https://github.com/microsoft/TypeScript/issues/13298).
+
+Use-case: You need to do validation but you don't have an enum, only an union type. Similar to the [`Extract` example in the readme](https://typescript-play.js.org/?target=6#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXzSwEdkQBJYACgEoAueVZAWwCMQYBuAKDDwGcM8MgBF4AXngBlAJ6scESgHIRi6ty5ZUGdoihgEABXZ888AN5d48ANoiAuvUat23K6ihMQ9ATE0BzV3goPy8GZjZOLgBfLi4Aejj4AEEICBwAdz54MAALKFQQ+BxEeAAHY1NgKAwoIKy0grr4DByEUpgccpgMaXgAaxBerCzi+B9-ZulygDouFHRsU1z8kKMYE1RhaqgAHkt4AHkWACt4EAAPbVRgLLWNgBp9gGlBs8uQa6yAUUuYPQwdgNpKM7nh7mMML4CgA+R5WABqUAgpDeVxuhxO1he0jsXGh8EoOBO9COx3BQPo2PBADckaR6IjkSA6PBqTgsMBzPsicdrEC7OJWXSQNwYvFEgAVTS9JLXODpeDpKBZFg4GCoWa8VACIJykAKiQWKy2YQOAioYikCg0OEMDyhRSy4DyxS24KhAAMjyi6gS8AAwjh5OD0iBFHAkJoEOksC1mnkMJq8gUQKDNttKPlnfrwYp3J5XfBHXqoKpfYkAOI4ansTxaeDADmoRSCCBYAbxhC6TDx6rwYHIRX5bScjA4bLJwoDmDwDkfbA9JMrVMVdM1TN69LgkTgwgkchUahqIA), but in that case the `changePersonData` function does not validate the `key` parameter at runtime, making it only safe inside TypeScript. Using a tuple and `includes` for validation gives us runtime safety for that kind of function.
+
+@example
+```
+type Union = 'A' | 'B';
+
+const tuple: UnionToTuple<Union> = ['A', 'B'];
+```
+
+@example
+```
+// This type may come from a library that you are using
+type ActionsInSomeLibrary = 'create' | 'read' | 'update' | 'delete' | 'aggregate';
+
+type AllowedActions = Exclude<ActionsInSomeLibrary, 'aggregate'>;
+
+const actions: UnionToTuple<AllowedActions> = ['create', 'read', 'update', 'delete'];
+
+actions.includes(requestedAction) // Validate
+```
+*/
+export type UnionToTuple<Union> =
+    UnionToIntersection<
+        // Distributive conditional trick.
+        // See the source for the `UnionToIntersection` type for more details.
+        Union extends unknown
+        ? (distributed: Union) => void
+        : never
+    > extends ((merged: infer Intersection) => void)
+        // Transforms ('A' | 'B') into [[[], "A"], "B"], but destructures the arrays
+        ? readonly [...UnionToTuple<Exclude<Union, Intersection>>, Intersection]
+        : [];
--- a/source/union-to-tuple.d.ts
+++ b/source/union-to-tuple.d.ts
@@ -1,7 +1,7 @@
 import {UnionToIntersection} from './union-to-intersection';
 
 /**
-Convert an explicit union type to a tuple.
+Convert a user defined union type to a tuple type.
 
 The element order of the tuple and of the union will be the same. Because of that, you should not use this if the union is computed instead of manually declared, because the TypeScript compiler does not guarantee the order of such union elements in that case.
 
--- a/test-d/union-to-tuple.ts
+++ b/test-d/union-to-tuple.ts
@@ -1,0 +1,6 @@
+import {expectAssignable} from 'tsd';
+import {UnionToTuple} from '..';
+
+declare type Union = 'A' | 'B';
+declare const tuple: UnionToTuple<Union>;
+expectAssignable<readonly ['A', 'B']>(tuple);
--- a/base.d.ts
+++ b/base.d.ts
@@ -25,6 +25,7 @@ export {ConditionalExcept} from './source/conditional-except';
 export {ConditionalKeys} from './source/conditional-keys';
 export {ConditionalPick} from './source/conditional-pick';
 export {UnionToIntersection} from './source/union-to-intersection';
+export {UnionToTuple} from './source/union-to-tuple';
 export {Stringified} from './source/stringified';
 export {FixedLengthArray} from './source/fixed-length-array';
 export {IterableElement} from './source/iterable-element';


sindresorhus (approved):

readme.md
```diff
@@ -80,6 +80,7 @@ Click the type names for complete docs.
 - [`ConditionalPick`](source/conditional-pick.d.ts) - Like `Pick` except it selects properties from a shape where the values extend the given `Condition` type.
 - [`ConditionalExcept`](source/conditional-except.d.ts) - Like `Omit` except it removes properties from a shape where the values extend the given `Condition` type.
 - [`UnionToIntersection`](source/union-to-intersection.d.ts) - Convert a union type to an intersection type.
+- [`UnionToTuple`](source/union-to-tuple.d.ts) - Convert an explicit union type to a tuple.
```
voxpelli, line 83: `UnionToIntersection` says `Convert a union type to an intersection type.`

I would expect this one to say something like:

```suggestion
- [`UnionToIntersection`](source/union-to-intersection.d.ts) - Convert a union type to an intersection type.
- [`UnionToTuple`](source/union-to-tuple.d.ts) - Convert a user defined union type to a tuple type. Order can not be guaranteed on computed union types.
```

source/union-to-tuple.d.ts
```diff
@@ -0,0 +1,41 @@
...
+@example
+```
+// This type may come from a library that you are using.
+type ActionsInSomeLibrary = 'create' | 'read' | 'update' | 'delete' | 'aggregate';
+
+type AllowedActions = Exclude<ActionsInSomeLibrary, 'aggregate'>;
+
+const actions: UnionToTuple<AllowedActions> = ['create', 'read', 'update', 'delete'];
+
+actions.includes(requestedAction) // Validate
```
voxpelli, line 28: What benefit does this provide over `AllowedActions[]`? It's likely just me missing something, but I'm thinking, if I'm missing something then someone else can as well, so then we should clarify it 🙂 

```typescript
const actions: AllowedActions[] = ['create', 'read', 'update', 'delete'];
```

I get the impression that `actions.includes(requestedAction)` will work better when the type of `actions` is `UnionToTuple<AllowedActions>` rather than `AllowedActions[]`? But I can't reproduce that: [TypeScript Playground example](https://www.typescriptlang.org/play?#code/PQKhCgGEHsDsDcCmAnALgAgIboK6wJZzqoCeADosdFrOvrKigM6IDGqhtpFuT9A5ugDaAE3xNUyfACMcHJOlZwxHOJgA2xcoiYBdABQALVKjJMAXMGAB3WwDpuO1lLKp1mWPzvRk-YCOhWJmBDDxFpaGgAa2BkRHVETBYAWlhoRmDHJmd8V2SAJmSADjtjAFt1AGIxCSlZeURkpVgVTg1krIBKO3BwAElYJjJ8OJF0aRJhVENxdABlVExWKPQAeSRkADN1aGsaJmsUA2NTCysJJajoDe3duyUy4ExgAFYABgBmAHYX-KKANmA+QAjF8QS8Pt1egABRAAD0wZTICXAAAN0eB8EifBgAN4AVQIcAAKtABoxkCx2JwAL7oTbIaBldAAckcyU2OlQLIA3L1HOhCZx0ABedC46aIfSdczoeDQfAiOkAH3F-DimFQ+kwvlltQEMrlCqV6FVuKcmAosoi0ASHhpfPAAvJzDYqloYqFJLJDFd1LgAB4vbAAHx8qwikPiyXS2XyxU89DqxJanX8PWSA1x42Ji1W8aRO2wB1ojHgACC6DKPkoluR+FYmGkCXQ8MRyMo1hmrEMimgOHUYzKmCilHwqCY6GsmEm9FQ1BI-eQ6HUMmQOsmShElE2tp21iYPXAsIRSJR6NRmOxaHFwdJLspbtp9MZzLZ2g5XN5vVY7iYk5gMphxaVZYEoXFwAASAeYCRAscUoMgzBgVlaVRSjPBt02ehEBEAAaRDpBQ9A0MjXAWkQbCwPwqCaXAOjwF-JIAKZWDiWsagIOg1iwngrikPyVC03LDMpE8Tp0PIrCcJoyDIOkQSSLTAAhUSDUkzDKJkgjILohjmgkcjhTFIQwL2QDYNAqVOjw9AzPQCywnY6BpV0OxhzIfR6AuWBWEoMjvMWXzEHuHiWiYTo+QFYNRS0ChoE2Iy4HDYAyNxZDYyNBNxmBTL4xNM1MHybVfBE9B9XE7NsoUkr+FU8rM0qrKlX5bR0AfKl3Viu8fQpTrOCDIlQxStKMsNfLEyIvKcywYrhLUpqJvGObfHqir+HG406IvcAQGAcB4TIHE4soHqOqfQNgyjMV9Cgqx0FReFGHC8iojSaxYFROhJw0acSEnfgFU8KhxkoSVFCSWsWm+3hcJB5oNgwSU7uAB7gy+udqGwURxEzep8AUZpWjUdQUZO44TDMSwbHsLIclcdxPG8Xx-ECYJQhaG0YjiO0UjSDJgDplxUAKYpSlQCpqlxuo5AJxoifHNp1A6bQIp6SCYqexAXrwN7dlgRD7uJQxKDwYUBVmHAWDGJJiBN9A4HUSY0xwMptaRrH6Twf1aD4YK7dNobDdRhKvd8rqXbdhhJ1mDw6F9R8ffVyCAH4SJqPG5Fw4NZWDCSyPy4P0GN2YONgNl0FCMgKFgZPZTAjYi4GTll3B1Fzp9r6BT9vzi+0OZ6YwOIyDiFho4D9Ajr4d0NCLyP3cnUOzbgRfEs2b33R+n7aDnP0utD8Hl9rqCJK1l79H0N3fFwjv3VlegW-ahP+rgfOo3yzpELT2-OEQ+vEA2I6UAEAYAIBQBgOOh1VysHHElLgbV5xYGIDgDsPRjaUHiIgKOGAfDbmXAfe2qAUEtjCA7RKh8hpTnwOoTQ0gwb2yYIiEK6BlJsEwFbSgBDNS2UXDgcqhh+yDjsukWGdtZj4HIfbI+MMHhkCzmMQKiQxih2AjgDQTt0DbiYqMWydDGwcInsSfug8+xImoSgTR0AdDCIwPwNR65fT0M1BPXBFjQ5MBwD2OBrYEjYJjlwUIGBGwsB6AMIYIw4YTCmDMGO-4cBjgCZQIxFAB7C3QMPaAFNTjU34OOQwOBpChUeGUBsjImAJVQMAZJiBUm5EqeIDxOhgDAg+PkAAnEUKE+IUjBMQLKAAmv2OyiA4aIICHKDQipNTCnqOgXhljy4YFCAoSBsBXa2Uds7Wg0jHB2HmFiahOoQbgyEKiAAonCSQSxUBfTbGeBJE8NQiDdlkqmVghZ1OSMiGcdgABWh4fB+BTosa+qART-EqFuRAwBID9PxNYOE-AABy-BDD8GJPgcsZzywAC9yy+FWHigAYtYAAUv0gAEqgAAGjiuY1gzkiCiAARWUuS8skB+BnOgOWeJAA1AAWuWAA6tYSAABZZl-TlI4HLAAaQACIKusAAcVYOKoo4q6pEoACzlmpZ4ZS6hyykv+NkM5cx+AUr6AAJXwP8UgLwBX4hVQEfAaLsXKWpQKoovryxIpeCIHVRR-UKl5fAfEmp8gfDlf8d1kq2nlmJGct4ykcV8o+IDAACiQIoKqBW-IFasAAMnVTYxb8B6sQL835eqzlnL6JAZS1hywiBxS8HV4ryzlmLXKolzLmUAGplJwjOYgbtFL+nAhRXK8sHE+hypIG8dUyAdUKuUiQM5+IyD8FYKwHd4rMDUv4OWTAcJWHIEgDi-Aw62nJAFQOEg-AFX9iJRSm1TB8TAhxUUKIcrxX9LOcCG1oQACOJ6KVRFQHqyDwqOXQexeWLN0g+U2v4MW4twqUXKTIG0-gKr1DKSYEUHAzLMD-BIOWfE+IcD9Kzcy6wIhkVkH-V8WAhgvhlHFeK4tOrOXlkHTal4cHlKgfxOWfgZAFWID5XCHAhg4SrGBCbN4-zqUqsMEUblqxlKrDaZAVYcIPjKWZVm6A+Qs3KXLAq5YmAbX-D6L8qIcxyz-Es6B4k-AmDirTVmvgrARDIDOZAL4qxSXCupXMZlSLrD9PgESs5J6+XEjmG00lxbqWrAVRQbLcrlICqJfwHVKqYDCswEUPlHK+ikpIFEeV+BmXCsXfkfpzLVjlkIP0-AUROVvFWGc8VCqSCGBtXMEga6SBwjmPkHVcrDDdvFb8kg9r+BzBDXO35hgXhZbePgZSb7yxRFJdAM5qwohMEgMCMosAojitJaBoo-B8TMsVUikwcqs3qFgAyWLZAPikpeNSzYykKXUtA9AOVZBNj9Lq6sPoOqPBMGJAiRACrrPVlG5AJt-TyzSDhIYSA-xiQKrhP8ZAsXrU2upS8aQA9fnlh1dIYtpKOIKrKMqhVURNjSHLGl8VyA+Xir5SIcVwJiRIv+G00tURPPWH4FEHskbDCgb6OWGy4w5Dxzts43pE9UQ9g8PwRAWbmBwAVZqTAX117h2FAEax-MJmrhEJqehlBUSjhIF9MgOomEUiwEPPAHA3a2WHFEAQdAcGwA0Ywzk8c+Dbj7ikweezunh+wEQjsNAxionxeuEglR6C-hwNuJgVufCO6mV1XJSBJxW3SYHrElAY+IFIPScv0xnFh+hqHa3Sdjx3I7KWS80VKFihZOWFkppWTKW-AZJGxC+mCiGqSYkC-BqcGusIcfLJbIshn7oPkO0TztnPBiI2MSTpVhnKYncr4kGrmkOuZAkwO8YHmTqU2fsvBOjauWH2TABg5gmREAK1H8NxYoWRnAUxEBJ9VQWQnlYDWQUFndGBECWRtwEhUCp8WRMB+Bkx+AXdvwBRywaFdhcI-9N5YoLki9twAwKDOAADYAgC3ZQCn8SBd9cD8DCCwxGIV4IF-9c5l9oBV8Ow6DSDDgRB6CV5N8hBICNRUDd8EDd9kDCDd8MCW8YCD9wBrkGC7BC91Bi8dB9A4hQN4kJByCfYJJ7o+VJkUDEAh9dp9pDpjoR9OAV818rpRRwB0AfCl83Deo94BpvDfCQj7oFVpYZBZZCZlBFYSZiApBlgegQjQjUY5gRkJ5ykcBkBe5dxW57ZUQzpn4LpPpL9ciqwaxNEW9MBqFDxgjki-CiAz44JXp3oDZ6ifC059AM4ZZGARBBDOA35mo6jkj-5G4QiowmjJwL4r4jc+j45H4f5X5JIP5hiUji4HEmBciygpjt9sCZ8JJMZhAhAhBdBbIAAicsM4049AM45SK43RTXEvSQTxIhUeCeHUPPJgVY3wtOJ5TZYQOwQEnqEQhIAMaggw2g4MWyRY0MEMaEoon2XQb4nw2UE4oBVGEuScAUYcTcYAl8Jke-NccA1-OZIZD-XgAQH-HgKQwYQA4A1g8AsfKAwg7AhA7AlQrAuA9Qzk1kTguIAg1Ax0Yg8QiwygsUcEwwug--Oklgok5-DgvA-k7gx0OfLAAQho2Adw0Qkg-cUUhgmQuQ6AnfVkJQpAsgOw409AzBVArQ8Ae6EgikWAaZFZauRkJYXsawPJfsDAYEtfEMXgwYfgzeRSHUsgyQ--E42KWQ5khQk0pRS0jkmAtQ60zQ9E9AUlK2DAOhB2DYDRX8RIZcJsb0ieeADQeJAMwyEwsw3omk2UQ0lkpk+QmAtMlVFvLEtqM4m0IsM4g6OEI6G8VU3cagMUHQlePQ3yCEowqsrkPU1+O01GEg8pJMNsy-TswsRIWAHs5wgcvgtvczWKUcwYfIccmgqcxAUwmc8M90ToIAA)

Also, I think that `requestedAction` needs to be defined in this example, else one doesn't know what type that one has at all, and that seems to be vital to the example. So add something like `const requestedAction = 'create';`?

source/union-to-tuple.d.ts
```diff
@@ -0,0 +1,41 @@
...
+export type UnionToTuple<Union> =
+    UnionToIntersection<
+        // Distributive conditional trick.
+        // See the source for the `UnionToIntersection` type for more details.
+        Union extends unknown
+        ? (distributed: Union) => void
+        : never
+    > extends ((merged: infer Intersection) => void)
+        // Transforms ('A' | 'B') into [[[], "A"], "B"], but destructures the arrays
+        ? readonly [...UnionToTuple<Exclude<Union, Intersection>>, Intersection]
```
voxpelli, line 40: Why is this set to `readonly`? Because it's expected to only be used with `string` only unions? And that way it would essentially not be possible to manipulate anyway? But then, wouldn't this `readonly` be fairly redundant as any meaningful mutation would anyhow already be rejected by the tuple values in itself?

test-d/union-to-tuple.ts
```diff
@@ -0,0 +1,6 @@
+import {expectAssignable} from 'tsd';
+import {UnionToTuple} from '..';
+
+declare type Union = 'A' | 'B';
+declare const tuple: UnionToTuple<Union>;
+expectAssignable<readonly ['A', 'B']>(tuple);
```
voxpelli, line 6: I think we need to extend these tests quite a bit. This only tests a "sunny day scenario", the simplest of cases, we should have some tests that tries to break it.

I eg. tried this:

```typescript
type AdvancedUnion = 'create' | number;

type AdvancedTuple = UnionToTuple<AdvancedUnion>;

const foo: AdvancedTuple = ['create', 123];
```

And that fails (see [playground](https://www.typescriptlang.org/play?#code/PQKhCgGEHsDsDcCmAnALgAgIboK6wJZzqoCeADosdFrOvrKigM6IDGqhtpFuT9A5ugDaAE3xNUyfACMcHJOlZwxHOJgA2xcoiYBdABQALVKjJMAXMGAB3WwDpuO1lLKp1mWPzvRk-YCOhWJmBDDxFpaGgAa2BkRHVETBYAWlhoRmDHJmd8V2SAJmSADjtjAFt1AGIxCSlZeURkpVgVTg1krIBKO3BwAElYJjJ8OJF0aRJhVENxdABlVExWKPQAeSRkADN1aGsaJmsUA2NTCysJJajoDe3duyUy4ExgAFYABgBmAHYX-KKANmA+QAjF8QS8Pt1egABRAAD0wZTICXAAAN0eB8EifBgAN4AVQIcAAKtABoxkCx2JwAL7oTbIaBldAAckcyU2OlQLIA3L1HOhCZx0ABedC46aIfSdczoeDQfAiOkAH3F-DimFQ+kwvlltQEMrlCqV6FVuKcmAosoi0ASHhpfPAAvJzDYqloYqFJLJDFd1LgAB4vbAAHx8qwikPiyXS2XyxU89DqxJanX8PWSA1x42Ji1W8aRO2wB1ojHgACC6DKPkoluR+FYmGkCXQ8MRyMo1hmrEMimgOHUYzKmCilHwqCY6GsmEm9FQ1BI-eQ6HUMmQOsmShElE2tp21iYPXAsIRSJR6NRmOxaHFwdJLspbtp9MZzLZ2g5XN5vVY7iYk5gMphxaVZYEoXFwAASAeYCRAscUoMgzBgVlaVRSjPBt02ehEBEAAaRDpBQ9A0MjXAWkQbCwPwqCaXAOjwF-JIAKZWDiWsagIOg1iwngrikPyVC03LDMpE8Tp0PIrCcJoyDIOkQSSLTAAhUSDUkzDKJkgjILohjmgkcjhTFIQwL2QDYNAqVOjw9AzPQCywnY6BpV0OxhzIfR6AuWBWEoMjvMWXzEHuHiWiYTo+QFYNRS0ChoE2Iy4HDYAyNxZDYyNBNxmBTL4xNM1MHybVfBE9B9XE7NsoUkr+FU8rM0qrKlX5bR0AfKl3Viu8fQpTrOCDIlQxStKMsNfLEyIvKcywYrhLUpqJvGObfHqir+HG406IvcAQGAcB4TIHE4soHqOqfQNgyjMV9Cgqx0FReFGHC8iojSaxYFROhJw0acSEnfgFU8KhxkoSVFCSWsWm+3hcJB5oNgwSU7uAB7gy+udqGwURxEzep8AUZpWjUdQUZO44TDMSwbHsLIclcdxPG8Xx-ECYJQhaG0YjiO0UjSDJgDplxUAKYpSlQCpqlxuo5AJxoifHNp1A6bQIp6SCYqexAXrwN7dlgRD7uJQxKDwYUBVmHAWDGJJiBN9A4HUSY0xwMptaRrH6Twf1aD4YK7dNobDdRhKvd8rqXbdhhJ1mDw6F9R8ffVyCAH4SJqPG5Fw4NZWDCSyPy4P0GN2YONgNl0FCMgKFgZPZTAjYi4GTll3B1Fzp9r6BT9vzi+0OZ6YwOIyDiFho4D9Ajr4d0NCLyP3cnUOzbgRfEs2b33R+n7aDnP0utD8Hl9rqCJK1l79H0N3fFwjv3VlegW-ahP+rgfOo3yzpELT2-OEQ+vEA2I6UAEAYAIBQBgOOh1VysHHElLgbV5xYGIDgDsPRjaUHiIgKOGAfDbmXAfe2qAUEtjCA7RKh8hpTnwOoTQ0gwb2yYIiEK6BlJsEwFbSgBDNS2UXDgcqhh+yDjsukWGdtZj4HIfbI+MMHhkCzmMQKiQxih2AjgDQTt0DbiYqMWydDGwcInsSfug8+xImoSgTR0AdDCIwPwNR65fT0M1BPXBFjQ5MBwD2OBrYEjYJjlwUIGBGwsB6AMIYIw4YTCmDMGO-4cBjgCZQIxFAB7C3QMPaAFNTjU34OOQwOBpChUeGUBsjImAJVQMAZJiBUm5EqeIDxOhgDAg+PkAAnEUKE+IUjBMQLKAAmv2OyiA4aIICHKDQipNTCnqOgXhljy4YFCAoSBsBXa2Uds7Wg0jHB2HmFiahOoQbgyEKiAAonCSQSxUBfTbGeBJE8NQiDdlkqmVghZ1OSMiGcdgABWh4fB+BTosa+qART-EqFuRAwBID9PxNYOE-AABy-BDD8GJPgcsZzywAC9yy+FWHigAYtYAAUv0gAEqgAAGjiuY1gzkiCiAARWUuS8skB+BnOgOWeJAA1AAWuWAA6tYSAABZZl-TlI4HLAAaQACIKusAAcVYOKoo4q6pEoACzlmpZ4ZS6hyykv+NkM5cx+AUr6AAJXwP8UgLwBX4hVQEfAaLsXKWpQKoovryxIpeCIHVRR-UKl5fAfEmp8gfDlf8d1kq2nlmJGct4ykcV8o+IDAACiQIoKqBW-IFasAAMnVTYxb8B6sQL835eqzlnL6JAZS1hywiBxS8HV4ryzlmLXKolzLmUAGplJwjOYgbtFL+nAhRXK8sHE+hypIG8dUyAdUKuUiQM5+IyD8FYKwHd4rMDUv4OWTAcJWHIEgDi-Aw62nJAFQOEg-AFX9iJRSm1TB8TAhxUUKIcrxX9LOcCG1oQACOJ6KVRFQHqyDwqOXQexeWLN0g+U2v4MW4twqUXKTIG0-gKr1DKSYEUHAzLMD-BIOWfE+IcD9Kzcy6wIhkVkH-V8WAhgvhlHFeK4tOrOXlkHTal4cHlKgfxOWfgZAFWID5XCHAhg4SrGBCbN4-zqUqsMEUblqxlKrDaZAVYcIPjKWZVm6A+Qs3KXLAq5YmAbX-D6L8qIcxyz-Es6B4k-AmDirTVmvgrARDIDOZAL4qxSXCupXMZlSLrD9PgESs5J6+XEjmG00lxbqWrAVRQbLcrlICqJfwHVKqYDCswEUPlHK+ikpIFEeV+BmXCsXfkfpzLVjlkIP0-AUROVvFWGc8VCqSCGBtXMEga6SBwjmPkHVcrDDdvFb8kg9r+BzBDXO35hgXhZbePgZSb7yxRFJdAM5qwohMEgMCMosAojitJaBoo-B8TMsVUikwcqs3qFgAyWLZAPikpeNSzYykKXUtA9AOVZBNj9Lq6sPoOqPBMGJAiRACrrPVlG5AJt-TyzSDhIYSA-xiQKrhP8ZAsXrU2upS8aQA9fnlh1dIYtpKOIKrKMqhVURNjSHLGl8VyA+Xir5SIcVwJiRIv+G00tURPPWH4FEHskbDCgb6OWGy4w5Dxzts43pE9UQ9g8PwRAWbmBwAVZqTAX117h2FAEax-MJmrhEJqehlBUSjhIF9MgOomEUiwEPPAHA3a2WHFEAQdAcGwA0Ywzk8c+Dbj7ikweezunh+wEQjsNAxionxeuEglR6C-hwNuJgVufCO6mV1XJSBJxW3SYHrElAY+IFIPScv0xnFh+hqHa3Sdjx3I7KWS80VKFihZOWFkppWTKW-AZJGxC+mCiGqSYkC-BqcGusIcfLJbIshn7oPkO0TztnPBiI2MSTpVhnKYncr4kGrmkOuZAkwO8YHmTqU2fsvBOjauWH2TABg5gmREAK1H8NxYoWRnAUxEBJ9VQWQnlYDWQUFndGBECWRtwEhUCp8WRMB+Bkx+AXdvwBRywaFdhcI-9N5YoLki9twAwKDOAADYAgC3ZQCn8SBd9cD8DCCwxGIV4IF-9c5l9oBV8Ow6DSDDgRB6CV5N8hBICNRUDd8EDd9kDCDd8MCW8YCD9wBrkGC7BC91Bi8dB9A4hQN4kJByCfYJJ7o+VJkUDEAh9dp9pDpjoR9OAV818rpRRwB0AfCl83Deo94BpvDfCQj7oFVpYZBZZCZlBFYSZiApBlgegQjQjUY5gRkJ5ykcBkBe5dxW57ZUQzpn4LpPpL9ciqwaxNEW9MBqFDxgjki-CiAz44JXp3oDZ6ifC059AM4ZZGARBBDOA35mo6jkj-5G4QiowmjJwL4r4jc+j45H4f5X5JIP5hiUji4HEmBciygpjt9sCZ8JJMZhAhAhBdBbIAAicsM4049AM45SK43RTXEvSQTxIhUeCeHUPPJgVY3wtOJ5TZYQOwQEnqEQhIAMaggw2g4MWyRY0MEMaEoon2XQb4nw2UE4oBVGEuScAUYcTcYAl8Jke-NccA1-OZIZD-XgAQH-HgVteADwPyEQGKMfKAwgqfNZMoOhZAR0YgkQWk4KEQEE-yBo2Adw0Qmkuk7OIaHg3gwYDAXcaAWUMUvkgU2KWQ5khQ9AYEaNLQoAA)) as for some reason the calculated tuple becomes `[number, 'create']` rather than the expected `['create', number]`.

I similarly tested `UnionToTuple<'create' | boolean>`, which results in `[false, true, 'create']`, so also it returns things in the wrong order, but it also extends `boolean` into `false` and `true` respectively.

If `UnionToTuple` only supports `string` unions, then it should be changed to `UnionToTuple<Union extends String>` + a test should be added with `expectError()` that ensures that it eg. `UnionToTuple<'create' | boolean>` then fails.

Alternatively, if `UnionToTuple` is supposed to support more than just strings, then tests should be added to ensure that + the implementation fixed so that it works as expected.

I would also like a test for:

```typescript
type Foo = 'abc' | 'def';
type Bar = Foo | 'xyz';
type GiveMeThatTuple = UnionToTuple<Bar>;
```

Which would highlight if the expected outcome is `['abc' | 'def', 'xyz']` or `['abc', 'def', 'xyz']`. It may be obvious, but yet not obvious.

voxpelli (changes requested):
Interesting addition! I tried to squeeze in a quick review, I hope all the comments of mine still makes sense 🙏

readme.md
joaopaulobdac: I think this is clearer.

source/union-to-tuple.d.ts
joaopaulobdac: > What benefit does this provide over `AllowedActions[]`?

With just `AllowedActions[]` you are allowed to do `const actions3: AllowedActions[] = [];` and `const actions2: AllowedActions[] = ['create', 'create', 'create'];` which are not be what is actually intended. Do you think this should be talked about in the description?

source/union-to-tuple.d.ts
joaopaulobdac: Without `readonly` you are allowed to do manipulations like
```ts
const actions: UnionToTuple<AllowedActions> = ['create', 'read', 'update', 'delete'];
actions.shift();
```

source/union-to-tuple.d.ts
```diff
@@ -0,0 +1,41 @@
+import {UnionToIntersection} from './union-to-intersection';
+
+/**
+Convert an explicit union type to a tuple.
```
sindresorhus, line 4:
```suggestion
Convert a user defined union type to a tuple type.
```

source/union-to-tuple.d.ts
```diff
@@ -0,0 +1,41 @@
...
+
+Use-case: You need to do validation but you don't have an enum, only an union type. Similar to the [`Extract` example in the readme](https://typescript-play.js.org/?target=6#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXzSwEdkQBJYACgEoAueVZAWwCMQYBuAKDDwGcM8MgBF4AXngBlAJ6scESgHIRi6ty5ZUGdoihgEABXZ888AN5d48ANoiAuvUat23K6ihMQ9ATE0BzV3goPy8GZjZOLgBfLi4Aejj4AEEICBwAdz54MAALKFQQ+BxEeAAHY1NgKAwoIKy0grr4DByEUpgccpgMaXgAaxBerCzi+B9-ZulygDouFHRsU1z8kKMYE1RhaqgAHkt4AHkWACt4EAAPbVRgLLWNgBp9gGlBs8uQa6yAUUuYPQwdgNpKM7nh7mMML4CgA+R5WABqUAgpDeVxuhxO1he0jsXGh8EoOBO9COx3BQPo2PBADckaR6IjkSA6PBqTgsMBzPsicdrEC7OJWXSQNwYvFEgAVTS9JLXODpeDpKBZFg4GCoWa8VACIJykAKiQWKy2YQOAioYikCg0OEMDyhRSy4DyxS24KhAAMjyi6gS8AAwjh5OD0iBFHAkJoEOksC1mnkMJq8gUQKDNttKPlnfrwYp3J5XfBHXqoKpfYkAOI4ansTxaeDADmoRSCCBYAbxhC6TDx6rwYHIRX5bScjA4bLJwoDmDwDkfbA9JMrVMVdM1TN69LgkTgwgkchUahqIA), but in that case the `changePersonData` function does not validate the `key` parameter at runtime, making it only safe inside TypeScript. Using a tuple and `Array#includes` for validation gives us runtime safety for that kind of function.
+
+@example
+```
+type Union = 'A' | 'B';
```
sindresorhus, line 14: Can you include the import statement? It's nice when examples are directly runnable.

sindresorhus (changes requested):

source/union-to-tuple.d.ts
joaopaulobdac: Just added import statements and declared the `requestedAction` variable as well.

AndrewMorsillo:
This is extremely useful for some cases, especially with template literal types now where you can get a string union from a tuple, derive another useful union from that, but then cannot get a tuple type for the derived union. For example:

```
const actions = ["create", "update", "delete"] as const
type ActionType = typeof actions[number] // "create" | "update" | "delete"

//derive some useful string union to use in other places in the app
type ErrorActionType = `error_${ActionsType}` // "error_create" | "error_update" | "error_delete"

//We can derive a union type with template literal types but there's no way to check the type of an errorActions tuple without UnionToTuple!
const errorActions: UnionToTuple<ErrorActionType> = ["error_create", "error_update", "error_delete"] as const
```

I don't fully understand the error cases other have pointed out here, but perhaps restricting UnionToTuple to operating only on string unions actually solves most of the use cases?

joaopaulobdac:
Yeah, this becomes even more powerful with template literal types as you demonstrated.

I would be okay with constraining this to only string unions since I don't know how to solve the issue with non-string values that @voxpelli encountered above.

joaopaulobdac:
Hey guys. While testing this some more, I found that there were multiple times where TS would randomly change the position of some values inside a string union, which would then break tuple types that were previously working. For example, there was a time where a tuple was assigned to `['abc', 'def', 'xyz']` inside VS Code, but then when running the tests in the terminal it would become `['abc', 'xyz', 'def']`, failing the test. There were also times where cutting and pasting code inside the editor would change tuple definitions.

I initially thought that only computed unions would be a problem as it talked about in the TS repo issue, but it seems like the order of manually declared unions are also not stable. Because of this, I'll have to close this PR.

If anyone still needs a way to convert string unions to tuples for validation, I posted this solution which uses `Object.keys()` and still works if the order of the elements is changed: https://github.com/microsoft/TypeScript/issues/13298#issuecomment-748340770

papb:
Maybe this PR can still be useful? You could just add a note that "the order in which the tuple is generated is not guaranteed".

joaopaulobdac:
The problem is that every time TS changes the order of your tuple, your code will not compile until you also change it.

One way to fix this would be to generate all possible permutations of the union so that any order you use would be valid. [So for a union of 10 elements, we would have to generate 10! (3628800) types](https://github.com/microsoft/TypeScript/issues/13298#issuecomment-514082646), which obliterates the performance.

papb:
> The problem is that every time TS changes the order of your tuple, your code will not compile until you also change it.

Only if your code depends on the order of the generated tuple, right? Since this type will be documented to not guarantee order, users of this type will know that they should only use it when the order does not matter.

alita-moore:
ya could put it in a __UNSAFE__ object or something like that

PR #170, opened 2020-12-15
Improve readability of `Merge` result

papb:
Consider this example:

```ts
type Foo = { foo: string };
type Bar = { bar: string };

type Merged = Merge<Foo, Bar>;
```

Before this PR:

image

After this PR:

image

### DIFF
--- a/source/merge.d.ts
+++ b/source/merge.d.ts
@@ -1,4 +1,7 @@
 import {Except} from './except';
+import {Simplify} from './simplify';
+
+type Merge_<FirstType, SecondType> = Except<FirstType, Extract<keyof FirstType, keyof SecondType>> & SecondType;
 
 /**
 Merge two types into a new type. Keys of the second type overrides keys of the first type.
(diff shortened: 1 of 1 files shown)


source/merge.d.ts
```diff
@@ -1,5 +1,9 @@
 import {Except} from './except';
 
+type ImproveIntersectionHoverReadabilityShallow<T> = T extends Record<string, unknown> ? { [K in keyof T]: T[K] } : T;
```
sindresorhus, line 3: This looks like https://gist.github.com/ClickerMonkey/a081b990b9b14215141fb6248cef4dc4#file-types-ts-L49-L54 Maybe we should do it like that one and call it `Simplify`? I think it might be useful for other types too, so can you move it to https://github.com/sindresorhus/type-fest/blob/master/source/utilities.d.ts

sindresorhus:
@mmkal Do you know of any way to test this? To ensure it doesn't regress in the future.

sindresorhus:
This is blocked by https://github.com/sindresorhus/type-fest/pull/157/files#diff-520a63fe400b4b92fa3ccc14cabc1d84153987354c3f242deddfaeff0a7deae2, which introduces such a type.

mmkal:
@sindresorhus I can't think of a way to do it without using the compiler API, but maybe that would be appropriate given this library is specifically for types.

sindresorhus:
@papb Can you update to use https://github.com/sindresorhus/type-fest/blob/master/source/simplify.d.ts ?

PR #188, opened 2021-03-14
Add `Passthrough` type

JasonHK:
Sometimes you might want to extend an interface with a type that TypeScript should accept, but it's not valid syntax, for example, the props of a JSX intrinsic element (`JSX.IntrinsicElements["a"]`). This new type solves this issue.

```ts
import {Passthrough} from 'type-fest';

// Syntax Error
interface EnhancedAnchorProps extends JSX.IntrinsicElements["a"] { ... }

// Valid Syntax
interface EnhancedAnchorProps extends Passthrough<JSX.IntrinsicElements["a"]> { ... }
```


source/passthrough.d.ts
```diff
@@ -0,0 +1,15 @@
+/**
+A passthrough type that allows some types to be used to extend an interface, which originally can't.
```
sindresorhus, line 2: It would be useful to document exactly what types.

readme.md
```diff
@@ -104,6 +104,7 @@ Click the type names for complete docs.
 - [`Entries`](source/entries.d.ts) - Create a type that represents the type of the entries of a collection.
 - [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`Asyncify`](source/asyncify.d.ts) - Create an async version of the given function type.
+- [`Passthrough`](source/passthrough.d.ts) - A passthrough type that allows some types to be used to extend an interface, which originally can't.
```
sindresorhus, line 107: `A passthrough` is moot as it's just repeating the type name. Try to make the description follow the format of the other descriptions.

sindresorhus (changes requested):

sindresorhus:
Make sure you adhere to: https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md

sindresorhus:
Can you propose some alternative names for this type? I'm not entirely sold on `Passthrough` name. It doesn't describe what actually happens.

Is this trick something we can rely on? I just want to make sure it's just not a bug in TypeScript that will get fixed at some point.

papb:
Maybe `AsExtendable` instead of `Passthrough`?

JasonHK:
> Can you propose some alternative names for this type? I'm not entirely sold on `Passthrough` name. It doesn't describe what actually happens.

I've originally named it `Type`, but it is more confusing than `Passthrough`. I named it `Passthrough` because it simply returns the type you passed to it, I'm not good at naming things so this is the best name I could think out of.

> Is this trick something we can rely on? I just want to make sure it's just not a bug in TypeScript that will get fixed at some point.

It is basically a shorthand for

```ts
type AnchorProps = JSX.IntrinsicElements["a"];
interface EnhancedAnchorProps extends AnchorProps {...}
```

with the advantage of doesn't require a `type` to be declared. This type won't you use something like a union type or primitive type to extend an interface, so this shouldn't break anything anytime soon.

source/passthrough.d.ts
JasonHK: I've added a [use-case](https://github.com/sindresorhus/type-fest/pull/188/files#diff-f7c49e21c2e22539ee8f81d4a42709bbd8700866e88aa1aa55a3d86cfaca322dR4-R5) (more if I can think out of later), hopefully it would easier to understand.

readme.md
JasonHK: Is the new one better?

sindresorhus:
> I'm not good at naming things so this is the best name I could think out of.

You will not get better at naming without doing it a lot.

sindresorhus:
The type looks useful, but it needs a better name.

papb:
> Maybe `AsExtendable` instead of `Passthrough`?

Any thoughts on this?

JasonHK:
> > Maybe `AsExtendable` instead of `Passthrough`?
> 
> Any thoughts on this?

`AsExtendable` did reflect the purpose of the type, I'll change to this name if @sindresorhus okay with it.

sindresorhus:
Alright. Let's go with `AsExtendable`.

JasonHK:
> Alright. Let's go with `AsExtendable`.

@sindresorhus The type was renamed.

readme.md
```diff
@@ -104,6 +104,7 @@ Click the type names for complete docs.
 - [`Entries`](source/entries.d.ts) - Create a type that represents the type of the entries of a collection.
 - [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`Asyncify`](source/asyncify.d.ts) - Create an async version of the given function type.
+- [`AsExtendable`](source/as-extendable.d.ts) - Wrap a type like `JSX.IntrinsicElements["a"]` so that it could be used to extend an interface.
```
sindresorhus, line 107: > like `JSX.IntrinsicElements["a"]`

This might make the user think only that kind of type works. I would prefer to make the description without an exact example.

source/as-extendable.ts
```diff
@@ -0,0 +1,18 @@
+/**
+A passthrough type that allows some types to be used to extend an interface, which originally can't.
+
+Use-cases:
+- A member of an interface (which is an object): `JSX.IntrinsicElements["a"]`
+
+@example
+```
+import {Passthrough} from 'type-fest';
```
sindresorhus, line 9: Outdated

source/as-extendable.ts
```diff
@@ -0,0 +1,18 @@
+/**
+A passthrough type that allows some types to be used to extend an interface, which originally can't.
+
+Use-cases:
+- A member of an interface (which is an object): `JSX.IntrinsicElements["a"]`
```
sindresorhus, line 5:
```suggestion
- A member of an interface (which is an object): `JSX.IntrinsicElements['a']`
```

Applies elsewhere too.

source/as-extendable.ts
```diff
@@ -0,0 +1,18 @@
+/**
+A passthrough type that allows some types to be used to extend an interface, which originally can't.
+
+Use-cases:
```
sindresorhus, line 4: Can you come up with some more examples of where it could be used?

test-d/as-extendable.ts
```diff
@@ -0,0 +1,20 @@
+import {expectType} from 'tsd';
+import {AsExtendable} from '../ts41';
```
sindresorhus, line 2: It's not in the `ts41` directory.

test-d/as-extendable.ts
```diff
@@ -0,0 +1,20 @@
...
+
+namespace JSX {
+	export interface IntrinsicElements {
```
sindresorhus, line 5: I think `export` is moot here.

sindresorhus:
Please spend some more effort on this PR.

sindresorhus:
The PR title is outdated.

sindresorhus:
Bump

sindresorhus:
https://github.com/sindresorhus/type-fest/issues/227

PR #191, opened 2021-03-19
Improve the doc comment for the Except type

shiftgeist:
> After much discussion, we think libraries providing their own "stricter" versions of Omit (as well as `Exclude` and other friends) is preferable.

https://github.com/microsoft/TypeScript/issues/30825#issuecomment-523668235

Since the question is decided and the issue closed, the comment may be removed.


sindresorhus:
I think it should still link to the TS issue, but instead the text should be the resolution; the fact that the TS team has declined a stricter version and why.

shiftgeist:
I changed the comment.

Issue #171, opened 2020-12-16
Rename `SetOptional` and `SetRequired`?


sindresorhus:
See: [`1af6d4a` (#157)](https://github.com/sindresorhus/type-fest/pull/157/commits/1af6d4a2cb0982ab76f611943d5f9c47a49d7e74#r533532327)

`SetOptional` would be renamed to `Optional`. However `Required` already exists, so not sure what to do about that? Can we make it an overload?

voxpelli:
Maybe `Required` could get an `ExtendedRequired` or such as a compromise? Somewhat less pretty, but also somewhat more explicit in what it does, that it extends the built in one

kainiedziela:
Pretty sure we can just overwrite `Required` to achieve what we want here.

papb:
By the way, recently I needed a different version of `Required` that turned `{ foo?: string }` into `{ foo: string | undefined }` instead of `{ foo: string }` - I had to write my own, and had trouble choosing a good name as well. It would be great to have a name that could clearly show which case is it (although I agree most people will want the current behavior)

sindresorhus:
> By the way, recently I needed a different version of Required that turned { foo?: string } into { foo: string | undefined } instead of { foo: string } 

I'm curious what you needed it for?

PR #199, opened 2021-04-16
Add `SetNonNullable` type

### LINKED ISSUE #144: [Proposal] SetNonNullable
I honestly thought `SetRequired` did this. example of expected behavior

```ts
type Optional<T> = T | undefined | null;
type Foo = {
   bar: Optional<string>;
   foo?: string | null;
   baz?: string;
};

type NonNullableFoo = SetNonNullable<Foo, 'bar' | 'foo'>;

// would be equivalent to

type ManualNonNullableFoo = {
   bar: string;
   foo: string;
   baz?: string;
};
```

Smtih:
Often, you will want to extract code the checks or asserts that specific fields of an object are present, this makes typing this much easier.

```typescript
function assertHasName(person: Person): asserts person is SetNonNullable<Person, "name"> {
  if(person.name == null) throw new Error("Person has no name");
}
```

Fixes #144

### DIFF
--- a/source/set-non-nullable.d.ts
+++ b/source/set-non-nullable.d.ts
@@ -2,7 +2,7 @@ import {Except} from './except';
 import {Simplify} from './simplify';
 
 /**
-Create a type that makes the given keys non-nullable. The remaining keys are kept as is.
+Create a type removes nullish (optional, undefined and null) types from the given keys. The remaining keys are kept as is.
 
 Use-case: You want to define a single model where the only thing that changes is whether or not some of the keys are required.
 
@@ -21,7 +21,7 @@ type Foo = {
 type SomeNonNullable = SetNonNullable<Foo, 'b' | 'c' | 'd' | 'e'>;
 // type SomeNonNullable = {
 // 	a?: number;
-// 	b: string; // Was already non-null and still is.
+// 	b: string; // Was already not nullish and still is.
 // 	c: boolean; // Is now not optional.
 //  d: number; // Is now not null
 //	e: string; // Is now not undefined
--- a/source/set-non-nullable.d.ts
+++ b/source/set-non-nullable.d.ts
@@ -1,0 +1,36 @@
+import {Except} from './except';
+import {Simplify} from './simplify';
+
+/**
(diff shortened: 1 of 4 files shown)


sindresorhus:
> I wonder if we could just overwrite the built-in NonNullable version with a more powerful one that also accepts which keys to apply to, as discussed in #171.

Any thoughts on this?

Smtih:
While we could do this, I'm unsure what the benefit is.

I think it would be pretty unexpected for a type named non-nullable to do this, especially as it's already a built in type Does it also make the entire object non-nullable as well as the keys?

I think having the word 'Set' makes it clear how this type operates.

source/set-non-nullable.d.ts
```diff
@@ -0,0 +1,36 @@
+import {Except} from './except';
+import {Simplify} from './simplify';
+
+/**
+Create a type removes nullish (optional, undefined and null) types from the given keys. The remaining keys are kept as is.
+
```
sindresorhus, line 6: Can you mention how it related to `SetRequired` and the inverse in the `SetRequired` doc comment?

source/set-non-nullable.d.ts
```diff
@@ -0,0 +1,36 @@
+import {Except} from './except';
+import {Simplify} from './simplify';
+
+/**
+Create a type removes nullish (optional, undefined and null) types from the given keys. The remaining keys are kept as is.
```
sindresorhus, line 5:
```suggestion
Create a type that removes nullish (optional, `undefined`, and `null`) types from the given keys. The remaining keys are kept as is.
```

sindresorhus (changes requested):

sindresorhus:
Please follow: https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md

sindresorhus:
Alright, let's go with `Set` like you have it.

sindresorhus:
Friendly bump :)

sindresorhus:
https://github.com/sindresorhus/type-fest/issues/228

PR #205, opened 2021-05-06
Add `PackageJson.PublishConfig` type

### LINKED ISSUE #204: PackageJson type does not contain known properties for "publishConfig"
publishConfig is set to:
https://github.com/sindresorhus/type-fest/blob/532e1da01650382e6989c407fc83918ab5e72fff/source/package-json.d.ts#L574

I believe the publishConfig should have `access`, `registry`, and `tag`.

ryansonshine:
This change adds the interface for the PackageJson.PublishConfig.

Closes #204

### DIFF
--- a/source/package-json.d.ts
+++ b/source/package-json.d.ts
@@ -601,20 +601,25 @@ declare namespace PackageJson {
 
 	export interface PublishConfig {
 		/**
+		Additional, less common properties from the [npm docs on publishConfig](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#publishconfig)
+		*/
+		[additionalProp: string]: unknown;
+
+		/**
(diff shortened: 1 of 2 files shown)


sindresorhus:
According to https://docs.npmjs.com/cli/v7/configuring-npm/package-json#publishconfig, it supports a lot more properties.

ryansonshine:
@sindresorhus Ah good catch, when I initially read that I thought the config options applied to other sections in package.json as a whole. I'll do some additional digging and submit an update

ryansonshine:
I added the type to allow for additional props in general, per https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/package.json#L576-L591

ryansonshine (commented):
Looks good to me

Issue #207, opened 2021-05-14
PartialDeep does not work with `Date` type


ChromeQ:
Hi,
I am having trouble making my deep partial type setting dates correctly.

Example (including it working with Partial<T>):
```typescript
import { PartialDeep } from 'type-fest';

type Foo = {
    a: Date;
    b: {
        c: Date;
    };
};

type Bar = Partial<Foo>;
const bar1: Bar = { a: 123 }; // Does error
const bar2: Bar = { b: { c: 123 } }; // Does error

type Baz = PartialDeep<Foo>;
const baz1: Baz = { a: 123 }; // Should error
const baz2: Baz = { a: new Date() }; // Only this should be allowed

const baz3: Baz = { b: { c: 123 } }; // Should error
const baz4: Baz = { b: { c: new Date() } }; // Only this should be allowed
```

I think the fix is easy enough so I will open a pull request.

Nr9:
Same issue here

voxpelli:
Have you figured out why it behaves like this?

Seems almost as if the `Date` after a `PartialDeep<Foo>` has been replaced with an `any`? Have you inspected what the resulting type looks like?

sindresorhus:
If anyone wants to work on this, see the feedback given in https://github.com/sindresorhus/type-fest/pull/208.

sdotson:
I'm giving this a try in https://github.com/sindresorhus/type-fest/pull/342. Happy to address any feedback.

PR #208, opened 2021-05-14
Adding Date handling to PartialDeep

ChromeQ:
Fixing the issue #207

### DIFF
--- a/source/partial-deep.d.ts
+++ b/source/partial-deep.d.ts
@@ -30,6 +30,8 @@ settings = applySavedSettings({textEditor: {fontWeight: 500}});
 */
 export type PartialDeep<T> = T extends Primitive
 	? Partial<T>
+	: T extends Date
+	? T
 	: T extends Map<infer KeyType, infer ValueType>
 	? PartialMapDeep<KeyType, ValueType>
 	: T extends Set<infer ItemType>
--- a/test-d/partial-deep.ts
+++ b/test-d/partial-deep.ts
@@ -9,6 +9,7 @@ const foo = {
 		string: 'waldo',
 		number: 1,
 		boolean: false,
+		date: new Date(),
 		symbol: Symbol('test'),
 		null: null,
 		undefined: undefined, // eslint-disable-line object-shorthand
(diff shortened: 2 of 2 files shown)


sindresorhus:
I would prefer a more general solution that doesn't special-case `Date`.

The error is:

```
test-d/partial-deep.ts:37:29
  ✖  37:29  Argument of type PartialObjectDeep<Date> | undefined is not assignable to parameter of type Date | undefined.
  Property [Symbol.toPrimitive] is missing in type PartialObjectDeep<Date> but required in type Date.
```

https://stackoverflow.com/questions/57100616/how-to-forward-well-known-symbols-in-a-type-wrapper-mapped-type/57101132

sindresorhus:
// @kainiedziela

ChromeQ:
I was not seeing that error when I was editing but it looks a bit more complex than I originally thought then. Is this something you will be able to fix fairly easily?

sindresorhus:
No, I unfortunately don't have time to look into this. Do you want to try to fix it or should I close to let someone else handle it?

ChromeQ:
I'll take another look and attempt to fix without failing any tests or errors.
Any pointers on the solution off the top of your head? I've lost the context and not as familiar with the code or TS

source/partial-deep.d.ts
```diff
@@ -30,6 +30,8 @@ settings = applySavedSettings({textEditor: {fontWeight: 500}});
 */
 export type PartialDeep<T> = T extends Primitive
 	? Partial<T>
+	: T extends Date
+	? T
```
voxpelli, line 34: Rather than setting `Date` here explicitly, could you maybe do something like:

```suggestion
export type PartialDeep<T, ExceptTypes = never> = T extends Primitive
	? Partial<T>
	: T extends ExceptTypes
	? T
```

Then someone can chose to exclude eg. `Date` when that's a good idea, or eg a collection like `PartialDeep<Foo, MyBuiltInTypes>`

voxpelli (changes requested):
Building on top of what Sindre said :)

Also, looks like the test run result has disappeared 🤔 It will hopefully appear if you do a new push here

PR #214, opened 2021-05-30
Add `ScreamingSnakeCase` type

### LINKED ISSUE #192: Proposal: `ScreamingSnakeCase` type
## Proposal

Add a type for the naming convention screaming snake.

## Background

The `SCREAMING_SNAKE` Case is a very popular naming convention for `enums` and for `constants`. It is just the snake case and uppercase combined.

Yash-Singh1:
Resolves #192

### DIFF
--- a/ts41/screaming-snake-case.d.ts
+++ b/ts41/screaming-snake-case.d.ts
@@ -1,4 +1,5 @@
-import {DelimiterCase, SplitIncludingDelimiters} from './delimiter-case';
+import {SplitIncludingDelimiters} from './delimiter-case';
+import {SnakeCase} from './snake-case';
 
 /**
 Checks if an array contains an item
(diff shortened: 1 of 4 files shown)


sindresorhus:
Why not just use the existing `SnakeCase` type with the built-in `Uppercase`?

ts41/screaming-snake-case.d.ts
```diff
@@ -0,0 +1,60 @@
+import {SplitIncludingDelimiters} from './delimiter-case';
+import {SnakeCase} from './snake-case';
+
+/**
+Returns a boolean on whether the given array contains a certain item
```
sindresorhus, line 5: End sentences with `.`

ts41/screaming-snake-case.d.ts
```diff
@@ -0,0 +1,60 @@
...
+*/
+type Includes<Value extends any[], Item> = {
+	[P in keyof Value & number as Value[P]]: true;
+}[Item] extends true
+	? true
+	: false;
```
sindresorhus, line 11: Do you think it would be useful to make this a public type?

sindresorhus (changes requested):

PR #217, opened 2021-06-02
Add `Includes` type

### LINKED ISSUE #216: Proposal: `Includes` type
https://github.com/sindresorhus/type-fest/pull/214/files#r642218008

Would anyone such type useful?

Yash-Singh1:
Resolves #216

The problem is that my solution to `Includes` works only with the `PropertyKey` type for `Item`, because it makes the values the key and the value a boolean, then checks if the value is true for the property with the key of `Item`. The other solution is:

```ts
type Includes<T extends readonly any[], U> = U extends T[number] ? true : false;
```

However as shown in type-challenges/type-challenges#1432, this solution does not work, because it uses `extends` allowing `[{a: 5}]` to include `{}`. Any ideas on supporting `Item` for types other than `PropertyKey` and at the same time not running into the `extends` problem (https://github.com/Yash-Singh1/type-fest/blob/main/test-d/includes.ts#L22)?

### DIFF
--- a/test-d/includes.ts
+++ b/test-d/includes.ts
@@ -8,14 +8,14 @@ const includesSingleItemArray: Includes<['colors'], 'colors'> = true;
 expectType<true>(includesSingleItemArray);
 
 const includesComplexMultiTypeArray: Includes<[
-  {
-    prop: 'value';
-    num: 5;
-    anotherArr: [1, '5', false];
-  },
-  true,
-  null,
-  'abcd'
+	{
+		prop: 'value';
+		num: 5;
+		anotherArr: [1, '5', false];
+	},
+	true,
+	null,
+	'abcd'
 ], 'abc'> = false;
 expectType<false>(includesComplexMultiTypeArray);
 
--- a/test-d/includes.ts
+++ b/test-d/includes.ts
@@ -22,6 +22,12 @@ expectType<false>(includesComplexMultiTypeArray);
 const noExtendsProblem: Includes<[boolean], true> = false;
 expectType<false>(noExtendsProblem);
 
+const objectIncludes: Includes<[{}], { a: 1 }> = false;
+expectType<false>(objectIncludes);
+
+const objectIncludesPass: Includes<[{ a: 1 }], { a: 1 }> = true;
+expectType<true>(objectIncludesPass);
+
 // @ts-expect-error
 const noArgsIncludes: Includes<> = false;
 
--- a/test-d/includes.ts
+++ b/test-d/includes.ts
@@ -1,5 +1,5 @@
-import {Includes} from '..';
 import {expectType} from 'tsd';
+import {Includes} from '..';
 
 const includesEmptyArray: Includes<[], 'abc'> = false;
 expectType<false>(includesEmptyArray);
--- a/test-d/includes.ts
+++ b/test-d/includes.ts
@@ -25,7 +25,7 @@ expectType<false>(noExtendsProblem);
 const objectIncludes: Includes<[{}], { a: 1 }> = false;
 expectType<false>(objectIncludes);
 
-const objectIncludesPass: Includes<[{ a: 1 }], { a: 1 }> = true;
+const objectIncludesPass: Includes<[{a: 1}], {a: 1}> = true;
 expectType<true>(objectIncludesPass);
 
 // @ts-expect-error
--- a/test-d/includes.ts
+++ b/test-d/includes.ts
@@ -1,4 +1,4 @@
-import {expectType} from 'tsd';
+import {expectError, expectType} from 'tsd';
 import {Includes} from '..';
 
 const includesEmptyArray: Includes<[], 'abc'> = false;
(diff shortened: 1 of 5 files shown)


test-d/includes.ts
```diff
@@ -0,0 +1,35 @@
...
+import {expectType} from 'tsd';
+
+const includesEmptyArray: Includes<[], 'abc'> = false;
+expectType<false>(includesEmptyArray);
+
+const includesSingleItemArray: Includes<['colors'], 'colors'> = true;
+expectType<true>(includesSingleItemArray);
+
+const includesComplexMultiTypeArray: Includes<[
+  {
```
sindresorhus, line 11: Use tab-indentation.

test-d/includes.ts
```diff
@@ -0,0 +1,35 @@
+import {Includes} from '..';
+import {expectType} from 'tsd';
```
sindresorhus, line 2: This should be first.

ts41/includes.d.ts
```diff
@@ -0,0 +1,19 @@
+/**
+Returns a boolean for whether the given array includes the given item.
+
+This can be useful if another type wants to make a decision based on whether the array includes that item.
+
+@example
+```
+import {Includes} from 'type-fest';
+
+type hasRed<arr extends any[]> = Includes<arr, 'red'>;
```
sindresorhus, line 10:
```suggestion
type hasRed<array extends any[]> = Includes<array, 'red'>;
```

sindresorhus (changes requested):

test-d/includes.ts
Yash-Singh1: Could you elaborate a bit? Do you mean:

```ts
import {Includes} from '..';
```

or

```ts
import {Includes} from '../ts41/includes';
```

sindresorhus: The former

sindresorhus:
How can we get this moving forward?

Yash-Singh1:
I did find a solution but it depends on the `Equal` type presented here: https://github.com/Microsoft/TypeScript/issues/27024#issuecomment-421529650. Should I change this PR to add `Includes` and `Equal` or should I simply update the `Includes` type to use it inline? Here it is by the way:

```ts
export type Includes<Value extends any[], Item> =
  (<G>() => G extends Value[0] ? 1 : 2) extends
  (<G>() => G extends Item ? 1 : 2)
	? true
	: Value extends [Value[0], ...infer rest]
		? Includes<rest, Item>
		: false;
```

sindresorhus:
Great. Make it a separate type called `IsEqual`, but put it inside the same file as `Includes`. And link to https://github.com/microsoft/TypeScript/issues/27024#issuecomment-421529650 in a code comment for context.

test-d/includes.ts
```diff
@@ -0,0 +1,41 @@
...
+const noExtendsProblem: Includes<[boolean], true> = false;
+expectType<false>(noExtendsProblem);
+
+const objectIncludes: Includes<[{}], { a: 1 }> = false;
+expectType<false>(objectIncludes);
+
+const objectIncludesPass: Includes<[{ a: 1 }], { a: 1 }> = true;
+expectType<true>(objectIncludesPass);
+
+// @ts-expect-error
```
sindresorhus, line 31: This should use the `expectError` assertion method instead.

test-d/includes.ts
```diff
@@ -0,0 +1,41 @@
...
+], 'abc'> = false;
+expectType<false>(includesComplexMultiTypeArray);
+
+const noExtendsProblem: Includes<[boolean], true> = false;
+expectType<false>(noExtendsProblem);
+
+const objectIncludes: Includes<[{}], { a: 1 }> = false;
+expectType<false>(objectIncludes);
+
+const objectIncludesPass: Includes<[{ a: 1 }], { a: 1 }> = true;
```
sindresorhus, line 28:
```suggestion
const objectIncludesPass: Includes<[{a: 1}], {a: 1}> = true;
```

sindresorhus (changes requested):

test-d/includes.ts
Yash-Singh1: > This should use the `expectError` assertion method instead.

When you say instead, do you mean without the `// @ts-expect-error`, because I couldn't figure out how to use the `expectError` assertion method works without it. Is this how I am supposed to do it?:

```ts
// @ts-expect-error
const oneArgIncludes: Includes<['my', 'array', 'has', 'stuff']> = false;
expectError<Includes<['my', 'array', 'has', 'stuff']>>(oneArgIncludes);
```

Removing the `// @ts-expect-error`, gives an error (as expected):

```console
✖  34:22  Generic type Includes requires 2 type argument(s).
```

sindresorhus: > When you say instead, do you mean without the // @ts-expect-error

Yes

sindresorhus: You have to put the type inside the method call to `expectError`. `tsd` only ignores TS error inside the `()`.

Yash-Singh1: a10c8b5277cb4e95c05ebc41b2eb8564f0e062b1

PR #218, opened 2021-06-05
add aliases for template literal case types

replygirl:
aliases exports of template literal types for identifier formats (`FooCase` types) with other common names of those wrd cases

#### Why would we care about the identifier name?
- **Familiarity:** I'm never going to remember that "screaming snake case" is a real thing
- **Avoiding frivolous or offensive language:** Kebabs... camels... the community is moving away from this type of labeling
- **Using terms for naming conventions that relate to their primary use case:** For example, it's much easier for some to remember that "param case" is used for params as opposed to "kebab case" (kebabs even aren't always on sticks)
- A team is likely to have a common, domain-agnostic language for naming conventions, that may be influenced by any of the above concerns

### DIFF
--- a/ts41/index.d.ts
+++ b/ts41/index.d.ts
@@ -2,19 +2,68 @@
 export * from '../base';
 
 // These are special types that require at least TypeScript 4.1.
-export {CamelCase} from './camel-case';
-export {CamelCasedProperties} from './camel-cased-properties';
-export {CamelCasedPropertiesDeep} from './camel-cased-properties-deep';
-export {KebabCase} from './kebab-case';
-export {KebabCasedProperties} from './kebab-cased-properties';
-export {KebabCasedPropertiesDeep} from './kebab-cased-properties-deep';
-export {PascalCase} from './pascal-case';
-export {PascalCasedProperties} from './pascal-cased-properties';
-export {PascalCasedPropertiesDeep} from './pascal-cased-properties-deep';
-export {SnakeCase} from './snake-case';
-export {SnakeCasedProperties} from './snake-cased-properties';
-export {SnakeCasedPropertiesDeep} from './snake-cased-properties-deep';
-export {ScreamingSnakeCase} from './screaming-snake-case';
+export {
+	CamelCase,
+	CamelCase as DromedaryCase,
+} from './camel-case';
+export {
+	CamelCasedProperties,
+	CamelCasedProperties as DromedaryCasedProperties,
+} from './camel-cased-properties';
+export {
+	CamelCasedPropertiesDeep,
+	CamelCasedPropertiesDeep as DromedaryCasedPropertiesDeep,
+} from './camel-cased-properties-deep';
+export {
+	KebabCase,
+	KebabCase as DashCase,
+	KebabCase as LispCase,
+	KebabCase as ParamCase,
+} from './kebab-case';
+export {
+	KebabCasedProperties,
+	KebabCasedProperties as DashCasedProperties,
+	KebabCasedProperties as LispCasedProperties,
+	KebabCasedProperties as ParamCasedProperties,
+} from './kebab-cased-properties';
+export {
+	KebabCasedPropertiesDeep,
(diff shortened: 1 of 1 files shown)


sindresorhus:
I appreciate the suggestion, but this is not something I'm interested in. I don't want multiple ways to do the same thing. That just creates inconsistency and confusion.

sindresorhus:
> Familiarity: I'm never going to remember that "screaming snake case" is a real thing

I agree, it's not the best name. We could consider renaming it to `UpperSnakeCase` or `ConstantSnakeCase`. You could open an issue about this if you want.

> Avoiding frivolous or offensive language: Kebabs... camels... the community is moving away from this type of labeling

There's nothing offensive about the current names.

> A team is likely to have a common, domain-agnostic language for naming conventions, that may be influenced by any of the above concerns

All teams are not going to be happy about these aliases either. This is just opening the floodgates for endless amount of aliases and churn.

It's better for *your* team to just import `type-fest` into an utility file and then re-export the methods with your preferred names. Then you're 100% in control.

sindresorhus:
I also do agree that `DashCase` would be better than `KebabCase`.

replygirl:
@sindresorhus Thank you for the feedback. Respectfully, I must have not done a good job of communicating the intent of the proposal, as the feedback implies things about the proposal (that the aliases are my preferred names; that I'm submitting this because a team I'm on wants a symbol renamed; that it's a good idea to rename any of the symbols in question) that I didn't intend for the proposal to represent

> I agree, it's not the best name. We could consider renaming it to `UpperSnakeCase` or `ConstantSnakeCase`. You could open an issue about this if you want.
>
> ...
>
> All teams are not going to be happy about these aliases either. This is just opening the floodgates for endless amount of aliases and churn.

So, sorry for the confusion, but this PR is specifically written to avoid preferential concerns like that—I didn't add _aliases I use that are missing,_ but _aliases that are global standards._ The list is compiled from [Wikipedia](https://en.wikipedia.org/wiki/Naming_convention_(programming)#Examples_of_multiple-word_identifier_formats) and [blakeembrey/change-case](https://github.com/blakeembrey/change-case), not based on brainstorming or surveys/interviews

Debates about what name is best for one convention are exactly what this PR attempts to preclude. I wouldn't submit an issue to _rename_ anything because I personally would strongly object to `UpperSnakeCase` for preferential reasons, and I would very strongly object to `ConstantSnakeCase` because I'm pretty sure you literally just made that up and this library is quite widely used. And that's just one person reacting to suggestions about one convention. The reason these conversations haven't happened yet, as I can see in the comments on #138 and #214, is that the names were determined on the fly without consideration for what was most commonly used—I mean, the author of #214 is in middle school, and while they're quite talented, they probably don't have much exposure to semantics standards. Instead of having the uncomfortable debate, I'm suggesting supporting all of the common names.

I won't be submitting individual issues per your suggestion because of the above & because it would ultimately inconvenience people.

> There's nothing offensive about the current names.

I mean, I'm not offended, but do keep in mind that the [guidance against using "master/slave" and "blacklist/whitelist"](https://tools.ietf.org/id/draft-knodel-terminology-00.html) was published fewer than three years ago. But the point of this PR is you don't need to discuss the merits of any individual term if you give people options

> It's better for your team to just import type-fest into an utility file and then re-export the methods with your preferred names. Then you're 100% in control.

I am not submitting a PR in response to something that's happening on a project or within any team I'm on, just trying to make the package better for more people for the reasons above.

_if this sounds rude it's because i'm autistic so like please be understanding and know i'm just trying to communicate clearly_

sindresorhus:
> but aliases that are global standards

There are no global standards for this. Just data points on which terms are more commonly used.

> just trying to make the package better for more people for the reasons above.

I think we just have different perspectives on what makes the package better for people. For you, it's adding choice, and for me it's limiting choice so people don't have to be bothered to choose. Both are totally valid perspectives.

> if this sounds rude it's because i'm autistic so like please be understanding and know i'm just trying to communicate clearly

For what it's worth, I didn't find this rude.

PR #220, opened 2021-06-15
Fix `CamelCase` behavior when value is in uppercase

My42:
When using `CamelCase` with uppercase value the behavior is incorrect

# Current Behavior

```javascript
const camelFromUppercase: CamelCase<'FOO'> = 'fOO';
```

# Expected Behavior (like [lodash](https://www.npmjs.com/package/lodash) or [camelcase](https://www.npmjs.com/package/camelcase))
```javascript
const camelFromUppercase: CamelCase<'FOO'> = 'foo';
```


sindresorhus:
Can you fix the listing?

My42:
> Can you fix the listing?

Done 👍

PR #226, opened 2021-07-04
Fix `Split` type

HunterKohler:
Consider the input:

```ts
Split<"a", "">
```

You would expect this to give the single length tuple `["a"]` , but it gives `["a", ""]`.

In addition, the default for the delimiter should be `""`, and that has been added.

This is necessary due to the statement in the doc comment about writing a signature for
String.prototype.split, which has `""` as the default delimiter.

> Use-case: Defining the return type of a method like `String.prototype.split`.

### DIFF
--- a/ts41/split.d.ts
+++ b/ts41/split.d.ts
@@ -18,7 +18,11 @@ array = split(items, ',');
 
 @category Template Literals
 */
-export type Split<S extends string, D extends string> =
-	S extends `${infer T}${D}${infer U}`
-		? [T, ...Split<U, D>]
-		: [S];
+export type Split<
+    S extends string,
+    Delimiter extends string = ""
+> = S extends `${infer Si}${Delimiter}${infer Sj}`
+    ? [Si, ...Split<Sj, Delimiter>]
+    : S extends Delimiter
+    ? []
+    : [S];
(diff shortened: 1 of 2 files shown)


sindresorhus:
Can you add some tests?

ts41/split.d.ts
```diff
@@ -18,7 +18,11 @@ array = split(items, ',');
...
-		? [T, ...Split<U, D>]
-		: [S];
+export type Split<
+    S extends string,
+    Delimiter extends string = ""
+> = S extends `${infer Si}${Delimiter}${infer Sj}`
+    ? [Si, ...Split<Sj, Delimiter>]
+    : S extends Delimiter
+    ? []
+    : [S];
```
sindresorhus, line 28: Tab-indentation

ts41/split.d.ts
```diff
@@ -18,7 +18,11 @@ array = split(items, ',');
...
 @category Template Literals
 */
-export type Split<S extends string, D extends string> =
-	S extends `${infer T}${D}${infer U}`
-		? [T, ...Split<U, D>]
-		: [S];
+export type Split<
+    S extends string,
+    Delimiter extends string = ""
+> = S extends `${infer Si}${Delimiter}${infer Sj}`
```
sindresorhus, line 24: Use more descriptive names than `Si` and `Sj`.

HunterKohler:
Added some tests and fixed the style errors.

PR #235, opened 2021-07-29
Add `Join` type

avaly:
This type is useful for advanced types when defining field paths in a nested object (e.g. for dot-notation fields in MongoDB queries).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -52,6 +52,7 @@ export {ScreamingSnakeCase} from './source/screaming-snake-case';
 export {DelimiterCase} from './source/delimiter-case';
 export {DelimiterCasedProperties} from './source/delimiter-cased-properties';
 export {DelimiterCasedPropertiesDeep} from './source/delimiter-cased-properties-deep';
+export {Join} from './source/join';
 export {Split} from './source/split';
 export {Trim} from './source/trim';
 export {Includes} from './source/includes';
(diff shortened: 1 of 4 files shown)


ts41/join.d.ts
```diff
@@ -0,0 +1,22 @@
...
+```
+
+@category Template Literals
+*/
+export type Join<
+	Strings extends string[],
+	Delimiter extends string
+> = Strings extends [] ? '' :
+	Strings extends [string] ? `${Strings[0]}` :
+	// @ts-expect-error `Rest` is inferred as `unknown` here
```
sindresorhus, line 20: Is this a TS limitation or bug?

sindresorhus:
> If there has been any discussion somewhere about this type, include a link to it. For example, a discussion on the TypeScript issue tracker. - https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md

ts41/join.d.ts
avaly: I opened https://github.com/microsoft/TypeScript/issues/45281 for this.

If this is confirmed as intentional and not a TS bug, I'll update the generic type in this type to `unknown[]` instead of `string[]`.

sindresorhus: Can you add a link to that issue as a code comment?

sindresorhus:
https://github.com/sindresorhus/type-fest/pull/235#issuecomment-890348210 

^

avaly:
Initially discussed in https://github.com/microsoft/TypeScript/pull/40336

PR #237, opened 2021-07-30
Add `ExactMatch` type

adrianbrs:
This type can be useful if another type wants to make a decision based on whether one type is exactly the same as the other, with the same keys of the same types.
If you just use the `extends` expression as equality, the result will be `true` even if one of the types has extra keys.

I needed this type to do something like the example below:
```ts
import {ExactMatch} from 'type-fest';

interface Foo {
  a: number;
  b: string;
}

interface Bar extends Foo {
  c: number;
}

type TestOptions<FooLike> = ExactMatch<FooLike, Foo> extends true ? { foo?: FooLike } : { foo: FooLike }

class Test<FooLike extends Foo = Foo> {
  private foo: FooLike = { a: 1, b: 'a' } as FooLike;

  constructor(options: TestOptions<FooLike>) {
    if (options.foo) {
      this.foo = options.foo;
    }
  }
}

// With the default type it is not necessary to specify the options.
const a = new Test({})

// With a custom type you need to pass the options to the constructor
// because the new type has the extra key `c`.
const b = new Test<Bar>({
  foo: {
    a: 1,
    b: 'a',
    c: 2
  }
})
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -119,6 +119,7 @@ Click the type names for complete docs.
 - [`Entries`](source/entries.d.ts) - Create a type that represents the type of the entries of a collection.
 - [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`Asyncify`](source/asyncify.d.ts) - Create an async version of the given function type.
+- [`ExactMatch`](source/exact-match.d.ts) - Returns a `boolean` type to determine if the two types match exactly.
 
 ### Template literal types
 
--- a/readme.md
+++ b/readme.md
@@ -119,7 +119,7 @@ Click the type names for complete docs.
 - [`Entries`](source/entries.d.ts) - Create a type that represents the type of the entries of a collection.
 - [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`Asyncify`](source/asyncify.d.ts) - Create an async version of the given function type.
-- [`ExactMatch`](source/exact-match.d.ts) - Returns a `boolean` type to determine if the two types match exactly.
+- [`ExactMatch`](source/exact-match.d.ts) - Returns a boolean for whether the given two types match exactly.
 
 ### Template literal types
 
--- a/source/exact-match.d.ts
+++ b/source/exact-match.d.ts
@@ -1,0 +1,55 @@
+/**
+Returns a `boolean` type to determine if the two types match exactly.
+
+This can be useful if another type wants to make a decision based on whether one type is exactly the same as the other, with the same keys of the same types.
+
+If you just use the `extends` expression as equality, the result will be `true` even if one of the types has extra keys.
+
+@example
+```
+import {ExactMatch} from 'type-fest';
+
+interface Foo {
+  a: number;
+  b: string;
+}
+
+interface Bar extends Foo {
+  c: number;
+}
+
+type TestOptions<FooLike> = ExactMatch<FooLike, Foo> extends true ? { foo?: FooLike } : { foo: FooLike }
+
+class Test<FooLike extends Foo = Foo> {
+  private foo: FooLike = { a: 1, b: 'a' } as FooLike;
+
+  constructor(options: TestOptions<FooLike>) {
+    if (options.foo) {
+      this.foo = options.foo;
+    }
+  }
+}
+
+// With the default type it is not necessary to specify the options.
+const a = new Test({})
+
+// With a custom type you need to pass the options to the constructor
+// because the new type has the extra key `c`.
+const b = new Test<Bar>({
+  foo: {
+    a: 1,
+    b: 'a',
+    c: 2
+  }
+})
+```
+
+@category Utilities
+*/
+export type ExactMatch<FirstType, SecondType> =
+  FirstType extends SecondType ?
+    Exclude<keyof FirstType, keyof SecondType> extends never ?
+      Exclude<keyof SecondType, keyof FirstType> extends never ? true
+      : false
+    : false
+  : false;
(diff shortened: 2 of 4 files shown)


sindresorhus:
> If there has been any discussion somewhere about this type, include a link to it. For example, a discussion on the TypeScript issue tracker. - https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md

test-d/exact-match.ts
```diff
@@ -0,0 +1,39 @@
...
+declare const exampleTypeWithMoreKeysReverse: ExactMatch<Bar, Foo>;
+expectType<false>(exampleTypeWithMoreKeysReverse);
+
+declare const exampleTypeWithFewerKeys: ExactMatch<Foo, Baz>;
+expectType<false>(exampleTypeWithFewerKeys);
+
+declare const exampleTypeWithFewerKeysReverse: ExactMatch<Baz, Foo>;
+expectType<false>(exampleTypeWithFewerKeysReverse);
+
+declare const exampleTypeExactMatch: ExactMatch<Foo & { c: boolean }, Bar>;
```
sindresorhus, line 29:
```suggestion
declare const exampleTypeExactMatch: ExactMatch<Foo & {c: boolean}, Bar>;
```

sindresorhus:
Since it returns a boolean, I think it should be named `IsExactMatch`.

source/exact-match.d.ts
```diff
@@ -0,0 +1,55 @@
...
+
+@category Utilities
+*/
+export type ExactMatch<FirstType, SecondType> =
+  FirstType extends SecondType ?
+    Exclude<keyof FirstType, keyof SecondType> extends never ?
+      Exclude<keyof SecondType, keyof FirstType> extends never ? true
+      : false
+    : false
+  : false;
```
sindresorhus, line 55: Use tab-indentation.

source/exact-match.d.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+})
+```
+
+@category Utilities
+*/
+export type ExactMatch<FirstType, SecondType> =
+  FirstType extends SecondType ?
+    Exclude<keyof FirstType, keyof SecondType> extends never ?
+      Exclude<keyof SecondType, keyof FirstType> extends never ? true
```
sindresorhus, line 52: Does it actually need this check? Because tests are passing without it. If it's needed, there should be a test to validate that.

```js
export type ExactMatch<FirstType, SecondType> =
  FirstType extends SecondType ?
    Exclude<keyof FirstType, keyof SecondType> extends never
      ? true
      : false
  : false;
```

source/exact-match.d.ts
```diff
@@ -0,0 +1,55 @@
...
+interface Foo {
+  a: number;
+  b: string;
+}
+
+interface Bar extends Foo {
+  c: number;
+}
+
+type TestOptions<FooLike> = ExactMatch<FooLike, Foo> extends true ? { foo?: FooLike } : { foo: FooLike }
```
sindresorhus, line 21:
```suggestion
type TestOptions<FooLike> = ExactMatch<FooLike, Foo> extends true ? {foo?: FooLike} : {foo: FooLike}
```

readme.md
```diff
@@ -119,6 +119,7 @@ Click the type names for complete docs.
 - [`Entries`](source/entries.d.ts) - Create a type that represents the type of the entries of a collection.
 - [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`Asyncify`](source/asyncify.d.ts) - Create an async version of the given function type.
+- [`ExactMatch`](source/exact-match.d.ts) - Returns a `boolean` type to determine if the two types match exactly.
```
sindresorhus, line 122:
```suggestion
- [`ExactMatch`](source/exact-match.d.ts) - Returns a boolean for whether the given two types match exactly.
```

source/exact-match.d.ts
```diff
@@ -0,0 +1,55 @@
+/**
+Returns a `boolean` type to determine if the two types match exactly.
+
+This can be useful if another type wants to make a decision based on whether one type is exactly the same as the other, with the same keys of the same types.
```
sindresorhus, line 4: It would be nice to elaborate more here on specific real-world situation where it could come in handy.

sindresorhus (changes requested):

sindresorhus:
How does this type compare to / differ from https://github.com/sindresorhus/type-fest/pull/217/files#diff-bd6e698dc881ed944085f03f6ab5eddfa0af159b33d9a52c12640439ab309d7bR1-R10 ?

---

// @Yash-Singh1

Yash-Singh1:
By the looks of this, it will fail in some scenarios. I tested and figured the following in the TypeScript Playground:

```ts
ExactMatch<boolean, true>          // true
ExactMatch<any, never>             // true
ExactMatch<'abc', string>          // true
ExactMatch<[] | 'a', []>           // true
ExactMatch<string & 'abc', 'abc'>  // true
ExactMatch<never, never>           // false
```

It might be able to be modified to work with this. However, the solution provided in https://github.com/microsoft/TypeScript/issues/27024#issuecomment-421529650, has only one fallback https://github.com/microsoft/TypeScript/issues/27024#issuecomment-568233987

sindresorhus:
I'm going to review this in detail soon. In the meantime, it would be great if you could help review some of the other open pull requests. 🙏

sindresorhus:
@adrianbrs Still interested in working on this or should I close?

adrianbrs:
I think that just exporting the type (IsEqual) you mentioned should be better. But I ended up using another solution for my problem, then is up to you whether to export it or not.

Anyway, thanks for your response, and congrats for this awesome project! 😀

sindresorhus:
https://github.com/sindresorhus/type-fest/issues/346

PR #242, opened 2021-08-02
Fix tuple index out of bounds type

### LINKED ISSUE #239: Get type works incorrectly with tuples, it treats tuples as arrays and thereby makes it possible to access indexes that are not defined on the tuple.
Get type works incorrectly with tuples, it treats tuples as arrays and thereby makes it possible to access indexes that are not defined on the tuple.

```ts
import { Get } from "type-fest";

type Type = [
    string,
    number
];

// Givs correct type string
type test1 = Get<Type, "0">;

// Givs correct type number
type test2 = Get<Type, "1">;

// Givs correct type unkown
type test3 = Get<Type, "-1">;

// Givs incorrect type string|number
// It should give unkown
type test4 = Get<Type, "2">;

// Givs incorrect type string|number
// It should give unkown
type test4 = Get<Type, "20">;
```

mmkal:
Fixes #239


sindresorhus:
Thanks for fixing ;)

Issue #228, opened 2021-07-08
Add `SetNonNullable` type


sindresorhus:
https://github.com/sindresorhus/type-fest/pull/199

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/sindresorhus) so you can upvote and help fund this issue.
- The funding will be given to active contributors.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/sindresorhus/type-fest/issues/228">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/228/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/sindresorhus/type-fest/issues/228/pledge.svg">
</picture>
</a>

falkenhawk:
looking forward to this to be included 👍

sindresorhus:
https://github.com/sindresorhus/type-fest/pull/245

sindresorhus:
@joealden It should work like `NonNullable`, so not deal with property optionality.

joealden:
I believe this can be closed now that https://github.com/sindresorhus/type-fest/pull/431 has been merged?

kachkaev:
yeah this can be closed I guess

PR #245, opened 2021-08-06
Add `SetNonNullable` type

Smtih:
Adds `SetNonNullable<Type, Keys>` which allows someone to remove nullish values at specific keys of an object. This is useful when defining type guards.

```ts
type Person {
	firstName?: string | null;
	lastName?: string | null;
}

function hasFirstName(person: Person) person is SetNonNullable<Person, "firstName"> {
	return person.firstName != null;
}
```

Continuation of https://github.com/sindresorhus/type-fest/pull/199
Addresses https://github.com/sindresorhus/type-fest/issues/228

If anything is still missing let me know, I've tried to fulfil the contributing guidelines.

### DIFF
--- a/source/set-non-nullable.d.ts
+++ b/source/set-non-nullable.d.ts
@@ -2,7 +2,7 @@ import {Except} from './except';
 import {Simplify} from './simplify';
 
 /**
-Create a type that makes the given keys non-nullable. The remaining keys are kept as is.
+Create a type removes nullish (optional, undefined and null) types from the given keys. The remaining keys are kept as is.
 
 Use-case: You want to define a single model where the only thing that changes is whether or not some of the keys are required.
 
@@ -21,7 +21,7 @@ type Foo = {
 type SomeNonNullable = SetNonNullable<Foo, 'b' | 'c' | 'd' | 'e'>;
 // type SomeNonNullable = {
 // 	a?: number;
-// 	b: string; // Was already non-null and still is.
+// 	b: string; // Was already not nullish and still is.
 // 	c: boolean; // Is now not optional.
 //  d: number; // Is now not null
 //	e: string; // Is now not undefined
--- a/source/set-non-nullable.d.ts
+++ b/source/set-non-nullable.d.ts
@@ -1,0 +1,36 @@
+import {Except} from './except';
+import {Simplify} from './simplify';
+
+/**
+Create a type that makes the given keys non-nullable. The remaining keys are kept as is.
+
+Use-case: You want to define a single model where the only thing that changes is whether or not some of the keys are required.
+
+@example
+```
+import {SetNonNullable} from 'type-fest';
+
+type Foo = {
+	a?: number;
+	b: string;
+	c?: boolean;
+	d: number | null;
+	e: string | undefined;
+}
+
+type SomeNonNullable = SetNonNullable<Foo, 'b' | 'c' | 'd' | 'e'>;
+// type SomeNonNullable = {
+// 	a?: number;
+// 	b: string; // Was already non-null and still is.
+// 	c: boolean; // Is now not optional.
+//  d: number; // Is now not null
+//	e: string; // Is now not undefined
+// }
+```
+*/
+export type SetNonNullable<BaseType, Keys extends keyof BaseType> = Simplify<
+	// Pick just the keys that are optional from the base type.
+	Except<BaseType, Keys> &
+		// For each 'Key' provided, make it not optional or nullable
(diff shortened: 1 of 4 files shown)


source/set-non-nullable.d.ts
```diff
@@ -0,0 +1,54 @@
+import {Except} from './except';
+import {Simplify} from './simplify';
+
+/**
+Create a type that removes nullish (optional, undefined and null) types from the given keys. The remaining keys are kept as is.
+This behaves in a similar manor to `SetRequired<T,K>` by removing optionals, however also applies the `NonNullable<T>` operator to the values at the specified keys.
```
BendingBender, line 6:
```suggestion
Create a type that removes nullish (optional, `undefined` and `null`) types from the given keys. The remaining keys are kept as is.
This behaves in a similar manner to `SetRequired<T,K>` by removing optionals, however also applies the `NonNullable<T>` operator to the values at the specified keys.
```

source/set-non-nullable.d.ts
```diff
@@ -0,0 +1,54 @@
...
+	a?: number;
+	b: string;
+	c?: boolean;
+	d: number | null;
+	e: string | undefined;
+}
+
+type SomeNonNullable = SetNonNullable<Foo, 'b' | 'c' | 'd' | 'e'>;
+// type SomeNonNullable = {
+// 	a?: number;
```
BendingBender, line 24:
```suggestion
// 	a?: number; // Kept as is, was not mentioned in `'b' | 'c' | 'd' | 'e'` type argument.
```

BendingBender (changes requested):
Other than the suggested changes this looks good to me. I have just one principled question: Wouldn't it make more sense to make this type have only one responsibility, just like `SetRequired`? I mean that this type should only remove the `null`/`undefined` from the object values but don't touch the optionality of the keys? This could make a difference especially in the context of the newly introduced [`--exactOptionalPropertyTypes` compiler option](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#exact-optional-property-types). When designed like this, users could mix and match `SetNonNullable`/`SetRequired` as they see fit and we could even introduce a combined type that makes both transformations. @sindresorhus

sindresorhus:
I agree with @BendingBender. The name is also misleading as it contains `NonNullable` in the name, but it does even more than `NonNullable`.

sindresorhus:
@Smtih If you have time, it would be great if you could help review some of the other open pull requests. 🙏

goodoldneon:
+1 to the name concerns. Maybe `SetNonNullish` instead?

source/set-non-nullable.d.ts
```diff
@@ -0,0 +1,54 @@
...
+// }
+
+function hasRequiredKeys(foo: Foo): foo is SomeNonNullable;
+```
+
+This is especially useful in cases where you want to write a type-guard that asserts one or more values of an object are not null.
+
+@example
+```
+type Person {
```
goodoldneon, line 38: Missing `=`

source/set-non-nullable.d.ts
```diff
@@ -0,0 +1,54 @@
...
+	firstName?: string | null;
+	lastName?: string | null;
+}
+
+function hasFirstName(person: Person) person is SetNonNullable<Person, "firstName"> {
```
goodoldneon, line 43:
```suggestion
function hasFirstName(person: Person): person is SetNonNullable<Person, "firstName"> {
```

sindresorhus:
> Maybe SetNonNullish instead?

I don't think the distinction between "nullable" and "nullish" are clear enough.

goodoldneon:
> I don't think the distinction between "nullable" and "nullish" are clear enough.

“Nullish” is the official term for “either `null` or `undefined`“. For example:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

sindresorhus:
@goodoldneon Yes, but TS uses "nullable" for null/undefined in `NonNullable`. So it just makes it slightly different without actually making it any clearer that it also deals with optionality.

goodoldneon:
@sindresorhus good point. It seems like the JavaScript community considers a "nullable" to be a something that can accept either `null` or `undefined`. It's unfortunate that `null` and `undefined` are conflated like that, but it is what it is.

If the goal of the proposed type is to apply the standard library `NonNullable` type over specific keys, then I think `SetNonNullable` is the most intuitive option

sindresorhus:
@Smtih Still interested in finishing this?

PR #247, opened 2021-08-11
Add `ExtendsOr`, `ExtendsAnd`, and `ExtendsThenElse` types

tyler1205:
A few basic utility types that act as `||` and `&&` shorthands. 

The third type is simply an if/then/else conditional which simply mirrors the behavior of typescript's `extends` ternary conditional expression (yeah I know not very useful but it allows for alternative syntax/style).

These utilities could also have other names such as:
`ExtendsOr` -> `Or`
`ExtendsAnd` -> `And`
`ExtendsThenElse` -> `Cond` or `Condition` or `Conditional` or `Ternary` etc.
...but I thought these names were ok.

These are all simple utilities, so how they work should be rather easy to figure out, although I did try to provide some examples (including an example that reflects a problem that originally inspired this idea having to do with Parameters<> erroring if it is not sure that it is receiving a Function).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -34,6 +34,11 @@ export {SetReturnType} from './source/set-return-type';
 export {Asyncify} from './source/asyncify';
 export {Simplify} from './source/simplify';
 export {Jsonify} from './source/jsonify';
+export {
+    ExtendsOr,
+    ExtendsAnd,
+    ExtendsThenElse,
+} from './source/extends-or-and-then-else';
 
 // Template literal types
 export {CamelCase} from './source/camel-case';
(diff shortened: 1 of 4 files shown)


sindresorhus:
@tyler1205 If you have time, it would be great if you could help review some of the other open pull requests. 🙏

sindresorhus:
I'm a little bit conflicted about this. It tries to simplify code with utilities, but we also have to learn the utilities. When I person reading the code sees `ExtendsOr`, they need to look it up, but using `extends` directly, they don't need to do that.

papb:
I am not sure either. I have also needed these helpers in the past in a few situations, but I just defined them myself so anyone reading my code will be able to understand without having to look it up.

sindresorhus:
I have decided to decline this as the benefits are not worth the learnability overhead.

PR #254, opened 2021-08-28
Add `RequireAllOrNone` type

goodoldneon:
Adding a type that requires all or none for given keys.

### DIFF
--- a/source/require-all-or-none.ts
+++ b/source/require-all-or-none.ts
@@ -29,7 +29,7 @@ const responder2: RequireAllOrNone<Responder, 'text' | 'json'> = {
 
 @category Utilities
 */
-export type RequireAllOrNone<ObjectType, KeysType extends keyof ObjectType = keyof ObjectType> = (
+export type RequireAllOrNone<ObjectType, KeysType extends keyof ObjectType = never> = (
     | Required<Pick<ObjectType, KeysType>> // Require all of the given keys.
     | Partial<Record<KeysType, never>> // Require none of the given keys.
 ) &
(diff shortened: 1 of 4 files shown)


source/require-all-or-none.ts
```diff
@@ -0,0 +1,36 @@
...
+};
+```
+
+@category Utilities
+*/
+export type RequireAllOrNone<ObjectType, KeysType extends keyof ObjectType = never> = (
+    | Required<Pick<ObjectType, KeysType>> // Require all of the given keys.
+    | Partial<Record<KeysType, never>> // Require none of the given keys.
+) &
+    Omit<ObjectType, KeysType>; // The rest of the keys.
```
sindresorhus, line 36: Would there be any benefit of using https://github.com/sindresorhus/type-fest/blob/main/source/except.d.ts here?

source/require-all-or-none.ts
```diff
@@ -0,0 +1,36 @@
...
+	json: () => '{"message": "ok"}',
+	secure: true
+};
+```
+
+@category Utilities
+*/
+export type RequireAllOrNone<ObjectType, KeysType extends keyof ObjectType = never> = (
+    | Required<Pick<ObjectType, KeysType>> // Require all of the given keys.
+    | Partial<Record<KeysType, never>> // Require none of the given keys.
```
sindresorhus, line 34: Tab-indentation.

sindresorhus (changes requested):

source/require-all-or-none.ts
goodoldneon: I don't think that'd make a difference since `KeysType extends keyof ObjectType` is already requiring all specified keys to be present

sindresorhus:
I'm going to review this in detail soon. In the meantime, it would be great if you could help review some of the other open pull requests. 🙏

Issue #256, opened 2021-08-29
`Jsonify` recursive cast properties of Date instead of cast to string


leosuncin:
The following code is not transformed in a correct form

```typescript
import type { Jsonify } from 'type-fest';

interface Product {
  title: string;
  price: number;
  category: string;
  description: string;
  image: string;
  createdAt: Date;
  updatedAt: Date;
}

type ProductJson = Jsonify<Product>;
```

![Example error](https://user-images.githubusercontent.com/4307697/131237457-147336b3-8467-401a-89e8-1beb9217ecea.png)

I will expect that properties of type `Date` are converted to `string`, and I suspect that the problem is here

https://github.com/sindresorhus/type-fest/blob/6de66eb8c26cd37adda7213cdbfb5e8246af8328/source/jsonify.d.ts#L52-L54

leosuncin:
What I want is a type when `stringify` a value then `parse` it, so the new value is still typed with the correct type  
  
```typescript
const product: Product = {
  title: 'Fjallraven - Foldsack No. 1 Backpack, Fits 15 Laptops',
  price: 109.95,
  description: 'Your perfect pack for everyday use and walks in the forest.',
  category: "men's clothing",
  image: 'https://fakestoreapi.com/img/81fPKd-2AYL._AC_SL1500_.jpg',
  createdAt: new Date(),
  updatedAt: new Date(),
};

const jsonValues: Jsonify<Product> = JSON.parse(JSON.stringify(product));
```

sindresorhus:
// @darcyparker

darcyparker:
`Jsonify` does not transform a `Date` type to `string type because `Date` is not valid `JSON`.

`Product` must be valid `JSON` to `Jsonify` it... And by inspection it is not JSON because it contains `Date`.  `Jsonify` applies to interfaces that are `JSON`.  It transforms an `interface` into a `type` so that it can be structurally compared to `JsonValue`.  Interfaces cannot be structurally compared to a type because an interface can be re-opened to add properties that may not be satisfy `JsonValue`.

While its true that `JSON.stringify()` can serialize `Date` to a string. `JSON.parse()` is not an inverse because it does not parse the ISO8601String back to `Date`.

For your use case, you either:
* need something different than `JSON.parse()` that understands to inspect `string` to see if it is an ISO8601 string and if so, parse it back to `Date`. (And then you don't need `Jsonify<Product>`)
[...]

PR #257, opened 2021-08-30
Accept non-JSON types with `.toJSON()` method in `Jsonify` type

darcyparker:
Fixes https://github.com/sindresorhus/type-fest/issues/256

@leosuncin pointed out the desire to `Jsonify` non-JSON that is Jsonable via `JSON.parse(JSON.stringify(v))`.  The example provided was an interface `Product` that was clearly not assignable to `JsonValue` because it contains `Date` values.  But `Date` implements a `toJSON()` method so the non-JSON value can be converted to JSON in a predictable way.

This pull request updates `Jsonify` to also transform non-JSON values that are Jsonable.  Where Jsonable means the non-JSON value implements  `.toJSON()` method that returns a value that is assignable to `JsonValue`.

In the tests, `parsedStringifiedX` uses a non-JSON type that currently exists in the tests and is similar to the example @leosuncin provided.  As well, I added a couple of classes of non-JSON objects. (One of the non JSON objects is Jsonable and the other is not)

Some items to reflect on and possibly discuss in this review:
* Should `Jsonify` strictly apply to JSON values only?  Or is it acceptable to include non-JSON values that are Jsonable with `toJSON()`?  
    * This pull request takes the view that it is acceptable to be more flexible and allow non-JSON values that are Jsonable.
       * I reflected on this after #256 was raised.  `Jsonify` was originally designed to transform a `JSON` `interface` to a `type` that is assignable to `JsonValue`.  It was not designed to transform non-JSON.  So my initial feeling was that #256 was not a bug.  But the idea of using `JSON.stringify()`, which leverages the inner object's `.toJSON()` methods, raises a compelling use case.
   * Another choice is to keep `Jsonify` as is... and have a new `JsonifyJsonable` type to handle non-JSON that is Jsonable.
      * But I think having 2 types would be confusing... and I can't think of a use case where it would be necessary to have 2 types.
      * Can others think of a use case where it is helpful to have a separate `JsonifyJsonable` type?  (This is my only doubt because as I approached thinking about #256 my initial feeling was that it should be a separate type... I can't create a strong argument for it though. And I am leaning towards relaxing `Jsonify` to also support non-Json but Jsonable input.)

### DIFF
--- a/source/jsonify.d.ts
+++ b/source/jsonify.d.ts
@@ -4,10 +4,11 @@ import {JsonPrimitive, JsonValue} from './basic';
 type NotJsonable = ((...args: any[]) => any) | undefined;
 
 /**
-Transform a type to one that is assignable to the `JsonValue` type. This includes:
+Transform a type to one that is assignable to the `JsonValue` type.
 
-1. Transforming JSON `interface` to a `type` that is assignable to `JsonValue`
-2. Transforming non JSON that is **Jsonable**, to a type that is assignable to JsonValue. Where **Jsonable** means the non JSON value implements `.toJSON()` method that returns a value that is assignable to `JsonValue`.
+This includes:
+1. Transforming JSON `interface` to a `type` that is assignable to `JsonValue`.
+2. Transforming non-JSON value that is *jsonable* to a type that is assignable to `JsonValue`, where *jsonable* means the non-JSON value implements the `.toJSON()` method that returns a value that is assignable to `JsonValue`.
 
 @remarks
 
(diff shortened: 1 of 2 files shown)


leosuncin:
Maybe call the new type as `ToJsonify` because it will transform the types

darcyparker:
@leosuncin - Is this the name you're suggesting if the existing `Jsonify` goes unchanged and `ToJsonify` is an alternate name to `JsonifyJsonable` that I proposed? And by suggesting this name, do you favour two separate types? Or is the revised `Jsonify` okay with you?

I am hoping people will like be amicable to a single `Jsonify` definition that handles:

1. transforming JSON interfaces to a type that is assignable to `JsonValue`
2. transforming non JSON interfaces that are Jsonable to a type that is assignable to `JsonValue`. Where Jsonable means the non-JSON value implements .toJSON() method that returns a value that is assignable to JsonValue.

If we agree to a single `Jsonify` definition, then we don't need to think about the right name for `JsonifyJsonable`.

But if we really need to separate handling the 2 cases into separate methods, then we need to think about a good name.  I am not sure about `ToJsonify`... It does hint at the Jsonable criteria of having a `toJSON()` method. But I find it hard to read/parse.  `JsonifyJsonable` is easier to read for me because it communicates JSON can be created from it.  If we do need 2 separate methods, I am curious to see what others think the name should be.  I am open to opinions.

sindresorhus:
I think it makes sense to support anything with a `.toJSON()` method, but the docs needs to be updated to make it clear how it works. If someone wants the previous strict behavior, they can complain, and we can then consider a strict version or something of this.

sindresorhus:
// @BendingBender In case you have any opinions. Feel free to ignore.

BendingBender:
I don't have an opinion on this.

darcyparker:
@sindresorhus - Please review the updated documentation.

PR #258, opened 2021-09-06
Support numbers in `Join` type

o-alexandrov:
Thank you for the `Join` type, really helpful.
Let's please support `number` parameters as well, so we could create results like the one below:
- `foo.0.bar`

On a side note, I'm using `Join` type to cast type override on `Array.join()` and this native JavaScript function supports `number` parameters, which leads to believe this `Join` type should support it too.

### DIFF
--- a/source/join.d.ts
+++ b/source/join.d.ts
@@ -20,7 +20,7 @@ const path: Join<[1, 2, 3], '.'> = [1, 2, 3].join('.');
 @category Template Literals
 */
 export type Join<
-	Strings extends (string | number)[],
+	Strings extends Array<string | number>,
 	Delimiter extends string,
 > = Strings extends [] ? '' :
 	Strings extends [string | number] ? `${Strings[0]}` :
--- a/source/join.d.ts
+++ b/source/join.d.ts
@@ -13,10 +13,10 @@ const path: Join<['foo', 'bar', 'baz'], '.'> = ['foo', 'bar', 'baz'].join('.');
 @category Template Literals
 */
 export type Join<
-	Strings extends string[],
+	Strings extends (string | number)[],
 	Delimiter extends string,
 > = Strings extends [] ? '' :
-	Strings extends [string] ? `${Strings[0]}` :
+	Strings extends [string | number] ? `${Strings[0]}` :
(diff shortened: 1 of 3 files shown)


o-alexandrov:
@avaly hello, sorry for the ping, you might want to comment on it, since `Join` type is your addition

sindresorhus:
Can you update the doc comment to indicate that it accepts numbers too?

o-alexandrov:
@sindresorhus thank you for taking time and I apologize for forgetting to update some of the related docs.
in the last commit above, I carefully reviewed missed docs comments and updated them.

source/join.d.ts
```diff
@@ -1,22 +1,22 @@
 /**
-Join an array of strings using the given string as delimiter.
+Join an array of strings and/or numbers using the given string as delimiter.
```
sindresorhus, line 2:
```suggestion
Join an array of strings and/or numbers using the given string as a delimiter.
```

source/join.d.ts
```diff
@@ -1,22 +1,22 @@
...
 
 Use-case: Defining key paths in a nested object. For example, for dot-notation fields in MongoDB queries.
 
 @example
 ```
 import {Join} from 'type-fest';
 
-const path: Join<['foo', 'bar', 'baz'], '.'> = ['foo', 'bar', 'baz'].join('.');
+const path: Join<['foo', '0', 'baz'], '.'> = ['foo', '0', 'baz'].join('.');
```
sindresorhus, line 10: Shouldn't it use number literals here?

source/join.d.ts
o-alexandrov: fixed https://github.com/sindresorhus/type-fest/pull/258/commits/ddf70235718d2e7753be483a23a385abdfdabfb0#diff-665c8ccce5a5adf0c05f3a4fc2ecdb0e16ca90812a3df2bd54fb05eed1ed94c7R2

source/join.d.ts
o-alexandrov: fixed here: https://github.com/sindresorhus/type-fest/pull/258/commits/ddf70235718d2e7753be483a23a385abdfdabfb0#diff-665c8ccce5a5adf0c05f3a4fc2ecdb0e16ca90812a3df2bd54fb05eed1ed94c7R11

Issue #156, opened 2020-11-26
Proposal: `Exact` and `ExactDeep`


stephenh:
Hi! Not sure if you take requests, but I've wanted an `Exact` type in TS for awhile, primarily to model wire calls (i.e. HTTP / DB / RPC calls) where the TS semantics of "including a few extra keys won't hurt" is (imo) not the best b/c it can make the caller _think_ it's data is going over the wire, when really unknown keys are being dropped.

 There is a TS issue about this:

https://github.com/Microsoft/TypeScript/issues/12936

And a few ways of doing a non-deep `Exact` that all look good / work great afaict.

However, at least to my initial efforts/attempts, a `DeepExact` / `ExactDeep` is more nuanced / not available / solved yet, where, similar to your impl of `PartialDeep`, values of map/array/set/etc. all need to be handled differently/recursively.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/sindresorhus) so you can upvote and help fund this issue.
- The funding will be given to active contributors.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/sindresorhus/type-fest/issues/156">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/156/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/sindresorhus/type-fest/issues/156/pledge.svg">
</picture>
</a>

sindresorhus:
Yeah, those sound like useful types for me. It's the kind of types I didn't know I needed.

kainiedziela:
I see no clean way to implement this type. In my mind the perfect solution would be `Exact<T>`, but as of today it would have to be implemented as `Exact<Expected, Actual>`, which I think lessers the usability of this type to be the equivalent of a helper function that strips unwanted props.

I'd love to see someone tackle this, godspeed my friend.

zorji:
Hi @stephenh do you have an example that you need to differentiate `Exact` and `ExactDeep`?

I was trying to implement an `ExactDeep` implementation in https://github.com/sindresorhus/type-fest/pull/259

But this type `Exact` is by default deep already. If there is a true use case of `Exact` instead of `ExactDeep`, I can split them into 2.

As @kainiedziela suggested I don't see there is a way to implement it without `Actual`. So the implementation is `Exact<Expected, Actual>`.

stephenh:
[...]

PR #259, opened 2021-09-08
Add `TypeName` type

zorji:
Exactly is a type to allow a function to reject excess properties.

TypeScript only performs excess property check on object literal only. (see explanation https://stackoverflow.com/a/62358260)

i.e. it does not complain when you assign an object that has more fields than the required type except when it's an object literal.

Here is the demo code

```typescript
type GuardType = {
  allowedField: string
}

type InputType = {
  excessField: string
  allowedField: string
}

const fn = (attr: GuardType) => {
  return attr
}

const aWithType: TypeA = {
  excessField: '',
  allowedField: '',
}

const aWithoutType = {
  excessField: '',
  allowedField: '',
}

// no complains
fn(aWithType)

// no complains
fn({
  ...aWithType,
})

// complains when it's an object literal
// complains excess field is not accepted
fn({
  name: 'name',
  code: 'code',
})

// no complains
fn(aWithoutType)
```

To have a better type check, we can apply this `Exactly<T>` to the above demo (inspired from https://stackoverflow.com/a/59230299)

My implementation also use `KeysOfUnion` from https://stackoverflow.com/a/49402091

```typescript

type GuardType = {
  allowedField: string
}

type InputType = {
  excessField: string
  allowedField: string
}

const aWithType: InputType = {
  excessField: '',
  allowedField: '',
}

const aWithoutType = {
  excessField: '',
  allowedField: '',
}

const fn2 = <T extends Exactly<GuardType, T>>(attr: T) => {
  return attr
}

// complains excess field is not accepted
fn2(aWithType)

// complains excess field is not accepted
fn2({
  ...aWithType,
})

// complains excess field is not accepted
fn2({
  excessField: '',
  allowedField: '',
})

// complains excess field is not accepted
fn2(aWithoutType)
```

The idea is to find all excess properties and mark them as `never`.

# Note

I am still learning how to write test cases with `tsd`. The included test cases in PR are written with `@ts-expect-error` which seems work better in the use case of `Exactly<>`. However, I can try to rewrite the test cases with `tsd` if this type is accepted.


sindresorhus:
Sounds like a useful type.

Please follow: https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md#submitting-a-new-type

sindresorhus:
I think `Exact` would be a better name.

sindresorhus:
The type description and docs text needs more work.

readme.md
```diff
@@ -124,6 +124,7 @@ Click the type names for complete docs.
 - [`Includes`](source/includes.d.ts) - Returns a boolean for whether the given array includes the given item.
 - [`Simplify`](source/simplify.d.ts) - Flatten the type output to improve type hints shown in editors.
 - [`Jsonify`](source/jsonify.d.ts) - Transform a type to one that is assignable to the `JsonValue` type.
+- [`Exactly`](source/exactly.d.ts) - Create a guard for function to reject excess properties.
```
sindresorhus, line 127: The type is not just about guarding functions.

source/exactly.d.ts
```diff
@@ -0,0 +1,45 @@
+import {Primitive} from './primitive';
+
+type KeysOfUnion<T> = T extends T ? keyof T : never;
```
sindresorhus, line 3: Needs a short description.

source/exactly.d.ts
```diff
@@ -0,0 +1,45 @@
...
+
+// however, when u try from a variable, things changed
+const inputWithRequiredPropertyOnly = { code: '' };
+const inputWithOptionalProperty = { code: '', name: '' };
+const inputWithExcessProperty = { code: '', name: '', excessProperty: '' };
+fn2(inputWithRequiredPropertyOnly); // allowed, works!
+fn2(inputWithOptionalProperty); // allowed, works!
+fn2(inputWithExcessProperty); // allowed, NO!!!
+
+ // now with Exactly<>
```
sindresorhus, line 33:
```suggestion
// now with Exactly<>
```

source/exactly.d.ts
```diff
@@ -0,0 +1,45 @@
...
+import {Exactly} from 'type-fest';
+
+// e.g. you want to accept this type in your function
+type ParamType = { code: string, name?: string };
+
+// then you have this function to only accept this type
+const fn1 = (args: ParamType) => args;
+
+// then you use it with object literals
+fn1({ code: '' }); // allowed, works!
```
sindresorhus, line 21:
```suggestion
fn1({code: ''}); // allowed, works!
```

source/exactly.d.ts
```diff
@@ -0,0 +1,45 @@
...
+ // now with Exactly<>
+const fn2 = <T extends Exactly<ParamType, T>>(args: T) => args;
+
+fn2(inputWithRequiredPropertyOnly); // allowed, works!
+fn2(inputWithOptionalProperty); // allowed, works!
+fn2(inputWithExcessProperty); // disallowed, works!
+```
+
+@category Utilities
+ */
```
sindresorhus, line 42:
```suggestion
*/
```

sindresorhus (changes requested):

sindresorhus:
Have you seen https://stackoverflow.com/questions/49580725/is-it-possible-to-restrict-typescript-object-to-contain-only-properties-defined ?

sindresorhus:
Looks like this partly fixes: https://github.com/sindresorhus/type-fest/issues/156

zorji:
Thanks @sindresorhus 

I'll review the other implementation in the SO link u suggested.

zorji:
Reviewing another `Exact` implementation from https://github.com/dotansimha/graphql-code-generator/pull/4194 but looks like it doesn't solve the issue.

zorji:
> Have you seen https://stackoverflow.com/questions/49580725/is-it-possible-to-restrict-typescript-object-to-contain-only-properties-defined ?

Hi @sindresorhus sorry I finally have some time to pick up the work on this again.

I have reviewed the solution in this SO post and surprisingly the concept is exactly the same as my solution in this PR.

So the solution is basically "Find all excess properties from the input type and mark them as never."

Are you suggesting you prefer to use the name from the SO post above?

readme.md
zorji: I can only think of guarding a function because it requires input type from a function input. Could you suggest other usages?

source/exactly.d.ts
zorji: description added.

zorji:
> Sounds like a useful type.
> 
> Please follow: https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md#submitting-a-new-type

Hi @sindresorhus 

I have tried to rewrite tests with `expectError` from `tsd` and functions from `expect-type`. However, I found that both the tests using `tsd` and `expect-type` can perform better tests and explain the intention better, therefore I feel it's more maintainable to keep the tests in the PR.

I have also verified that the tests are executing correctly when running `npx tsd`. e.g. if I remove this line [here](https://github.com/sindresorhus/type-fest/pull/259/files#diff-3733fbf164ea4cdc682f5490dc391fb0fd1a98a69774316a0e9501164ef7a5b4R14) it does trigger an error.

Do you help review and see what else should be improved? Much appreciated.

sindresorhus:
> Are you suggesting you prefer to use the name from the SO post above?

No, `Exact` should be the name.

readme.md
sindresorhus: The descriptions describes *one* of the uses of the type, but the description should describe what it does.

sindresorhus:
> Please help review the other open pull requests. If there are no open pull requests, provide some feedback on some of the open issues. - https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md#submitting-a-new-type

readme.md
```diff
@@ -124,6 +124,7 @@ Click the type names for complete docs.
 - [`Includes`](source/includes.d.ts) - Returns a boolean for whether the given array includes the given item.
 - [`Simplify`](source/simplify.d.ts) - Flatten the type output to improve type hints shown in editors.
 - [`Jsonify`](source/jsonify.d.ts) - Transform a type to one that is assignable to the `JsonValue` type.
+- [`Exact`](source/exact.d.ts) - Create a type from type A and B and convert all keys exclusive to type B as `never`.
```
zorji, line 127: Thanks for the code review @sindresorhus 
I have updated the description to describe what it does instead of what it can do.

zorji:
> > Please help review the other open pull requests. If there are no open pull requests, provide some feedback on some of the open issues. - https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md#submitting-a-new-type

I'll try my best. Currently trying to review #156

source/exact.d.ts
```diff
@@ -0,0 +1,59 @@
...
+> only be the common keys.
+@link https://stackoverflow.com/a/49402091
+*/
+type KeysOfUnion<T> = T extends T ? keyof T : never;
+
+/**
+Create a type from type A and B and convert all keys exclusive to type B as `never`.
+
+This is useful when function wants to guard the arguments to only accept the
+exact defined keys and reject any excess properties.
```
sindresorhus, line 17: Don't hard-wrap.

readme.md
sindresorhus: You are now describing the actual logic of the code. The "what it does" needs to be more high level. Look at the other descriptions for inspiration.

source/exact.d.ts
```diff
@@ -0,0 +1,59 @@
+import {Primitive} from './primitive';
+
+/**
+Returns the accessible keys that also works for union type.
+
+> The reason a simple keyof Union does not work is because keyof always
```
sindresorhus, line 6:
```suggestion
> The reason a simple `keyof Union` does not work is because `keyof` always
```

source/exact.d.ts
```diff
@@ -0,0 +1,59 @@
...
+
+This is useful when function wants to guard the arguments to only accept the
+exact defined keys and reject any excess properties.
+
+Due to the nature of TypeScript, it does not complains if a superset is provided
+as the input of a type unless the provided value is an object literal.
+
+@example
+```
+type OnlyAcceptName = { name: string };
```
sindresorhus, line 24:
```suggestion
type OnlyAcceptName = {name: string};
```

source/exact.d.ts
```diff
@@ -0,0 +1,59 @@
...
+
+// TypeScript complains this because it's an object literal
+onlyAcceptName({ name: "name", id: 1 }); // compilation error
+
+// TypeScript does not complain because it's not an object literal
+const invalidInput = { name: "name", id: 1 };
+onlyAcceptName(invalidInput); // no errors
+```
+
+ Having `Exact` allows TypeScript to reject excess properties
```
sindresorhus, line 35:
```suggestion
Having `Exact` allows TypeScript to reject excess properties.
```

sindresorhus (changes requested):

sindresorhus:
Friendly bump

sindresorhus:

gulsharan:
@sindresorhus can you give an idea when this will be merged/released?

sindresorhus:
> can you give an idea when this will be merged/released?

No, but if you want to see it released, you can help review.

sindresorhus:
I do intend to merge this, but it still requires many docs improvements, and I don't have to do them right now.

zorji:
> I do intend to merge this, but it still requires many docs improvements, and I don't have to do them right now.

Hi @sindresorhus do you mind help point out which doc i can improve on, i will do it asap.

e.g.

- is it the doc missing more examples
- or missing something to explain how it works

the following comments have been addressed but if it still requires more improvement, pls let me know
 - You are now describing the actual logic of the code. The "what it does" needs to be more high level. Look at the other descriptions for inspiration.
 - This type could be moved into internal.d.ts as it could be useful for other types here too.
 - The descriptions describes one of the uses of the type, but the description should describe what it does.

sindresorhus:
> You are now describing the actual logic of the code. The "what it does" needs to be more high level. Look at the other descriptions for inspiration.

You are still describing the code. I think the Stack Overflow answers describes it more clearly:

> which properties are restricted only to those explicitly declared

source/exact.d.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const invalidInput = { name: "name", id: 1 };
+onlyAcceptNameImproved(invalidInput); // compilation error
+```
+
+The solution of `Exact` is
+- take both the preferred type and actual provided type as input.
+- generates the list of keys that exist in the provided type but not in the
+ defined type.
+- mark these excess keys as `never`
```
sindresorhus, line 45: This should be an internal code comment, not in the doc comment.

source/internal.d.ts
```diff
@@ -42,3 +42,12 @@ export type Subtract<A extends number, B extends number> = BuildTuple<A> extends
 Matches any primitive, `Date`, or `RegExp` value.
 */
 export type BuiltIns = Primitive | Date | RegExp;
+
+/**
+Returns the accessible keys that also works for union type.
```
sindresorhus, line 47: No idea what this is trying to say. But from reading the SO answer, I guess you meant something like this:

> Get all the keys from types in a union type.

source/exact.d.ts
```diff
@@ -0,0 +1,52 @@
...
+properties. Due to the nature of TypeScript, it does not complain if excess properties are
+provided unless the provided value is an object literal.
+
+@example
+```
+type OnlyAcceptName = {name: string};
+
+function onlyAcceptName(args: OnlyAcceptName) {}
+
+// TypeScript complains this because it's an object literal.
```
sindresorhus, line 17: Typo

zorji: > > TypeScript complains this because it's an object literal.
> Typo

I am not sure which word has typo but I rewrite this sentence and hopefully this makes more sense.

I have updated the other comments as suggested.

sindresorhus: I would recommend using a grammar checker like Grammarly. It would have caught this.

sindresorhus:
I finally had time to look at this.

Thanks ;)

zorji:
@adam-thomas-privitar Do you mind create a PR for this? I think it can be verified by a test.

JasonShin:
hey @adam-thomas-privitar and @zorji. Sorry first time here

I was playing around with the suggested update but it seems like there's a circular constraint issue. Something weird is following is fine

```
export type Exact<ParameterType, InputType extends ParameterType> = ParameterType extends Primitive
	? ParameterType
	: ParameterType extends [infer U]
		? InputType extends [infer B]
			? Array<Exact<U[0], B[0]>>
			: never
		: { [Key in keyof ParameterType]: Exact<ParameterType[Key], InputType[Key]> } & Record<Exclude<keyof InputType, KeysOfUnion<ParameterType>>, never>;
```

The only significant change that I can see is I'm matching on `[infer T]` instead `Array<any>` and `Array<infer B>`. When I try to match based on `Array` TS would throw a circular constraint error. I'm not sure why TS dislike `Array` pattern in this

Testing with `[infer T]` :
Screen Shot 2022-07-07 at 4 50 33 pm

Testing with matching on `Array<infer T>`

Screen Shot 2022-07-07 at 4 52 18 pm

Also checking exactness on array types wouldn't work in above scenario

zorji:
Hi @JasonShin @adam-thomas-privitar

Do you have a case where the type does not work?

I am trying it with array and it looks like it's working for me.

Screen Shot 2022-07-07 at 7 30 52 pm

```typescript
import { Exact } from 'type-fest'

type Type = Array<{ code: string; name: string }>
const fn = <T extends Exact<Type, T>>(args: T) => args

// correct input
fn([{
  code: 'test',
  name: '1'
}])

// missing field
fn([{
  code: 'test'
}])

// incorrect type
fn([{
  code: 'test',
  name: 1
}])

// excess field
fn([{
  code: 'test',
  name: '1',
  excessField: '',
}])
```

Also, I would suggest open a new issue and link to this PR.

PR #262, opened 2021-09-10
Add `Url2Json` type

Asarua:
Is a type that gets the query string from the URL

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -59,6 +59,7 @@ export {Trim} from './source/trim';
 export {Includes} from './source/includes';
 export {Get} from './source/get';
 export {LastArrayElement} from './source/last-array-element';
+export {Url2Json} from './source/url2json';
 
 // Miscellaneous
 export {PackageJson} from './source/package-json';
--- a/readme.md
+++ b/readme.md
@@ -149,6 +149,7 @@ Click the type names for complete docs.
 - [`Trim`](source/trim.d.ts) - Remove leading and trailing spaces from a string.
 - [`Get`](source/get.d.ts) - Get a deeply-nested property from an object using a key path, like [Lodash's `.get()`](https://lodash.com/docs/latest#get) function.
 - [`LastArrayElement`](source/last-array-element.d.ts) - Extracts the type of the last element of an array.
+- [`Url2Json`](source/url2json.d.ts) - Get query object from string.
 
 ### Miscellaneous
 
--- a/source/url2json.d.ts
+++ b/source/url2json.d.ts
@@ -1,0 +1,43 @@
+/**
+ * Get keys from target
+ */
+type GetQueryKeys<S extends string, D = ''> = S extends `${infer T}=${any}`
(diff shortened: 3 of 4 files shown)


sindresorhus:
What's the use-case?

Asarua:
> What's the use-case?

Sorry, let me complete it

Required: Typescript > 4.1

Here is the demo code

```typescript
const url = 'https://google.com?a=1&b=2'

type QueryString = Url2Json<typeof url>
```

it will be

```typescript
{
    a: "1",
    b: "2"
}
```

sindresorhus:
It's a cute idea, but I don't see how it will work in real-life scenarios. I doesn't and cannot handle escaping, which most URLs have:

```js
const url = 'https://google.com?a=1%20foo&b=2'
```

Also, with most request libraries, you pass the query parameters as an object, not in the actual URL.

Asarua:
I use this method to obtain URL parameters, because many projects I contact need this step. I think I should not be the only one with this requirement.

Similarly, it can also make the type of our request library more perfect.

voxpelli:
I think question is:

Why in that use case is a URL string literal the basis for the types?

Take fastify [as an example](https://github.com/fastify/fastify/blob/main/docs/TypeScript.md#using-generics):

```typescript
interface IQuerystring {
  a: string;
  b: number;
}

server.get<{
  Querystring: IQuerystring,
}>('/auth', async (request, reply) => {
  const { a, b } = request.query
  // The a and b will have the correct types inferred
})
```

The only thing you can infer from a URL string literal query parameters that themselves are string literals, and static query parameters are not that different from the path sections in the URL?

Most query parameters are dynamic, so an approach like Fastify's seems like the more typical approach to this for me (but I'll happily hear more examples to prove me wrong)

(For the sake of completeness, my own approach with Fastify was actually a bit more complex, I sent in a JSON Schema to validate the query parameters and then used [`json-schema-to-typescript`](https://www.npmjs.com/package/json-schema-to-typescript) to generate types from those schemas, thus actually knowing for a fact that my routes never got any other data than the ones my types expected)

sindresorhus:
We are unfortunately not going to accept this type.

Next time, I recommend opening an issue first for discussion so you don't have to have your effort go in vain.

skarab42:
> (For the sake of completeness, my own approach with Fastify was actually a bit more complex, I sent in a JSON Schema to validate the query parameters and then used [`json-schema-to-typescript`](https://www.npmjs.com/package/json-schema-to-typescript) to generate types from those schemas, thus actually knowing for a fact that my routes never got any other data than the ones my types expected)

FI: @voxpelli I have made a [Fastify tRPC adapter](https://trpc.io/docs/fastify) to make this process more easier ;)

Space-yg:
Hi @Asarua, I have been inspired by your request, so for the past couple of weeks, I have been working on it and I made it real here:
- Github: https://github.com/Space-yg/space.types
- npm: https://www.npmjs.com/package/space.types

_(Pssst, it can escape characters too)_

PR #263, opened 2021-09-10
Fix `Except` not working with generic properties

### LINKED ISSUE #232: Except is broken when used with generic property
```ts
class Example {
    foo: number;
    bar: string;

    constructor() {
        this.foo = 1;
        this.bar = '';
    }

    [key: string]: unknown;
}

type Except<ObjectType, KeysType extends keyof ObjectType> = Pick<ObjectType, Exclude<keyof ObjectType, KeysType>>;
const test: Except<Example, 'bar'> = new Example();

// `foo` is now `unknown`
console.log(test.foo + 5);
```

[Playground link](https://www.typescriptlang.org/play?ts=4.4.0-beta#code/MYGwhgzhAECiAeYC2AHEBTaBvAUNf0AZgPbEBc0AdgK5IBG6ATgNx4F1iMUQAujAlpQDmrNvmDFKvRtWA9ijABQBKbGIL4eAC34QAdCWLQAvNACMrDRu269HRiegByJ5YIBfHOoDaAa3QAntx8gkIAuhTUlL6UxADulKyeODwBKJgIwOgoPAA8APJ0AFbocgAqaegANNAA0oEQFenQ6PA86JQAJjD+AcSE0IUl5ZUAfI4ACvzAvgXFpTxN1XDwoNSd6Lm9-YPzI+k19QGNY6OsElI80O28FJnZeQjIaMtO9k7jppTocSvPGCpRAB6IHQAAGhjB0F0VHi4KiMXilDBOAuEGIGD0IGIQkUNx4BlI0AA1NAAKzKZhAA)

szmarczak:
Fixes #232

### DIFF
--- a/source/except.d.ts
+++ b/source/except.d.ts
@@ -1,7 +1,7 @@
 import {IsEqual} from './internal';
 
 /**
-Used to filter keys out from an object.
+Filter out keys from an object.
 
 Returns `never` if `Exclude` is strictly equal to `Key`.
 Returns `never` if `Key` extends `Exclude`.
--- a/source/except.d.ts
+++ b/source/except.d.ts
@@ -21,4 +21,12 @@ type FooWithoutA = Except<Foo, 'a' | 'c'>;
 
 @category Utilities
 */
-export type Except<ObjectType, KeysType extends keyof ObjectType> = Pick<ObjectType, Exclude<keyof ObjectType, KeysType>>;
+
+type EqualsTest<T> = <A>() => A extends T ? 1 : 0;
+type Equals<A1, A2> = EqualsTest<A2> extends EqualsTest<A1> ? 1 : 0;
+
+type Filter<K, I> = Equals<K, I> extends 1 ? never : (K extends I ? never : K);
+
+export type Except<T, I> = {
+	[K in keyof T as Filter<K, I>]: T[K];
+};
--- a/source/except.d.ts
+++ b/source/except.d.ts
@@ -1,6 +1,32 @@
 import {IsEqual} from './includes';
 
 /**
+Returns `never` if `Exclude` is strictly equal to `Key`.
+Returns `never` if `Key` extends `Exclude`.
+Returns `Key` otherwise.
+
+@example
+```
+type Filtered = Filter<'foo', 'foo'>;
+//=> never
+```
+
+@example
+```
+type Filtered = Filter<'bar', string>;
+//=> never
+```
+
+@example
+```
+type Filtered = Filter<'bar', 'foo'>;
+//=> 'bar'
+```
+*/
+
+type Filter<Key, Exclude> = IsEqual<Key, Exclude> extends true ? never : (Key extends Exclude ? never : Key);
+
+/**
 Create a type from an object type without certain keys.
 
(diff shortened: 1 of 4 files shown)


source/except.d.ts
```diff
@@ -21,4 +21,12 @@ type FooWithoutA = Except<Foo, 'a' | 'c'>;
 
 @category Utilities
 */
-export type Except<ObjectType, KeysType extends keyof ObjectType> = Pick<ObjectType, Exclude<keyof ObjectType, KeysType>>;
+
+type EqualsTest<T> = <A>() => A extends T ? 1 : 0;
+type Equals<A1, A2> = EqualsTest<A2> extends EqualsTest<A1> ? 1 : 0;
```
szmarczak, line 26: Should this be a separate type? E.g. `EqualsStrict`?

szmarczak: Simplified

sindresorhus: Looks like this type: https://github.com/sindresorhus/type-fest/blob/70fe5242645071808bc556e8aaa0d2ba9e12991b/source/includes.d.ts#L1-L10

Can you move that one out into `source/internal` and use it for now. We can consider exposing it publicly later on.

sindresorhus: Any reason you're using `1` and `0` here instead of `true` and `false`?

source/except.d.ts
```diff
@@ -21,4 +21,12 @@ type FooWithoutA = Except<Foo, 'a' | 'c'>;
...
+
+type Filter<K, I> = Equals<K, I> extends 1 ? never : (K extends I ? never : K);
```
sindresorhus, line 28: Needs a doc comment.

source/except.d.ts
```diff
@@ -21,4 +21,12 @@ type FooWithoutA = Except<Foo, 'a' | 'c'>;
...
+
+export type Except<T, I> = {
```
sindresorhus, line 30: Descriptive generic type names.

source/except.d.ts
sindresorhus: Also link to https://stackoverflow.com/a/68963796/64949 for explanation of the type.

sindresorhus:
I think you missed some of my comments in https://github.com/sindresorhus/type-fest/pull/263#discussion_r706084325

source/except.d.ts
```diff
@@ -1,3 +1,31 @@
+import {IsEqual} from './includes';
+
+/**
+Returns `never` if `Exclude` is strictly equal to `Key`.
```
sindresorhus, line 4: A description of the purpose of the type would be useful before the technical explanation.

source/except.d.ts
```diff
@@ -1,3 +1,31 @@
...
+//=> never
+```
+
+@example
+```
+type Filtered = Filter<'bar', 'foo'>;
+//=> 'bar'
+```
+*/
+
```
sindresorhus, line 26:
```suggestion
```

source/except.d.ts
szmarczak: Done.

xiaoyu-tamu:
@szmarczak the editor autocomplete feature is disappeared after this commit 

#267

PR #269, opened 2021-09-22
Fix deep properties of `PackageJson` and `TsConfigJson`

zardoy:
All top-level properties are marked as optional. This makes sense and with TS defaults it is possible to assign `undefined`:

```ts
const package: PackageJson = {
    bin: undefined,
}
```
But not when it comes to `Record<string, string>`
```ts
const package: PackageJson = {
    bin: {
        path: undefined // it is required by design, but should be optional
    },
}
```
I've just replaced `Record` types with `Partial<Record>`, because there is [no short way](https://stackoverflow.com/questions/53276792/define-a-list-of-optional-keys-for-typescript-record) of doing that

Maybe we should introduce `OptionalRecord` type or rewrite everything as `{[k: string]?: type}`?

The worst thing happens in `scripts` property. Despite of having `?` in properties, we can't specify `undefined` e.g.

```ts
const b: PackageJson = {
    scripts: {
        install: undefined
    }
}
```
Even VSCode displays that it's possible to assign undefined here, weird.
image

---

Also, I can't see strict errors in ide for test files.
Why `test-d` was excluded from `tsconfig`?

P.S. I understand that as a workaround I could use `PartialDeep<PackageJson>`, but it wouldn't fix individual types such as `Dependency`.


zardoy:
Also, maybe it's time to drop `;`?

voxpelli (commented):
This is somewhat related to #79 and the fact that both `PackageJson` and `TsConfigJson` types should both really be valid `JsonValue` types, as else they are kind of lying about being `Json` types.

And as can be seen in the `JsonValue` definition, `undefined` is not a valid value there: https://github.com/sindresorhus/type-fest/blob/6b6d81ce894517268391db63b3d6538e971c8126/source/basic.d.ts#L15-L45

I guess it can be exemplified by this:

```javascript
const value = {
	xyz: undefined
};

if (value.xyz === undefined) {
	console.log('value.xyz is undefined!');
}
if (Object.prototype.hasOwnProperty.call(value, 'xyz') === false) {
	console.log('value.xyz property does not exist');
}
if (value.abc === undefined) {
	console.log('value.abc is undefined!');
}
if (Object.prototype.hasOwnProperty.call(value, 'abc') === false) {
	console.log('value.abc property does not exist');
}
```

Which prints:

```
value.xyz is undefined!
value.abc is undefined!
value.abc property does not exist
```

And thus show cases that there's a difference between a property being evaluated as `undefined` and the object actually having that property.

And the output of `JSON.parse` can never be like `value.xyz` in that a property exists, but its value is `undefined`.

I'm not sure what @sindresorhus thinks here? I'm leaning more towards moving in the direction of #79

zardoy:
> And the output of JSON.parse can never be like value.xyz in that a property exists, but its value is undefined.

I agree, but on the other hand this is the limitation for `JSON.stringify`.

Just because by TypeScript design there is no difference between property that exist with `undefined` value and property that doesn't exist, that's why `?` accepts both:
```ts
const obj: {a?: string} = {a: undefined}
```

And it makes sense, since `undefined` values would be omitted from `JSON.stringify` output (but only in case if they are values of object, not array)

So, I'm using this trick to remove values that come from other sources e.g. I use `return { ...deps, tsd: undefined }` instead of `delete deps.tsd;return deps`.

Of course, I can use `PartialDeep<PackageJson>` as a workaround here.

---

> And the output of JSON.parse can never be like value.xyz in that a property exists, but its value is undefined.

is that bad? I didn't explicitly add `undefined` to types, I just made them optional (and also undefined-assignable by TS design).

zardoy:
> And as can be seen in the JsonValue definition, undefined is not a valid value there:

Sorry, misread your message. `undefined` is valid here 

```ts
const a: JsonObject = {
    bugs: undefined
}
```

voxpelli:
> by TypeScript design there is no difference between property that exist with `undefined` value and property that doesn't exist, that's why `?` accepts both:

This was true up until TypeScript 4.4 was recently released. 

TypeScript 4.4 introduced [`--exactOptionalPropertyTypes`](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#exact-optional-property-types) which no longer infers `undefined` from optional properties.

Due to #65 it doesn't apply correctly to `JsonObject` right now unfortunately, which I would consider a bug.

You can see it in more [proper action on TS Playground here](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcBiEJwLxwN4DaA1sIgFxwDOMUAlgHYDmAuhdXUwL4DcAULwGMI9anACGFDFlx5eceXABGAV0aUKy+gBNgAMwbAtvTvyEiIAG2AA6CxEYAKMQEpuQA).

voxpelli:
> > And the output of JSON.parse can never be like value.xyz in that a property exists, but its value is undefined.
> 
> is that bad? I didn't explicitly add `undefined` to types, I just made them optional (and also undefined-assignable by TS design).

I think that as long as `PackageJson` is a valid `JsonValue` then I'm fine with it, though until #79 gets fixed that becomes hard to test here.

I guess there might also be a case where we need to ensure that its tested in the `--exactOptionalPropertyTypes` mode, as that's the strictest one in this regard.

sindresorhus:
> Due to #65 it doesn't apply correctly to JsonObject right now unfortunately, which I would consider a bug.

Can you elaborate? It's unclear to me why #65 is a problem for this pull request?

sindresorhus:
@voxpelli Are we good to merge this?

voxpelli (approved):
@sindresorhus Sorry, I've dropped things completely here I see:

My trail of thought was:

1. `PackageJson` should be possible to use wherever a `JsonObject` is requested. That's currently not possible according to #79 
2.  As #272 points out, `{ bugs: undefined }` is not a possible value of a `JsonObject` (though it is a valid object when using `JSON.stringify()`, but will be treated the same as if it was left out completely)
3. Considering #272, this PR would bring us further from #79 than we are today.

Though I now elaborated in https://github.com/sindresorhus/type-fest/issues/272#issuecomment-1075015337 that its like more complex than that, so lets merge this and deal with the #79 and #272 problems in those respective issues.

(As I mentioned in #272, it kind of depends on whether one see the `JsonObject`/ `PackageJson` value to represent something that is valid to go into a `JSON.stringify()` or whether its more strictly something that one could expect back from `JSON.parse()`, but lets leave that to those issues)

source/tsconfig-json.d.ts
```diff
@@ -713,7 +713,7 @@ declare namespace TsConfigJson {
 		/**
 		Specify path mapping to be computed relative to baseUrl option.
 		*/
-		paths?: Record<string, string[]>;
+		paths?: Partial<Record<string, string[]>>;
```
privatenumber, line 716: Why is this changed?


VSCode doesn't allow `undefined` to be set:
Screen Shot 2022-06-07 at 5 00 50 PM


tsc requires a non-empty array value:

Screen Shot 2022-06-07 at 5 01 49 PM

privatenumber: Revert PR: https://github.com/sindresorhus/type-fest/pull/404

PR #277, opened 2021-10-04
Add `MultidimensionalArray` type 

Menecats:
MultidimensionalArray type allows you to type multidimensional array declaratively.

This is useful when a funciton returns a multidimentional array were it's dimensions depends on the input of the functions.
e.g.
```typescript
function createNDimensionalGrid<Depth extends number>(depth: Depth): MultidimensionalArray<number, Depth> {
  // [...]
}

const myGrid = createNDimensionalGrid(2);
const myOtherGrid = createNDimensionalGrid(9);

myGrid[0][5] = 3;
myOtherGrid[0][1][2][3][4][5][6][7][8] = 9;
```

### DIFF
--- a/source/multidimensional-array.d.ts
+++ b/source/multidimensional-array.d.ts
@@ -19,7 +19,9 @@ function emptyMatrix<T extends number>(dimensions: T): MultidimensionalArray<unk
 	const matrix: unknown[] = [];
 
 	let subMatrix = matrix;
-	for (let i = 1; i < dimensions; ++i) {
+	for (let dimension = 1; dimension < dimensions; ++dimension) {
+		console.log(`Initializing dimension #${dimension}`);
+
 		subMatrix[0] = [];
 		subMatrix = subMatrix[0] as unknown[];
 	}
--- a/test-d/multidimensional-array.ts
+++ b/test-d/multidimensional-array.ts
@@ -5,7 +5,9 @@ function createArray<T extends number>(dimensions: T): MultidimensionalArray<unk
 	const root: unknown[] = [];
 
 	let array = root;
-	for (let i = 1; i < dimensions; ++i) {
+	for (let dimension = 1; dimension < dimensions; ++dimension) {
+		console.log(`Initializing dimension #${dimension}`);
+
 		array[0] = [];
 		array = array[0] as unknown[];
 	}
--- a/source/multidimensional-readonly-array.d.ts
+++ b/source/multidimensional-readonly-array.d.ts
@@ -19,9 +19,11 @@ function emptyMatrix<T extends number>(dimensions: T): MultidimensionalReadonlyA
 	const matrix: unknown[] = [];
 
 	let subMatrix = matrix;
-	for (let i = 1; i < dimensions; ++i) {
+	for (let dimension = 1; dimension < dimensions; ++dimension) {
+		console.log(`Initializing dimension #${dimension}`);
+
 		subMatrix[0] = [];
-		if (i < dimensions - 1) {
+		if (dimension < dimensions - 1) {
 			subMatrix = subMatrix[0] as unknown[];
 		} else {
 			subMatrix[0] = 42;
--- a/test-d/multidimensional-readonly-array.ts
+++ b/test-d/multidimensional-readonly-array.ts
@@ -5,9 +5,11 @@ function createArray<T extends number>(dimensions: T): MultidimensionalReadonlyA
 	const root: unknown[] = [];
 
 	let array = root;
-	for (let i = 1; i < dimensions; ++i) {
+	for (let dimension = 1; dimension < dimensions; ++dimension) {
+		console.log(`Initializing dimension #${dimension}`);
+
 		array[0] = [];
-		if (i < dimensions - 1) {
+		if (dimension < dimensions - 1) {
 			array = array[0] as unknown[];
 		} else {
 			array[0] = '42';


source/multidimensional-array.d.ts
```diff
@@ -0,0 +1,51 @@
...
+type Recursive<T> = Array<Recursive<T>>;
+
+// The MultidimensionalArray depends on the following types that come from a blogpost from Ryan Dabler where he explains
+// how we can implement simple arithmetic operations within types.
+// https://itnext.io/implementing-arithmetic-within-typescripts-type-system-a1ef140a6f6f
+type Length<T extends any[]> = T extends {length: infer L}
+	? L
+	: never;
+
+type BuildTuple<L extends number, T extends any[] = []> = T extends {length: L}
```
sindresorhus, line 41:
```suggestion
type BuildTuple<L extends number, T extends readonly unknown[] = []> = T extends {length: L}
```

source/multidimensional-array.d.ts
```diff
@@ -0,0 +1,51 @@
...
+	? T
+	: BuildTuple<L, [...T, any]>;
```
sindresorhus, line 43: Possible to use `unknown` here?

source/multidimensional-array.d.ts
```diff
@@ -0,0 +1,51 @@
...
+	: EQ<Dimensions, 0> extends true
+		? Element
+		: Array<MultidimensionalArray<Element, Subtract<Dimensions, 1>>>;
+
+type Recursive<T> = Array<Recursive<T>>;
+
+// The MultidimensionalArray depends on the following types that come from a blogpost from Ryan Dabler where he explains
+// how we can implement simple arithmetic operations within types.
+// https://itnext.io/implementing-arithmetic-within-typescripts-type-system-a1ef140a6f6f
+type Length<T extends any[]> = T extends {length: infer L}
```
sindresorhus, line 37:
```suggestion
type Length<T extends readonly unknown[]> = T extends {length: infer L}
```

test-d/multidimensional-array.ts
```diff
@@ -0,0 +1,22 @@
+import {expectType} from 'tsd';
+import {MultidimensionalArray} from '../index';
+
+function createArray<T extends number>(_dimensions: T): MultidimensionalArray<unknown, T> {
+	return [];
+}
+
```
sindresorhus, line 7: Can you add a test to ensure it works with readonly arrays?

source/multidimensional-array.d.ts
```diff
@@ -0,0 +1,51 @@
...
+
+@category Utilities
+ */
+export type MultidimensionalArray<Element, Dimensions extends number> = number extends Dimensions
+	? Recursive<Element>
+	: EQ<Dimensions, 0> extends true
+		? Element
+		: Array<MultidimensionalArray<Element, Subtract<Dimensions, 1>>>;
+
+type Recursive<T> = Array<Recursive<T>>;
```
sindresorhus, line 32:
```suggestion
type Recursive<T> = ReadonlyArray<Recursive<T>>;
```

source/multidimensional-array.d.ts
```diff
@@ -0,0 +1,51 @@
+/**
+Create a type that represents a multidimensional array of the given type and dimension.
+
+Use-cases:
+- Declaring n-dimensional array easily and declaratively.
```
sindresorhus, line 5:
```suggestion
- Declaring a n-dimensional array easily and declaratively.
```

source/multidimensional-array.d.ts
```diff
@@ -0,0 +1,51 @@
...
+- Creating n-dimensional array in functions.
```
sindresorhus, line 6:
```suggestion
- Creating a n-dimensional array in functions.
```

sindresorhus: This item is not much different from the previous one. I would try to list some practical specific use-cases.

source/multidimensional-array.d.ts
```diff
@@ -0,0 +1,51 @@
...
+
+@example
+```
+import {MultidimensionalArray} from 'type-fest';
+
+function emptyMatrix<T extends number>(dimensions: T): MultiDimensionalArray<number, T> {
+  // Initialize the matrix in somehow
```
sindresorhus, line 13: "in somehow"?

sindresorhus:
Thanks for contributing a type. Make sure you read and follow: https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md#submitting-a-new-type

Menecats:
I think that having the mutable version of this array is a useful option, in fact in a project that I'm currently working on I need it to be mutable.

If you prefer to have only one type definition instead of two, I've just came up with the following:

```typescript
export type MultidimensionalArray<Element, Dimensions extends number, Readonly extends boolean = true> = number extends Dimensions
	? Recursive<Element, Readonly>
	: IsEqual<Dimensions, 0> extends true
		? Element
		: true extends Readonly
			? ReadonlyArray<MultidimensionalArray<Element, Subtract<Dimensions, 1>, Readonly>>
			: Array<MultidimensionalArray<Element, Subtract<Dimensions, 1>, Readonly>>;

type Recursive<T, Readonly extends boolean> = true extends Readonly
	? ReadonlyArray<Recursive<T, Readonly>>
	: Array<Recursive<T, Readonly>>;


// An example of how a user can use this type

declare const a: MultidimensionalArray<number, 3>; // Readonly
declare const b: MultidimensionalArray<number, 4, true>; // Readonly
declare const c: MultidimensionalArray<number, 5, false>; // Mutable

type MyMutableMultidimensionalArray<T, D extends number> = MultidimensionalArray<T, D, false>;
declare const d: MyMutableMultidimensionalArray<number, 6>; // Mutable
```

This will allow a user to specify weather or not it wants the MultidimensionalArray to be Readonly.
If you think it's a good idea I can proceed changing my pull request by removing the `MultidimensionalReadonlyArray` and updating the `MultidimensionalArray` with these changes

sindresorhus:
Nah. I think it's better with two separate types than a boolean argument.

PR #281, opened 2021-10-10
Improve reference links

kidonng:
- Replace GitHub links (`https://github.com/Microsoft/TypeScript`) with documentation links (`https://www.typescriptlang.org/docs`)

 The GitHub links doesn't provide any other info except source code, while the documentation links contains links to their release and (possibly more updated) examples.

- Improve existing `https://www.typescriptlang.org/` links

  This includes fixing deprecated links and replacing release notes links with their documentation counterpart, for reasons similar to above.


sindresorhus:
Great. Thank you :)

Issue #222, opened 2021-06-17
A strict version of Extract


Vages:
I have been looking for a strict version of `Extract<T, U>`. For simplicity, let's call it `ExtractStrict<T, U>`. `ExtractStrict` would have one additional feature compared to `Extract`: `ExtractStrict` would throw an error if some element in `U` is not part of `T`.

I know of an implementation by @ChristianIvicevic which has been proposed for inclusion in Typescript itself: https://github.com/microsoft/TypeScript/issues/31474 I've asked him if he'd be willing to submit it here instead.

sindresorhus:
Sounds useful. A pull request would be welcome.

Vages:
I've found another use case in my code for a strict version of Exclude too, so I've added it to the title.

Vages:
I'll wait until next Thursday for an answer from @ChristianIvicevic; he deserves the chance to  get some credit for his contribution.

Otherwise, I'll write my own implementation and submit a PR.

ChristianIvicevic:
Hey folks, I'll be taking a look at it this week. Quick question about tests @sindresorhus: I noticed that the tests are used to assert what resulting types look like. Does the project use anything to handle certain test cases that are expected to yield invalid results in order to test erroneous type arguments?

sindresorhus:
Yes, there's an `expectError` assertion: https://github.com/SamVerschueren/tsd#expecterrorfunction

Vages:
Here's my implementation of ExcludeStrict, @ChristianIvicevic (a derivative of your ExtractStrict): `export type ExcludeStrict<T, U extends T> = Exclude<T, U>;`.

younho9:
> Here's my implementation of ExcludeStrict, @ChristianIvicevic (a derivative of your ExtractStrict): `export type ExcludeStrict<T, U extends T> = Exclude<T, U>;`.

Could ExtractStrict be implemented in the same way? `export type ExtractStrict<T, U extends T> = Extract<T, U>;`

Vages:
I'll put up a PR for this in a few hours.

NN---:
You can use ts-essentials package which has these definitions.

https://github.com/ts-essentials/ts-essentials#StrictExtract

Vages:
Thank you for the tip, @NN--- . In that case, I don't see any point in fixing PR #291 :/

sindresorhus:
[...]

PR #291, opened 2021-10-13
Add `ExtractStrict` and `ExcludeStrict` types

Vages:
Solves #222 

I could not figure out how to make test cases to demonstrate that these types fail when you supply an ExcludedUnion or Union that includes an element which is not found in Type. If someone knows how, a contribution is very welcome!

### DIFF
--- a/source/exclude-strict.d.ts
+++ b/source/exclude-strict.d.ts
@@ -1,0 +1,18 @@
+/**
+Constructs a type by excluding from Type all union members that are assignable to ExcludedUnion.
+
+This type is a stricter version of [`Exclude`](https://www.typescriptlang.org/docs/handbook/utility-types.html#excludetype-excludedunion). The `Exclude` type does not restrict the omitted keys to be keys present on the given type, while `ExcludeStrict` does. The benefits of a stricter type are avoiding typos and allowing the compiler to pick up on rename refactors automatically.
+
+@example
+```
+import {ExcludeStrict} from 'type-fest';
+
+type ShirtSizes = 'xxxl' | 'xxl' | 'xl' | 'l' | 'm' | 's' | 'xs' | 'xxs';
+
+type OnlyLargeShirts = ExcludeStrict<ShirtSizes, 'm' | 's' | 'xs' | 'xxs'>;
+//=> 'xxxl' | 'xxl' | 'xl' | 'l';
+```
+
+@category Utilities
+*/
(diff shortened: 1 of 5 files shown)


Vages:
> I think this would be a requirement for this to be merged as otherwise it's not proven that these types actually work

I agree one hundred percent 😅 

> Did you try using expectError?

Yeah, but I did not succeed after trying for 30 minutes. I'll just try again; I'm sure I'll manage if I give it enough time.

I'll come back to fix this this sometime within the next few months.

source/exclude-strict.d.ts
```diff
@@ -0,0 +1,18 @@
...
+```
+import {ExcludeStrict} from 'type-fest';
+
+type ShirtSizes = 'xxxl' | 'xxl' | 'xl' | 'l' | 'm' | 's' | 'xs' | 'xxs';
+
+type OnlyLargeShirts = ExcludeStrict<ShirtSizes, 'm' | 's' | 'xs' | 'xxs'>;
+//=> 'xxxl' | 'xxl' | 'xl' | 'l';
+```
+
+@category Utilities
```
sindresorhus, line 16: I think we should also have a category for utilities that are better versions of builtins. Maybe called `Improved builtin`?

```suggestion
@category Improved builtin
```

Vages: I think that's a great idea. 👍

Vages:
Please excuse me for taking so long to incorporate your suggestions, @sindresorhus. I'm currently finishing a book, and I do not expect to have time to fix it before the middle of February.

sindresorhus:
> and I do not expect to have time to finish this PR before the middle of February.

No worries. I'll leave this open until you have some time.

sindresorhus:
Friendly bump :)

sindresorhus:
https://github.com/sindresorhus/type-fest/issues/222#issuecomment-1190157120

nigeltroy:
Hey y'all, I know that this issue is pretty stale, but I have implemented `ExtractStrict` in [this PR](https://github.com/sindresorhus/type-fest/pull/1119), but I did change some functionality that was provided here: for `ExtractStrict<T, U>`, instead of requiring `U extends T`, I now have `U extends Partial<Exact<T, U>>`. This accomplishes the following:

1. Still validates that the fields in U are present in T (like `U extends T` does)
2. You don't have to provide the ENTIRE subtype to extract a union member, you just have to provide a partial of the subtype that `Extract` can use to discriminate.

I have added examples and test cases in the PR. Let me know what you think!

PR #287, opened 2021-10-12
Add OneOrMore and NOrMore types

darcyparker:
Depends on https://github.com/sindresorhus/type-fest/issues/284 being fixed.

This is still a draft. You can see I inlined @jonahsnider's improved `FixedLengthArray`. (It does not work with existing `FixedLengthArray`)

PR #294, opened 2021-10-15
Fix #284 - Stricter FixedLengthArray

### LINKED ISSUE #284: Stricter `FixedLengthArray`
`FixedLengthArray` could be improved to be stricter. Take this snippet for example:

```ts
const array: FixedLengthArray<number, 3> = [1, 2, 3];

array[4];
```

The final line should not compile since index `4` doesn't exist on the array.

Instead a type something like this could be used:

```ts
type Helper<Element, Length extends number, Rest extends Element[]> = Rest['length'] extends Length ? Rest : Helper<Element, Length, [Element, ...Rest]>;
export type FixedLengthArray<Element, Length extends number> = Length extends Length ? (number extends Length ? Element[] : Helper<Element, Length, []>) : never;
```

This uses TypeScript's tuple types directly instead of `[index: number]: Element` which is much safer.

[Playground for current behavior](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBDAnmApnA3nAYsAHigEwBkUA7AcxgAsBBKKAQ0TgF84AzKCEOAciVQBadigDOMXgG4AUNIDGEUuLgN6TAFzY8hEhWp1GiADykAriABGKKABo4AZgB8cALxwA2gEY7AJjv2AXRlpVUN3ABYgoA)
[Playground for proposed behavior](https://www.typescriptlang.org/play?#code/C4TwDgpgBAEhA2kBOAeAovCBbCA7YANFADJ4DmwAFlBAB7B4AmAzlLgK5YBGESRAShGbAa9JqwzY8wANoBdAHxQAvFEHCZAcky4KlTXNENcLEuSpQA-GqEiAXLATJ0mHPiKldVIjMlvCUAB0werAigDcAFB0YAD2SCKgkFAAYgCWtBCMnnoAgkhIAIYgLlLuZl7UdMamHNy8Sqo5FtXiFXpWUAAUdTxIRm3N1NZ+0vJQDnCIvKX+HuaUPooAlBNsEABuvFGRAMaxuMJQhQXFDumZ2Qv5RSW9vEQAzI1QMgCMRABMT3I7J7cyAAsvyAA)

Menecats:
Hi @sindresorhus, I've implemented the changes requested by the issue #284, but I still have some doubts about them.

There's a "problem" with this implementation, it's the fact that all the mutation methods are now present again in the `FixedLengthArray` type, that said, also in the standard Typescript tuple all those methods are available:
```typescript
const a: [number, number] = [0, 1];
a.push(2);
a[2] = 2; // Error
```
So this may not be a problem.

If instead we want to remove them we can still use the `Omit` utility from typescript as follows:
```typescript
type ArrayLengthMutationKeys = 'splice' | 'push' | 'pop' | 'shift' | 'unshift';

export type FixedLengthArray<Element, Length extends number> = Omit<
	BuildTuple<Length, Element>,
	ArrayLengthMutationKeys
> & {readonly length: Length};
```
The problem with that is that each type you merge a tuple type with another type in Typescript, the tuple usually becomes an array of the tuple type so with these changes we lose the initial strictness:
```typescript
const a: FixedLengthArray<number, 2> = [0, 1];
a.push(2); // Error
a[2] = 2;
```

There is still another option and that's to exclude also the `number` type from our tuple:
```typescript
type ArrayLengthMutationKeys = 'splice' | 'push' | 'pop' | 'shift' | 'unshift';

export type FixedLengthArray<Element, Length extends number> = Omit<
	BuildTuple<Length, Element>,
	number | ArrayLengthMutationKeys
> & {readonly length: Length};
```

This prevents accessing all the mutation methods and keys that are not present in our tuple, but it has another disadvantage, you can't reference tuple elements by numerical index, so that's prevent you to use for loops (unless you specify explicitly the type of the index variable as `0 | 1 | ...`
```typescript
const a: FixedLengthArray<number, 2> = [0, 1];
a[0] = 1; // Ok
a[1] = a[0]; // Ok
a.push(2); // Error
a[2] = 2; // error

for (let index = 0; index < a.length; ++index) {
  a[index] = index; // Error
}
```

What do you think about these changes?

(@darcyparker **fyi**)

### DIFF
--- a/source/fixed-length-array.d.ts
+++ b/source/fixed-length-array.d.ts
@@ -1,10 +1,7 @@
-/**
-Methods to exclude.
-*/
-type ArrayLengthMutationKeys = 'splice' | 'push' | 'pop' | 'shift' | 'unshift';
+import {BuildTuple} from './internal';
 
 /**
-Create a type that represents an array of the given type and length. The array's length and the `Array` prototype methods that manipulate its length are excluded in the resulting type.
+Create a type that represents an array of the given type and length.
 
 Please participate in [this issue](https://github.com/microsoft/TypeScript/issues/26223) if you want to have a similiar type built into TypeScript.
 
@@ -30,11 +27,4 @@ guestFencingTeam.push('Sam');
 
 @category Utilities
 */
-export type FixedLengthArray<Element, Length extends number, ArrayPrototype = [Element, ...Element[]]> = Pick<
-	ArrayPrototype,
-	Exclude<keyof ArrayPrototype, ArrayLengthMutationKeys>
-> & {
-	[index: number]: Element;
-	[Symbol.iterator]: () => IterableIterator<Element>;
-	readonly length: Length;
-};
+export type FixedLengthArray<Element, Length extends number> = BuildTuple<Length, Element>;
--- a/source/fixed-length-array.d.ts
+++ b/source/fixed-length-array.d.ts
@@ -1,4 +1,12 @@
-import {BuildTuple} from './internal';
+/**
+ * Creates a ReadonlyArray<Element> of length Length (aka a tuple)
+ * @private
+ * @see FixedLengthArray which is safer because it tests if `Length` is a specific finite number
+ */
+type BuildTupleHelper<Element, Length extends number, Rest extends Element[]> =
+	Rest['length'] extends Length ?
+		readonly [...Rest] : // Terminate with readonly array (aka tuple)
+		BuildTupleHelper<Element, Length, [Element, ...Rest]>;
 
 /**
 Create a type that represents an array of the given type and length.
@@ -27,4 +35,8 @@ guestFencingTeam.push('Sam');
 
 @category Utilities
 */
-export type FixedLengthArray<Element, Length extends number> = BuildTuple<Length, Element>;
+export type FixedLengthArray<Element, Length extends number> =
+	number extends Length ?
+		// Because `Length extends number` and `number extends Length`, then Length is not a specific finite number
+		Element[] : // It's not fixed length
+		BuildTupleHelper<Element, Length, []>; // Otherwise it is fixed length tuple
--- a/source/internal.d.ts
+++ b/source/internal.d.ts
@@ -18,14 +18,13 @@ Infer the length of the given array `<T>`.
 type TupleLength<T extends readonly unknown[]> = T extends {readonly length: infer L} ? L : never;
 
 /**
-Create a tuple type of the given length `<Length>` and type type `<Element>`.
+Create a tuple type of the given length `<Length>`.
 
 @link https://itnext.io/implementing-arithmetic-within-typescripts-type-system-a1ef140a6f6f
-@link https://github.com/sindresorhus/type-fest/pull/286#discussion_r727799560
 */
-export type BuildTuple<Length extends number, Element = unknown, InputTuple extends readonly unknown[] = []> = InputTuple extends {readonly length: Length}
-	? InputTuple
-	: BuildTuple<Length, Element, [...InputTuple, Element]>;
+type BuildTuple<Length extends number, Rest extends readonly unknown[] = []> = Rest['length'] extends Length
+	? Rest
+	: BuildTuple<Length, [...Rest, unknown]>;
 
 /**
 Create a tuple of length `A` and a tuple composed of two other tuples,
--- a/source/internal.d.ts
+++ b/source/internal.d.ts
@@ -18,13 +18,14 @@ Infer the length of the given array `<T>`.
 type TupleLength<T extends readonly unknown[]> = T extends {readonly length: infer L} ? L : never;
 
 /**
-Create a tuple type of the given length `<L>`.
+Create a tuple type of the given length `<Length>` and type type `<Element>`.
 
 @link https://itnext.io/implementing-arithmetic-within-typescripts-type-system-a1ef140a6f6f
+@link https://github.com/sindresorhus/type-fest/pull/286#discussion_r727799560
 */
-type BuildTuple<L extends number, T extends readonly unknown[] = []> = T extends {readonly length: L}
-	? T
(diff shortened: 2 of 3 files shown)


darcyparker:
This seems reasonable to me:
```typescript
const a: FixedLengthArray<number, 2> = [0, 1];
a.push(2); // Error
a[2] = 2;
```
I expect `FixedLengthArray` will not change length... so `push` etc should be omitted.  And if I wanted to create a new array to push something to, then I would need a new type.  For example:

```typescript
const b: [...typeof a, number] = [...a, 2]; //or concat etc.
```

Also, I am not sure `BuildTuple` is the right way to do it...
The proposal in https://github.com/sindresorhus/type-fest/issues/284 is better in my opinion.
* `type Helper<Element, Length extends number, Rest extends Element[]>` is better than `BuildTuple<Length extends number, Element = unknown, InputTuple extends readonly unknown[] = []>` because
  * `Element` and `Rest` types are aligned.  When you have the default `unknown`, they can get misaligned if say `Element` is provided and `InputTuple` is not.
  * In this case I think having default generics can cause unexpected results if not used correctly. (It is a risk because `BuildTuple` is exported for internal use. If its internal, we don't need defaults if they could be applied incorrectly.)
* I think `Rest['length'] extends Length` reads easier than `InputTuple extends {readonly length: Length}`...  And because `InputTuple` already extends `readonly Element[]`, you don't need to check if `InputTuple` has a length property... it does.
* It's subtle... but breaking out `Helper` is an improvement.  In https://github.com/sindresorhus/type-fest/issues/284 proposed implementation of `FixedLengthArray` has additional checks such as `Length extends Length ? (number extends Length...` which handles the case where `Length` is not finite.  If it is not finite, then FixedLengthArray is basically `Element[]`.
* `Rest` is a better name than `InputTuple` in my opinion... 

Rather than trying to make `BuildTuple` be like helper, why not just keep `Helper` private to `FixedLengthArray`, and for replace `BuildTuple` in `internal.ts` with `FixedLengthArray`.

```typescript
type Helper<Element, Length extends number, Rest extends Element[]> = Rest['length'] extends Length ? Rest : Helper<Element, Length, [Element, ...Rest]>;
export type FixedLengthArray<Element, Length extends number> = Length extends Length ? (number extends Length ? Element[] : Helper<Element, Length, []>) : never;
```

More tests are required...

Here's my take: (Notice I am relying on typescript's tuples and not using `Omit`...
```typescript
/**
 * Creates a ReadonlyArray<Element> of length Length (aka a tuple)
 * @private
 */
type Helper<Element, Length extends number, Rest extends Element[]> =
    Rest['length'] extends Length ?
        // Recursion terminates with readonly tuple.
        // => Note if `Rest extends ReadonlyArray<Element>`, the result is not a tuple here.
        //    I found I had to explicitly destructure it and mark it readonly in this context for
        //    Help<> result to be readonly (ie a tuple)
        readonly [...Rest] :
        Helper<Element, Length, [Element, ...Rest]>; //recurse

export type FixedLengthArray<Element, Length extends number> =
    Length extends Length ?
        (
            number extends Length ?
                // It's not fixed length
                // * Length extends number && number extends Length. Which means Length is not a specific finite number.
                Element[] :
                Helper<Element, Length, []>
        ) :
        never;

const a: FixedLengthArray<number, 2> = [1, 2];
a.push(3); // As expected: typescript reporst Property 'push' does not exist on type 'readonly [number, number]'
```

darcyparker:
@Menecats 
- See suggestions https://github.com/darcyparker/type-fest/commit/8b423c9cfe6f6c799529fd85fd5f4c3ead564c11 if you want to merge that into this pull request
   - I removed the tests `expectAssignable<typeof a.length>(2);  expectNotAssignable<typeof a.length>(3);`, because the assignability of different length tuples is already tested... 
  - It's not necessary to just address this change to `FixedLengthArray`.  Maybe in a separate pull request, `BuildTuple` in `internal.d.ts` should be removed and `Subtract` should use `FixedLengthArray` because it checks `Length` value to make sure its finite.  But doing so causes some tests in `multidimensional-array.ts` and `multidimensional-readonly-array` to fail. You probably have a better understanding of those tests.

darcyparker:
@Menecats - https://github.com/darcyparker/type-fest/commit/bd026d8061fec48001b1bf1a65a3a3388732b930 too... (presuming you like the previous suggestions)

darcyparker:
@Menecats - sounds like we are in agreement about https://github.com/Menecats/type-fest/pull/1 ?  (You'll notice I removed the `Length extends Length ? ... : never` in it too.  It might be easier for @sindresorhus to follow if it is merged into your branch?

darcyparker:
@Menecats - That is a good point about the readonly array.  Length is readonly, but not the tuple's items.  I am thinking about how to deal with that...

sindresorhus:
> Maybe it would make sense to keep FixedLengthArray non-readonly and then also introduce a ReadonlyTuple type?

Alright. Let's go with this.

// @Menecats

sindresorhus:
@Menecats Still interested in finishing this?

sindresorhus:
This is currently blocking https://github.com/sindresorhus/type-fest/pull/287.

Menecats:
Hi @sindresorhus, sorry for the delay I lost the notification for this message..
> > Maybe it would make sense to keep FixedLengthArray non-readonly and then also introduce a ReadonlyTuple type?
> 
> Alright. Let's go with this.
> 
> // @Menecats

Sure, no problem, only one question, should we keep the changes made so far on the FixedLengthArray or it's better to rollback these changes and only implement the new ReadonlyTuple?

darcyparker:
@Menecats - my 2 cents: Leave `FixedLengthArray` as is (non-readonly).  And create new `ReadonlyTuple`.   I think this is what @sindresorhus was saying Jan 8.  Perhaps close this, and create a new pull request for `ReadonlyTuple` that references this pull request.

Issue #213, opened 2021-05-20
Feature Request: add `Paths` type


zhaoyao91:
```ts
//  by @金鑫

type Ensure<X, Y> = X extends Y ? X : never;
type SafeKey<T> = T extends string | number ? `${T}` : never;
type SafeValue<O, K> = K extends keyof O ? O[K] : never;

type Paths<O, P = ""> = O extends object
  ? {
      [K in keyof O]:
        | (P extends "" ? K : `${SafeKey<P>}.${SafeKey<K>}`)
        | Paths<O[K], P extends "" ? K : `${SafeKey<P>}.${SafeKey<K>}`>;
    }[keyof O]
  : never;

type Get<O, P extends Paths<O>> = P extends `${infer X}.${infer XS}`
  ? Get<SafeValue<O, X>, Ensure<XS, Paths<SafeValue<O, X>>>>
  : SafeValue<O, P>;
```

These types help us generate all possible paths of an object, which help build library like `lodash.get` and could be considered as companions of the already exists `Get`.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/sindresorhus) so you can upvote and help fund this issue.
- The funding will be given to active contributors.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/sindresorhus/type-fest/issues/213">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/213/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/sindresorhus/type-fest/issues/213/pledge.svg">
</picture>
</a>

Beraliv:
@zhaoyao91 hey 👋

Do you have any specific requirements for `Get` to have?

I also created an issue in `ts-essentials` and maybe you're interested in this solution and I can start working on it there

WDYT?

sindresorhus:
If anyone wants to work on this, see the initial attempt in https://github.com/sindresorhus/type-fest/pull/300.

sindresorhus:
Thank you @Emiyaaaaa for contributing to close this issue! ⭐

The rewards from this issue, totalling $100, has been shared with you.

**What now?**

1. Create a [Polar](https://polar.sh) account
2. See incoming rewards & setup Stripe to receive them
3. Get payouts as backers finalize their payments

_If you already have a Polar account setup, you don't need to do anything._

[...]

PR #300, opened 2021-10-22
Add `Paths` type

LinusU:
This is loosely based on the code in #213


sindresorhus:
> Hmmm, I'm running in to some problems with this approach, specifically when an object contains Date fields.

I guess this could exclude `Date` and `RegExp` by default. See: https://github.com/sindresorhus/type-fest/pull/342/files

sindresorhus:
Closing as this is not moving forward.

Issue #299, opened 2021-10-21
Title Case


grumpyTofu:
Would you be open to a PR for Title Case?

Ex.
- fooBar -> Foo Bar
- foo-bar -> Foo Bar
- bar_baz -> Bar Baz

PR #303, opened 2021-10-28
Add TitleCase and Titleize

grumpyTofu:
#299 TitleCase

---

Adds TitleCase and Titleize types:
- typeOfAnimal -> Type of Animal (Title Case APA Compliant)
- typeOfAnimal -> Type Of Animal (Titleize)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -57,6 +57,8 @@ export {DelimiterCasedProperties} from './source/delimiter-cased-properties';
 export {DelimiterCasedPropertiesDeep} from './source/delimiter-cased-properties-deep';
 export {Join} from './source/join';
 export {Split} from './source/split';
+export {TitleCase} from './source/title-case';
+export {Titleize} from './source/titleize';
 export {Trim} from './source/trim';
 export {Includes} from './source/includes';
 export {Get} from './source/get';
--- a/readme.md
+++ b/readme.md
@@ -150,6 +150,8 @@ Click the type names for complete docs.
 - [`DelimiterCasedPropertiesDeep`](source/delimiter-cased-properties-deep.d.ts) – Convert object properties to a custom string delimiter casing recursively.
 - [`Join`](source/join.d.ts) - Join an array of strings and/or numbers using the given string as a delimiter.
 - [`Split`](source/split.d.ts) - Represents an array of strings split using a given character or character set.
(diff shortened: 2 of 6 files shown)


grumpyTofu:
@sindresorhus

sindresorhus:
It's cool that you managed to implement APA style titleization, but are there any realistic use-cases for that? The contention in the issue was really the naming, not whether or not to add APA style titleization.

grumpyTofu:
@sindresorhus I think the use case is the same. It will be for data display. I would think it would depend on the data you are displaying. For example, if you had book titles as keys in a hashmap structure in any of the other types of casing, you could type the conversion back to APA format. Otherwise the first one would work fine. I did also make the Title Case option extendable so, it could be configured to skip certain words if desired. 

Either way, just let me know if you like it, or would like something changed and I can make the changes.

sindresorhus:
Since there's no standard for title case, if we decide to keep it, it should be named `TitleCaseAPA` or something similar.

sindresorhus:
So it seems like a common naming for the case when you uppercase the first characters in each word is commonly known as "StartCase", "CapitalizedCase", or "Capitalize".

https://en.wikipedia.org/wiki/Letter_case#Stylistic_or_specialised_usage

sindresorhus:
@voxpelli @jonahsnider Any opinions whether to keep `TitleCaseAPA` and what to call the other simple capitalize case type?

voxpelli:
@sindresorhus I'm myself getting a bit hesitant in adding too many of these capitalization helpers to `type-fest`, at some point it becomes the dominant part of this project and then would maybe be better of added as some kind of `type-capitalization` project?

Especially exotic ones like `TitleCaseAPA` I would rather like to see outside of this project.

We have Camel, Kebab, Pascal and Snake – those are the four most common – and we have two additional helpers for each, `Properties` and `PropertiesDeep` + we have the general helper `Delimiter` which can be used to easily construct even more types of helpers locally in a project.

Especially when seeing the complexity that some of these add, like eg. https://github.com/sindresorhus/type-fest/issues/326

sindresorhus:
After thinking about this for a while, I agree, it's a bit obscure. I have often needed camel-case conversion, etc. But never title case. I recommend publishing this as a separate package and I'd be happy to link to it from the type-fest readme ;)

aglemann:
fwiw this is useful to type the output of lodash's `startCase`.. common use-case is mapping enum values to more friendly representation..

PR #310, opened 2021-11-04
Jsonify: handle undefined in array.

### LINKED ISSUE #309: `Jsonify` removes all optional properties
I assume this isn't the intended behavior.

```ts
type T = Jsonify<{ x?: string }>;
// type T = { x?: undefined }
```

So far I'm working around it by using `Partial<Jsonify<Required<T>>>` rather than `Jsonify<T>`, but I can imagine situations where that's still not desirable (i.e. only some keys are optional, so `Partial` doesn't work).

I haven't looked too far into this, but if I had to guess what's happening:
1. TypeScript automatically adds `undefined` to `x`'s type, giving `string | undefined`.
2. Since `undefined` isn't valid JSON, it fails the `NotJsonable` check, so `Jsonify` returns `{ x?: never }`
3. TypeScript again adds `undefined`, but `undefined | never` collapses to just `undefined`.

It's also worth noting that while `undefined` is not normally valid in JSON, `undefined` in an array becomes `null`.

darcyparker:
Fixes #309

@bbrk24 - Please review and suggest tests if you see anything I missed.

I did not address `Jsonify<{ x?: string }>`.  This seems to be a harder problem.

I experimented with

```typescript
type RequiredKeys<T> = {
	[K in keyof T]-?: (
		{} extends {[P in K]: T[K]} ? // {} extends optional, but not required
		never :
		K
	)
}[keyof T];

type OptionalKeys<T> = {
	[K in keyof T]-?: (
		{} extends {[P in K]: T[K]} ? // {} extends optional, but not required
		K :
		never
	)
}[keyof T];

type JsonifyObject<T extends object> =
	{[P in RequiredKeys<T>]: Jsonify<T[P]>} &
	{[P in OptionalKeys<T>]?: Jsonify<Exclude<T[P], undefined>>;};

```
And adding `JsonifyObject<T` to `Jsonify<T>`...

But as I started to write tests, I realized I have an issue with `JsonObject`'s implementation
```typescript
// Special cases of Object with optional member
declare const objectWithUndefinedValue: {x: undefined};
declare const parsedStringifiedObjectWithUndefinedValue: Jsonify<{x: undefined}>;
expectNotAssignable<JsonValue>(objectWithUndefinedValue); //I expected this to pass
expectAssignable<JsonValue>(parsedStringifiedObjectWithUndefinedValue);
expectAssignable<{}>(parsedStringifiedObjectWithUndefinedValue);

declare const objectWithOptionalValue: {x?: string};
expectAssignable<JsonValue>(objectWithOptionalValue);
```

I feel https://github.com/sindresorhus/type-fest/pull/65 had the right idea and should allow optional key-value pairs, but not a key-value pair where the value is `undefined`.

I need to come back to this... but wanted to leave the above notes if someone else wants to take a crack at it.


test-d/jsonify.ts
```diff
@@ -122,3 +122,15 @@ class NonJsonWithInvalidToJSON {
 const nonJsonWithInvalidToJSON = new NonJsonWithInvalidToJSON();
 expectNotAssignable<JsonValue>(nonJsonWithInvalidToJSON);
 expectNotAssignable<JsonValue>(nonJsonWithInvalidToJSON.toJSON());
+
+// Special cases of Array with `undefined` member
+// `[undefined]` is not JSON because it contains non JSON value `undefined`
+// However `JSON.parse(JSON.stringify())` transforms array members of `undefined` to `null`
+expectNotAssignable<JsonValue>([undefined]);
+declare const parsedStringifiedArrayWithUndefined1: Jsonify<[undefined]>; // = JSON.parse(JSON.stringify([undefined]));
+expectType<null[]>(parsedStringifiedArrayWithUndefined1);
```
bbrk24, line 131: Since it's expecting `null[]`, I think it should test `Jsonify<undefined[]>` for consistency.

bbrk24 (commented):
One minor suggestion for the test, other than that everything looks good.

test-d/jsonify.ts
darcyparker: I am not sure what you mean.

I am verifying `Jsonify<undefined[]>` is the expected `null[]`.

`expectType<Jsonify<undefined[]>(parsedStringifiedArrayWithUndefined1);` would be true if the implementation of `Jsonify` was wrong.

bbrk24: > I am verifying `Jsonify<undefined[]>` is the expected `null[]`.

I am saying that it should do that. However, it currently doesn't test `Jsonify<undefined[]>`, it tests `Jsonify<[undefined]>`.

Ultimately it isn't a big deal, I just didn't see any point in using a tuple.

darcyparker: Ah - I get what you're saying now.  Thanks.  I will make that correction when I get back to this problem.

sindresorhus:
Note that https://github.com/sindresorhus/type-fest/pull/312 was merged.

sindresorhus:
It's unclear to me whether this is still needed after #312?

darcyparker:
I think this is still relevant.  It handles undefined in array correctly.  See #309 where it was noted that "undefined in an array becomes null".  I am sorry this sat in draft.  #312 does solve the other part of #309.

https://github.com/sindresorhus/type-fest/pull/310/files#r743986229 has a small comment about the tests.  I tested `[undefined]` and not the more general `undefined[]`.

darcyparker:
@sindresorhus - Updated and ready for review.

test-d/jsonify.ts
darcyparker: Addressed in [68474d1](https://github.com/sindresorhus/type-fest/pull/310/commits/68474d1fd432c0ee16a94ab0a76bc65da179dba8)

PR #317, opened 2021-11-10
Add strict mode for `Get`

### LINKED ISSUE #313: Proposal: Strict `Get`
The `Get` type is incredibly helpful, so thank you for that!

There is, however, one thing that I wish for. I'm a big fan of defensive typing, and in that regard the `Get` type has the same shortcoming as native `Object[Key]` type expressions do: the resulting type doesn't encode failure when dealing with arrays/dictionaries, even when `noUncheckedIndexedAccess` is enabled. F ex. `Get<string[], "[1000000000000000000000000]">` has the type `string` instead of the more defensive `string | undefined`.

In general this is easy to get around in userland, but looking at the implementation for `Get` it doesn't seem like a local proxy type is a feasible way of solving this.

Would you be open to changing the `Get` type or to adding something like a `GetStrict`? If so then I'd be happy to do the actual work

asbjornh:
Fixes #313 

As discussed in #313 the strict mode is opt-in.
~~I discovered one unfortunate edge case, and I believe that the strict mode should probably be kept as opt-in even in the next major release. The edge case is that if a type has both required named properties _and_ an index signature, the evaluated type includes `undefined`: `Get<{ a: string; [key: string]: string }, "a", true>` incorrectly evaluates to `string | undefined`. I haven't found a way around this and I don't know if it's even possible. It seems like there's no difference between `"a" | string` and `string`, so the compiler possibly just collapses the key type into `string`. If we're unable to fix it we should document it somewhere, but I'm not sure where?~~
Edit: There's one known case where strictness doesn't work: when accessing properties of an object with an index signature, `undefined` is _not_ included in the type when accessing the indexed property (`Get<{ a: string; [key: string]: string }, "foo", { strict: true }>` evaluates to `string`).

I also discovered to my surprise that accessing properties of `Record<number, any>` always evaluates to `unknown`, even before my changes. ~~I've left a failing test that demonstrates this~~. There's a comment on line 104 (in my fork) that says something about being lax with records, which I don't fully understand. Is there a particular reason that `Record<number, any>` shouldn't be allowed?

Edit: I also discovered that array brackets couldn't be used at the start of a path, and also could be repeated (like `"[0][0]"`), so I've included a fix for that too.

### DIFF
--- a/source/get.d.ts
+++ b/source/get.d.ts
@@ -19,18 +19,20 @@ type GetWithPath<BaseType, Keys extends readonly string[], Options extends GetOp
 	>
 	: never;
 
-/** If `Options["strict"]` is `true`, includes `undefined` in the returned type when accessing dictionary properties
+/** Adds `undefined` to `Type` if `strict` is enabled */
+type Strictify<Type, Options extends GetOptions> =
+	Options['strict'] extends true ? Type | undefined : Type;
+
+/** If `Options["strict"]` is `true`, includes `undefined` in the returned type when accessing properties on `Record<string, any>`
  *
  *  Known limitations:
-	- Returns `T | undefined` for required properties on object types with an index signature (f ex. `{ a: string; [key: string]: string }`)
+	- Does not include `undefined` in the type on object types with an index signature (f ex. `{ a: string; [key: string]: string }`)
 */
 type StrictPropertyOf<BaseType, Key extends keyof BaseType, Options extends GetOptions> =
-	Options['strict'] extends true
-	? Record<string, any> extends BaseType
-		? string extends keyof BaseType
-			? BaseType[Key] | undefined // Record<string, any>
-			: BaseType[Key] // Record<"a" | "b", any> (Records with a string union as keys have required properties)
-		: BaseType[Key]
+	Record<string, any> extends BaseType
+	? string extends keyof BaseType
+		? Strictify<BaseType[Key], Options> // Record<string, any>
+		: BaseType[Key] // Record<"a" | "b", any> (Records with a string union as keys have required properties)
 	: BaseType[Key];
 
 /**
@@ -115,9 +117,7 @@ type PropertyOf<BaseType, Key extends string, Options extends GetOptions = {}> =
 	}
 	? (
 		ConsistsOnlyOf<Key, StringDigit> extends true
-		? Options['strict'] extends true
-			? Item | undefined
-			: Item
+		? Strictify<Item, Options>
 		: unknown
 	)
 	: Key extends keyof WithStringKeys<BaseType>
@@ -158,8 +158,8 @@ const getName = (apiResponse: ApiResponse) =>
 	//=> Array<{given: string[]; family: string}>
 
 // Strict mode:
-Get<string[], "3", true> //=> string | undefined
-Get<Record<string, string>, "foo", true> // => string | undefined
+Get<string[], "3", { strict: true }> //=> string | undefined
+Get<Record<string, string>, "foo", { strict: true }> // => string | undefined
 ```
 
 @category Template Literals
--- a/source/get.d.ts
+++ b/source/get.d.ts
@@ -16,7 +16,7 @@ type GetWithPath<BaseType, Keys extends readonly string[], Strict extends boolea
 	: never;
 
 /** If `Strict` is `true`, includes `undefined` in the returned type when accessing dictionary properties
- * 
+ *
  *  Known limitations:
 	- Returns `T | undefined` for required properties on object types with an index signature (f ex. `{ a: string; [key: string]: string }`)
 */
--- a/source/get.d.ts
+++ b/source/get.d.ts
@@ -19,20 +19,23 @@ type GetWithPath<BaseType, Keys extends readonly string[], Options extends GetOp
 	>
 	: never;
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
> The edge case is that if a type has both required named properties and an index signature, the evaluated type includes undefined: Get<{ a: string; [key: string]: string }, "a", true> incorrectly evaluates to string | undefined.

That's quite unfortunate. That means opt-ing into strict mode makes it less strongly typed.

Would there maybe be a way to fix this if we made it only strict by depending on `noUncheckedIndexedAccess`?

asbjornh:
Yes, it's not good.

I don't think this edge case is dependent on or solvable by `noUncheckedIndexedAccess`, because that setting doesn't actually do anything on the type level (it only affects value-level property access). Is there even a way of accessing the ts config without using the file system?

I did some experiments, and it seems like the keys type of an object with an index signature is `string | number`:

```ts
type Simplify<T> = {[KeyType in keyof T]: T[KeyType]};

type A = { a: string; [key: string]: string };
type Keys = Simplify<keyof A>; // string | number
```

Because of this, I don't think there exists a way of identifying objects with named properties and index signatures directly. If there's no way of inspecting the the named properties then I don't see a way of doing conditional stuff here.

I did, however, discover that it's possible to get the opposite behavior (without affecting the behavior for plain records), which to me seems nicer:

```ts
type A = { a: string; [key: string]: string };
type B = Get<A, "a", true>; // string
type C = Get<A, "foo", true>; // string
```

In this case, the strict mode doesn't work when accessing `"foo"` (should be `string | undefined`), but it does return the correct type when accessing `"a"` which is probably better? At least to me it seems like a strict mode that isn't perfectly strict is better than a strict mode that's so strict that it returns incorrect types in some cases

asbjornh:
@mmkal Is there a specific reason that getting properties of `Record<number, any>` always results in `unknown`?

I see that there's a comment about `avoid being too lax with records types using number keys` but I don't really understand why

mmkal:
@asbjornh I might be misunderstanding the question, is this what you mean?

```ts
type X = Get<Record<number, any>, '123'>
```

If so, I guess that evaluates to `unknown` because the type guarantees that a number key will give you the value type of the record (which happens to be `any`), but `Get` is using a string key technically. So we don't know if the property exists, and the default for not found paths is `unknown`.

source/get.d.ts
```diff
@@ -4,13 +4,29 @@ import {Split} from './split';
 /**
 Like the `Get` type but receives an array of strings as a path parameter.
 */
-type GetWithPath<BaseType, Keys extends readonly string[]> =
+type GetWithPath<BaseType, Keys extends readonly string[], Strict extends boolean = false> =
```
mmkal, line 7: Might be worth this being a type level "options bag" so it doesn't become a mess if more options are needed

```ts
interface GetOptions {
  strict?: boolean
}

type GetWithPath<BaseType, Keys extends readonly string[], Options extends GetOptions = {}> =
```

mmkal (commented):
Left a comment about switching to a type level options bag. But it would be much better IMO if this behaviour were automatic with noUncheckedIndexAccess set to true. I haven't tried this and am on mobile rn, but maybe you could evaluate a global variable to replace the passed in `Strict` typearg?

```ts
type NoUncheckedIndexAccess = undefined extends Record<string, string>[string] ? true : false
```

asbjornh:
> @asbjornh I might be misunderstanding the question, is this what you mean?

Sorry, I realize that my example was very clumsy. I'll try again!

```ts
type A = Get<Record<number, string>, "1"> // unknown
type B = Record<number, string>["1"] // string
```

Is the fact that `Get` returns `unknown` in this case (while the native equivalent evaluates to `string`) on purpose?
Also, `Get` works as I would expect when using specific numbers instead of `number`:

```ts
type C = Get<Record<1 | 2 | 3, string>, "2"> // string
```

Just wondering whether this is an uncovered edge case or a feature, really

source/get.d.ts
asbjornh: Yeah good point I agree

asbjornh:
> Left a comment about switching to a type level options bag. But it would be much better IMO if this behaviour were automatic with noUncheckedIndexAccess set to true. I haven't tried this and am on mobile rn, but maybe you could evaluate a global variable to replace the passed in `Strict` typearg?

A solution like that would be ideal, yes, but sadly it's not possible :/
`noUncheckedIndexedAccess` currently doesn't have _any_ effect on the type level. It only does something when you access properties of a _value_ (I was recently very surprised to learn this). This is discussed in multiple issues on the TS GitHub like https://github.com/microsoft/TypeScript/issues/46273 or https://github.com/microsoft/TypeScript/issues/42471. The TS maintainers seem quite adamant that this is the correct approach, so I'm not expecting this to change anytime soon.

source/get.d.ts
asbjornh: This is done now :)

asbjornh:
@mmkal I tweaked the `FixPathSquareBrackets` to support repeated array brackets and array brackets at the beginning of the path. Does it look sensible to you?

sindresorhus (approved):

sindresorhus:
Thanks ;)

asbjornh:
Takk sjæl :)

PR #318, opened 2021-11-13
Add `Nullish` type

jonahsnider:
Implements a type for the nullish values described in the ECMAScript Language Specification:

> “nullish” values (`null` or `undefined`)

([source](https://tc39.es/ecma262/#:~:text=%E2%80%9Cnullish%E2%80%9D%20values%20(null%20or%20undefined)))

### DIFF
--- a/source/basic.d.ts
+++ b/source/basic.d.ts
@@ -43,3 +43,10 @@ Matches any valid JSON value.
 @category Basic
 */
 export type JsonValue = JsonPrimitive | JsonObject | JsonArray;
+
+/**
+Matches a nullish value.
+
+@category Basic
+*/
+export type Nullish = null | undefined;


sindresorhus:
What would you use it for? I'm not sure I want to add this as I generally try to encourage people to move away from using `null` whenever possible: https://github.com/sindresorhus/meta/discussions/7

jonahsnider:
It's only really useful when you're treating `null` or `undefined` as "missing value" and don't want to write `null | undefined` over and over.

```ts
function nullish(value: unknown): value is Nullish {
	return value === null || value === undefined;
}

const data = [1, null, 2, undefined, 3];

if (data.some(nullish)) {
	throw new TypeError('bad data');
}

data.reduce((a, b) => a + b);
```

sindresorhus:
I'm unfortunately going to pass on this type.

sadokmtir:
I dont think, it makes sense to move away from null and just rely on undefined. 

From graphql persepective, having null as explicit value for some input would mark it as deleted and leaving it as undefined would mean no updated happened.
You can see the proposal here:
https://github.com/graphql/graphql-spec/issues/476

So as graphql user, we rely on the differentiation between null and undefined.

sindresorhus:
@sadokmtir There are often better ways to solve problems like this than to use `null`. https://github.com/sindresorhus/meta/discussions/7#discussioncomment-1989583

sadokmtir:
@sindresorhus Can you show me an example or point me to some resources that show what you are proposing ? Because I don't think there are better solutions, otherwise this won't be requested to add it into the graphql-spec.

Saying that other devs are misusing null is never an argument. As inexperienced could misuse anything. So using null for all kind of purposes in a wild is never a valid reason. Null points to something intentional set and undefined means the value was not provided. Those rules are mostly valid to all programming languages.

shanebdavis:
I strongly agree with @sadokmtir. I wish JavaScript didn't have both null and undefined (like Ruby has nil), but since we DO have null and undefined, just ignoring one isn't very helpful. Further, JSON supports both null and undefined conceptually. Null is a critical part of the JSON schema.

The canonical way to use null and undefined correctly in JavaScript is as @sadokmtir says:

> "Null points to something intentional set and undefined means the value was not provided."

It's a common pattern in APIs with an `update` method to allow support passing values with `null` to clear the values in the database OR support not passing the values at all (e.g. `undefined`) to not update them. 

As such, I'd add my support to requesting that 'type-fest', and awesome library BTW, to support ways of working with types that are "nullish" (don't care what the actual word is) - that are either null or undefined.

runninggator:
> @sadokmtir There are often better ways to solve problems like this than to use `null`. [sindresorhus/meta#7 (reply in thread)](https://github.com/sindresorhus/meta/discussions/7#discussioncomment-1989583)

Type-fest not supporting nullish types is quite the disappoint. I see no need to discourage developers from using nulls and [sadokmtir](https://github.com/sadokmtir) makes a great case for its usage. It's simply a matter of fact that null vs undefined is an important distinction for many TS applications. A util like `SetNull` or `SetNullish` would be greatly appreciated by many TS devs. Hopefully this issue can be revisited.

PR #319, opened 2021-11-13
Add numeric range types

jonahsnider:
This PR adds "numeric range" types, `number`/`bigint` types for validating "ranges" of numbers (ex. integers, positive numbers)

Inspired by the Nim programming language's [`Natural`](https://nim-lang.org/docs/system.html#Natural) and [`Positive`](https://nim-lang.org/docs/system.html#Positive) range types.

They can be useful for validating and documenting parameters.
For example, the `BigInt` function could be typed as

```ts
declare function BigInt<T extends number>(number: Integer<T>): bigint;
```

and prevent the following code from being a runtime error:
```js
BigInt(1.5);
//=> throws RangeError
```

Other examples are included in the doc comments.


source/numeric-range.d.ts
```diff
@@ -0,0 +1,46 @@
...
+A `number` that is an integer.
+
+Use-case: Validating and documenting parameters.
+
+@example
+```
+declare function setYear<T extends number>(length: Integer<T>): void;
+```
+*/
+export type Integer<T extends number> = `${T}` extends `${bigint}` ? T : never;
```
sindresorhus, line 11: This is really cool! I have been wanting in integer type for a long time. It would be nice with a short code comment explaining how this works.

sindresorhus:
I think each type should be in separate files.

You also need to add them to the readme.

source/numeric-range.d.ts
```diff
@@ -0,0 +1,46 @@
...
+
+@example
+```
+declare function setQuantity<T extends number>(length: Positive<T>): void;
+```
+*/
+export type Positive<T extends Numeric> = T extends Zero ? never : Negative<T> extends never ? T : never;
+
+/**
+An natural `number`/`bigint` (`[0, ∞)`).
```
sindresorhus, line 37: I would explain for explicitly what a natural number is.

source/numeric-range.d.ts
sindresorhus: Maybe add a caveat to the docs that it will not detect overflowing integer literal.

source/numeric-range.d.ts
jonahsnider: "`[0, ∞)`" is supposed to describe it as an [interval](https://en.wikipedia.org/wiki/Interval_(mathematics)).
I could replace that or add some text of "positive numbers and 0".

sindresorhus: I think it's fine to have it like that, but add another sentence below with:

> A natural number is a positive number or zero.

source/numeric-range.d.ts
jonahsnider: I updated the types to make `Integer` return `never` when given [positive or negative infinity as a type](https://github.com/microsoft/TypeScript/issues/32277).

sindresorhus:
Zero is technically not a positive number, so maybe we should rename `Positive` to `NonNegative`?

source/numeric-range.d.ts
```diff
@@ -0,0 +1,68 @@
+type Infinities = 'Infinity' | '-Infinity';
+
+/**
+A `number` that is an integer.
```
sindresorhus, line 4: I think we should explicitly mention that `BigInt` will not be accepted.

source/numeric-range.d.ts
```diff
@@ -0,0 +1,68 @@
...
+@example
+```
+declare function setYear<T extends number>(length: Integer<T>): void;
+```
+
+@category Utilities
+*/
+// `${bigint}` is a type that matches a valid bigint literal without the `n` (ex. 1, 0b1, 0o1, 0x1)
+// Because T is a number and not a string we can effectively use this to filter out any numbers containing decimal points
+export type Integer<T extends number> = `${T}` extends Infinities ? never : `${T}` extends `${bigint}` ? T : never;
```
sindresorhus, line 17: I wonder how this technique compares to https://stackoverflow.com/a/69413070/64949

sindresorhus:
Would you be willing to add a `NonNegativeInteger` too? I think it's quite a common need.

sindresorhus:
These types are not just enforcing ranges though, but also types. Maybe it would be more correct to name the file `numeric.d.ts`?

sindresorhus:
Note to self:
- `isInteger` method with type-guard for https://github.com/sindresorhus/ts-extras
- `isSafeInteger`

sindresorhus:
Should we add a `FiniteNumber` type too?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite

sindresorhus:
I'm also wondering whether we should be explicit in the types: `Positive` => `PositiveNumber`?

sindresorhus (changes requested):

source/numeric-range.d.ts
```diff
@@ -0,0 +1,68 @@
+type Infinities = 'Infinity' | '-Infinity';
+
+/**
+A `number` that is an integer.
+
+Use-case: Validating and documenting parameters.
+
```
sindresorhus, line 7: Are there any open TS issues about adding an integer type that we could link to here?

source/numeric-range.d.ts
```diff
@@ -0,0 +1,68 @@
...
+@example
+```
+declare function setYear<T extends number>(length: Integer<T>): void;
```
sindresorhus, line 10: Include import statement in all the examples.

source/numeric-range.d.ts
```diff
@@ -0,0 +1,68 @@
...
+
+@category Utilities
+*/
+// `${bigint}` is a type that matches a valid bigint literal without the `n` (ex. 1, 0b1, 0o1, 0x1)
+// Because T is a number and not a string we can effectively use this to filter out any numbers containing decimal points
+export type Integer<T extends number> = `${T}` extends Infinities ? never : `${T}` extends `${bigint}` ? T : never;
+
+type Numeric = number | bigint;
+
+type Zero = 0 | 0n;
```
sindresorhus, line 21: Put all the utility types at the top.

source/numeric-range.d.ts
jonahsnider: https://github.com/microsoft/TypeScript/issues/43505 seems to be the best fit

jonahsnider:
> Zero is technically not a positive number, so maybe we should rename `Positive` to `NonNegative`?

`Positive<0>` is `never`.

> Would you be willing to add a `NonNegativeInteger` too? I think it's quite a common need.

I have added `PositiveInteger<T>` and `NegativeInteger<T>` types.

> These types are not just enforcing ranges though, but also types. Maybe it would be more correct to name the file `numeric.d.ts`?

Renamed.

> Should we add a `FiniteNumber` type too?
> 
> [developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite)

I have added a `Finite<T>` type which accepts a `number` (since `bigint`s are always finite).

> I'm also wondering whether we should be explicit in the types: `Positive` => `PositiveNumber`?

I would advise against this as those types can accept a `number | bigint`, putting `Number` in the name might be misleading.

sindresorhus:
> Positive<0> is never.

When do you ever need only a positive number though? I always want zero too. I also fear the type will be confusing to users as not everyone knows that a "positive" number does not include zero.

sindresorhus:
> Positive<0> is never.

Also, then the docs you have provided are confusing as you show the range `(0, ∞)`, which is read as inclusive, meaning from zero to infinity.

sindresorhus:
> I have added PositiveInteger<T> and NegativeInteger<T> types.

I don't really see the usefulness of these. `NonNegativeInteger` is useful as it includes zero. I don't think it's common to need an integer that is from 1 and up, so I think that's better left to composition with the `Integer` and `Positive` type.

sindresorhus:
Also, the exact definition of "natural number" seems to be unclear:

> Some definitions, including the standard ISO 80000-2,[3][a] begin the natural numbers with 0, corresponding to the non-negative integers 0, 1, 2, 3, ..., whereas others start with 1, corresponding to the positive integers 1, 2, 3, ...[4][b] Texts that exclude zero from the natural numbers sometimes refer to the natural numbers together with zero as the whole numbers, while in other writings, that term is used instead for the integers (including negative integers).[5]
> - https://en.wikipedia.org/wiki/Natural_number

sindresorhus:
> I have added a Finite<T> type which accepts a number

Needs to be added to the readme. There are some other missing types in the readme too.

sindresorhus:
I suggest we rename `Natural` to `NonNegative`. It will then be coherent with `NonNegativeInteger`.

source/numeric.d.ts
```diff
@@ -0,0 +1,149 @@
...
+@see Negative
+
+@example
+```
+import {Natural} from 'type-fest';
+
+declare function setLength<T extends number>(length: Natural<T>): void;
+```
+
+@category Utilities
```
sindresorhus, line 148: Should we maybe make a new `Numeric` category?

jonahsnider:
> When do you ever need only a positive number though? I always want zero too. I also fear the type will be confusing to users as not everyone knows that a "positive" number does not include zero.

I think there are some use cases for 1 or more vs 0 or more (mostly just enforcing a minimum value).

> Also, then the docs you have provided are confusing as you show the range `(0, ∞)`, which is read as inclusive, meaning from zero to infinity.

That's not correct, in range syntax a `(` is exclusive while a `[` is inclusive. Probably would be best to just rewrite this as an inequality since they're harder to misinterpret (ex. `0 < x < ∞`).

> I suggest we rename `Natural` to `NonNegative`. It will then be coherent with `NonNegativeInteger`.

I'm going to rename the types to have more useful names rather than follow strict mathematical definitions. It's clear that trying to apply those concepts to software ends up being unnecessarily confusing.

sindresorhus:
> I think there are some use cases for 1 or more vs 0 or more (mostly just enforcing a minimum value).

Sure, but it's not a common case.

> That's not correct, in range syntax a ( is exclusive while a [ is inclusive.

I don't think there's any common standard for this. Personally, I'm mostly familiar with range syntax from Swift which is inclusive: `0...4` meaning 0 to 4, including 0 and 4. But yes, let's make it completely clear.

jonahsnider:
> Sure, but it's not a common case.

Fair enough, I have removed the `Positive` and `PositiveInteger` types.

sindresorhus:
Very nice work. Thanks for adding these 👍

PR #320, opened 2021-11-17
Fix `DelimiterCase` type incorrectly converting string to empty string

### LINKED ISSUE #293: SnakeCasedPropertiesDeep Incorrectly Converts Record<String, _>
When using `SnakeCasedPropertiesDeep` type, there is an apparent bug if an interface has a property of `Record<string, anyType>`.  It will enforce the string key is an empty string `""`.  The below codesandbox is a reproducible example with type-fest  version `2.5.0`.  I spent a while looking through the source code but couldn't find the source of the bug.  

https://codesandbox.io/s/fancy-wave-jmb4t?file=/src/index.ts

By the way, thanks for an awesome library!

chentsulin:
Fix #293

截圖 2021-11-17 下午3 27 40


voxpelli (approved):
Looks good to me, thanks 🙏

PR #321, opened 2021-11-17
Add `Schema` type

### LINKED ISSUE #103: Proposal: DeepRecord
A while back while working on a project we needed a generic function to slice all strings of the provided object.
```ts
const stringSlicer = <T, U> (data: T, checker: U): T => {};
```
We wanted the `U` type the have the same type as the first argument but it had to flatten all the arrays. 
So if T was of type
```Typescript
interface T {
	name: string;
	items: {
 		id: string
	}[]
}
```
U had to be of type
```Typescript
interface U {
	name: number;
	items: {
		id: number;
	}
}
```
With interface `U` defining the length of each string. But it shouldn't provide an array for `items`

**proposal**
```Typescript
export type DeepRecord<K, T> = {
	[P in keyof K]:
		Exclude<K[P], undefined> extends Array<infer E> ? DeepRecord<E, T> :
		Exclude<K[P], undefined> extends object ? DeepRecord<K[P], T> :
		T;
};
```
This way we can write the function as following
```ts
const stringSlicer = <T> (data: T, checker: DeepRecord<T, number>): T => {};
```

Tellios:
`Schema` allows re-using the structure of an interface and applying new values to it. This is commonly seen in validation where libraries like [joi](https://www.npmjs.com/package/joi) and [yup](https://www.npmjs.com/package/yup) when used with TypeScript replaces interface values with validation schema types.

The intention of `Schema` is to make it easier to provide similar workflows which also support other use-cases like masking data or providing config to parsing.

The schema is recursive/deep which means that it will dig down objects to provide the same value on all properties.

`Schema` might be better named as `RecordDeep` since it is very similar to a TypeScript `Record<TObject, ValueType>`. But compared to a `Record` the `Schema` traverses the objects in the tree.

**Example:**

```ts
import { Schema } from 'type-fest';

interface Foo {
  array: any[];
  primitive: boolean;
  object: {
    primitive: boolean;
  }
}

type FooSchema = Schema<Foo, 'A'>;

// FooSchema is now the same as this interface
interface FooSchema {
  array: 'A';
  primitive: 'A';
  object: 'A' | { primitive: 'A' }; // For an ordinary Record this would only be 'A'
};
```

**Real life use-case:**

I use it to apply behavior for how data should be outputted in logs. So the below code is somewhat similar to actual code.

```ts
interface HttpRequest {
  url: string;
  headers: {
    authorization: string;
    contentType: string;
  };
  body: {
    userId: string;
    settings: {
      foo: boolean;
      bar: boolean;
    };
  };
};

type SchemaOptions =
  'hide'
  | 'hash' 
  | 'show' // default behavior

const schema: PartialDeep<Schema<HttpRequest, 'hide' | 'hash' | 'show'>> = {
  headers: {
    authorization: 'hide',
  },
  body: {
    userId: 'hash',
  },
};

const request: HttpRequest = { /* create request */ };

log.debug(sanitizeObject(request, schema));
```

---

Fixes #103 
Closes #354

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -118,7 +118,7 @@ Click the type names for complete docs.
 - [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`Simplify`](source/simplify.d.ts) - Useful to flatten the type output to improve type hints shown in editors. And also to transform an interface into a type to aide with assignability.
 - [`Get`](source/get.d.ts) - Get a deeply-nested property from an object using a key path, like [Lodash's `.get()`](https://lodash.com/docs/latest#get) function.
-- [`Schema`](source/schema.d.ts) - Recursively replaces all the property values into a given value type for an object.
+- [`Schema`](source/schema.d.ts) - Create a deep version of another object type where property values are recursively replaced into a given value type.
 
 ### JSON
 
@@ -201,6 +201,7 @@ Click the type names for complete docs.
 *If you know one of our types by a different name, add it here for discovery.*
 
 - `PartialBy` - See [`SetOptional`](https://github.com/sindresorhus/type-fest/blob/main/source/set-optional.d.ts)
+- `RecordDeep`- See [`Schema`](https://github.com/sindresorhus/type-fest/blob/main/source/schema.d.ts)
 
 ## Tips
 
--- a/readme.md
+++ b/readme.md
@@ -118,7 +118,7 @@ Click the type names for complete docs.
 - [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`Simplify`](source/simplify.d.ts) - Useful to flatten the type output to improve type hints shown in editors. And also to transform an interface into a type to aide with assignability.
 - [`Get`](source/get.d.ts) - Get a deeply-nested property from an object using a key path, like [Lodash's `.get()`](https://lodash.com/docs/latest#get) function.
-- [`Schema`](source/schema.d.ts) - Deeply changes the values for all properties on a type to whatever specified. Useful for code e.g validation or configuration that relies on using the same structure of a type but the values are different.
+- [`Schema`](source/schema.d.ts) - Recursively replaces all the property values into a given value type for an object.
 
 ### JSON
 
--- a/readme.md
+++ b/readme.md
@@ -129,6 +129,7 @@ Click the type names for complete docs.
 - [`Includes`](source/includes.d.ts) - Returns a boolean for whether the given array includes the given item.
 - [`Simplify`](source/simplify.d.ts) - Useful to flatten the type output to improve type hints shown in editors. And also to transform an interface into a type to aide with assignability.
 - [`Jsonify`](source/jsonify.d.ts) - Transform a type to one that is assignable to the `JsonValue` type.
+- [`Schema`](source/schema.d.ts) - Deeply changes the values for all properties on a type to whatever specified. Useful for code e.g validation or configuration that relies on using the same structure of a type but the values are different.
 
 ### Template literal types
 
--- a/source/schema.d.ts
+++ b/source/schema.d.ts
@@ -40,28 +40,28 @@ const userMaskSettings: UserMask = {
 @category Utilities
 */
 export type Schema<TObject, ValueType> = TObject extends string
-	? ValueType
-	: TObject extends Map<unknown, unknown>
-	? ValueType
-	: TObject extends Set<unknown>
-	? ValueType
-	: TObject extends ReadonlyMap<unknown, unknown>
-	? ValueType
-	: TObject extends ReadonlySet<unknown>
-	? ValueType
+  ? ValueType
+  : TObject extends Map<unknown, unknown>
+  ? ValueType
+  : TObject extends Set<unknown>
+  ? ValueType
+  : TObject extends ReadonlyMap<unknown, unknown>
+  ? ValueType
+  : TObject extends ReadonlySet<unknown>
+  ? ValueType
   : TObject extends readonly unknown[]
   ? ValueType
-	: TObject extends unknown[]
-	? ValueType
-	: TObject extends (...arguments: unknown[]) => unknown
-	? ValueType
+  : TObject extends unknown[]
+  ? ValueType
+  : TObject extends (...arguments: unknown[]) => unknown
+  ? ValueType
   : TObject extends Date
   ? ValueType
   : TObject extends Function
   ? ValueType
-	: TObject extends object
-	? SchemaObject<TObject, ValueType>
-	: ValueType;
+  : TObject extends object
+  ? SchemaObject<TObject, ValueType>
+  : ValueType;
 
 /**
 Same as `Schema`, but accepts only `object`s as inputs. Internal helper for `Schema`.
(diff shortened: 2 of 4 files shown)


Tellios:
#103 This pull request shares similarities with RecordDeep proposal. The discussion ended with a question on which path to take for objects in the tree.

This pull request selects a combination of both options.

source/schema.d.ts
```diff
@@ -0,0 +1,71 @@
+/**
+Creates a type that mimics the structure of the specified type but replaces all the property values with a string-type.
```
sindresorhus, line 2: It doesn't just do string type.

source/schema.d.ts
```diff
@@ -0,0 +1,71 @@
...
+  ? ValueType
+  : TObject extends ReadonlySet<unknown>
+  ? ValueType
+  : TObject extends readonly unknown[]
+  ? ValueType
+  : TObject extends unknown[]
+  ? ValueType
+  : TObject extends (...arguments: unknown[]) => unknown
+  ? ValueType
+  : TObject extends Date
```
sindresorhus, line 58: It should ignore `RegExp` too.

source/schema.d.ts
```diff
@@ -0,0 +1,71 @@
...
+  created: 'show',
+  active: 'show',
+  passwordHash: 'hide',
+}
+
+```
+
+@category Utilities
+*/
+export type Schema<TObject, ValueType> = TObject extends string
```
sindresorhus, line 42:
```suggestion
export type Schema<ObjectType, ValueType> = ObjectType extends string
```

source/schema.d.ts
```diff
@@ -0,0 +1,71 @@
...
+const userMaskSettings: UserMask = {
+  id: 'show',
+  name: {
+    firstname: 'show',
+    lastname: 'mask',
+  },
+  phoneNumbers: 'mask',
+  created: 'show',
+  active: 'show',
+  passwordHash: 'hide',
```
sindresorhus, line 35: Use tab indentation.

readme.md
```diff
@@ -129,6 +129,7 @@ Click the type names for complete docs.
 - [`Includes`](source/includes.d.ts) - Returns a boolean for whether the given array includes the given item.
 - [`Simplify`](source/simplify.d.ts) - Useful to flatten the type output to improve type hints shown in editors. And also to transform an interface into a type to aide with assignability.
 - [`Jsonify`](source/jsonify.d.ts) - Transform a type to one that is assignable to the `JsonValue` type.
+- [`Schema`](source/schema.d.ts) - Deeply changes the values for all properties on a type to whatever specified. Useful for code e.g validation or configuration that relies on using the same structure of a type but the values are different.
```
sindresorhus, line 132:
```suggestion
- [`Schema`](source/schema.d.ts) - Deeply change the type of values for all properties on a type to the given value.
```

sindresorhus: I think it could still be made clearer though. If you have any suggestions.

source/schema.d.ts
```diff
@@ -0,0 +1,71 @@
...
+  ? ValueType
+  : TObject extends Function
+  ? ValueType
+  : TObject extends object
+  ? SchemaObject<TObject, ValueType>
+  : ValueType;
+
+/**
+Same as `Schema`, but accepts only `object`s as inputs. Internal helper for `Schema`.
+ */
```
sindresorhus, line 68:
```suggestion
*/
```

sindresorhus:
> This pull request selects a combination of both options.

What do you mean it's a combination of both options?

sindresorhus (changes requested):

sindresorhus:
Which name would you pick? `Schema` or `RecordDeep`?

sindresorhus:

Tellios:
> > This pull request selects a combination of both options.
> 
> What do you mean it's a combination of both options?

@sindresorhus I'm referring to this question: https://github.com/sindresorhus/type-fest/issues/103#issuecomment-627250369.

Sorry this is very philosophical but very important when it comes to naming this type. If this should be called `RecordDeep` it creates assumptions on the behavior provided.

So using the question's example type:

```
interface MyType {
    keyOne: 'key1',
    foo: {
        keyTwo: 'key2'
    }
}
``` 

This implementation would create this:

```
// This
type NewType = Schema<MyType, string>

// Equals this
interface NewType {
  keyOne: string;
  foo: string | { keyTwo: string }
}
``` 

The interesting part here is the property `foo` which is now either a `string` or an object mimicking the original structure. Depending on your use-case this is either great or bad.

This kind of boil downs into what expectations you have when a `Record` would be nested. If we look at real world scenarios where similar types have been used I have so far only seen validation and configuration. Both of these use the above.

`joi` and `yup` are two validation packages where we essentially do the above:

```
type MyTypeValidationSchema = RecordDeep<MyType, Schema>;
```

And that allows us to provide validation in two ways:

```
const schemaWithoutNesting: MyTypeValidationSchema = {
  foo: object({
    keyTwo: string().required(),
  }),
};

const schemaWithNesting: MyTypeValidationSchema = {
  foo: {
    keyTwo: string().required(),
  },
};
``` 

Any preference one way or another?

readme.md
Tellios: @sindresorhus Made changes to it. What do you think about it now?

sindresorhus:
> This is commonly seen in validation where libraries like joi and yup when used with TypeScript replaces interface values with validation schema types.

Thought. Why not include this type in the validation library you're using?

Tellios:
> > This is commonly seen in validation where libraries like joi and yup when used with TypeScript replaces interface values with validation schema types.
> 
> Thought. Why not include this type in the validation library you're using?

The above are just examples on use-cases that are very similar. In our case we use this type for something else entirely. We use it for configuring logging behavior of objects. For that we use the schema type to convert an existing interface into a configuration interface for the original interface where we can apply specific behavior on different properties, e.g keep, hide, remove, hash, etc.

I have been in the situation that I needed the type contributed here enough times that I felt that it was worth doing it.

readme.md
```diff
@@ -118,6 +118,7 @@ Click the type names for complete docs.
 - [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`Simplify`](source/simplify.d.ts) - Useful to flatten the type output to improve type hints shown in editors. And also to transform an interface into a type to aide with assignability.
 - [`Get`](source/get.d.ts) - Get a deeply-nested property from an object using a key path, like [Lodash's `.get()`](https://lodash.com/docs/latest#get) function.
+- [`Schema`](source/schema.d.ts) - Recursively replaces all the property values into a given value type for an object.
```
sindresorhus, line 121: The description makes it sound like it mutates the existing type, but it creates a new type. See the description of the other "deep" types.

sindresorhus:
Alright. I'm going to accept this.

sindresorhus:
Could you add the type here too, by mentioning `RecordDeep`: https://github.com/sindresorhus/type-fest#alternative-type-names So we have some reference if anyone is looking for a deep version of `Record`.

source/schema.d.ts
```diff
@@ -0,0 +1,72 @@
+/**
+Recursively replaces all the property values into a given value type for an object.
+
+Use-cases:
+- Form validation, specify on every field how it should be validated.
+- Form settings, specify configuration for input field.
+- Provide types that allow specifying settings/behavior to types when e.g masking or parsing data.
```
sindresorhus, line 7: The text here could use some tweaks.

For example:

> Form validation: Define how each field should be validated.

sindresorhus (changes requested):

Issue #326, opened 2021-11-23
Cannot copy Date type via CamelCasedPropertiesDeep


David245M:
I'm currently using `CamelCasedPropertiesDeep` type and I found that it's not copy Date type, instead it creates copy of it. I've got this error
```
Argument of type '{ toString: () => string; toDateString
string; toLocaleString: { (): string; (locales?: string | string[] | undefined, options?: DateTimeFormatOptions | undefineON: (key?: any) => string; }' is not assignable to parameter of type 'number | Date'.
```
So I tried it with another standard build-in object types and it works with none of them.
```
interface ExampleType {
    bool: Boolean
    sym: Symbol
    error: Error
    num: Number
    bigInt: BigInt
    math: Math
    date: Date
    str: String
    regExp: RegExp
    array: Int8Array
    map: Map<any,any>
    weakMap: WeakMap<any, any>
    weakSet: WeakSet<any>
    json: JSON
    promise: Promise<any>
    generator: Generator
    dateTimeFormat: Intl.DateTimeFormat
}

type CamelizedExample = CamelCasedPropertiesDeep<ExampleType>
```
[...]

PR #327, opened 2021-11-25
Prevent deep copy for standard build-in types

David245M:
`CamelCasedPropertiesDeep` recursively copy all types including standard build-it object types. You can pass type or union of type and they will no recursively copy more. Also included for some build-it types.

### DIFF
--- a/source/camel-cased-properties-deep.d.ts
+++ b/source/camel-cased-properties-deep.d.ts
@@ -1,4 +1,5 @@
 import {CamelCase} from './camel-case';
+import {StandartBuildInTypes} from './utilities';
 
 /**
 Convert object properties to camel case recursively.
@@ -40,8 +41,9 @@ const result: CamelCasedPropertiesDeep<UserWithFriends> = {
 
 @category Template Literals
 */
-export type CamelCasedPropertiesDeep<Value> = Value extends Function
-	? Value
+export type CamelCasedPropertiesDeep<Value, Exclude = never>
+  = Value extends Exclude | StandartBuildInTypes
+  ? Value
 	: Value extends Array<infer U>
 	? Array<CamelCasedPropertiesDeep<U>>
 	: Value extends Set<infer U>
--- a/source/camel-cased-properties-deep.d.ts
+++ b/source/camel-cased-properties-deep.d.ts
@@ -1,5 +1,4 @@
 import {CamelCase} from './camel-case';
-import {StandartBuildInTypes} from './utilities';
 
 /**
 Convert object properties to camel case recursively.
@@ -42,7 +41,7 @@ const result: CamelCasedPropertiesDeep<UserWithFriends> = {
 @category Template Literals
 */
 export type CamelCasedPropertiesDeep<Value, Exclude = never>
-  = Value extends Exclude | StandartBuildInTypes
+  = Value extends Exclude
   ? Value
 	: Value extends Array<infer U>
 	? Array<CamelCasedPropertiesDeep<U>>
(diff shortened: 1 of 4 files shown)


source/camel-cased-properties-deep.d.ts
```diff
@@ -40,8 +41,9 @@ const result: CamelCasedPropertiesDeep<UserWithFriends> = {
 
 @category Template Literals
 */
-export type CamelCasedPropertiesDeep<Value> = Value extends Function
-	? Value
+export type CamelCasedPropertiesDeep<Value, Exclude = never>
+  = Value extends Exclude | StandartBuildInTypes
```
voxpelli, line 45: Skip adding the `StandartBuildInTypes` for now, when/if it becomes evident that we need, then we can extract it

source/utilities.d.ts
```diff
@@ -3,3 +3,7 @@ export type UpperCaseCharacters = 'A' | 'B' | 'C' | 'D' | 'E' | 'F' | 'G' | 'H'
 export type WordSeparators = '-' | '_' | ' ';
 
 export type StringDigit = '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9';
+
+export type StandartBuildInTypes =
+  | Function
+  | Date;
```
voxpelli, line 9: Not sure we should include `Date` up front, lets get the PR merged without it at first, then it won't be able to cause any side-effects and we can get this feature out for you asap.

voxpelli (changes requested):
Can you include an update of `DelimiterCasedPropertiesDeep` as well? And of the ones using it? So that the two implementations remain consistent? https://github.com/sindresorhus/type-fest/blob/8028b63e3fa8e3e65735dbe685d86feac5b3d120/source/delimiter-cased-properties-deep.d.ts#L43-L56

And thanks a lot for the PR 🙏

source/utilities.d.ts
sindresorhus: > Not sure we should include Date up front

Why not? I cannot imagine anyone would want to camel-case the `Date` methods.

sindresorhus: This is similar to https://github.com/sindresorhus/type-fest/pull/208. Would be better to come up with a general solution than adding `Date`, `RegExp`, etc, explicitly.

source/camel-cased-properties-deep.d.ts
David245M: OK, remove it at all or just remove its union with `Exclude`?

David245M: Should I return `extends Function` to support backward combability?

source/utilities.d.ts
voxpelli: I guess `Date` is unique enough to include, but I wouldn't include `Function` in here I think, as `Date` is a built in `interface`, whereas `Function` isn't an interface?

We can't add eg. `Boolean` though I think, because of: https://github.com/sindresorhus/type-fest/issues/326#issuecomment-976913486

David245M:
@voxpelli is it still needed?

source/camel-cased-properties-deep.d.ts
```diff
@@ -40,11 +40,12 @@ const result: CamelCasedPropertiesDeep<UserWithFriends> = {
 
 @category Template Literals
 */
-export type CamelCasedPropertiesDeep<Value> = Value extends Function
-	? Value
+export type CamelCasedPropertiesDeep<Value, Exclude = never>
+  = Value extends Exclude | Function
```
sindresorhus, line 44: I'm not sure `Function` is correct here. A function can act as an object.

sindresorhus:
See: https://github.com/sindresorhus/type-fest/issues/347

sindresorhus:
Closing for lack of activity.

PR #330, opened 2021-11-30
Add `Float` type

### LINKED ISSUE #329: Float type support
The following seems to work for enforcing a float constrain on a number:

```ts
type Float<T extends number> =  `${T}` extends `${bigint}` ? never : T;

const onlyFloat = <T extends number> ( x: Float<T> ): void => {}

onlyFloat ( 123 ); // Error
onlyFloat ( 123.123 ); // OK
```

It's basically just Integer but with the results flipped.

jonahsnider:
Adds a `Float` and `NegativeFloat` type.
Not really sure what the use case for `NegativeFloat` would be though.

Also: Should these be called `NonInteger` instead of `Float`? Might be a more accurate description of what the type is, especially considering JS `number`s are always float64s.

Fixes #329

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -42,6 +42,7 @@ export {
 	NegativeInfinity,
 	Finite,
 	Integer,
+	Float,
 	Negative,
 	NonNegative,
 	NegativeInteger,
--- a/index.d.ts
+++ b/index.d.ts
@@ -46,6 +46,7 @@ export {
 	Negative,
 	NonNegative,
 	NegativeInteger,
+	NegativeFloat,
 	NonNegativeInteger,
 } from './source/numeric';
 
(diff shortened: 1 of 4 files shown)


sindresorhus:
> Also: Should these be called NonInteger instead of Float? Might be a more accurate description of what the type is, especially considering JS numbers are always float64s.

Nah, I think `Float` is fine.

sindresorhus:
If you have time and interest, I could use some help with reviewing the open pull requests. No worries if not though.

jonahsnider:
> If you have time and interest, I could use some help with reviewing the open pull requests. No worries if not though.

Sure! Happy to help

sindresorhus:
🙌

PR #331, opened 2021-11-30
Improve `@category` doc comments

jonahsnider:
Nearly every type was `@category Utilities`, not very helpful.

This PR changes all types to have one or more of the following categories:
- Objects
- Classes
- Maps
- Sets
- Arrays
- Async
- Functions
- Types
- JSON
- Numeric
- Observables

<details>
  <summary>TypeDoc output before:</summary>

  typedoc_before
</details>

<details>
  <summary>TypeDoc output after:</summary>

  typedoc_after
</details>

AFAIK Paka does not support multiple `@category` comments, it just uses the first one it sees.
Because Paka is closed source the only way to test this PR would be to publish this to npm and see what Paka renders.

### DIFF
--- a/source/async-return-type.d.ts
+++ b/source/async-return-type.d.ts
@@ -20,6 +20,6 @@ async function doSomething(value: Value) {}
 asyncFunction().then(value => doSomething(value));
 ```
 
-@category Utilities
+@category Async
 */
 export type AsyncReturnType<Target extends AsyncFunction> = PromiseValue<ReturnType<Target>>;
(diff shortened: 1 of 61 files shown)


sindresorhus:
Great. I think we should make the category names singular. Same with the existing `Template Literals`. I also don't think we should bother with title casing, so `Template Literals` => `Template literal`.

sindresorhus:
I also think we should move most of the camel case etc types to a new category, both in the TS docs and in the readme. Maybe `Change Case`? Open to a better name.

PR #332, opened 2021-12-03
`LiteralUnion`: Replace _ key with empty Record

a2br:
I think it's cleaner, I feel a bit uncomfortable defining a key for nothing. It's subjective though.


sindresorhus:
Have you manually confirmed that autocomplete still works correctly in VS Code after this change?

a2br:
Yes, it does work. That's what I usually use.

PR #342, opened 2022-01-05
`PartialDeep`: Ignore `Date` and `RegExp` types

sdotson:
Addresses issue https://github.com/sindresorhus/type-fest/issues/207, using feedback/code from closed pr https://github.com/sindresorhus/type-fest/pull/208 as a starting point.

This pr adds an additional `ExceptTypes` argument to `PartialDeep`, allowing consumers to state that certain types like `Date` or perhaps custom types written by somebody else, should be treated as exceptions.

Super excited to contribute and open to whatever feedback. Glad to help.

### DIFF
--- a/source/partial-deep.d.ts
+++ b/source/partial-deep.d.ts
@@ -33,49 +33,55 @@ settings = applySavedSettings({textEditor: {fontWeight: 500}});
 @category Set
 @category Map
 */
-export type PartialDeep<T> = T extends Primitive
+export type PartialDeep<T, ExceptTypes = never> = T extends Primitive
 	? Partial<T>
+	: T extends ExceptTypes
+	? T
 	: T extends Map<infer KeyType, infer ValueType>
-	? PartialMapDeep<KeyType, ValueType>
+	? PartialMapDeep<KeyType, ValueType, ExceptTypes>
 	: T extends Set<infer ItemType>
-	? PartialSetDeep<ItemType>
+	? PartialSetDeep<ItemType, ExceptTypes>
 	: T extends ReadonlyMap<infer KeyType, infer ValueType>
-	? PartialReadonlyMapDeep<KeyType, ValueType>
+	? PartialReadonlyMapDeep<KeyType, ValueType, ExceptTypes>
 	: T extends ReadonlySet<infer ItemType>
-	? PartialReadonlySetDeep<ItemType>
+	? PartialReadonlySetDeep<ItemType, ExceptTypes>
 	: T extends ((...arguments: any[]) => unknown)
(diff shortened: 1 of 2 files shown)


sindresorhus:
I think it should always ignore `Date | RegExp`. So instead of extending `Primitive`. Make a new local type called `Builtins` and make it `Primitive | Date | RegExp`.

source/partial-deep.d.ts
```diff
@@ -33,49 +33,55 @@ settings = applySavedSettings({textEditor: {fontWeight: 500}});
...
 /**
 Same as `PartialDeep`, but accepts only `object`s as inputs. Internal helper for `PartialDeep`.
 */
-type PartialObjectDeep<ObjectType extends object> = {
-	[KeyType in keyof ObjectType]?: PartialDeep<ObjectType[KeyType]>
+type PartialObjectDeep<ObjectType extends object, ExceptTypes> = {
+	[KeyType in keyof ObjectType]?: PartialDeep<ObjectType[KeyType], ExceptTypes>
 };
+
+type AddSymbolToPrimitive<T> = T extends
```
sindresorhus, line 85: This is unused

sdotson:
@sindresorhus Thank you for the feedback. I've created the `BuiltIns` as you suggested, removed the `ExceptionTypes` parameter, and updated the tests.

source/partial-deep.d.ts
```diff
@@ -33,8 +33,8 @@ settings = applySavedSettings({textEditor: {fontWeight: 500}});
 @category Set
 @category Map
 */
-export type PartialDeep<T> = T extends Primitive
-	? Partial<T>
```
sindresorhus, line 37: Intentional that you're removing the `Partial`?

sdotson: Yes. Couldn't really get `RegExp` and `Date` to work with it there. Are there any drawbacks to removing it? I'll admit I don't fully understand the advantages of doing `Partial` with primitives and `RegExp`/`Date`.

sindresorhus: > Are there any drawbacks to removing it?

I don't know. I recommend using Git blame to figure out who added it and then mentioning them here.

sdotson: @kainiedziela I see you made the most recent change to this, which was actually adding this 2 years ago. Any concerns?

kainiedziela: No concerns

sindresorhus:
In case you're interested, we should apply this fix to the other deep types too: https://github.com/sindresorhus/type-fest/issues/347

PR #344, opened 2022-01-06
Add `StringKeyOf` type

### LINKED ISSUE #341: Proposal: `StringKeyof`
`keyof` object could include number type. ([It is intended.](https://github.com/microsoft/TypeScript/issues/44577))

However, sometimes we want to get only string keys. (https://github.com/sindresorhus/type-fest/blob/main/source/get.d.ts#L68-L70)

How about provide `StringKeyof` type for this use cases?

```ts
type StringKeyof<BaseType> = `${Extract<keyof BaseType, string | number>}`
```

See also:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_Accessors#property_names

younho9:
Fixes #341

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -118,7 +118,7 @@ Click the type names for complete docs.
 - [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`Simplify`](source/simplify.d.ts) - Useful to flatten the type output to improve type hints shown in editors. And also to transform an interface into a type to aide with assignability.
 - [`Get`](source/get.d.ts) - Get a deeply-nested property from an object using a key path, like [Lodash's `.get()`](https://lodash.com/docs/latest#get) function.
-- [`StringKeyOf`](source/string-key-of.d.ts) - Get string keys from type.
+- [`StringKeyOf`](source/string-key-of.d.ts) - Get the string keys of the given type.
 
 ### JSON
 
--- a/readme.md
+++ b/readme.md
@@ -118,7 +118,7 @@ Click the type names for complete docs.
 - [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`Simplify`](source/simplify.d.ts) - Useful to flatten the type output to improve type hints shown in editors. And also to transform an interface into a type to aide with assignability.
 - [`Get`](source/get.d.ts) - Get a deeply-nested property from an object using a key path, like [Lodash's `.get()`](https://lodash.com/docs/latest#get) function.
-- [`StringKeyOf`](source/string-key-of.d.ts) - Get the string keys of the given type.
+- [`StringKeyOf`](source/string-key-of.d.ts) - Get keys of the given type as strings.
 
 ### JSON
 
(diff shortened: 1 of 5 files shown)


source/string-key-of.d.ts
```diff
@@ -0,0 +1,19 @@
+/**
+Get string keys from type.
+
```
sindresorhus, line 3: Should mention some use-cases.

younho9:
How about extracting `WithStringKeys` type which is internal type for `Get`?  

https://github.com/sindresorhus/type-fest/blob/main/source/get.d.ts#L85-L101

readme.md
```diff
@@ -118,6 +118,7 @@ Click the type names for complete docs.
 - [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`Simplify`](source/simplify.d.ts) - Useful to flatten the type output to improve type hints shown in editors. And also to transform an interface into a type to aide with assignability.
 - [`Get`](source/get.d.ts) - Get a deeply-nested property from an object using a key path, like [Lodash's `.get()`](https://lodash.com/docs/latest#get) function.
+- [`StringKeyOf`](source/string-key-of.d.ts) - Get string keys from type.
```
sindresorhus, line 121:
```suggestion
- [`StringKeyOf`](source/string-key-of.d.ts) - Get the string keys of the given type.
```

Same with the description in the file.

source/get.d.ts
```diff
@@ -1,5 +1,6 @@
 import {StringDigit} from '../source/utilities';
 import {Split} from './split';
+import {StringKeyOf} from './string-key-of.js';
```
sindresorhus, line 3:
```suggestion
import {StringKeyOf} from './string-key-of';
```

The `.js` extension is only required when a package is ESM.

source/string-key-of.d.ts
```diff
@@ -0,0 +1,23 @@
...
+Use-cases:
+- Get string keys from a type which may have number keys.
+- Making it possible to index using strings retrieved from template types.
```
sindresorhus, line 6:
```suggestion
- Makes it possible to index using strings retrieved from template types.
```

sindresorhus:

sindresorhus:
> How about extracting WithStringKeys type which is internal type for Get?

Extracting as in exporting it as a user-facing type? Or do you mean something else?

younho9:
> Extracting as in exporting it as a user-facing type? Or do you mean something else?

Exporting it as user-facing type.

source/string-key-of.d.ts
```diff
@@ -0,0 +1,23 @@
+/**
+Get string keys from type.
```
sindresorhus, line 2: Should be in sync with the readme.

sindresorhus: I think it should also explicitly mention that number keys are converted to strings. I don't think that's obvious. I first assumed it would **only** get the keys that are of string type.

sindresorhus:
> Exporting it as user-facing type.

Yeah, I think that one would be useful. Should be a separate pull request after this one though.

source/string-key-of.d.ts
younho9: How about change description to "Convert a type which may have number keys to one with string keys, and get the string keys." ?

sindresorhus: You text does not say that the keys will be coerced. Use the readme text first and then do a second paragraph where it explicitly says that fact.

younho9:
ping?

// @sindresorhus

Issue #347, opened 2022-01-09
Exclude `Date` and `RegExp` in all "deep" methods


sindresorhus:
See: https://github.com/sindresorhus/type-fest/pull/342

chris-amaya:
The title should've been "Include Date and RegExp in all deep methods" isnt this right? 

basically, what i understand its that deep methods are extending `Primitive` type and now instead of
```ts
 T extends Primitive
```

is going to be 
```ts
type BuiltIns = Primitive | Date | RegExp;
```

to finally extend the new type `BuiltIns`
```ts
 T extends BuiltIns 
```

is this correct?

chris-amaya:
basically, what i understand its that deep methods are extending `Primitive` type and now instead of
```ts
 T extends Primitive
```

is going to be 
```ts
type BuiltIns = Primitive | Date | RegExp;
```

to finally extend the new type `BuiltIns`
```ts
 T extends BuiltIns 
```

is this correct?

PR #349, opened 2022-01-09
Skip `Date` and `RegExp` in "deep" types

sdotson:
Adding Date/RegExp support to `PascalCasedPropertiesDeep`, `DelimiterCasedPropertiesDeep`, `ReadonlyDeep`, `CamelCasedPropertiesDeep`, `KebabCasedPropertiesDeep`, `SnakeCasePropertiesDeep`.

Fixes https://github.com/sindresorhus/type-fest/issues/347.

### DIFF
--- a/source/built-ins.d.ts
+++ b/source/built-ins.d.ts
@@ -1,0 +1,7 @@
+import {Primitive} from './primitive';
+/**
+Matches any [primitive value](https://developer.mozilla.org/en-US/docs/Glossary/Primitive).
+
+@category Type
+*/
+export type BuiltIns = Primitive | Date | RegExp;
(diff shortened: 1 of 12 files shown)


source/built-ins.d.ts
```diff
@@ -0,0 +1,7 @@
+import {Primitive} from './primitive';
+/**
+Matches any primitive, date, or regexp value
+
+@category Type
+*/
+export type BuiltIns = Primitive | Date | RegExp;
```
sindresorhus, line 7: Move it to https://github.com/sindresorhus/type-fest/blob/main/source/internal.d.ts

sindresorhus (changes requested):

source/built-ins.d.ts
sdotson: sure thing

Issue #292, opened 2021-10-14
Proposal: `Indexify`


younho9:
## Why?

Plain interface is not assignable to Record type because [Index signature is missing in type (only on interfaces, not on type alias)](https://github.com/microsoft/TypeScript/issues/15300)

## How?

Convert interface to a type with an index signature using the keys of the interface.

```typescript
/**
Add index signature to interface
*/
export declare type Indexify<O extends object> = {
  [P in keyof O]: O[P];
};

/**
Constructs a index signature to interface
*/
export declare const indexify: <O extends object>(object: O) => Indexify<O>;
```

[Here](https://github.com/younho9/ts-indexify/blob/main/index.test-d.ts) is a comparison with default behavior.

```typescript
type Dict = Record<string, unknown>;

/**
Default behavior
*/
expectAssignable<Dict>(someObject);
expectAssignable<Dict>(someObject as SomeType);
expectNotAssignable<Dict>(someObject as SomeInterface); // Index signature is missing in interface
expectNotAssignable<Dict>(someInstance); // Index signature is missing in class

/**
Indexify
*/
expectAssignable<Dict>(indexify(someObject));
expectAssignable<Dict>(indexify(someObject as SomeType));
expectAssignable<Dict>(indexify(someObject as SomeInterface));
expectAssignable<Dict>(indexify(someInstance));
```

younho9:
Thanks for answer!

Then, can we consider adding only `Indexify` type to type-fest? Or should I make a Pull Request for adding a `Indexify` type and type constructor together to `ts-extra`?

darcyparker:
Why not use https://github.com/sindresorhus/type-fest/blob/main/source/simplify.d.ts#L24?

An `interface` can be re-opened.  Whereas a `type` is sealed.  So a solution for https://github.com/microsoft/TypeScript/issues/15300 is to map the interface (which can be defined in many places) to a `type`.

The proposed `Indexify` is very similar the existing `Simplify`.  I think you've just pointed out a new use case for `Simplify`.

younho9:
@darcyparker That's true. I didn't find it. `Simplify` and `Indexify` are the same.

Could we consider ways to improve the document for this use case?

Which would be better, adding comments to use cases or adding type aliases? (`type Indexify<T> = Simplify<T>`)

darcyparker:
[...]

PR #352, opened 2022-01-17
Add `InvariantOf` type

### LINKED ISSUE #288: Proposal: `InvariantOf`
Inspired by [InvariantOf](https://flow.org/en/docs/lang/variance/#toc-invariance).

It could be useful to make object invariant. It is similar `opaque` type without brand or tag.

```typescript
import {Opaque} from './opaque';

export type InvariantOf<Type> = Opaque<Type, keyof Type>;
```

I published this utility type as [package](https://github.com/younho9/invariant-of), but I wish to add it to this project.

[Here](https://github.com/younho9/invariant-of/blob/main/index.test-d.ts) are some test cases.

I'm willing to do PR. 😃

younho9:
resolve #288

### DIFF
--- a/source/invariant-of.d.ts
+++ b/source/invariant-of.d.ts
@@ -1,9 +1,4 @@
-declare const tag: unique symbol;
-
-declare type InvariantSignature<Type> = {
-	// Make invariant function signature.
-	readonly [tag]: (argument: Type) => Type;
-};
+import {Opaque} from './opaque';
 
 /**
 Create an [invariant type](https://basarat.gitbook.io/typescript/type-system/type-compatibility#footnote-invariance), which is a type that does not accept supertypes and subtypes.
@@ -74,4 +69,4 @@ keyOfInvariantFooBar(invariantOf(fooBarBaz)); // Error: Argument of type 'Invari
 
 @category Type
 */
-export type InvariantOf<Type> = Type & InvariantSignature<Type>;
+export type InvariantOf<Type> = Opaque<Type, (argument: Type) => Type>;
--- a/source/invariant-of.d.ts
+++ b/source/invariant-of.d.ts
@@ -51,7 +51,7 @@ function keyOfFooBar(fooBar: FooBar) {
 }
 
 keyOfFooBar(fooBar) //=> (keyof FooBar)[]
-keyOfFooBar(fooBarBaz) // Allowed but, error at runtime
+keyOfFooBar(fooBarBaz) //=> (keyof FooBar)[] but, (keyof FooBarBaz)[] at runtime
 
 // In invariance
 
--- a/source/invariant-of.d.ts
+++ b/source/invariant-of.d.ts
@@ -14,8 +14,13 @@ Use-case:
 
 @example
 ```
-class Animal { constructor(public name: string){} }
-class Cat extends Animal { meow() { } }
+class Animal {
+	constructor(public name: string){}
+}
+
+class Cat extends Animal {
+	meow() {}
+}
 
 let animalArray: Animal[] = [animal];
 let catArray: Cat[] = [cat];
--- a/source/invariant-of.d.ts
+++ b/source/invariant-of.d.ts
@@ -1,0 +1,72 @@
+declare const tag: unique symbol;
+
+declare type InvariantSignature<Type> = {
+	// Make invariant function signature.
+	readonly [tag]: (argument: Type) => Type;
+};
+
+/**
+Create an [invariant type](https://basarat.gitbook.io/typescript/type-system/type-compatibility#footnote-invariance), which is a type that does not accept supertypes and subtypes.
+
+Use-case:
+- Prevent runtime errors that may occur due to assigning subtypes to supertypes.
+- Improve type signature of object methods like [`Object.keys()` or `Object.entries()`](https://github.com/microsoft/TypeScript/pull/12253#issuecomment-263132208) by sealing the object type.
+
+@example
+```
+class Animal { constructor(public name: string){} }
+class Cat extends Animal { meow() { } }
+
+let animalArray: Animal[] = [animal];
+let catArray: Cat[] = [cat];
+
+animalArray = catArray; // Okay if covariant
+animalArray.push(new Animal('another animal')); // Pushed an animal into catArray
+catArray.forEach(c => c.meow()); // Allowed but, error at runtime
+
+let invariantAnimalArray: InvariantOf<Animal>[] = [animal] as InvariantOf<Animal>[];
+let invariantCatArray: InvariantOf<Cat>[] = [cat] as InvariantOf<Cat>[];
+
+invariantAnimalArray = invariantCatArray; // Error: Type 'InvariantOf<Cat>[]' is not assignable to type 'InvariantOf<Animal>[]'.
+```
+
+@example
+```
+// In covariance (default)
+
+interface FooBar {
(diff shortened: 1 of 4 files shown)


papb:
In (your package)[https://github.com/younho9/invariant-of], I see that you also monkey-patched the global `Object.keys` typings to behave better with your custom type. Also, I noticed that [your example Playground](https://www.typescriptlang.org/play?#code/PTAEEkDtQZwFwE4FcDGckIIYBtRwJ4AOAlpAOYA0okmCCA9gO6lmgICm6C0Bh7o9AGagA8gCMAVuzQA6ANbt8oYjFAATTOXYMkMGQCh9gpJDTF60BfhGCAggCEAFPUnS4AfQBcoBwEpQAN76oGycGNDiUrJWMM6uaO7+mKqOVkI+9r4A2gC6+gC+hqRw2oKYKPwOgcGgmN6QSAC2YtoA3DVi3vAILAVFkCUIZRUZAMKg7AAeJZBqqlVBISjeYvT02Oyaffpq0ti0-CgW8LWdGTt7B6BHkCeYYstjhlY2Do73-iCgqYrpfrn6F52Jz3FCfMA-fB-TK5GR4IjsGAoHqEODqeiIyAAcjRjFokAM+i+AFUYCwIJAAG60YiaOA2Qy7FD7DjXY5ouCYMjeEzEACOSH4MHwzXW7X0vH4UGpPTpAAUGHwEAQADwAFQRAD5QABeb60bmgDV8fw67XG9jiyUUmW0gYAZWIZBoXHY6q1uuqIQ4mDUFmwSiynLIOW80pp8sV2lVFs17Xy4qmhHoyvhfBtEYGNndfG1notoAAZBnZQ6nS6MG7Y4nJsnU8ZTHBzNBSLa6dnY45qdhBd4Lb4w1TM-TBDn2NrFqEuNBu4Laqpw6WR2O49sG2YLKAgYu7XA3i4oh5B22s6OHJr-JOONPRPE4PJFLED25EvPvmlhP88oV9DcTq3hyqPUAKXGx3jEXx2j-NEQN3BxxmAodQMEcCwXFbckLgpxYLpPxWlAL5IWhbI8gwk892wzDcPsUZIIIsBJSRFFcXxAwgA) defines an `invariantOf` helper function which I like and even think necessary for the ergonomics of your proposed type.

To me it feels like both things (`invariantOf` and monkey-patching the Object.keys typings) are necessary for this type to really shine, so maybe it should remain a separate package?

younho9:
> Also, I noticed that [your example Playground](https://www.typescriptlang.org/play?#code/PTAEEkDtQZwFwE4FcDGckIIYBtRwJ4AOAlpAOYA0okmCCA9gO6lmgICm6C0Bh7o9AGagA8gCMAVuzQA6ANbt8oYjFAATTOXYMkMGQCh9gpJDTF60BfhGCAggCEAFPUnS4AfQBcoBwEpQAN76oGycGNDiUrJWMM6uaO7+mKqOVkI+9r4A2gC6+gC+hqRw2oKYKPwOgcGgmN6QSAC2YtoA3DVi3vAILAVFkCUIZRUZAMKg7AAeJZBqqlVBISjeYvT02Oyaffpq0ti0-CgW8LWdGTt7B6BHkCeYYstjhlY2Do73-iCgqYrpfrn6F52Jz3FCfMA-fB-TK5GR4IjsGAoHqEODqeiIyAAcjRjFokAM+i+AFUYCwIJAAG60YiaOA2Qy7FD7DjXY5ouCYMjeEzEACOSH4MHwzXW7X0vH4UGpPTpAAUGHwEAQADwAFQRAD5QABeb60bmgDV8fw67XG9jiyUUmW0gYAZWIZBoXHY6q1uuqIQ4mDUFmwSiynLIOW80pp8sV2lVFs17Xy4qmhHoyvhfBtEYGNndfG1notoAAZBnZQ6nS6MG7Y4nJsnU8ZTHBzNBSLa6dnY45qdhBd4Lb4w1TM-TBDn2NrFqEuNBu4Laqpw6WR2O49sG2YLKAgYu7XA3i4oh5B22s6OHJr-JOONPRPE4PJFLED25EvPvmlhP88oV9DcTq3hyqPUAKXGx3jEXx2j-NEQN3BxxmAodQMEcCwXFbckLgpxYLpPxWlAL5IWhbI8gwk892wzDcPsUZIIIsBJSRFFcXxAwgA) defines an `invariantOf` helper function which I like and even think necessary for the ergonomics of your proposed type.

Having actual functions here is a non-goal. This package is for types only, meaning nothing ends up compiled into actual JS code. https://github.com/sindresorhus/type-fest/issues/292#issuecomment-944272247

If `invariantOf` type is accepted, it would be possible to propose a helper function to the [ts-extras](https://github.com/sindresorhus/ts-extras).

> In (your package)[https://github.com/younho9/invariant-of], I see that you also monkey-patched the global `Object.keys` typings to behave better with your custom type.

I think it's convenient for me to monkey-patch global typing, but I think discussion is necessary.

sindresorhus:
I think this type can be useful for your own types.

sindresorhus:
> If invariantOf type is accepted, it would be possible to propose a helper function to the ts-extras.

I don't think we should override globals in a package, but it could maybe used to improve https://github.com/sindresorhus/ts-extras/blob/main/source/object-keys.ts

sindresorhus:
Could this one be used to make the `Opaque` type?

source/invariant-of.d.ts
```diff
@@ -0,0 +1,72 @@
...
+	// Make invariant function signature.
+	readonly [tag]: (argument: Type) => Type;
+};
+
+/**
+Create an [invariant type](https://basarat.gitbook.io/typescript/type-system/type-compatibility#footnote-invariance), which is a type that does not accept supertypes and subtypes.
+
+Use-case:
+- Prevent runtime errors that may occur due to assigning subtypes to supertypes.
+- Improve type signature of object methods like [`Object.keys()` or `Object.entries()`](https://github.com/microsoft/TypeScript/pull/12253#issuecomment-263132208) by sealing the object type.
```
sindresorhus, line 13: Can you think of any more specific use-cases?

source/invariant-of.d.ts
```diff
@@ -74,4 +69,4 @@ keyOfInvariantFooBar(invariantOf(fooBarBaz)); // Error: Argument of type 'Invari
 
 @category Type
 */
-export type InvariantOf<Type> = Type & InvariantSignature<Type>;
+export type InvariantOf<Type> = Opaque<Type, (argument: Type) => Type>;
```
younho9, line 72: > Could this one be used to make the `Opaque` type?

Improve to use the `Opaque` type.

source/invariant-of.d.ts
```diff
@@ -51,7 +51,7 @@ function keyOfFooBar(fooBar: FooBar) {
 }
 
 keyOfFooBar(fooBar) //=> (keyof FooBar)[]
-keyOfFooBar(fooBarBaz) // Allowed but, error at runtime
+keyOfFooBar(fooBarBaz) //=> (keyof FooBar)[] but, (keyof FooBarBaz)[] at runtime
```
younho9, line 54: There was a mistake in the document here.

source/invariant-of.d.ts
younho9: It's hard to think of any more specific use-cases for me.

Issue #305, opened 2021-10-31
Make `GetWithPath` public


stefanprobst:
[`Get`](https://github.com/sindresorhus/type-fest/blob/main/source/get.d.ts#L7) internally uses [`GetWithPath`](https://github.com/sindresorhus/type-fest/blob/main/source/get.d.ts#L7-L12), which could be helpful in cases where the key path is already an array.

would you be open to making `GetWithPath` public?

thanks!

sindresorhus:
Yeah, makes sense. Would be nice if we could just put the ability in the `Get` type instead of having to export another type just for this.

asbjornh:
@sindresorhus 

I'm not sure that #353 really fixes this. As far as I can tell this only works on the type level right now, not using value-level arrays:

```ts
const get = <O, K extends string[]>(obj: O, keys: K): Get<O, K> => implementation;
get({ foo: "bar" }, ["foo"]); // never - because `Get<{ foo: string }, string[]>` doesn't make sense
get({ foo: "bar" }, ["foo"] as const); // type error because readonly array can't be assigned to mutable array
```

In my personal opinion it should be possible to use `Get` with value-level arrays, just as you can with value-level strings. I was able to get the bottom example (with `as const`) to evaluate to the correct type locally, so if you're open to reopening this I could submit a PR

PR #353, opened 2022-01-18
Extend `Get` to accept `Path` as `string[]`

sdotson:
Fixes https://github.com/sindresorhus/type-fest/issues/305


sindresorhus:
Array support should be documented in the doc comments for the type.

sdotson:
sure thing

PR #354, opened 2022-01-19
Add `TransformDeep` type

JuanM04:
## `TransformDeep<Base, Value>`

Create a deeply transformed version of another type where each value (`BuiltIns` or functions) are replaced by an arbitrary type.

```ts
import {TransformDeep} from 'type-fest';

type Example = TransformDeep<{
  foo: number
  bar: { qux: boolean }
}, string>;

// => { foo: string; bar: { qux: string } }
```

### Use case

When validating an object, return an object with the same key disposition where each value is a string with a description of the error.

```ts
import {PartialDeep, TransformDeep} from 'type-fest';
const formValues: Values = {
  name: "John Doe",
  address: {
    street: "Av. de Mayo",
    number: 505,
    city: "Buenos Aires",
    country: "Argentina"
  }
  subscribeToNewsletter: true
};

const validate = (values: Values): PartialDeep<TransformDeep<Values, string>> => {
  // ...
}

const errors = validate(formValues);
// => 
// { 
//   name?: string
//   address?: {
//     street?: string
//     number?: string
//     city?: string
//     country?: string
//   }
//   subscribeToNewsletter?: string
// }
```

### DIFF
--- a/source/transform-deep.d.ts
+++ b/source/transform-deep.d.ts
@@ -1,0 +1,91 @@
+import {BuiltIns} from './internal';
+
+/**
+Create a type from another type with all keys and nested keys set to an arbitrary value.
+
+Use-cases:
+- When validating an object, return an object with the same key disposition where each value is a string with a description of the error.
+
+@example
+```
+import {PartialDeep, TransformDeep} from 'type-fest';
+
+const formValues: Values = {
+	name: "John Doe",
+	address: {
+		street: "Av. de Mayo",
+		number: 505,
+		city: "Buenos Aires",
+		country: "Argentina"
+	}
+	subscribeToNewsletter: true
+};
+
+const validate = (values: Values): PartialDeep<TransformDeep<Values, string>> => {
+	// ...
+}
+
+const errors = validate(formValues);
+```
+
+@category Object
+@category Array
+@category Set
+@category Map
+*/
+export type TransformDeep<Base, Value> = Base extends
+	| BuiltIns
+	| ((...arguments: any[]) => unknown)
+	? Value
+	: Base extends Map<infer KeyType, infer ValueType>
+	? TransformMapDeep<KeyType, ValueType, Value>
+	: Base extends Set<infer ItemType>
+	? TransformSetDeep<ItemType, Value>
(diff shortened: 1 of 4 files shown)


source/transform-deep.d.ts
```diff
@@ -0,0 +1,93 @@
+import {BuiltIns} from './internal';
+
+type IsAny<T> = 0 extends 1 & T ? true : false; // https://stackoverflow.com/a/49928360/3406963
```
sindresorhus, line 3: This should be moved to the `internal` file.

source/transform-deep.d.ts
```diff
@@ -0,0 +1,93 @@
...
+
+Use-cases:
+- When validating an object, return an object with the same key disposition where each value is a string with a description of the error.
+
+@example
+```
+import {PartialDeep, TransformDeep} from 'type-fest';
+
+const formValues: Values = {
+	name: "John Doe",
```
sindresorhus, line 16: Use single-quotes.

sindresorhus (changes requested):

sindresorhus:
I'm not sure about this one. It seems a bit niche. It seems like a type that should be part of the validation library you're using. I cannot think of any other use-case for this type than what you have presented.

---

// @papb @younho9

younho9:
> It seems like a type that should be part of the validation library you're using. I cannot think of any other use-case for this type than what you have presented.

I'm with @sindresorhus on this one.

JuanM04:
> I'm not sure about this one. It seems a bit niche. It seems like a type that should be part of the validation library you're using. I cannot think of any other use-case for this type than what you have presented.

That's fair. Do you want me to close this PR?

sindresorhus:
I think this is very similar to https://github.com/sindresorhus/type-fest/pull/321, which provided some additional use-cases, and is accepted. Would that solve your need? And if yes, would you be able to help review it?

JuanM04:
> I think this is very similar to #321

That's basically the same goal. Sorry I didn't notice it

sindresorhus:
https://github.com/sindresorhus/type-fest/releases/tag/v2.11.0

PR #359, opened 2022-01-29
`ReadonlyDeep`: Fix handling of objects with call signatures

### LINKED ISSUE #337: `ReadonlyDeep` does not work for objects with call signatures
Hi,
I have recently stumbled upon an issue with the `ReadonlyDeep<>` type. I have a type which has both members as well as call signatures (in my case the `JQueryStatic` type from `@types/jquery`). When I tried to create `ReadonlyDeep<JQueryStatic>`, what I got back was just `JQueryStatic` - no change in type...

Eventually, I tracked it down to the definition:
https://github.com/sindresorhus/type-fest/blob/f45033c5449652143261cbeaba6f4eff3f73c2a0/source/readonly-deep.d.ts#L37-L45
In this case, because the type has a call signature, `T extends Primitive | ((...arguments: any[]) => unknown)` evaluates to `true` and therefore what you get back is the original type...

## Trying to fix the issue

I tried to fix the issue by changing the first check to be just `T extends Primitive`, but then the type would lose all of its call signatures, as it would return `ReadonlyObjectDeep<>`...

So I tried changing `ReadonlyObjectDeep` from 
https://github.com/sindresorhus/type-fest/blob/f45033c5449652143261cbeaba6f4eff3f73c2a0/source/readonly-deep.d.ts#L62-L64
to
```ts
type ReadonlyObjectDeep<ObjectType extends object> = {
    readonly [KeyType in keyof ObjectType]: ReadonlyDeep<ObjectType[KeyType]>
    (...parameters: Parameters<ObjectType>): ReturnType<ObjectType>
};
```
but that just got me `TS7061: A mapped type may not declare properties or methods.`

Is there any way to fix this? Also, as far as I know TS objects have members, call signatures, construct signatures and index infos - the current `ReadonlyObjectDeep` does only members and index infos, I stumbled here upon call signatures, but construct signatures should probably be solved as well...

RebeccaStevens:
partial fix of #337

Makes objects with a call signatures readonly.

**Limitations:**

- An object with both properties and multiple call signatures cannot be mapped to another type without losing the call signatures.
  - see https://github.com/microsoft/TypeScript/issues/29732

Current this PR assumes that any object that has an call signatures is not a map or set. I can extends this PR to handle maps and sets with call signatures.

### DIFF
--- a/source/readonly-deep.d.ts
+++ b/source/readonly-deep.d.ts
@@ -34,8 +34,12 @@ data.foo.push('bar');
 @category Set
 @category Map
 */
-export type ReadonlyDeep<T> = T extends BuiltIns | ((...arguments: any[]) => unknown)
+export type ReadonlyDeep<T> = T extends BuiltIns
 	? T
+	: T extends (...arguments: any[]) => unknown
+	? {} extends ReadonlyObjectDeep<T>
+		? T
+		: ((...arguments: Parameters<T>) => ReturnType<T>) & ReadonlyObjectDeep<T>
 	: T extends Readonly<ReadonlyMap<infer KeyType, infer ValueType>>
 	? ReadonlyMapDeep<KeyType, ValueType>
 	: T extends Readonly<ReadonlySet<infer ItemType>>
--- a/source/readonly-deep.d.ts
+++ b/source/readonly-deep.d.ts
@@ -70,9 +70,11 @@ type ReadonlyObjectDeep<ObjectType extends object> = {
 };
 
 /**
-Test if the given function has multiple call signatures. Internal helper for `ReadonlyDeep`.
+Test if the given function has multiple call signatures.
+
 Needed to handle the case of a single call signature with properties.
-Multipile call signatures cannot currently be supported due to a TypeScript limitation.
+
+Multiple call signatures cannot currently be supported due to a TypeScript limitation.
 @see https://github.com/microsoft/TypeScript/issues/29732
 */
 type HasMultipleCallSignatures<T extends (...arguments: any[]) => unknown> =
--- a/source/readonly-deep.d.ts
+++ b/source/readonly-deep.d.ts
@@ -39,6 +39,8 @@ export type ReadonlyDeep<T> = T extends BuiltIns
 	: T extends (...arguments: any[]) => unknown
 	? {} extends ReadonlyObjectDeep<T>
 		? T
+		: HasMultipleCallSignatures<T> extends true
+		? T // Limitation - can't be made readonly
 		: ((...arguments: Parameters<T>) => ReturnType<T>) & ReadonlyObjectDeep<T>
 	: T extends Readonly<ReadonlyMap<infer KeyType, infer ValueType>>
 	? ReadonlyMapDeep<KeyType, ValueType>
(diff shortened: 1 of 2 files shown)


source/readonly-deep.d.ts
```diff
@@ -62,3 +68,14 @@ Same as `ReadonlyDeep`, but accepts only `object`s as inputs. Internal helper fo
 type ReadonlyObjectDeep<ObjectType extends object> = {
 	readonly [KeyType in keyof ObjectType]: ReadonlyDeep<ObjectType[KeyType]>
 };
+
+/**
+Test if the given function has multiple call signatures. Internal helper for `ReadonlyDeep`.
+Maybe wants to me exported as it's own until type.
+*/
+type HasMultipleCallSignatures<T extends (...arguments: any[]) => unknown> =
+	T extends {(...args: infer A1): any; (...args: any): any}
```
sindresorhus, line 77: Is it possible to use `unknown` for the return values here? I try to use `unknown` whenever possible.

source/readonly-deep.d.ts
```diff
@@ -62,3 +68,14 @@ Same as `ReadonlyDeep`, but accepts only `object`s as inputs. Internal helper fo
 type ReadonlyObjectDeep<ObjectType extends object> = {
 	readonly [KeyType in keyof ObjectType]: ReadonlyDeep<ObjectType[KeyType]>
 };
+
+/**
+Test if the given function has multiple call signatures. Internal helper for `ReadonlyDeep`.
```
sindresorhus, line 73: Can you explain the limitation here and link to the TypeScript issue?

source/readonly-deep.d.ts
```diff
@@ -34,8 +34,14 @@ data.foo.push('bar');
...
 @category Map
 */
-export type ReadonlyDeep<T> = T extends BuiltIns | ((...arguments: any[]) => unknown)
+export type ReadonlyDeep<T> = T extends BuiltIns
 	? T
+	: T extends (...arguments: any[]) => unknown
+	? {} extends ReadonlyObjectDeep<T>
+		? T
+		: HasMultipleCallSignatures<T> extends true
+		? T // Limitation - can't be made readonly
```
sindresorhus, line 43: Drop this comment. It's better explained on line 74.

sindresorhus (changes requested):

sindresorhus:
> Current this PR assumes that any object that has an call signatures is not a map or set. I can extends this PR to handle maps and sets with call signatures.

Do you think it should be handled?

RebeccaStevens:
@sindresorhus It's a rare case but probably should be handled for completeness.

sindresorhus:
> It's a rare case but probably should be handled for completeness.

I agree. It would be great if you could handle it then.

RebeccaStevens:
@sindresorhus I've made that change locally to support maps and sets but then I realized that with how this type currently works, only native maps and sets are supported.
That is to say if we have the type `Set<string> & { customProp: string }` the `ReadonlyDeep` version of this would just be the same as `ReadonlyDeep<Set<string>>`.

As native map and set can't be made callable (as far as I know), thus my local change will have no benefit.
If we want to support non-native maps and sets, that should probably be done in it's own PR.

I'd say the PR is currently ready in its current state.

sindresorhus:
Thanks :)

PR #368, opened 2022-02-24
Add `NullableToOptional` and `NonNullableKeys` type

1nVitr0:
This adds the new type [`NullableToOptional`](source/nullable-to-optional.d.ts) and the type [`NonNullableKeys`](source/non-nullable-keys.d.ts) which is used internally by `NullableToOptional`.

###  [`NullableToOptional`](source/nullable-to-optional.d.ts)

Make the keys of a type where the value type of the key include `null` optional.

```
import {NullableToOptional} from 'type-fest';

interface Example {
  a: string;
  b: number | null;
  c: null;
}

type NullIsOptional = NullableToOptional<Example>;
//=> { a: string, b?: number | null, c?: null }
```

### [`NonNullableKeys`](source/non-nullable-keys.d.ts)

Extract the keys from a type where the value type of the key include `null`.

Internally this is used for the `NullableToOptional` type.

```
import {NullableKeys} from 'type-fest';

interface Example {
	a: string;
	b: string | number;
	c: string | null;
	d: {} | null;
}

type NullableKeys = ConditionalKeys<Example>;
//=> 'c' | 'd'
```

### Use cases

A use case I recently came across is models that are generated by [`typeorm`](https://github.com/typeorm/typeorm). Nullable fields are typed like `'propertyName': string | null` instead of `'propertyName'?: string | null`. This can be problematic when creating new entries, when typescript (rightfully) complains about unassigned nullable properties.

***

I'm not 100% sure on the naming, feel free to suggest different ones. I have also noticed some weird editor behaviour where `NonNullableKeys` resolves to `never` in `.ts` files while working in the `.d.ts` file. If this is not only for me the type should probably be revised somewhat.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -54,6 +54,8 @@ export {
 	NonNegativeInteger,
 } from './source/numeric';
 export {StringKeyOf} from './source/string-key-of';
+export {NonNullableKeys} from './source/non-nullable-keys';
+export {NullableToOptional} from './source/nullable-to-optional';
 
 // Template literal types
 export {CamelCase} from './source/camel-case';
--- a/readme.md
+++ b/readme.md
@@ -122,6 +122,8 @@ Click the type names for complete docs.
 - [`Get`](source/get.d.ts) - Get a deeply-nested property from an object using a key path, like [Lodash's `.get()`](https://lodash.com/docs/latest#get) function.
 - [`StringKeyOf`](source/string-key-of.d.ts) - Get keys of the given type as strings.
 - [`Schema`](source/schema.d.ts) - Create a deep version of another object type where property values are recursively replaced into a given value type.
+- [`NonNullableKeys`](source/non-nullable-keys.d.ts) - Extract the keys from a type where the value type of the key include `null`.
+- [`NullableToOptional`](source/nullable-to-optional.d.ts) - Make the keys of a type where the value type of the key include `null` optional.
 
 ### JSON
 
--- a/source/non-nullable-keys.d.ts
+++ b/source/non-nullable-keys.d.ts
@@ -1,0 +1,26 @@
+/**
+Extract the keys from a type where the value type of the key include `null`.
+
+Internally this is used for the `NullableToOptional` type.
+
+@example
+```
+import {NullableKeys} from 'type-fest';
+
+interface Example {
+	a: string;
+	b: string | number;
+	c: string | null;
+	d: {} | null;
+}
+
+type NullableKeys = ConditionalKeys<Example>;
+//=> 'c' | 'd'
+```
+
+@category Object
+*/
+export type NonNullableKeys<Base> = {
+	// Pick all keys where the value type of the key include `null`
+	[K in keyof Base]: null extends Base[K] ? never : K
+}[keyof Base];
--- a/source/nullable-to-optional.d.ts
+++ b/source/nullable-to-optional.d.ts
@@ -1,0 +1,38 @@
+import {NonNullableKeys} from './non-nullable-keys';
+import {Simplify} from './simplify';
+import {UnionToIntersection} from './union-to-intersection';
+
+/**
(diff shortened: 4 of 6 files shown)


sindresorhus:
Thanks for the suggestion. The feels more like a type that is only for working around `typeorm` types than something that is generally useful. Wouldn't it be better to try to fix the `typeorm` types?

1nVitr0:
> Thanks for the suggestion. The feels more like a type that is only for working around `typeorm` types than something that is generally useful. Wouldn't it be better to try to fix the `typeorm` types?

Probably a good idea, will open an issue there and see what the response is. I definitely only came across this using `typeorm`, although I could see `NonNullableKeys` useful in other contexts as well.

aaronccasanova:
> although I could see `NonNullableKeys` useful in other contexts as well

Likewise 👍  Though I suggest updating the implementation to use mapped type conditionals:
```ts
export type NonNullableKeys<Base> = {
  // Pick all keys where the value type of the key include `null`
  [K in keyof Base as null extends Base[K] ? never : K]: Base[K]
}[keyof Base];
```

1nVitr0:
I removed the `NullableToOptional` type as it is too specialized and added types including `undefined` to the key extraction to stay in line with the `NonNullable` utility type.

@aaronccasanova For me, your suggestions always resolved to `unknown`, so I kept it as it was for now.

aaronccasanova:
@1nVitr0 I probably should have verified before commenting 🙂 

[Here are the small adjustments](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcByEB2aCuAbPAQwCM9gBpYRAZwB4AhQ64APjgF44BrKiAMzgBvAFBw4AenFwACgEsAxlziEC3KtTgB3ABbAoqGLrgA3FTgPJU-BEZ6I4szPLw4AJqgAGmfHg+i4ANrkDphqiNaMzMoa3qqgMMCYrhqRwEEAunAA-HCYwMZ6cABccOTpJakZwgC+ANzCwkgocAAqwNTwnBjYPiRklDS0ImJifBAQJQAM-mLEhFAlAIxwAD65PjNwcwBeJQBMNSwNkhxsAERjEGercGc7Z0A) needed to use mapped type conditionals! Note: It doesn't account for `optional?: properties` so there's a bit more to work out. Although, there doesn't appear to be much traction on this issue so the juice may not be worth the squeeze.

aaronccasanova:
The latest commit updated the implementation, per my POC, without addressing the noted caveat:

> Note: It doesn't account for `optional?:` properties so there's a bit more to work out.

I'm not sure if there is actual interest in adding `NonNullableKeys` to the library, but regardless I wanted to at least provide a [working solution](https://www.typescriptlang.org/play?#code/PTAEFEA8EsGcBdoDsDmp4E8AOBTUBXRAG2k1GVAHIA6YWHeAWgCcH9mlHNdqATa+LEoBuAFDc8ASVgA5HADcczADwAVAHygAvKADaqgLqgckeDiS9YepAqVGA-Omb48ALlAAzAIZF6Y0SCgjMGi4thSsvhERF4ARkQ4apo60nKKKlDwzF4AxvBqADSgSFFEoAA+BBY4Hsg4vOqaJmYWVlkuoI7evm5OLv4SoDIA9kgypXEJANI4GLDKAEJe9MmgANazwx6gAN6ioHpT5Ejrm9tL9KDLoKkT8YkXOLpTBk2m5pZ9eI426aDuL3cj2eBjEAF9QsgzMxvDk8FAvABbLAJXb7K7uBDMZAoMQHWKYrI4irFfCI2JKPGgHKE7GoEklaJU3juHZghmlKk4ezuErkymiCFhXCgGZzbRDUbjaKTHBi+YI5EJdRiEBaTSULyUEmUWKUUJAA).

sindresorhus:
@1nVitr0 Still interested in finishing this?

PR #372, opened 2022-03-07
Document rejection of `PackageJsonExtras` and add tip for extending `PackageJson`

### LINKED ISSUE #371: Add `eslintConfig` to `PackageJson` type
ESLint allows their [configuration to be stored](https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-file-formats) within the `package.json`. Seeing as TypeScript and Yarn configs are added, I think ESLint deserves a spot there as well.

philting:
Fixes #371

Initial PR to enhance the `PackageJson` type. Added a new type as proposed, named `PackageJsonExtras`.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -185,7 +185,7 @@ Click the type names for complete docs.
 
 ### Miscellaneous
 
-- [`PackageJson`](source/package-json.d.ts) - Type for [npm's `package.json` file](https://docs.npmjs.com/creating-a-package-json-file).
+- [`PackageJson`](source/package-json.d.ts) - Type for [npm's `package.json` file](https://docs.npmjs.com/creating-a-package-json-file). It also includes support for [TypeScript Declaration Files](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) and [Yarn Workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/).
 - [`TsConfigJson`](source/tsconfig-json.d.ts) - Type for [TypeScript's `tsconfig.json` file](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) (TypeScript 4.4).
 
 ## Declined types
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -33,6 +33,7 @@
 	],
 	"devDependencies": {
 		"@sindresorhus/tsconfig": "~0.7.0",
+		"@types/eslint": "^8.4.1",
```
voxpelli, line 36: For this to work, `@types/eslint` would have to be a dependency rather than a dev dependency.

Though I don't like the idea of this module starting to rely on lots of other type modules. I think it should be self-contained.

We don't know if the version of `eslint` used in the project is `8.x`, `7.x` or something else. We don't even know if they use `eslint` at all.

source/package-json.d.ts
```diff
@@ -305,6 +308,16 @@ declare namespace PackageJson {
...
 	}
 
+	/**
+	An alternative configuration for ESLint.
+	*/
+	export interface EslintConfig {
+		/**
+		ESLint configuration options.
+		*/
+		eslintConfig?: Linter.Config;
```
voxpelli, line 318: There is a hack where we don't have to include the ESLint module in our dependencies:

```suggestion
    // @ts-ignore
		eslintConfig?: import('eslint').Linter.Config;
```

Then it will use `import('eslint').Linter.Config` if it correctly finds it, else it will just fallback to any.

I used that pattern a lot through [`ts-ignore-import`](https://github.com/Sydsvenskan/node-ts-ignore-import) and its the one pattern I think we maybe could use to still include this here. And if we do, then we can include it in  `PackageJson`  straight away rather than in a `PackageJsonExtras`. Thoughts @sindresorhus?

voxpelli (changes requested):
I'm skeptical in general.

Instead of introducing anything like this, you can simply in your own project do:

```typescript
type MyPackageJson = PackageJson & { eslintConfig: import('eslint').Linter.Config };
```

Not sure adding it in here makes for any simpler or better implementation.

source/package-json.d.ts
sindresorhus: I don't think this is a good pattern. Falling back to `any` is never something I would want. It could accidentally allow specifying anything there without a warning. Maybe if it fell back to `never`.

sindresorhus:
I think this is a very good reason not to add this:

> We don't know if the version of eslint used in the project is 8.x, 7.x or something else.

sindresorhus:
I suggest we just document:

```ts
import type {PackageJson as BasePackageJson} from 'type-fest';
import type {Linter} from 'eslint';

type PackageJson = BasePackageJson & {eslintConfig: Linter.Config};
```

Then it's clear to user how to support other tools too.

source/package-json.d.ts
philting: That's a really interesting pattern. I couldn't figure out how to override the `any` type with `never` (or anything other than `any` for that matter).

Side note, apparently `@types/eslint` is included in `@typescript-eslint` so I had to manually delete the package to test it. But yeah, probably best to leave versioning to the consumer.

philting:
@sindresorhus Where would that go? Under "Declined Types"?

philting:
Reverted my changes and added some documentation based on the discussion.

sindresorhus:
Thanks :)

PR #383, opened 2022-03-24
Add `ReadonlyTuple` type

Menecats:
This pull-request adds an alternative solution for issue #284 as already discussed in PR #294

// @sindresorhus 
// @darcyparker


darcyparker:
Looks good to me.

source/fixed-length-array.d.ts
```diff
@@ -28,6 +28,8 @@ guestFencingTeam.push('Sam');
 //=> error TS2339: Property 'push' does not exist on type 'FencingTeam'
 ```
 
+@deprecated This type doesn't prevent you to access values outside the array length via index (see issue#284),
```
sindresorhus, line 31: I'm not sure if it makes sense to deprecate this. Someone might want a mutable version. I think we should document how this differs from `ReadonlyTuple` though.

PR #384, opened 2022-03-27
Add `watchOptions`field  to`TsConfig` type

privatenumber:
Added [`watchOptions`](https://www.typescriptlang.org/tsconfig#watchOptions).

Previously, there were watch options in `compilerOptions` that were not reflected in the docs so I removed them.

To prevent a breaking change, I can also add them back and mark them `@deprecated`.

### DIFF
--- a/source/tsconfig-json.d.ts
+++ b/source/tsconfig-json.d.ts
@@ -193,6 +193,30 @@ declare namespace TsConfigJson {
 			| 'remove'
 			| 'preserve'
 			| 'error';
+
+		export type FallbackPolling =
+			| 'fixedPollingInterval'
+			| 'priorityPollingInterval'
+			| 'dynamicPriorityPolling'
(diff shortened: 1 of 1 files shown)


source/tsconfig-json.d.ts
```diff
@@ -990,6 +934,71 @@ declare namespace TsConfigJson {
 		explainFiles?: boolean;
 	}
 
+	namespace WatchOptions {
+		export type WatchFileKind =
```
privatenumber, line 938: FYI I chose these names to be consistent with:
https://github.com/microsoft/TypeScript/blob/f9e6ba38f1cf38f7e5956396f548fce37934f9cc/lib/protocol.d.ts#L1121

sindresorhus:
> To prevent a breaking change, I can also add them back and mark them @deprecated.

That would be great.

PR #388, opened 2022-03-30
Update get.d.ts for TS 4.7

weswigham:
Callers of `WithStringKeys` pass in an unconstrained type parameter, and `StringKeyOf` uses an unconstrained type parameter, so `WithStringKeys` should handle an unconstrained type parameter. TS 4.7 calls out this inconsistency.


sindresorhus:
Thanks :)

PR #400, opened 2022-05-26
Add `recurseIntoArrays` option for `PartialDeep`

### LINKED ISSUE #367: Discussion/proposal: PartialDeep<Foo, {recurseIntoArrays: false}>
In https://github.com/sindresorhus/type-fest/issues/356#issuecomment-1030748321, @sindresorhus proposed an option `PartialDeep<Foo, {recurseIntoArrays: false}>` which would skip making array element types `undefined`. 

Per my earlier [comment](https://github.com/sindresorhus/type-fest/issues/356#issuecomment-1022114231):
 
> In some sense I think `PartialDeep` is going a little too far. `{foo: string[]}` becomes `{foo?: (string|undefined)[]}` but the user probably just intended `{foo?: string[]}`

I think this might be a sensible default to consider.

The only caveat I can think of is whether it's possible/wise to handle [tuple types](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types) differently. So `foo: [string,number]` vs `foo: (string|number)[]`.

Thoughts?

bbrk24:
Closes #367

This PR adds the `{recurseIntoArrays: false}` option as discussed in that issue. The property is optional (i.e. you may pass `{}` for the options) for forwards-compatibility: should there be more options in the future, such as `recurseIntoSets`, you may specify any one and not the rest.

### DIFF
--- a/source/partial-deep.d.ts
+++ b/source/partial-deep.d.ts
@@ -2,7 +2,7 @@ import type {BuiltIns} from './internal';
 
 /**
 The options for `PartialDeep`.
-@property recurseIntoArrays Whether to affect the individual elements of an array or tuple. (Default: true)
+@property recurseIntoArrays - Whether to affect the individual elements of arrays and tuples. (Default: true)
 */
 export interface PartialDeepOptions {
 	recurseIntoArrays?: boolean;
--- a/source/partial-deep.d.ts
+++ b/source/partial-deep.d.ts
@@ -4,7 +4,7 @@ import type {BuiltIns} from './internal';
 The options for `PartialDeep`.
 @property recurseIntoArrays Whether to affect the individual elements of an array or tuple. (Default: true)
 */
-interface PartialDeepOptions {
+export interface PartialDeepOptions {
 	recurseIntoArrays?: boolean;
 }
 
--- a/source/partial-deep.d.ts
+++ b/source/partial-deep.d.ts
@@ -1,10 +1,14 @@
 import type {BuiltIns} from './internal';
 
 /**
-The options for `PartialDeep`.
-@property recurseIntoArrays - Whether to affect the individual elements of arrays and tuples. (Default: true)
+@see PartialDeep
 */
 export interface PartialDeepOptions {
+	/**
+	Whether to affect the individual elements of arrays and tuples.
+	
+	@default true
+	*/
 	recurseIntoArrays?: boolean;
 }
 
@@ -37,6 +41,7 @@ settings = applySavedSettings({textEditor: {fontWeight: 500}});
 ```
 
 By default, this also affects array and tuple types:
+
 ```
 import type {PartialDeep} from 'type-fest';
 
@@ -48,7 +53,8 @@ const partialSettings: PartialDeep<Settings> = {
 	languages: [undefined]
 };
 ```
-If this is undesirable, you may pass `{recurseIntoArrays: false}` as the second type argument.
+
+If this is undesirable, you can pass `{recurseIntoArrays: false}` as the second type argument.
 
 @category Object
 @category Array
--- a/source/partial-deep.d.ts
+++ b/source/partial-deep.d.ts
@@ -9,7 +9,7 @@ export interface PartialDeepOptions {
 
 	@default true
 	*/
-	recurseIntoArrays?: boolean;
+	readonly recurseIntoArrays?: boolean;
 }
 
 /**
--- a/source/partial-deep.d.ts
+++ b/source/partial-deep.d.ts
@@ -6,7 +6,7 @@ import type {BuiltIns} from './internal';
 export interface PartialDeepOptions {
 	/**
 	Whether to affect the individual elements of arrays and tuples.
-	
+
 	@default true
 	*/
 	recurseIntoArrays?: boolean;
--- a/source/partial-deep.d.ts
+++ b/source/partial-deep.d.ts
@@ -1,6 +1,14 @@
 import type {BuiltIns} from './internal';
 
 /**
+The options for `PartialDeep`.
+@property recurseIntoArrays Whether to affect the individual elements of an array or tuple. (Default: true)
+*/
+interface PartialDeepOptions {
+	recurseIntoArrays?: boolean;
+}
+
+/**
 Create a type from another type with all keys and nested keys set to optional.
 
 Use-cases:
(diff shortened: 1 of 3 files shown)


source/partial-deep.d.ts
```diff
@@ -1,5 +1,13 @@
 import type {BuiltIns} from './internal';
 
+/**
+The options for `PartialDeep`.
+@property recurseIntoArrays Whether to affect the individual elements of an array or tuple. (Default: true)
+*/
+interface PartialDeepOptions {
```
sindresorhus, line 7: You need to export this here and in the https://github.com/sindresorhus/type-fest/blob/main/index.d.ts

source/partial-deep.d.ts
```diff
@@ -1,5 +1,13 @@
 import type {BuiltIns} from './internal';
 
+/**
+The options for `PartialDeep`.
+@property recurseIntoArrays Whether to affect the individual elements of an array or tuple. (Default: true)
```
sindresorhus, line 5:
```suggestion
@property recurseIntoArrays - Whether to affect the individual elements of arrays and tuples. (Default: true)
```

sindresorhus (changes requested):

source/partial-deep.d.ts
bbrk24: Oh yeah, I don't know how I missed that. I'll add that later this week.

Nokel81:
This change does not seem to be backwards compatible even though it was released in a minor version.

sindresorhus:
@bbrk24 Can you look into this ^

bbrk24:
@Nokel81 I'm unable to reproduce this in the playground, based on what information you've provided me. Could you provide either a playground link or a repo with a MRE so I can investigate this?

Nokel81:
Sure, [here is a link](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcAKUCWBbTNMBuqAvAFACQAPnAHYCuANgxdXTQCbABmmNw7LOAGcYWGgHNB9bACNgUQTIgQGwAIY1BQxLJWLM43jADcpUJFgJkqAEJ1MDGAEkaQuMXRZc+InGoARNRhUagAlYHEAUXBTUgB6ACoE0gABIWBUNDVYTDUGfwywADUAJlIEuLNwaHgjeS41AGNM7Pw8guAwAHkwfAhXOABvCkTk8gB1AAtgGGmoBAg4NS4uYEb4WdRedkJMdjo8uGBVbGAaGDcILiWoKDVENw12BDowVSEAOlIKFM4Gxg2UDowAoFQoUHU7H6DEQcAhjToUHSLhgEAAgrd7kIAPwALjgShU6hopgAvt9RqQAMIQoKoNRWFBwLhQCDYJY0CCbeZIJkAdzwkyWTDgAGtgA8Oc8+CJ+GKJW50htFhBeph+nkvqQAKrpAC0jTU6SEuNIergAFl5IYJEs4H81ADhDN8BIhHFGv0eOI4BAZAArNbwAWzDlcua+gNBgA0CGmzs9HEjgfWcD5EEYzzkds4nTgYFauQYvuum3t3EdjmTQa+5otEEaot4PqeCGAImbcE92DeoCO5zwmHbsb5c1QkDAjCCnY0-fwEOrqZDQrwbnFkvTmYJqAh+w4GlqNDbUGwl1L8ewDdFvp57ZgWpSoDUPdUpAABh-SDgLBtrEMsjk7SFKSzKsuyADkvLAHqqwiOBsSJiIzowK64gmnAADKLrNm4HjDOQwQgDAkQ7KiUD4vh5BcP0MAYZgABewD4gAjAALKY5BUTRVIqNA+LgQAxAADCJonwRQXHnOMwAGJMMD4qxIkceS5COqi3a9sE+INAw6QcWpEBCGoRD4qIwJkgh-RIWoYBvIgGHGfwWEoTh7hwAAFEZRDsM5qHoQBbT5IUAA8vk4QAfAAlO44VDOCMyIkegwfClSp+bGKUfF5TnYW6pJkt8aWuR4Nl2Q53lhW67mDIRxGkXxQzUVJMniHJ+IAKwiaSpKRaYH5vt8Niwg6AKxrMmCPLpizLKs6yPJisKtjAryqIy7amu+n7fjUa3-oWQGdCBLJsnAkHWDBd7iV+5z1E0qCVWhcXkAwGjiAc4jrcIojNgA2gAugVpCIfABaAQwD3+ftQWdKFuVobFeEUC9EjvZ9P1sH8vD8H9pD5ZtA2kE454TXAJMY+2mB3DIqixogGZdrOBZCG4b6DPCiLIuc6ILehOnpKSb5LG4ZbpImzxQTcb2nOcD6GsE4jQLCXRRusqRyxEitwBidyIGrdIK1AsLOXr8uaxaNnlJU5g7RLAVFh0xQlMFAAqsY9H0AygMEHBuHbB3dGqVluYMpII3AztHERZzsG4dgOM4rgUNi4cUPiEde9Hbjm2AwW8Ks8wANISs71ixnn8hwEUeTAiXKDhUn6BQ9nDulMFReILXwCxlXDA16XcDu+qrj1+QaeR97MeYTMuc0PncBOME2CdyPyd++DMwt07C-AEv-eD1ZI9jxnPtwOEahQjQMLZzPc-t53ZezxXPd93XDdr2fF9XzZm9t8X-fP8Ae+A9A7D1TuHcemdT6QmhPZae5d5jb13q-cgq8oYfxgc5H+iCgH71AaPcBx9J7uXcplbIaF8QaEQP9aKxBYpsFFJyPkNBIoNwjqwDg3AsYCHwenKOJ8-QphgKwiBJ90GX0QNre4N8K7YOsLFOIcRw53mZNAG4Ot3TLV7EIWMKB5hyRgGAE0CjDCzDoDID43Y4i4EaKyIQVwYBxE7hhGxmBehxAmkIYE7oADM7VmLtQAGwSWTrgoQP1wLsyRMAFE3N1HgT+iIyefNUAKPnueCUvpei+joPAK4aj7htg7BIWM-o6BITKagMs0ADC8EOFBCSKCU6cXwbIlA-1EluAjqk52yjqLzGyOojynIaB6k0e8aKQgUCNEwDwQ0TBdbNOTrST+sJWnAHaYQzpcBUn+BmfnM4bRghwmgeIqUcBsA5LUNTekPMGmNLETCSRiBgprywYvTufg4Dkx4HwdgbsQFCHCiPZp+InkvKhm8neHz2GY1+f8j2gLgX4LXsrQRP9XbAIRfIxRzsVrtm3JMRy+ZWS6JhBJfEKKVYwHRfCoegKwH0MYSSCkSRSAOVOELOAb5XmFDfLGGQOSliNGaL0S4py3zZzfI8JMRpSY0EnBcD488bpQBoIcaYDBdEqPmNyiFvKvhgjqFABozRG5g2boUVud9-7V0AXvAFHSzWBQdqE2KmzLQ2XBWDH+1qUC0oPrGHlnRW4AJwQCoFQxyTxFZey+kLMg1gD5QSQVTQRUXF9OK5yUrOW8AVZ8ZVwRVXquOFqvpXKE1vgNZUI1JqWhg0wZap2GLQmOoTa6x1zkvXOsbS7f1w9YohxZckWNnLdXet5fylNwrOjppgVyh5iBJWPDcLmnJ+aUTyDVcWDVpbVFju7Z0StltrqFtrU6osC6LXBqdr6ruldbVhoRa2vVnR23usvZ6hNVq-5+sxXS8KgaX2O2CqG+1WKB1RspCO2V+77YTuTfAVNM6xUwnnScmEWbl1yrzUqjdRbt0lormW2D-sj2GpVWe9+6HYHUp7c2h17q23hsdQuztX6m19sRZGodbLnxxvLUBpNArEPTtFRm1Db4BFBmzbK1diqC2buLZqoje6K1VtILbKGqKgw-20+sD57qpPrE48+8dr7mOI3ID9W9cr5R02uHpmAnc-p4nPf7VujnO7WZ-cAP6nH6542+NbSwmmzPFGYi7MOvCJ6x3sI4FwQhWFgOi5A6+8C4C3ofnPUDyDUHmu-paiLmX729ztcgo+fDJ6dvS2slebn160eDRF2ryWWPUbS4-QuPmstPwfXIt+aD2sFaa7-DuNrSvL1a++6j1XOvz3ef1xpVHz4YI3oV4KLWeGOuIaQqA5CORUL+jQuhNAGEQCYSwxpbCvkcJ+fwKblXLhUuEe6sFNWFt122Ti3pqiBlYjiGM4c+YK76MMbiYxgozEWLZFYzANjDL2McdYZxWA3EeK8XEXx-igmcWTmsjZj3wHdJ+-0nmQz+ijLxUICZUyZlw7yGSxZWsFpdrgyNtZnzvlcIjak8INj1BHIhKoAgB58mwglqiYlwAiDnCOGoD68xpfV2nP0EsXZKYIhevMCE+yaDNAaRSrTVKW4Redti8OVOCVErACS+QjPkVG7Rets3DLTtMtiDWu69Wr3hdG0AnLwA3WE+vl+orPmAP1ZNyBvrdcIM8eg-GoDRRmJCanWmlDsI3yYZzfKtduGVVbrgDulTOrQ9kerRRr3a8G0jbNx26eofItx+jcOvjo6y+TpE+n8TmeP2Jqw3J9dBelO7tL0nlP6nPemuWysn3ye-fjZfoHtrK3xEh-HwvuugGwvz4DxGwdLfeMcpgx3hDQru9zrfKxmYMmV25-k3hwvxf5jEbL5Pyv0-Bur4w2t2vQeYtQLf40as7uam776QYxpt4n7j6p5d7IY95cpGYwC37YZ54Kb4ZF6EYv6qYwHqahaBSOZR6eZ-iGZUphz4TeZjZMi8B2Z5LEEoAuaG474Rb0HrK3p-QBaxDBbwBnB0DsgFxmLACObOKqioDDBwBwAAByfGkyd00YpAEhVIvcsoUA8h5I3Bu0IhKA7A0hpwshzQwUuh7YBYzQzw7qAhcgwhnon2Hg7kChUhMhJhco5hghVhohHwRh+hIIEhEhycIgYgkgPhcA+InhThZhhOFhQhVKWhwAHwShZSha9hQRycXOvySRPh+I-hnYMKnCaRvUQW1QlgU+qAXQTC8g4Quupq4hSwYAmARQ8gQgQ8mR30EgpgEhTYHAzRARbRtAfGXRzYPR9g7A-RrR9hiYogKgqgUArmhIqgGgPR1MV4pRfAUABQqgHsMxygcxzKUaxRcANgRowAkRAAUnYjQGiLUY5laDAOfEEGoIYY4XdOEQAZEW4Z9tURIRoJyDcQiq5mvMFHztAOwMFFkcUl9AEUCj0RIY0MoYWkYa5qCeIFCcckAV2LSB7M7DgHeM+GAAiS0UiekcsnOpwOsUPAAOJ3DNBoDyDqg+RrD9AxyubSByBQDIlEmnIkkzBDyYl6E3E9h4ndHpEfQrF0jwkjEElBHsmobCnyAq40BMl8EsnIkvRyC6R-FQwAn0lQDAmImxiImQnpHYAaDy78AABiQ4DAjJ+IjK52NA-0bJ1GvRpw4pDpqJaqehThrmMROhjxBhoRTx4UrpKyxydiiIzQ9RSIQ8ApAxhJjp6QDAXAAAMrwKKNGaMZKY6UMWmRKT4baeUdwPIGcM0DiPiMsfmZUesgDOkT9OuOKX9NaW7raYDBLCcWcRcZgFcTMLcTcQ8R6U8Y6q8dEdYSQBlq4UOaIWHAcekK2f0O2Z2TcewHcb2cYQGXAAAGQeT+mmEDljmCIxEeG+moDJwfEonBnukrnNAum4zBGRr5EaZ-iDmCLXHdn3FbnOERG7lBgxFuSPlfnDlhwznnGXFUrPmLk9lvnsCxQbknlSmwjxlJkplXkZmolDFIU+GwVOlMTgkxnIXBkQihlQDhkNFNHYWjGBbdKTAkwkxljbSwCi71IwlGhuC-nrDBSgV3E7mWEgVdlgX3EsW0T-k-mfnrDsU9n8UxER4YQ-FlJuQ2lMKxgYRTKyWNlMKxTcZVA-hyqnpe6nGzm1GBA3FxRRoaF7GAXtkGX3GiUMguFyC6VAUdncULkcUeBmXAVPk8V3EKXSW4Q3ZnbyWYRKUeByU0D-6QJ2XmUcUnkdHDGkU5mfG1ERmNH9BoUSGnBOU3H4hWXIkiBBBlJeneXZVTJelTLIk-ThjyDtzoSIn1m+Xu64wFGaWcCMULgtmCHhX6V3GmnQBtwAWcVRGCLSLzBWU9bzBSW5XaJypzyKVrCxTJyuWYAWVsUeU3FeXjUKVTKxT4h8BECskNU7R7HhAEXNAFBCAuK9CqInmUhBFwAJDhzxjnlq5liiiCGqozD4r4UZiEXeFBFghBHnkpXbKsrXW3XOz3WHlxhBBxg7jtifWmoMCED4o0HuQzJQ0hmw1bBuDnleHsAsLA3pE3VSFchYUo3fHEoQAEB7ByhLTxgfVhmoDRWkyY2Hl-JQ1HjjRuAjSODCyLBvgABEnNMAvNA0wNlQf1h52ZgMTVWuqAjFzMo5cg7ZwUjmfVytLlwlMAsYFlfV7VC1nV3VjmAFbVbZHVNxXVUAStZBA6Yx6JwA1UmFsYWNThcApITBgUmpR1wAJ1Z1ZEEeoMzqdx6pO+TsBtkUFKYEE0wAFtginy9ATAgZ9VQML0ctVIdxDAEA4gaIaATgLsKtZBcU8VmA+IkRitdeORd27AsQ0JNt7Z7k0UJ57NHwNkmAbkfAfI8twANd+REhUaEhH0MAnd+dPhKN7kAAhA3U3XXfjXGBNBYtXbUbXciT3ehQlKqtPZ8E3ZysXbUZFj0UvUaIgHrmiQLrbf9bFTotkHxoWpDMwZFnqU2GAAABLqCOCTBUjTCNhuTJKT0ZnLSr0N192d2z3H120PWkjn13BpUNFd3O24xAA) to a playground

bbrk24:
Oh I see, this only occurs when you have two different versions of type-fest present at the same time. I think that can cause errors even from a patch-level change: as an obvious example, say a type mistakenly produces `string` where `number` is expected; changing the behavior to correctly give `number` is semver-patch, but loading the pre-patch and post-patch versions at the same time will cause a type mismatch. So, I'm not sure we should even support this use case.

Nokel81:
I am not quite sure that I agree that is a semver-patch change in the strictest sense because we are dealing with a type level library here.

From the definition of semver-patch and semver-minor, both are supposed to be resilient to multiple versions, I thought.

bbrk24:
> From the definition of semver-patch and semver-minor, both are supposed to be resilient to multiple versions, I thought.

That's in reference to upgrading. If you have a codebase that works with 1.2.3, it should still work with 1.2.4. I don't believe the standard says anything about having different parts of the project use both at the same time.

PR #403, opened 2022-06-02
feat: introduce UnwrapOpaque type

### LINKED ISSUE #165: Add Transparent to reverse Opaque
Opaque nominal types are incredibly useful, but there are times where you need to un-opaque a type ignorer to use it (e.g. as a key in an object).

I propose the addition of
```ts
export type Transparent<T> = T extends Opaque<infer O, any> ? O : T
```

fails now:
```ts
type MyId = Opaque<string, "MyId">
type Nums = Record<MyId, number>
```
> Element implicitly has an 'any' type because expression of type 'Opaque<string, "MyId">' can't be used to index type 'Record<Opaque<string, "MyId">, number>'

but this would fix that error without completely making Opaque types useless or overly compromising their contract:
```ts
type MyId = Opaque<string, "MyId">
type Nums = Record<Transparent<MyId>, number>
```

jmike:
This PR introduces a new `UnwrapOpaque ` type to remove the `[tag]` from Opaque values.

Why is this necessary?

1. Use an _opaque_ type as object keys
2. Prevent stupid `TS4058` error, i.e. _Return type of exported function has or is using name X from external module Y but cannot be named_.

Fixes #165

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -172,6 +172,7 @@ Click the type names for complete docs.
 - [`ReadonlyDeep`](source/readonly-deep.d.ts) - Create a deeply immutable version of an `object`/`Map`/`Set`/`Array` type. Use [`Readonly<T>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#readonlytype) if you only need one level deep.
 - [`LiteralUnion`](source/literal-union.d.ts) - Create a union type by combining primitive types and literal types without sacrificing auto-completion in IDEs for the literal type part of the union. Workaround for [Microsoft/TypeScript#29729](https://github.com/Microsoft/TypeScript/issues/29729).
 - [`Opaque`](source/opaque.d.ts) - Create an [opaque type](https://codemix.com/opaque-types-in-javascript/).
+- [`UnwrapOpaque`](source/opaque.d.ts) - Revert an [opaque type](https://codemix.com/opaque-types-in-javascript/) back to its original type.
 - [`InvariantOf`](source/invariant-of.d.ts) - Create an [invariant type](https://basarat.gitbook.io/typescript/type-system/type-compatibility#footnote-invariance), which is a type that does not accept supertypes and subtypes.
 - [`SetOptional`](source/set-optional.d.ts) - Create a type that makes the given keys optional.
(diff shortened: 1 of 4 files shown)


sindresorhus:
Any thoughts on the discussion in https://github.com/sindresorhus/type-fest/issues/165?

jmike:
As far as naming goes, I do not have a preference. I can amend this PR to anything you decide.

Re: `protected`, I don't like classes but they seem to do the trick here. I cannot say - needs extensive testing.

I just thought this discussion is ongoing for quite some time, so why not do something about it since I am also affected by the same problems at work.

sindresorhus:
I'm fine with adding this, but it should be called `UnwrapOpaque`.

jmike:
I have renamed the new type (as well as the PR title and description) to `UnwrapOpaque` per @sindresorhus suggestion.

source/opaque.d.ts
```diff
@@ -72,3 +72,36 @@ type Person = {
 @category Type
 */
 export type Opaque<Type, Token = unknown> = Type & Tagged<Token>;
+
+/**
+Remove readonly [tag] from an Opaque type, essentially reverting it back to a generic (i.e. non-Opaque) type.
```
sindresorhus, line 77: I would start this sentence with **what** it does, and then explain in another sentence how it does it.

> Reverts an `Opaque` type back to its original type. [...]

sindresorhus:
Looks good.

You need to add it to the readme.

source/opaque.d.ts
jmike: Done.

jmike:
Cool, the code has been updated accordingly.

readme.md
```diff
@@ -130,19 +130,19 @@ PR welcome for additional commonly needed types and docs improvements. Read the
...
+import type { Except } from "type-fest";
 
 type Foo = {
 	unicorn: string;
 	rainbow: boolean;
 };
 
-type FooWithoutRainbow = Except<Foo, 'rainbow'>;
+type FooWithoutRainbow = Except<Foo, "rainbow">;
 //=> {unicorn: string}
```
sindresorhus, line 146: Don't make unrelated changes

jmike: Apologies. This was unintentional.

Issue #402, opened 2022-06-01
`Merge` type does not handle correctly optional properties


adridavid:
The problem happens when `Merge` is used with two types that share one same key, and when the key is optional in the second type. The returned type for this property should be a union of the two initial types.

### Example

Given these two interfaces :

```ts
interface Foo {
    baz?: "a";
    qux: number
}

interface Bar {
    baz?: "b",
    qux: string
}
```

If we define two values typed as `Foo` and `Bar` :

```ts
const foo: Foo = {
    baz: "a",
    qux: 2
}

const bar: Bar = {
    qux: "qux"
}
```

After merging these two objects, the inferred type is :

```ts
const result = {
    ...foo,
    ...bar
}
// const result: {
//    baz?: "b" | "a" | undefined;
//    qux: string;
// }
```

The optional property `baz` is typed as `"b" | "a" | undefined`.

Using `Merge`, the optional property `baz` is typed as `"b" | undefined` :

```ts
type Result = Merge<Foo, Bar>
// type Result = {
//    baz?: "b" | undefined;
//    qux: string;
// }
```

`"a"` is missing in the union here.

I don't know if `Merge` should be changed to handle this use case or if a new type should be created ?

skarab42:
It's very strange, I don't expect that when I do an intersection. And TypeScript also seems to handles the case differently depending on whether you use the spread operator or an `Object.assign`. [REPL](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgGIHt3IN4FgBQyRyARnAF4D8AXMgERx0DcBxyAjgK4AetInAWxLQCAXwIFQkWIhQAhOFBytiZKrTok6AGhVEuvZAGcwUUAHMxE-AQToQJ5DEy0MWALzLCqihsa7vfR5aACYrG3w7BzBSRVoFJU88QI5g+gM6cNt7RygII04AGxikvWQAOkrndAC2SvKyKCzInJi8guKQ5E8AeRIAKwgEMHK4IyNgcxAACj7B4dHxyZnsUW0nTABKdcbNlgiwAE8ABxQ3BO60TGQAMmQE-ezo5HaisABmanPFS+S6qswtWI9UaYiAA)

```ts
const result2 = Object.assign(Object.assign({}, foo), bar); // infered as "Foo & Bar"

type FooBar = Foo & Bar;

const result3:FooBar = { // <- ERROR Type '"b"' is not assignable to type '"a"'
    ...foo,
[...]

PR #408, opened 2022-06-27
Add `MergeDeep` type

skarab42:
Merge two types recursively into a new type.

Properties set to `undefined` value are skipped when `strict` option is set to `true` (default).
Array and plain object properties are merged recursively.

```ts
type Foo = {foo: string; bar: {id: string; label: string}};
type Bar = {foo: number; bar: {id: number; nop: undefined}};

type FooBar = MergeDeep<Foo, Bar>;

// {
// 	foo: number;
// 	bar: {
// 		id: number;
// 		label: string;
// 	}
// }

type FooBarLazy = MergeDeep<Foo, Bar, {strict:false}>;

// {
// 	foo: number;
// 	bar: {
// 		id: number;
// 		label: string;
//		nop: undefined;
// 	}
// }
```

@ilchenkoArtem @sindresorhus `SetProp` is comming soooon!

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -9,7 +9,7 @@ export {Except} from './source/except';
 export {Mutable} from './source/mutable';
 export {Writable} from './source/writable';
 export {Merge} from './source/merge';
-export {MergeDeep} from './source/merge-deep';
+export {MergeDeep, MergeDeepOptions} from './source/merge-deep';
 export {MergeExclusive} from './source/merge-exclusive';
 export {RequireAtLeastOne} from './source/require-at-least-one';
 export {RequireExactlyOne} from './source/require-exactly-one';
--- a/index.d.ts
+++ b/index.d.ts
@@ -9,6 +9,7 @@ export {Except} from './source/except';
 export {Mutable} from './source/mutable';
 export {Writable} from './source/writable';
 export {Merge} from './source/merge';
+export {MergeDeep} from './source/merge-deep';
 export {MergeExclusive} from './source/merge-exclusive';
 export {RequireAtLeastOne} from './source/require-at-least-one';
 export {RequireExactlyOne} from './source/require-exactly-one';
--- a/source/internal.d.ts
+++ b/source/internal.d.ts
@@ -65,3 +65,13 @@ Extracts the type of an array minus the first element.
 export type ArrayTail<TArray> = TArray extends readonly [unknown, ...infer TTail]
   ? TTail
   : [];
+
+/**
+Matches any Array.
+*/
+export type AnyArray = unknown[] | readonly unknown[];
+
+/**
+Matches any Record.
+*/
+export type AnyRecord = Record<PropertyKey, unknown>;
--- a/source/internal.d.ts
+++ b/source/internal.d.ts
@@ -55,7 +55,7 @@ export type KeysOfUnion<T> = T extends T ? keyof T : never;
 /**
 Extracts the type of the first element of an array.
 */
-export type ArrayHead<TArray> = TArray extends readonly [infer THead, ...unknown[]]
+export type FirstArrayElement<TArray> = TArray extends readonly [infer THead, ...unknown[]]
   ? THead
   : never;
 
--- a/source/internal.d.ts
+++ b/source/internal.d.ts
@@ -51,3 +51,17 @@ The reason a simple `keyof Union` does not work is because `keyof` always return
 @link https://stackoverflow.com/a/49402091
 */
 export type KeysOfUnion<T> = T extends T ? keyof T : never;
+
+/**
+Extracts the type of the first element of an array.
+*/
+export type ArrayHead<TArray> = TArray extends readonly [infer THead, ...unknown[]]
+  ? THead
+  : never;
+
+/**
+Extracts the type of an array minus the first element.
+*/
+export type ArrayTail<TArray> = TArray extends readonly [unknown, ...infer TTail]
+  ? TTail
+  : [];
--- a/source/merge-deep.d.ts
+++ b/source/merge-deep.d.ts
@@ -1,3 +1,5 @@
+import {ConditionalExcept} from './conditional-except';
+
 /**
 MergeDeep options.
 
@@ -58,19 +60,13 @@ type GetValue<
   ? Destination[Key]
   : never;
 
-type ExtractType<TType, TUnion> = {
-  [Key in keyof TType]-?: TUnion extends TType[Key] ? Key : never;
-}[keyof TType];
-
-type OmitType<TType, TUnion> = Omit<TType, ExtractType<TType, TUnion>>;
-
 type MergeDeepLazy<Destination, Source, Options> = {
   [Key in Keyof<Destination, Source>]: Unwrap<
     GetValue<Destination, Source, Key, Options>
   >;
 };
 
-type MergeDeepStrict<Destination, Source, Options> = OmitType<
+type MergeDeepStrict<Destination, Source, Options> = ConditionalExcept<
   MergeDeepLazy<Destination, Source, Options>,
   undefined
 >;
--- a/source/merge-deep.d.ts
+++ b/source/merge-deep.d.ts
@@ -1,58 +1,100 @@
-type MergeDeepOptions = {
-	strict?: boolean;
-};
+interface MergeDeepOptions {
+  strict?: boolean;
+}
 
 type UnknownRecord = Record<string, unknown>;
 
 type Unwrap<Type> = Type extends UnknownRecord
-	? {[Key in keyof Type]: Type[Key]}
-	: Type;
+  ? {[Key in keyof Type]: Type[Key]}
+  : Type;
 
 type MergeValue<Destination, Source, Key, Options> =
-	Key extends keyof Destination
-		? Key extends keyof Source
-			? Destination[Key] extends UnknownRecord
-				? MergeDeep<Destination[Key], Source[Key], Options>
-				: Source[Key]
-			: never
-		: never;
+  Key extends keyof Destination
+    ? Key extends keyof Source
+      ? Destination[Key] extends UnknownRecord
+        ? MergeDeep<Destination[Key], Source[Key], Options>
+        : Source[Key]
+      : never
+    : never;
 
 type Keyof<Destination, Source> = keyof Destination | keyof Source;
 
+type MergeArrayValue<Destination, Source, Options> =
+  Destination extends UnknownArray
+    ? Source extends UnknownArray
+      ? ArrayMerge<Destination, Source, Options>
+      : Source
+    : Source;
+
 type GetValue<
-	Destination,
-	Source,
-	Key extends Keyof<Destination, Source>,
-	Options,
+  Destination,
+  Source,
+  Key extends Keyof<Destination, Source>,
+  Options,
 > = Key extends keyof Source
-	? Source[Key] extends UnknownRecord
-		? MergeValue<Destination, Source, Key, Options>
-		: Source[Key]
-	: Key extends keyof Destination
-	? Destination[Key] | null
-	: never;
+  ? Source[Key] extends UnknownRecord
+    ? MergeValue<Destination, Source, Key, Options>
+    : Source[Key] extends UnknownArray
+			? Key extends keyof Destination
+				? MergeArrayValue<Destination[Key], Source[Key], Options>
+				: Source[Key]
+			: Source[Key]
+  : Key extends keyof Destination
+  ? Destination[Key]
+  : never;
 
 type ExtractType<TType, TUnion> = {
-	[Key in keyof TType]-?: TUnion extends TType[Key] ? Key : never;
+  [Key in keyof TType]-?: TUnion extends TType[Key] ? Key : never;
 }[keyof TType];
 
 type OmitType<TType, TUnion> = Omit<TType, ExtractType<TType, TUnion>>;
 
 type MergeDeepLazy<Destination, Source, Options> = {
-	[Key in Keyof<Destination, Source>]: Unwrap<
-		GetValue<Destination, Source, Key, Options>
-	>;
+  [Key in Keyof<Destination, Source>]: Unwrap<
+    GetValue<Destination, Source, Key, Options>
+  >;
 };
 
 type MergeDeepStrict<Destination, Source, Options> = OmitType<
-	MergeDeepLazy<Destination, Source, Options>,
-	undefined
+  MergeDeepLazy<Destination, Source, Options>,
+  undefined
 >;
+
+type ArrayHead<TArray> = TArray extends readonly [infer THead, ...unknown[]]
+  ? THead
+  : never;
+
+type ArrayTail<TArray> = TArray extends readonly [unknown, ...infer TTail]
+  ? TTail
+  : [];
+
+type UnknownArray = readonly unknown[];
+
+type ArrayMergeValue<Destination, Source, Options> =
+  Destination extends UnknownRecord
+    ? Source extends UnknownRecord
+      ? Unwrap<MergeDeep<Destination, Source, Options>>
+      : Source
+    : Source;
+
+type ArrayMerge<
+  Destination extends UnknownArray,
+  Source extends UnknownArray,
+  Options extends MergeDeepOptions = {strict: true},
+> = Destination extends []
+  ? Source
+  : Source extends []
+  ? Destination
+  : [
+      ArrayMergeValue<ArrayHead<Destination>, ArrayHead<Source>, Options>,
+      ...ArrayMerge<ArrayTail<Destination>, ArrayTail<Source>>,
+    ];
 
 /**
 Merge two types recursively into a new type.
 
-Properties set to `undefined` value are skipped when strict is true (true by default).
+Properties set to `undefined` value are skipped
+when `strict` option is set to `true` (default).
 
 @example
 ```
@@ -84,9 +126,9 @@ Type FooBarLazy = MergeDeep<Foo, Bar, {strict:false}>;
 @category Object
 */
 export type MergeDeep<
-	Destination,
-	Source,
-	Options extends MergeDeepOptions = {strict: true},
+  Destination,
+  Source,
+  Options extends MergeDeepOptions = {strict: true},
 > = Options['strict'] extends true
-	? MergeDeepStrict<Destination, Source, Options>
-	: MergeDeepLazy<Destination, Source, Options>;
+  ? MergeDeepStrict<Destination, Source, Options>
+  : MergeDeepLazy<Destination, Source, Options>;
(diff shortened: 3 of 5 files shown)


sindresorhus:
Make sure you look at our existing "deep" types and that this new types handles all the cases handled by them.

For example, https://github.com/sindresorhus/type-fest/blob/2f418dbbb6182c53cdac26dc3421f7e8806c789b/source/readonly-deep.d.ts#L80

source/merge-deep.d.ts
```diff
@@ -0,0 +1,136 @@
+interface MergeDeepOptions {
+  strict?: boolean;
```
sindresorhus, line 2: This needs a doc comment and the interface should be exported.

source/merge-deep.d.ts
```diff
@@ -0,0 +1,136 @@
...
+}
+
+type UnknownRecord = Record<string, unknown>;
+
+type Unwrap<Type> = Type extends UnknownRecord
```
sindresorhus, line 7: Needs a short code comment about its purpose.

source/merge-deep.d.ts
```diff
@@ -0,0 +1,136 @@
...
+    : Source[Key] extends UnknownArray
+			? Key extends keyof Destination
+				? MergeArrayValue<Destination[Key], Source[Key], Options>
+				: Source[Key]
+			: Source[Key]
+  : Key extends keyof Destination
+  ? Destination[Key]
+  : never;
+
+type ExtractType<TType, TUnion> = {
```
sindresorhus, line 48: Move it to https://github.com/sindresorhus/type-fest/blob/main/source/internal.d.ts and add doc comment.

source/merge-deep.d.ts
```diff
@@ -0,0 +1,136 @@
...
+    GetValue<Destination, Source, Key, Options>
+  >;
+};
+
+type MergeDeepStrict<Destination, Source, Options> = OmitType<
+  MergeDeepLazy<Destination, Source, Options>,
+  undefined
+>;
+
+type ArrayHead<TArray> = TArray extends readonly [infer THead, ...unknown[]]
```
sindresorhus, line 65: Move it to https://github.com/sindresorhus/type-fest/blob/main/source/internal.d.ts and add doc comment.

source/merge-deep.d.ts
```diff
@@ -0,0 +1,136 @@
...
+  ? THead
+  : never;
+
+type ArrayTail<TArray> = TArray extends readonly [unknown, ...infer TTail]
```
sindresorhus, line 69: Move it to https://github.com/sindresorhus/type-fest/blob/main/source/internal.d.ts and add doc comment.

sindresorhus (changes requested):

source/merge-deep.d.ts
skarab42: Replaced with `ConditionalExcept` which already exists...

source/merge-deep.d.ts
skarab42: For the moment I have moved `ArrayHead` and `ArrayTail` to internal. But as the `LastArrayElement` type exists, I could add them as complementary types.

- FirstArrayElement (aka ArrayHead)
- RestArrayElements (aka ArrayTail)

And maybe later (if it's possible, I haven't thought about it):
- NFirstArrayElements
- NLastArrayElements

source/merge-deep.d.ts
skarab42: See https://github.com/sindresorhus/type-fest/pull/408#discussion_r909233665

skarab42:
> Make sure you look at our existing "deep" types and that this new types handles all the cases handled by them.
> 
> For example,
> 
> https://github.com/sindresorhus/type-fest/blob/2f418dbbb6182c53cdac26dc3421f7e8806c789b/source/readonly-deep.d.ts#L80

Sorry, I'm not sure what I'm supposed to test. Can you give me a concrete example?

source/internal.d.ts
```diff
@@ -51,3 +51,17 @@ The reason a simple `keyof Union` does not work is because `keyof` always return
 @link https://stackoverflow.com/a/49402091
 */
 export type KeysOfUnion<T> = T extends T ? keyof T : never;
+
+/**
+Extracts the type of the first element of an array.
+*/
+export type ArrayHead<TArray> = TArray extends readonly [infer THead, ...unknown[]]
```
sindresorhus, line 58: I think these have confusing names as they sound like the same for head/tail of the array, but actually do different things. The names should probably be something like `FirstArrayElement` and `ExceptFirstArrayElement`.

source/merge-deep.d.ts
```diff
@@ -0,0 +1,182 @@
+import {ConditionalExcept} from './conditional-except';
+import {ArrayHead, ArrayTail} from './internal';
+
+/**
+MergeDeep options.
+
+@param strict Properties set to `undefined` value are skipped when set to `true` (default).
```
sindresorhus, line 7:
```suggestion
@param strict - Properties set to `undefined` value are skipped when set to `true` (default).
```

source/merge-deep.d.ts
```diff
@@ -0,0 +1,182 @@
...
+}
+
+/**
+Represents an unknown record.
+*/
+type UnknownRecord = Record<string, unknown>;
+
+/**
+`Unwrap` is like `Simplify` it flatten the type output to improve type hints shown in editors.
+Unfortunately Simplify does not support the case where the type cannot be flattened.
```
sindresorhus, line 22: How does this handle types that cannot be flattened differently than `Simplify`?

index.d.ts
```diff
@@ -9,6 +9,7 @@ export {Except} from './source/except';
 export {Mutable} from './source/mutable';
 export {Writable} from './source/writable';
 export {Merge} from './source/merge';
+export {MergeDeep} from './source/merge-deep';
```
sindresorhus, line 12:
```suggestion
export {MergeDeep, MergeDeepOptions} from './source/merge-deep';
```

source/merge-deep.d.ts
```diff
@@ -0,0 +1,182 @@
...
+Determines the value to be returned according to the type of the `Source` and `Destination`.
+
+If both are array, returns the result of the merge, otherwise returns the `Source`.
+*/
+type MergeDeepArray<Destination, Source, Options> =
+  Destination extends UnknownArray
+    ? Source extends UnknownArray
+      ? ArrayMergeDeep<Destination, Source, Options>
+      : Source
+    : Source;
```
sindresorhus, line 57: Use tab-indentation.

source/merge-deep.d.ts
```diff
@@ -0,0 +1,182 @@
...
+type MergeDeepStrict<Destination, Source, Options> = ConditionalExcept<
+  MergeDeepLazy<Destination, Source, Options>,
+  undefined
+>;
+
+/**
+Merge two types recursively into a new type.
+
+Properties set to `undefined` value are skipped
+when `strict` option is set to `true` (default).
```
sindresorhus, line 145: Don't hard-wrap

sindresorhus (changes requested):

sindresorhus:
> Make sure you look at our existing "deep" types and that this new types handles all the cases handled by them.

Some other examples:

- Should it recurse into arrays?
- Should it recurse into Sets/Maps?

source/internal.d.ts
skarab42: This is in line with what I said in this comment https://github.com/sindresorhus/type-fest/pull/408#discussion_r909233665

skarab42: `FirstArrayElement` seems to be the best choice as [LastArrayElement](https://github.com/sindresorhus/type-fest/blob/main/source/last-array-element.d.ts#L21) works in the same way.

source/merge-deep.d.ts
skarab42: If you can't iterate in the type then it can't be flattened, a function for example. Here is the simplest case [TS Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAyglgWzAGzgMxAHgCoD4oC8UA3gNoDSEI240cAdlANZUD2aU2AugFycVUakLgF8A3AFgAUNNCQoAVXoB3AE4BDMDlr4iQ6BAAewCPQAmAZygAlCAGNWqs5gvBVDAOYAaKAFd6TPSsyvS40gCQAPwkAiBQDMxsHPq8nLSxohF8+pIyUnLQAMLqyMgARup2TDi6UAAUAJSE+Ni5srRQAIKEsIgo6FjFpRVVmG6+ELhh+R0AQj1KapqYQ+WV1eOTuEA)

skarab42:
> * Should it recurse into arrays?
> * Should it recurse into Sets/Maps?

At first I just wanted to contribute to a project I 💜 and improve my skills in the process by adding the [SetProp](https://github.com/sindresorhus/type-fest/pull/409) type. While prototyping, I realized that I would need a [Replace](https://github.com/sindresorhus/type-fest/blob/main/source/replace.d.ts) type to normalize the dotted paths. Then in several discussions I saw the need for the `DeepMerge` type, which would also simplify the creation of the `SetProp` type. Now, what I would love is to be able to type [lodash.merge](https://lodash.com/docs/4.17.15#merge), but this is not necessarily what you need in type-fest.

In any case I think I have to start all over again and define before going any further the needs for this type. 

If we want to go in the direction of lodash here is a non-exhaustive list  cases I have been able to define:

```ts
// If the first type (the destination) is [number, string, boolean] return the destination.
_.merge(42, 24); // 42 
_.merge("42", "24"); // "42" 
_.merge(true, false); // true 

// [undefined, null, symbol, Set, Map] are always treated as empty objects regardless of location.
_.merge(undefined, null); // {} 
_.merge(null, undefined); // {} 
_.merge(Symbol(42), Symbol(24)); // {}
_.merge(new Set([42]), new Set([24])); // {} 
_.merge(new Map([["life", 42]]), new Map([["life", 42]])); // {}  

// If the first type is [Class, Function] return undefined.
_.merge(() => 42, {life:42}); // undefined 
_.merge(Math.floor, {life:42}); // undefined 
_.merge(Buffer, {life:42}); // undefined 

// If the first type is an object and the second is an array, the array is treated as an indexed object.
_.merge({"id":42}, [42]); // {"0":42,"id":42} 
_.merge(null, [42]); // {"0":42} // <- remeber null is treated as empty object

// If the first type is an array and the second is an object, the object is treated as an numerical indexed array stripping any string indexed properties.
_.merge([42], {"0":24,"id":42}); // [24] 
_.merge([42], null); // [42] 

// Instance of class is treated as an plain object.
class Foo {
  constructor() {
    this.id = "foo";
    this.foo = true;
  }
}

class Bar {
  constructor() {
    this.id = "bar";
    this.bar = true;
  }
}

_.merge(new Foo(), new Bar()); // {"id":"bar","foo":true,"bar":true}
```

source/merge-deep.d.ts
skarab42: Perfect, I will make a PR ;)

source/merge-deep.d.ts
```diff
@@ -0,0 +1,181 @@
+import {ConditionalExcept} from './conditional-except';
+import {FirstArrayElement, ArrayTail} from './internal';
+
+/**
+MergeDeep options.
+
+@param strict - Properties set to `undefined` value are skipped when set to `true` (default).
```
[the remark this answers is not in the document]
skarab42: Good point, "loose" look good to me.

sindresorhus:
Relevant: https://github.com/sindresorhus/type-fest/pull/400

source/merge-deep.d.ts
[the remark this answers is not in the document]
sindresorhus: > name the option **specifically** for what it does

Loose is not specific either.

sindresorhus:
Also:

> Please help review the other open pull requests. - https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md#submitting-a-new-type

🙏

source/merge-deep.d.ts
```diff
@@ -0,0 +1,181 @@
...
+@see MergeDeep
+*/
+export interface MergeDeepOptions {
+	strict?: boolean;
+}
+
+/**
+Represents an unknown record.
+*/
+type UnknownRecord = Record<string, unknown>;
```
sindresorhus, line 18: Shouldn't this be:

```suggestion
type UnknownRecord = Record<PropertyKey, unknown>;
```

?

sindresorhus:
Maybe you have an opinion on https://github.com/sindresorhus/type-fest/issues/402 (write it in the issue) No worries if not.

skarab42:
I'm setting up the new signature for `DeepMerge` and I need you all to define two slightly confusing cases.

```ts
// The value "42" represents any type that is neither an record nor an array.

MergeDeep<{}, {}>; // {}
MergeDeep<[], []>; // never

MergeDeep<{}, []>; // never
MergeDeep<[], {}>; // never

MergeDeep<42, {}>; // never
MergeDeep<42, []>; // never

MergeDeep<{}, 42>; // never
MergeDeep<[], 42>; // never

MergeDeep<{}, {}, {recurseIntoArrays: true}>; // {}
MergeDeep<[], [], {recurseIntoArrays: true}>; // []

MergeDeep<{}, [], {recurseIntoArrays: true}>; // ???
MergeDeep<[], {}, {recurseIntoArrays: true}>; // ???

MergeDeep<42, {}, {recurseIntoArrays: true}>; // never
MergeDeep<42, [], {recurseIntoArrays: true}>; // never

MergeDeep<{}, 42, {recurseIntoArrays: true}>; // never
MergeDeep<[], 42, {recurseIntoArrays: true}>; // never
```

I think the two cases should be treated differently.

For the first case, If the first type (the destination) is an object then the second type is treated as an object.

```ts
MergeDeep<{life: 42, "1": 42}, [true, false], {recurseIntoArrays: true}>;
// {life: 42, "0": true, "1": false}
```

The second case is a little more ambiguous. I think you have to treat the second type as an array, but how do you handle indexes that are not defined?

```ts
MergeDeep<[true, false], {life: 42, "1": 42}, {recurseIntoArrays: true}>;
// [true, 42] or [null, 42]
```

I lean more towards the second answer `[null, 42]` because if you break it down you see this:

```ts
// {life: 42, "1": 42} -> {"0": null, "1": 42} -> [null, 42]
MergeDeep<[true, false], [null, 42], {recurseIntoArrays: true}>;
// [null, 42]
```

skarab42:
And how should "recurseIntoArrays" influence the merge of properties that are arrays?

```ts
MergeDeep<{items: [1, 2, 3, 4, 5, 6]}, {items: ['a', 'b', 'c']}>;
// 1. {items: ['a', 'b', 'c']}
// 2. {items: ['a', 'b', 'c', 4, 5, 6]}
// 3. {items: never}
// 4. never

MergeDeep<{items: [1, 2, 3]}, {items: [1, 2, 3, 4, 5, 6]}, {recurseIntoArrays: true}>;
// 1. {items: ['a', 'b', 'c']}
// 2. {items: ['a', 'b', 'c', 4, 5, 6]}
```

For me "recurseIntoArrays" should only influence the first call and should traverse and merge all child arrays regardless of whether "recurseIntoArrays" is true or false. This means that in all cases, proposal n°2 is applied.

sindresorhus:
> MergeDeep<{}, [], {recurseIntoArrays: true}>; // ???
MergeDeep<[], {}, {recurseIntoArrays: true}>; // ???

Both should be `never`

> For the first case, If the first type (the destination) is an object then the second type is treated as an object.

Arrays are techically objects, but it's better to treat them as not in general. I don't think anyone is gonna do this in practice and it's more likely it's a mistake, which we should catch.

> // 1. {items: ['a', 'b', 'c']} <---- my choice

:+1:

> // 2. {items: ['a', 'b', 'c', 4, 5, 6]} <---- my choice

I would have expected it to just concatenate the arrays.

skarab42:
> I would have expected it to just concatenate the arrays.

In fact I have thought wrongly, I should rather think in terms of types and not values.

```ts
MergeDeep<{items: [1, 2, 3, 4, 5, 6]}, {items: ['a', 'b', 'c']}>;
// -> {items: ['a', 'b', 'c']}
// -> {items: string[]}

MergeDeep<{items: [1, 2, 3, 4, 5, 6]}, {items: ['a', 'b', 'c']}, {recurseIntoArrays: true}>;
// -> {items: [1, 2, 3, 4, 5, 6, 'a', 'b', 'c']}
// -> {items: (string|number)[]}
```

Any suggestions?

sindresorhus:
> In fact I have thought wrongly, I should rather think in terms of types and not values.

I agree with your examples.

---

> // -> {items: string[]}

👍 

> // -> {items: (string|number)[]}

👍

Uzlopak:
This is really interesting. If you like you could provide some useful typings to @fastify/deepmerge :)

test-d/merge-deep.ts
```diff
@@ -0,0 +1,178 @@
...
+// Undefied properties must be skipped
+
+const merge4 = {a: 'a', b: 'b', e: 42};
+declare const type4: MergeDeep<typeof dest3, typeof src3, {stripUndefinedValues: true}>;
+
+expectType<typeof merge4>(type4);
+
+// Should replace arrays
+
+const src5 = {a: 'a', items: ['4', '5', '6']};
```
sindresorhus, line 89: Don't use abbreviations.
```suggestion
const source5 = {a: 'a', items: ['4', '5', '6']};
```

test-d/merge-deep.ts
```diff
@@ -0,0 +1,178 @@
...
+
+expectType<typeof merge5>(prout5);
+
+// Should concat arrays
+
+const merge6 = {a: 'a', b: 'b', items: [1, 2, 3, '4', '5', '6']};
+declare const prout6: MergeDeep<typeof dest5, typeof src5, {recurseIntoArrays: true}>;
+
+expectType<typeof merge6>(prout6);
+
```
sindresorhus, line 103: Can you add a test to ensure readonly arrays are supported as input?

source/merge-deep.d.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  @default false
+  */
+  stripUndefinedValues?: boolean;
+}
+
+/**
+Merge two types recursively into a new type.
+
+By default arrays are not supported, you have to explicitly enable it with the `recurseIntoArrays` option if you want it.
```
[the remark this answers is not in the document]
skarab42: Sorry for the long silence, too many yaks to shave.

> ```ts
> MergeDeep<{}, {}>; // {}
> MergeDeep<[], []>; // never
> ```
This is not what we decided above, but I am ok with this change.

sindresorhus:
@skarab42 Bump :)

skarab42:
@sindresorhus I'm working on a better version (I hope), it will take a few more days.

source/merge-deep.d.ts
```diff
@@ -0,0 +1,159 @@
...
+
+@see MergeDeepOptions
+
+@example
+```
+MergeDeep<[], []>;
+//=> never
+
+MergeDeep<[], [], {recurseIntoArrays: true}>;
+//=> []
```
sindresorhus, line 149: This is just showing a couple of edge-cases. It needs to show a real-world example usage.

skarab42:
I am closing this PR due to the new proposal #443.

PR #409, opened 2022-06-27
Add `SetProperty` type

### LINKED ISSUE #373: Add `Set` type
Add `Set` type. Looks like `Get`

Example: 
```typescript
type NewResult = Set<'level1.level2.level3',  number>

result:  {
 level1: {
   level2: {
     level3: number
   }
 }
}
```

skarab42:
Related to #408 
Close #373

Set a deeply-nested property to an object using a key path, like Lodash's `.set()` function.

```ts
import set from "lodash/set";

const data = {name: 'nyan', items: {life: 24}};

set(data, 'items.life', '42');
set(data, 'items.levels', [1, 2, 3]);

type SetLifeString = SetProp<typeof data, 'items.life', '42'>;
type CreateLevels = SetProp<SetLifeString, 'items.levels', number[]>;

export const newData = data as unknown as CreateLevels;
```

---

Fixes #373


source/merge-deep.d.ts
```diff
@@ -0,0 +1,136 @@
+interface MergeDeepOptions {
+  strict?: boolean;
+}
+
+type UnknownRecord = Record<string, unknown>;
```
sindresorhus, line 5: Shouldn't this be:

```suggestion
type UnknownRecord = Record<PropertyKey, unknown>;
```

?

sindresorhus: Any maybe it should be named `AnyRecord` and moved to internal.d.ts?

source/set-prop.d.ts
```diff
@@ -0,0 +1,50 @@
+import {MergeDeep, Unwrap} from './merge-deep';
+import {Replace} from './replace';
+import {Split} from './split';
+
+type ReplaceByDot<Path extends string, Search extends string> = Replace<Path, Search, '.', {all: true}>;
+type NormalizePath<Path extends string> = ReplaceByDot<ReplaceByDot<Path, '['>, '].'>;
+type SplitPath<S extends string> = Split<NormalizePath<S>, '.'>;
+
+type PathHead<List extends string[]> = List extends [infer Head, ...string[] ]
```
sindresorhus, line 9: Same as the MergeDeep comment regarding naming

source/set-prop.d.ts
```diff
@@ -0,0 +1,50 @@
...
+  : never;
+
+type SetPropArray<Path extends string[], Value> = Unwrap<Record<PathHead<Path>, SetPropDeep<PathTail<Path>, Value>>>;
+
+/**
+Set a deeply-nested property to an object using a key path, like Lodash's `.set()` function.
+
+@example
+```
+import set from "lodash/set";
```
sindresorhus, line 34: Single-quotes

sindresorhus:
Maybe you can find some inspiration in the implementation shared in https://github.com/sindresorhus/type-fest/issues/373#issuecomment-1156695584

sindresorhus:
I would prefer to name this `SetProperty`.

sindresorhus:
Hmm, actually, how about the name `SetKeyPath`?

skarab42:
Hum, `SetDotPath` ?

sindresorhus:
I think "key path" is a more established term for it.

However, not sure whether it should be `SetKeyPath` or `KeyPathSet`.

Also open to other alternatives.

skarab42:
I'm not very good at finding coherent names, I think I'm better at implementing them. By the way, I need to refactor it to use `MergeDeep`.

sindresorhus:
Let's go with `KeyPathSet`.

My reasoning. You're not setting the key path itself. You're using the key path to set a property. It's even more apparent with a "get". `GetKeyPath` would sound like you're getting the key path for a object, not the property at a key path.

sindresorhus:
We should probably also rename the `Get` method to `KeyPathGet` (while keeping an alias for `Get`).

sindresorhus:
Hmm, actually, thinking more about this. This goal of the type is to set a property. So I think maybe `SetProperty` is the best name for it after all.

skarab42:
@sindresorhus The refacto is almost done, but before documenting I would like some feedback on the tests, if you see anything wrong. The only doubt I have at the moment is whether to allow modification of readonly objects.

sindresorhus:
> The only doubt I have at the moment is whether to allow modification of readonly objects.

Why are you having doubts about that? If it's readonly, isn't it clear it should not allow modification? Or is there something I'm missing?

ehoogeveen-medweb:
I guess the question is whether the `readonly` parts of a *type* should be (deeply) immutable. Since the purpose of this utility is to generate a new type from an old one, I don't think it makes sense to limit the modifications it can do. Typescript will still do type checking based on the resulting types anyway.

sindresorhus:
> I guess the question is whether the readonly parts of a type should be (deeply) immutable. Since the purpose of this utility is to generate a new type from an old one, I don't think it makes sense to limit the modifications it can do. Typescript will still do type checking based on the resulting types anyway.

👍

Issue #406, opened 2022-06-25
`NaN` problems


frontsideair:
Hi, I just found out about this project while writing [my own `Jsonify` helper](https://gist.github.com/frontsideair/dfd696bc927514ba4a78e1c5ed4acd1c).

I wanted to document a few limitations regarding the unfortunate situation with `NaN`. Maybe this can be added to the readme.

Unfortunately, [`NaN` is not a TypeScript literal](https://github.com/microsoft/TypeScript/issues/28682) and cannot be detected at the type-level. I'm surprised that you found a way to detect `Infinity` and `-Infinity` so I still have some hope. :)

The affected helpers are `Jsonify` and `Finite`, while this list may not be exhaustive.

In case of `Jsonify`, a `NaN` value is encoded as `null` a `number` can be incorrectly encoded as `number` when it's a `NaN`.

In case of `Finite`, since `NaN` is not included in finite numbers (according to `Number.isFinite` check), it can also return an incorrect type in case of `NaN`.

For context, I needed a `Jsonify` utility to have typed loader data in Remix, after a JSON encode/decode chain. I also opened an issue (https://github.com/microsoft/TypeScript/issues/48697) in TypeScript repo to request a first-class helper type.

I hope this issue will be helpful and I can try to help in any way I can.

ikeyan:
@fregante While there is no `NaN` literal type in TS, you can define a NaN enum member, and it behaves like a NaN enum literal.

for example:

https://www.typescriptlang.org/play/?#code/KYOwrgtgBAgiCi5oG8BQUNQHJQLxQAYoB6QgGlQF9VUBjAexAGcAXKR4ALigEY9YESAHRYA3CVIAVAJ4AHYFADkcRJEVQAlkygh6bAIZMmGgOYh9AIwA2ClvSgs5CxT0VC6jVlH1ZIF4ABO3Ej+AfwqwmKojvLYPgDKLGH4AAYAJMgRkCKUKeLEpNL0YFC0+iBQACbALMC0BnE4MQoW0lCsARogJhoAZhplLBqMqEA
```typescript
enum AnEnum {
    N = 0 / 0,
}

const one: 1 = AnEnum.N; // Type 'AnEnum' is not assignable to type '1'.
const aNumber: number = AnEnum.N;
type NaNStr = `${AnEnum.N}`; // you can detect a NaN type by stringification
```

PR #410, opened 2022-06-28
Jsonify correctness

### LINKED ISSUE #407: `Jsonify` problems
I mentioned in #406 that I was building my own `Jsonify`. I have a small test suite that I based on [the spec](https://tc39.es/ecma262/#sec-json.stringify) and there's a [simplified version on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#description).

I put the definition in here and it fails a few cases, namely these:

- `Number`, `String` and `Boolean` should be turned into primitives
- `undefined`, `Function` and `symbol` should make keys to be omitted if they are values in an object
- `Function` and `symbol` in an array should turn into `null`
- `symbol` keys should be omitted in an object

I can help with these if it aligns with the project goals. My plan would be first add these tests, then fix the current definition to make the tests pass. [My current implementation](https://gist.github.com/frontsideair/dfd696bc927514ba4a78e1c5ed4acd1c) passes the tests, but is not exhaustively tested and it's not as modular as the one in this repo, so I would rather improve this implementation.

frontsideair:
This is my first attempt at fixing Jsonify problems I mentioned in #407.

I started with tests first, then proceeded to fix those tests. Tests are based on [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#description) and [my own tests](https://gist.github.com/frontsideair/dfd696bc927514ba4a78e1c5ed4acd1c#file-jsonify-spec-ts).

I had to delete a faulty test case, where an object had a function value. The spec tells that those entries are omitted from the output.

I checked the validity of the types with the provided `test` script. It also output a lot of lint warnings that were there before I picked it up, so I left them alone. I can fix those errors if you want me to, but it will end up destroying the git blame.

I'm not sure if I documented everything I did as best as I could, since this was a first attempt. I can fix anything where you thing it needs more explanation.

I also commented out the Infinity and NaN checks. Infinity is detectable with this library, but it seemed like a minor edge case. NaN is not detectable as far as I can understand, see #406.

Also missing is `TypedArray`s. I'm not sure how to approach them, any guidance is welcome, or we can get this merged and leave it as future work.

I can also add a warning about NaNs in the readme, or leave it as future work as well.

Sorry for a lot of text, I just wanted to cover everything.

---

Fixes #407

### DIFF
--- a/test-d/jsonify.ts
+++ b/test-d/jsonify.ts
@@ -123,17 +123,85 @@ const nonJsonWithInvalidToJSON = new NonJsonWithInvalidToJSON();
 expectNotAssignable<JsonValue>(nonJsonWithInvalidToJSON);
 expectNotAssignable<JsonValue>(nonJsonWithInvalidToJSON.toJSON());
 
-// Special cases of Array with `undefined` member
-// `[undefined]` is not JSON because it contains non JSON value `undefined`
-// However `JSON.parse(JSON.stringify())` transforms array members of `undefined` to `null`
-expectNotAssignable<JsonValue>([undefined]);
-declare const parsedStringifiedArrayWithUndefined1: Jsonify<undefined[]>; // = JSON.parse(JSON.stringify([undefined]));
-expectType<null[]>(parsedStringifiedArrayWithUndefined1);
-expectAssignable<JsonValue>(parsedStringifiedArrayWithUndefined1);
-expectNotAssignable<JsonValue>([undefined, 1]);
-declare const parsedStringifiedArrayWithUndefined2: Jsonify<[undefined, number]>;
-expectType<Array<number | null>>(parsedStringifiedArrayWithUndefined2);
-expectAssignable<JsonValue>(parsedStringifiedArrayWithUndefined2);
+// Not jsonable types; these types behave differently when used as plain values, as members of arrays and as values of objects
+declare const undef: undefined
+expectNotAssignable<JsonValue>(undef);
+
+declare const fn: (_: any) => void;
+expectNotAssignable<JsonValue>(fn);
+
+declare const sym: symbol
+expectNotAssignable<JsonValue>(sym);
+
+// Plain values fail JSON.stringify()
+declare const plainUndef: Jsonify<typeof undef>
+expectType<never>(plainUndef)
+
+declare const plainFn: Jsonify<typeof fn>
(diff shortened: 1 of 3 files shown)


test-d/jsonify.ts
```diff
@@ -123,17 +123,85 @@ const nonJsonWithInvalidToJSON = new NonJsonWithInvalidToJSON();
...
-expectNotAssignable<JsonValue>([undefined]);
-declare const parsedStringifiedArrayWithUndefined1: Jsonify<undefined[]>; // = JSON.parse(JSON.stringify([undefined]));
-expectType<null[]>(parsedStringifiedArrayWithUndefined1);
-expectAssignable<JsonValue>(parsedStringifiedArrayWithUndefined1);
-expectNotAssignable<JsonValue>([undefined, 1]);
-declare const parsedStringifiedArrayWithUndefined2: Jsonify<[undefined, number]>;
-expectType<Array<number | null>>(parsedStringifiedArrayWithUndefined2);
-expectAssignable<JsonValue>(parsedStringifiedArrayWithUndefined2);
+// Not jsonable types; these types behave differently when used as plain values, as members of arrays and as values of objects
+declare const undef: undefined
```
sindresorhus, line 127: Don't use abbreviations for the variable names. Applies to all of them.

sindresorhus:
You need to fix the lint errors.

sindresorhus:
> Also missing is TypedArrays. I'm not sure how to approach them, any guidance is welcome

What's the difficulty with supporting TypedArrays?

sindresorhus:
> I can also add a warning about NaNs in the readme, or leave it as future work as well.

Both

test-d/jsonify.ts
frontsideair: Fixed all of them except `fn`, can't use `function` since it's a reserved keyword.

frontsideair:
Sorry for the lint errors point, I thought they were there before but I was mistaken. I fixed them all. I'll investigate the typed arrays and update the readme about limitations.

frontsideair:
@sindresorhus I think it should be ready for review. I fixed newly introduced lint errors as well.

sindresorhus:
Thanks for contributing 🙏

PR #411, opened 2022-06-29
Add `UnionToArray` type

santosmarco:
# Overview

- Added the `UnionToArray` type, which takes a literal union as a parameter and converts it to an array containing _exactly_ the elements of the union.

For example,

```ts
type Union = 'a' | 'b' | 'c' | 'd' | 'e' | 'f';

const array: UnionToArray<Union> = ['a', 'b', 'c', 'd', 'e', 'f']; // Requires all the elements of the union, exactly once.
```

### DIFF
--- a/source/union-to-array.d.ts
+++ b/source/union-to-array.d.ts
@@ -1,0 +1,22 @@
+type _UnionToArray<T extends string> = {
+	[K in T]: [...(Exclude<T, K> extends never ? [] : _UnionToArray<Exclude<T, K>>[Exclude<T, K>]), K]
+}
+
+/**
+Creates an array of string literals that matches exactly the union type passed in.
+
+Use-case:
+- When you have a union of string literals but need an array containing all the elements of the union.
+
+@example
+```
+import type {UnionToArray} from 'type-fest';
+
+type Union = 'a' | 'b' | 'c' | 'd' | 'e' | 'f';
+
+const array: UnionToArray<Union> = ['a', 'b', 'c', 'd', 'e', 'f'];
+```
+
+@category Array
+*/
+export type UnionToArray<T extends string> = _UnionToArray<T>[T]


sindresorhus:
Thanks for contributing. Please adhere to the contribution guidelines: https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md

skarab42:
For me it's just a tuple, what's the difference with that:
``` ts
const array2: Union[] = ['a', 'b', 'c', 'd', 'e', 'f'];
```

[REPL](https://www.typescriptlang.org/play?#code/C4TwDgpgBA+gqgOwJYHsEBUUEEBOOCGIAPOlBAB7AQIAmAzlHcDkggOYB8UAvFAN4BYAFABIANoBpKKyjoAugC4oYgHRqAFAFFyAYwA2AVxoQSAGigSuFKrQYIIANwg4oAfmVyoS+MjSZcBMTa+kYm6OaWHGLBhsZmFhxyAJQRcsIAvsLCFGAoOMBQoJBQiKgY2HiEJGSU1PSMzKycPLClfhWBJFHyWUJF0G0ILQDk+MNQAD5QwwBG41PDOvPTNMvDEGsAZsMA3L06aExQ+JUgSoP+p0SDXLxio8Pms4-TS0+rTxtP23I7UL3CA4II4nQIAJnOvgQYk8dweTzmTzeKxe61R23MABYwb9-kJhEA)

santosmarco:
In your example, if you add a new literal to the union in the future, TypeScript will never let you know that you have forgotten to add that new literal to `array2`.

Suppose you have an array that you use to validate if a string is one of the items in that array. Every time you add a new literal to the union, you should update that array; otherwise, the validation will fail. If you type it as you did, TS will not complain; on the other hand, if you type it with my suggestion, it will not compile until you add the new literal to the validation array.


>

sindresorhus:
Does it return a tuple? Just wondering whether it should match the naming (in inverse) of https://github.com/sindresorhus/type-fest/pull/440

sindresorhus:
There are some lint issues.

sindresorhus:
You also need tests and adding it to the readme.

source/union-to-array.d.ts
```diff
@@ -0,0 +1,22 @@
...
+@example
+```
+import type {UnionToArray} from 'type-fest';
+
+type Union = 'a' | 'b' | 'c' | 'd' | 'e' | 'f';
+
+const array: UnionToArray<Union> = ['a', 'b', 'c', 'd', 'e', 'f'];
+```
+
+@category Array
```
sindresorhus, line 20:
```suggestion
@see TupleToUnion
@category Array
```

And also add this type as `@see` to `TupleToUnion`.

sindresorhus:
Bump :)

skarab42:
@santosmarco It seems that it only works with strings [Playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwL5wGZQiOAiGBPMAUwFpVCBnGbAKGvyLgH0BVAO2AlYBUIBBKKAEM8AHi5xCADxiFWAE3JxKUYKwDmAPjgBeRNQCQAbQDScVXC4BdAFxxDAOkcAKAKKSAxgBsArnMJiAGjhjLSkZeUVWQgA3Qig4AH47SzhbFnZOHn4hUTcvX38uIJCNQzyfP0DgjUsASmLLaiRaekI4Ng5uPgFhMQlpWQUlGBV1LV10zqye0S5Sq1o6AjaOzh04AHJBDbgAH02AIx39jfdjzblzjcIr1A2Ablp3Tko4QRnbVa7s3q-xuy2GyCGyOwLOwMuwJuwLulkeSwYAFlgJJCHIvutAXs4ABGbGnc4AJnxN2xAGZ4c9WK93jlCZ8Mt8ZiJkaj0Yz-oZAUEcWCgXBCVD+WS4UA).

@sindresorhus In fact this subject is more complex than it seems, and the challenge of keeping the tuple order is still not solved to my knowledge.

Some articles that provide solutions:
- https://catchts.com/union-array
- https://github.com/microsoft/TypeScript/issues/13298
- https://stackoverflow.com/questions/50374908/transform-union-type-to-intersection-type/50375286#50375286

NOTE: We already have [UnionToIntersection](https://github.com/sindresorhus/type-fest/blob/main/source/union-to-intersection.d.ts) which is part of the solution

sindresorhus:
Moving to an issue as this is not moving forward: https://github.com/sindresorhus/type-fest/issues/534

PR #413, opened 2022-07-01
Fix `Includes` null includes undefined

JonghwanWon:
I modified what passes when null includes undefined.

### DIFF
--- a/source/includes.d.ts
+++ b/source/includes.d.ts
@@ -15,8 +15,8 @@ type hasRed<array extends any[]> = Includes<array, 'red'>;
 @category Array
 */
 export type Includes<Value extends readonly any[], Item> =
-	IsEqual<Value[0], Item> extends true
-		? true
-		: Value extends [Value[0], ...infer rest]
-			? Includes<rest, Item>
-			: false;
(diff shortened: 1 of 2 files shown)


sindresorhus:
Can you explain your change?

JonghwanWon:
> Can you explain your change? Why does changing the order fix the issue?

Hello, @sindresorhus thank your comment. First at all, My English is not good. so, I attached source code. Please forgive me if there are any mistakes.

I've distinguished the old 'Includes' from the changed 'Includes' for comparison.

```
type T1= IsEqual<null, null>  // true
type T2= IsEqual<undefined, undefined>  // true
type T3= IsEqual<null, undefined>  // false

type T4 = OldIncludes<[null], null>  // true
type T5 = NewIncludes<[null], null>  // true

type T6 = OldIncludes<[undefined], undefined>  // true
type T7 = NewIncludes<[undefined], undefined>  // true

type T8 = OldIncludes<[null], undefined>  // Expected to be "false". but type is "true"
type T9 = NewIncludes<[null], undefined>  // false
```

I thought it was strange that the result was different when I ended up using `isEqual` depending on the internal behavior in `Includes`.

PR #414, opened 2022-07-01
Fix: `Simplify` handle all types

skarab42:
Related to https://github.com/sindresorhus/type-fest/pull/408#discussion_r909523586

- Fix the case where the type to be simplified is a function. In this case we return the type without modification.

### DIFF
--- a/source/simplify.d.ts
+++ b/source/simplify.d.ts
@@ -1,3 +1,8 @@
+import {Class, Constructor} from './basic';
+
+// Flatten a type without worrying about the result.
+type Flatten<AnyType> = {[KeyType in keyof AnyType]: AnyType[KeyType]};
+
 /**
 Useful to flatten the type output to improve type hints shown in editors. And also to transform an interface into a type to aide with assignability.
 
@@ -55,4 +60,4 @@ fn(someInterface as Simplify<SomeInterface>); // Good: transform an `interface` 
 
 @category Object
 */
-export type Simplify<T> = T extends Function ? T : {[KeyType in keyof T]: T[KeyType]};
+export type Simplify<AnyType> = Flatten<AnyType> extends AnyType ? Flatten<AnyType> : AnyType;
--- a/source/simplify.d.ts
+++ b/source/simplify.d.ts
@@ -1,7 +1,23 @@
 import {Class, Constructor} from './basic';
 
+/**
+Simplify options.
+
+@param deep - Do the simplification recursively (default: false).
+
+@see Simplify
+*/
+export interface SimplifyOptions {
+	deep?: boolean;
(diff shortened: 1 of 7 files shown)


test-d/simplify.ts
```diff
@@ -42,3 +42,25 @@ expectType<SomeInterfaceAsTypeWrittenByHand>(valueAsInterface);
 expectAssignable<Record<string, unknown>>(valueAsLiteral);
 expectAssignable<Record<string, unknown>>(valueAsSimplifiedInterface);
 expectNotAssignable<Record<string, unknown>>(valueAsInterface); // Index signature is missing in interface
+
+// Should return the original type if it is not simplifiable, like a function.
+type SomeFunction = (type: string) => string;
+expectType<Simplify<SomeFunction>>((type: string) => type);
+
+// Sould simplify to public interface (what it is currently doing)? Return the Class type? Or something else?
```
skarab42, line 50: I also wonder about classes. Currently it returns a simplified public version of the class interface.
What do you expect? the public interface? complete interface? or just the original type?

sindresorhus: I would say, just the original type. The point of the utility is to simplify nested and complicated types.

skarab42:
@sindresorhus I think I have found a trick to support any type without testing all cases. The strategy is if the simplified type extends the source type then we return the simplified type otherwise the source.

I also added the `deep` option to apply the simplification recursively.

source/simplify.d.ts
```diff
@@ -1,3 +1,24 @@
+import {Class, Constructor} from './basic';
+
+/**
+Simplify options.
+
+@param deep - Do the simplification recursively (default: false).
```
sindresorhus, line 6: This should be attached to the `deep` property itself. And use `@default` for the default value

source/simplify.d.ts
```diff
@@ -55,4 +76,9 @@ fn(someInterface as Simplify<SomeInterface>); // Good: transform an `interface`
 
 @category Object
 */
-export type Simplify<T> = {[KeyType in keyof T]: T[KeyType]};
+export type Simplify<
+	AnyType,
+	Options extends SimplifyOptions = {},
+> = Flatten<AnyType> extends AnyType
```
sindresorhus, line 82:
```suggestion
> = Flatten<AnyType, Options> extends AnyType
```

?

sindresorhus (changes requested):

source/simplify.d.ts
skarab42: No need to pass the options here, it's just for the test, if it doesn't pass without the deep there's no point in going any further

PR #416, opened 2022-07-02
Fix `Replace` and add `recursive` option.

JonghwanWon:
If specify the `{all:true}` option for the current type `Replace`, we can expect such as `.replaceAll` in Javascript.

But the current `Replace` is not. Constructs the result of `replace` recursively execution.

```ts
declare function replaceAll<
	Input extends string,
	Search extends string,
	Replacement extends string,
>(
	input: Input,
	search: Search,
	replacement: Replacement
): Replace<Input, Search, Replacement, {all: true}>;

replaceAll('foobarfoobar', 'ob', 'b'); //  expected to be "fobarfobar". but is "fbarfbar"
```

Therefore, the `recursive` option is added so that it can be accurately distinguished from `all`.

```ts
declare function replaceAll<
	Input extends string,
	Search extends string,
	Replacement extends string,
>(
	input: Input,
	search: Search,
	replacement: Replacement
): Replace<Input, Search, Replacement, {all: true}>;

declare function replaceRecursively<
	Input extends string,
	Search extends string,
	Replacement extends string,
>(
	input: Input,
	search: Search,
	replacement: Replacement
): Replace<Input, Search, Replacement, {recursive: true}>;

replaceAll('foobarfoobar', 'ob', 'b'); // 'fobarfobar'
replaceRecursively('foobarfoobar', 'ob', 'b'); //  'fbarfbar'


```


sindresorhus:
This just sounds like q bug in the `all` option.

// @skarab42

skarab42:
> Therefore, the `recursive` option is added so that it can be accurately distinguished from `all`.

I think that we should not add complexity, but simply fix the bug. For me "all" and "recursive" are the same thing.

skarab42:
In any case, thank you for poiting this bug. Your solution seems very good (without the complexity) ;)

sindresorhus:
@JonghwanWon Can you drop the recursive option and just make this a bug fix?

JonghwanWon:
> @JonghwanWon Can you drop the recursive option and just make this a bug fix?

Reverted recursive option and simply fixed the bug.

sindresorhus:
Can you include a test that would fail before this fix? To ensure it doesn't regress in the future.

Issue #118, opened 2020-07-07
Proposal: ArrayItem


scvnathan:
We've found this type handy in our codebase and thought this would be useful in `type-fest` too.

## Proposal
`ArrayItem` extracts the type of the item(s) in the array:

```ts
type ArrayItem<TArray extends any[]> = TArray extends (infer E)[] ? E : TArray;
```

Possibly could be improved by replacing the fallback conditional with `unknown` instead of `TArray`.

playground example: https://www.typescriptlang.org/play/#code/C4TwDgpgBAggTnAhiAksCBbAPAFXkkKCAD3QDsATAZykTJAG0BdAPigF4o8FkjSJKNABQBLMgDMIcKAFEAlMygB+WVABcXfMgDcAKF2hIUNJhqcA3lHEB7axqrA4YgObaoAI0RwNZAK4Z3KSgAX2Y9Q2gAWRFiCAoTDDMoIUsbOygHJzJXDy8ff0DpYKgAMihLTwAvDQAhEWcUMnRnIOCFJj0DcGgRdAwOWB5UPqwEqhZw7qgMGLiEga1hzCxo2Pi+8cmjMQA3RAAbEQoFoYSsPwCpCaA

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/sindresorhus) so you can upvote and help fund this issue.
- The funding will be given to active contributors.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/sindresorhus/type-fest/issues/118">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/118/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/sindresorhus/type-fest/issues/118/pledge.svg">
</picture>
</a>

sindresorhus:
> We've found this type handy in our codebase

I'm curious what *exactly* you use it for? Maybe you could provide an example of how you use it in your codebase?

scvnathan:
> > We've found this type handy in our codebase
> 
> I'm curious what _exactly_ you use it for? Maybe you could provide an example of how you use it in your codebase?

Sure, I'll provide some examples as soon as I can

scvnathan:
[...]

PR #421, opened 2022-07-09
fix(Exact): support array union

### LINKED ISSUE #419: `Exact` intersection type support
I was playing with @zorji 's `Exact` type and found that it errors for intersection types, but specifically when you construct the intersection type like `Array<{ a: string }> & Array<{ b: string }>`

I can only reproduce the error in the following scenario:

```typescript
type Type = Array<{ test: string }> & Array<{ z: number }>
// Can't construct Exact type with errors saying it's trying to also match on array method 'pop' and etc etc
const fn = <T extends Exact<Type, T>>(args: T) => args;

// correct input
fn([{
  code: 'test',
  name: '1'
}])

// missing field
fn([{
  code: 'test'
}])

// incorrect type
fn([{
  code: 'test',
  name: 1
}])

// excess field
fn([{
  code: 'test',
  name: '1',
  excessField: '',
}])
```

As per https://www.reddit.com/r/typescript/comments/m5wvrx/how_to_extract_proper_intersection_type_from_an/ and https://github.com/microsoft/TypeScript/issues/43267 

It seems like a workaround for those who have control over type declarations would be writing the intersection type like following:

```typescript
type Type = ({ code: string } & { name: string })[]
const fn = <T extends Exact<Type, T>>(args: T) => args;

// correct input
fn([{
  code: 'test',
  name: '1'
}])

// missing field
fn([{
  code: 'test'
}])

// incorrect type
fn([{
  code: 'test',
  name: 1
}])

// excess field
fn([{
  code: 'test',
  name: '1',
  excessField: '',
}])
```

The only problem is when the user doesn't have control over the underlying module's type declaration but they are providing the weird intersection type that goes `Array<{ ... }> & Array<{...}>`, then they will get the error when trying to use `Exact`

zorji:
Fixes #419

### DIFF
--- a/source/exact.d.ts
+++ b/source/exact.d.ts
@@ -5,7 +5,7 @@ Extract the element of an array that also works for array union.
 Return `never` if T is not an array.
 It creates a type-safe way to access the element type of `unknown` type.
 */
-type ArrayElement<T> = T extends unknown[] ? T[0] : never;
+type ArrayElement<T> = T extends readonly unknown[]? T[0] : never;
 
 /**
 Extract the object field type if T is an object and K is a key of T, return `never` otherwise.
@@ -64,5 +64,7 @@ onlyAcceptNameImproved(invalidInput); // Compilation error
 export type Exact<ParameterType, InputType> =
 	// Converting union of array to array of union. i.e. A[] & B[] => (A & B)[]
 	ParameterType extends unknown[] ? Array<Exact<ArrayElement<ParameterType>, ArrayElement<InputType>>>
+	// In TypeScript, Array is a subtype of ReadonlyArray, so always test Array before ReadonlyArray.
+	: ParameterType extends readonly unknown[] ? ReadonlyArray<Exact<ArrayElement<ParameterType>, ArrayElement<InputType>>>
 	: ParameterType extends object ? ExactObject<ParameterType, InputType>
 	: ParameterType;
--- a/source/exact.d.ts
+++ b/source/exact.d.ts
@@ -1,5 +1,25 @@
-import type {Primitive} from './primitive';
 import type {KeysOfUnion} from './internal';
+
+/**
+Extract the element of an array that also works for array union.
+Return `never` if T is not an array.
+It creates a type-safe way to access the element type of `unknown` type.
+*/
+type ArrayElement<T> = T extends unknown[] ? T[0] : never;
+
+/**
+Extract the object field type if T is an object and K is a key of T, return `never` otherwise.
+It creates a type-safe way to access the member type of `unknown` type.
+*/
+type ObjectValue<T, K> = K extends keyof T ? T[K] : never;
+
+/**
+Create a type from `ParameterType` and `InputType` and change keys exclusive to `InputType` to `never`.
+- Generate a list of keys that exists in `InputType` but not in `ParameterType`.
+- Mark these excess keys as `never`.
+*/
+type ExactObject<ParameterType, InputType> = {[Key in keyof ParameterType]: Exact<ParameterType[Key], ObjectValue<InputType, Key>>}
+	& Record<Exclude<keyof InputType, KeysOfUnion<ParameterType>>, never>;
 
 /**
 Create a type that does not allow extra properties, meaning it only allows properties that are explicitly declared.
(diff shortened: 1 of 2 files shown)


zorji:
@JasonShin could you help review this PR? including grammar issues lol

test-d/exact.ts
```diff
@@ -143,3 +143,69 @@ import type {Exact} from '../index';
...
+			y: '',
+			z: 1,
+		}];
+		// @ts-expect-error
+		fn(input);
+	}
+}
+
+{ // Spec - union of array with nested fields
+	type Type = Array<{x: string}> & Array<{z: number; d: {e: string}}>;
```
JasonShin, line 196: Thanks @zorji !! I was testing this again and found that union of 2 ReadOnlyArray breaks the type. But I think creating a union of 2 read-only arrays is already broken in TS... 

Screen Shot 2022-07-11 at 9 09 11 am

Unless you think this is also important to handle in this PR 👍

test-d/exact.ts
```diff
@@ -143,3 +143,69 @@ import type {Exact} from '../index';
...
+			x: '',
+			z: 1,
+			d: {
+				e: 'test',
+				f: true, // Excess field
+			},
+		}];
+		// @ts-expect-error
+		fn(input);
+	}
```
JasonShin, line 210: Possibly we can include more test coverages to check excess, exact and missing fields for the union array test

source/exact.d.ts
```diff
@@ -41,11 +61,8 @@ onlyAcceptNameImproved(invalidInput); // Compilation error
...
-export type Exact<ParameterType, InputType extends ParameterType> = ParameterType extends Primitive
-	? ParameterType
-	/*
-	Create a type from `ParameterType` and `InputType` and change keys exclusive to `InputType` to `never`.
-	- Generate a list of keys that exists in `InputType` but not in `ParameterType`.
-	- Mark these excess keys as `never`.
-	*/
-	: {[Key in keyof ParameterType]: Exact<ParameterType[Key], InputType[Key]>} & Record<Exclude<keyof InputType, KeysOfUnion<ParameterType>>, never>;
+export type Exact<ParameterType, InputType> =
+	// Converting union of array to array of union. i.e. A[] & B[] => (A & B)[]
```
JasonShin, line 65: 👍 
Screen Shot 2022-07-11 at 9 13 55 am

JasonShin:
Otherwise looks good, I will leave grammar and spelling checks to others 😊

test-d/exact.ts
```diff
@@ -143,3 +143,69 @@ import type {Exact} from '../index';
 		fn(input);
 	}
 }
+
```
JasonShin, line 146: A little bit out of the scope of this PR but can we also include exhaustive tests against ReadonlyArray as well? I just want to ensure ReadonlyArray will work consistently in the future along with the regular mutable arrays.

Screen Shot 2022-07-11 at 9 22 56 am

We can just copy & paste Array spec for the ReadonlyArray test

zorji:
@JasonShin looks like ReadOnly is an easy fix but I am not sure whether this is a breaking change. I'll put them in 2 separate PRs. https://github.com/zorji/type-fest/pull/1/files

zorji:
@JasonShin actually I think I found a way

JasonShin (approved):
Looks good from my end, thanks for the updates!

JasonShin:
@sindresorhus Please review this thanks 👍

source/exact.d.ts
```diff
@@ -1,6 +1,26 @@
-import type {Primitive} from './primitive';
 import type {KeysOfUnion} from './internal';
 
+/**
+Extract the element of an array that also works for array union.
+Return `never` if T is not an array.
+It creates a type-safe way to access the element type of `unknown` type.
+*/
+type ArrayElement<T> = T extends readonly unknown[]? T[0] : never;
```
sindresorhus, line 8: This should be moved to internal.d.ts

zorji:
> Could this use the trick mentioned in https://github.com/sindresorhus/type-fest/issues/118#issuecomment-657259814 ?

@sindresorhus I have tested and found the suggested implementation doesn't work for union of array

PR #426, opened 2022-07-19
Add PartialOnUndefinedDeep type

### LINKED ISSUE #399: Add `PartialOnUndefined` type
What it does is simple: it makes all keys whose type allows `undefined` optional.

This is similar to `ConditionalExcept<T, undefined> & Partial<T>`, except the condition is reversed: instead of `T[keyof T] extends undefined`, this checks `undefined extends T[keyof T]`.

[Sample implementation and use case](https://stackoverflow.com/a/72320731).

Maybe also a deep/recursive variant for consistency with existing types?

clemclx:
Closes #399

### DIFF
--- a/source/partial-on-undefined-deep.d.ts
+++ b/source/partial-on-undefined-deep.d.ts
@@ -17,7 +17,7 @@ export interface PartialOnUndefinedDeepOptions {
 Create a deep version of another type where all keys accepting `undefined` type are set to optional.
 
 This utility type is recursive, transforming at any level deep.
-By default, this type does not affects array and tuple types unless you explicitly pass `{recurseIntoArrays: true}` as the second type argument.
+By default, this type does not affects items type of arrays and tuples unless you explicitly pass `{recurseIntoArrays: true}` as the second type argument.
 
 Use-cases:
 - Make all properties of a type that can be undefined optional to not have to specify keys with undefined value
--- a/source/partial-on-undefined-deep.d.ts
+++ b/source/partial-on-undefined-deep.d.ts
@@ -11,6 +11,7 @@ https://github.com/sindresorhus/type-fest/issues/399
 @example
 ```
 import type {PartialOnUndefinedDeep} from 'type-fest';
+
 interface Settings {
 	optionA: string;
 	optionB: number | undefined;
@@ -19,6 +20,7 @@ interface Settings {
 		subOptionB: boolean | undefined;
 	}
 };
+
 const testSettings: PartialOnUndefinedDeep<Settings> = {
 	optionA: "foo",
 	// 👉 optionB is now optional and can be omited
--- a/source/partial-on-undefined-deep.d.ts
+++ b/source/partial-on-undefined-deep.d.ts
@@ -38,7 +38,7 @@ export type PartialOnUndefinedDeep<T> = T extends Record<any, any> | undefined
 	: T;
 
 /**
- * Utility type to get the value type by key, recursively call PartialOnUndefinedDeep to treat subObjects
+ * Utility type to get the value type by key and recursively call `PartialOnUndefinedDeep` to transform sub-objects.
  */
 type PartialOnUndefinedDeepValue<T> = T extends BuiltIns | ((...arguments: any[]) => unknown)
 	? T
--- a/source/partial-on-undefined-deep.d.ts
+++ b/source/partial-on-undefined-deep.d.ts
@@ -2,9 +2,22 @@ import type {BuiltIns} from './internal';
 import type {Merge} from './merge';
 
 /**
+@see PartialOnUndefinedDeep
+*/
+export interface PartialOnUndefinedDeepOptions {
+	/**
+	Whether to affect the individual elements of arrays and tuples
+
+	@default false
+	*/
+	readonly recurseIntoArrays?: boolean;
+}
+
+/**
 Create a deep version of another type where all keys accepting `undefined` type are set to optional.
 
-This utility type is recursive, transforming at any level deep, and recurses into arrays too.
+This utility type is recursive, transforming at any level deep.
+By default, this type does not affects array and tuple types unless you explicitly pass `{recurseIntoArrays: true}` as the second type argument.
 
 Use-cases:
 - Make all properties of a type that can be undefined optional to not have to specify keys with undefined value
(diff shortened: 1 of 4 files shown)


test-d/partial-on-undefined-deep.ts
```diff
@@ -0,0 +1,46 @@
...
+  map: Map<string, string> | undefined;
+  set: Set<string> | undefined;
+  array: any[] | undefined;
+  tuple: ['test1'] | undefined;
+  readonly: readonly string[] | undefined;
+}>;
+
+expectType<{
+  func?: typeof foo['func'];
+  obj?: typeof foo['obj'];
```
sindresorhus, line 28: Don't use abbreviations. Applies everywhere.
```suggestion
  object?: typeof foo['object'];
```

sindresorhus:
Can you explain why you went with a different implementation than what's in https://stackoverflow.com/a/72320731/64949 ? Just curious how they compare and whether it was an intentional choice.

sindresorhus:
Thanks for contributing. Make sure you read and adhere to: https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md#submitting-a-new-type

sindresorhus:
You need to add the type to the readme.

source/partial-on-undefined-deep.d.ts
```diff
@@ -0,0 +1,47 @@
...
+    subOptionA: boolean;
+    subOptionB: boolean | undefined;
+  }
+};
+const testSettings: PartialOnUndefinedDeep<Settings> = {
+  optionA: "foo",
+  // 👉 optionB is now optional and can be omited
+  subOption: {
+    subOptionA: true,
+    // 👉 subOptionB is now optional aswell and can be omited
```
sindresorhus, line 27: Use tab indentation.

source/partial-on-undefined-deep.d.ts
```diff
@@ -0,0 +1,47 @@
...
+*/
+export type PartialOnUndefinedDeep<T> = T extends Record<any, any> | undefined
+  ? {[KeyType in keyof T as undefined extends T[KeyType] ? KeyType : never]?: PartialOnUndefinedDeepValue<T[KeyType]>} extends infer U // Make a partial type with all value types accepting undefined (and set them optional)
+    ? Merge<{[KeyType in keyof T as KeyType extends keyof U ? never : KeyType]: PartialOnUndefinedDeepValue<T[KeyType]>}, U> // Join all remaining keys not treated in U
+    : never // Should not happen
+  : T;
+
+/**
+ * Utility type to get the value type by key, recursively call PartialOnUndefinedDeep to treat subObjects
+ */
```
sindresorhus, line 42:
```suggestion
/**
Utility type to get the value type by key and recursively call `PartialOnUndefinedDeep` to transform sub-objects.
*/
```

sindresorhus (changes requested):

clemclx:
> Can you explain why you went with a different implementation than what's in https://stackoverflow.com/a/72320731/64949 ? Just curious how they compare and whether it was an intentional choice.

Hey,
I ended with a different implementation for some reasons :

- more comprehensive (subjective)
- avoid the intersection with the "subtype" keeping only required props and a Partial of the whole type (imo, it could lead to some weird behavior with some complex types)
- recursion : the solution on StackOverflow could be Deep too, but the intersection type with the Partial would mean that recursion would happen two times (in the first type keeping only non optional props and in the Partial retreating the whole type)

This way, I finally implemented it from scratch as I would have done

source/partial-on-undefined-deep.d.ts
```diff
@@ -0,0 +1,47 @@
+import type {BuiltIns} from './internal';
+import type {Merge} from './merge';
+
+/**
+Create a type from another type with all keys and nested keys accepting undefined type set to optional.
+
+Use-cases:
+- Make all properties of a type that can be undefined optional to not have to specify keys with undefined value
```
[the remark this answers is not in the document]
clemclx: i'm not sure to have more use-cases that would be pertinent/understandable for users, let me know if something seems pertinent for you :)

source/partial-on-undefined-deep.d.ts
```diff
@@ -0,0 +1,51 @@
+import type {BuiltIns} from './internal';
+import type {Merge} from './merge';
+
+/**
+Create a deep version of another type where all keys accepting `undefined` type are set to optional.
+
+This utility type is recursive, transforming at any level deep, and recurses into arrays too.
```
sindresorhus, line 7: Is it really expected/wanted to recurse into arrays by default? Maybe we should make it opt-in instead? Like a `recurseIntoArrays` option. Similar to https://github.com/sindresorhus/type-fest/commit/37774694ba8c9ce83a998b317ee560bcff08eb2e but should be `false` by default.

clemclx: I see, both solutions seems fine for me, I will update the type to be consistent with `PartialDeep` asap then :blush:

source/partial-on-undefined-deep.d.ts
```diff
@@ -1,10 +1,23 @@
 import type {BuiltIns} from './internal';
 import type {Merge} from './merge';
 
+/**
+@see PartialOnUndefinedDeep
+*/
+export interface PartialOnUndefinedDeepOptions {
```
sindresorhus, line 7: This needs to be reexported in the index file.

source/partial-on-undefined-deep.d.ts
```diff
@@ -0,0 +1,72 @@
+import type {BuiltIns} from './internal';
+import type {Merge} from './merge';
+
+/**
+@see PartialOnUndefinedDeep
+*/
+export interface PartialOnUndefinedDeepOptions {
+	/**
+	Whether to affect the individual elements of arrays and tuples
```
sindresorhus, line 9:
```suggestion
	Whether to affect the individual elements of arrays and tuples.
```

source/partial-on-undefined-deep.d.ts
```diff
@@ -0,0 +1,72 @@
...
+	@default false
+	*/
+	readonly recurseIntoArrays?: boolean;
+}
+
+/**
+Create a deep version of another type where all keys accepting `undefined` type are set to optional.
+
+This utility type is recursive, transforming at any level deep.
+By default, this type does not affects array and tuple types unless you explicitly pass `{recurseIntoArrays: true}` as the second type argument.
```
sindresorhus, line 20: It's confusing to say that it would "affect" arrays, as it's for objects. It's more correct to say that the option makes it affect items of arrays/tuples.

sindresorhus:
Thanks :)

PR #433, opened 2022-08-15
feat(jsonify): do not reduce tuples to arrays

qwelias:
Tuples are kind of nice, would be a shame to reduce them to generic arrays when `Jsonify`ing.
Also refactored a bit, moved instanced 'primitives' under `object` case


qwelias:
@sindresorhus would be fantastic to get some feedback here, we're a bit blocked by tuples being converted to arrays

sindresorhus:
https://github.com/sindresorhus/type-fest/releases/tag/v2.18.1

PR #440, opened 2022-08-19
Add `TupleToUnion` type

mattmahn:
This type makes it easy to define an array of allowed values, and then use that same list in a type union.  All without needing to repeat the allowed values.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -193,6 +193,7 @@ Click the type names for complete docs.
 - [`MultidimensionalArray`](source/multidimensional-array.d.ts) - Create a type that represents a multidimensional array of the given type and dimensions.
 - [`MultidimensionalReadonlyArray`](source/multidimensional-readonly-array.d.ts) - Create a type that represents a multidimensional readonly array of the given type and dimensions.
 - [`ReadonlyTuple`](source/readonly-tuple.d.ts) - Create a type that represents a read-only tuple of the given type and length.
+- [`TupleToUnion`](source/tuple-to-union.d.ts) - Convert a tuple into a union type of its elements.
 
 ### Numeric
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -63,6 +63,7 @@ export {HasOptionalKeys} from './source/has-optional-keys';
 export {RequiredKeysOf} from './source/required-keys-of';
 export {HasRequiredKeys} from './source/has-required-keys';
 export {Spread} from './source/spread';
+export {TupleToUnion} from './source/tuple-to-union';
 
 // Template literal types
 export {CamelCase} from './source/camel-case';
--- a/source/tuple-to-union.d.ts
+++ b/source/tuple-to-union.d.ts
@@ -26,6 +26,26 @@ function verifyRequestBody(body: unknown): body is RequestBody {
 }
 ```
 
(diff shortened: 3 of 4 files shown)


source/array-to-union.d.ts
```diff
@@ -0,0 +1,29 @@
+/**
+Convert a fixed-size array into a union type of its elements.
```
sindresorhus, line 2: "fixed size array", isn't this a tuple? And then the name should probably be `TupleToUnion`.

mattmahn: Yeah, I went back-and-forth on using tuple vs array

readme.md
```diff
@@ -193,6 +193,7 @@ Click the type names for complete docs.
 - [`MultidimensionalArray`](source/multidimensional-array.d.ts) - Create a type that represents a multidimensional array of the given type and dimensions.
 - [`MultidimensionalReadonlyArray`](source/multidimensional-readonly-array.d.ts) - Create a type that represents a multidimensional readonly array of the given type and dimensions.
 - [`ReadonlyTuple`](source/readonly-tuple.d.ts) - Create a type that represents a read-only tuple of the given type and length.
+- [`TupleToUnion`](source/tuple-to-union.d.ts) - Convert a fixed-size array into a union type of its elements.
```
sindresorhus, line 196: Description is outdated

skarab42:
I'm not sure to see the difference from this:

```
const destinations = ['a', 'b', 'c'] as const;

type Destination = typeof destinations[number]; // "a" | "b" | "c"
```

mattmahn:
@skarab42 That's neat; I never thought of doing it that way.  I think it's nice to have a type like `TupleToUnion<T>` to create this type for you.  If you know `typeof destinations[number]` works, you probably wouldn't be looking at type-fest for something like `TupleToUnion<T>` in the first place, so I think having `TupleToUnion<T>` is more useful for those that don't already know TypeScript very well.

It looks like there is only a difference if the input is not a tuple: strings become `string`, and other types get an error like "Type 'Number' has no matching index signature for type 'number'".  I'm not convinced whether `never` or a compile error is better in this case.

sindresorhus:
> If you know typeof destinations[number] works,

Can you mention `typeof destinations[number]` in the docs an how it differs?

sindresorhus:
Thanks :)

jsbrain:
What about UnionToTuple?

```ts
/**
 * Create a type that is the tuple of all the types in the union.
 * Useful for creating an strongly typed tuple array for filtering and sorting.
 *
 * Usage:
 *
 * ```ts
 * const actions = { create: true, read: true, update: true, delete: false};
 * type Action = keyof Omit<typeof actions, "delete">;
 * type AvailableActions = UnionToTuple<Action>;
 * const availableActions: AvailableActions = ["create", "read", "delete"];
 * //       Type '"delete"' is not assignable to type '"update"'. ^^^^^^^
 * ```
 */
export type UnionToTuple<T> = (
  (T extends any ? (t: T) => T : never) extends infer U
    ? (U extends any ? (u: U) => any : never) extends (v: infer V) => any
      ? V
      : never
    : never
) extends (_: any) => infer W
  ? [...UnionToTuple<Exclude<T, W>>, W]
  : [];
```

Probably needs some refinement though ...

sindresorhus:
@jsbrain https://github.com/sindresorhus/type-fest/pull/411

Issue #436, opened 2022-08-16
Breaking change to `Simplify<>` introduced in v2.16.0


andyjy:
The change to `Simplify<>` introduced in https://github.com/sindresorhus/type-fest/pull/414 looks great, but turns out it breaks my codebase vs the prior definition of `Simplify<>`, in two ways:

1. `Type instantiation is excessively deep and possibly infinite` errors in the VSCode typescript checker
2. `RangeError: Maximum call stack size exceeded` fatal error when running `tsc`

I would love to create a Minimum Reproducible Example but unfortunately the types involved span a bunch of files and types (as suggested by the nature of these errors) so that would be quite a challenge. 

I've confirmed 100% this is the cause and created a workaround by patching back to the prior `Simplify<>` definition using [patch-package](https://www.npmjs.com/package/patch-package) and the patch below.

I also appreciate this error isn't caused by `type-fest` alone - only occurring in combination with my own code - so I can see some may not consider this strictly a breaking change. But I figured the fact the breakage only came with the update to `type-fest` seemed worth sharing - and potentially revertible.

In terms of a potential solution - I see that:

1) the change in https://github.com/sindresorhus/type-fest/pull/414 is to improve the behaviour of `Simplify<>` when applied to non-object types, and
2) `Simplify<>` is used "internally" within other `type-fest` types (e.g. `Merge<>`, `SetOptional<>`, `Spread<>`, `Writable<>`) where the input types are expected to be objects and thus the improvement implemented by https://github.com/sindresorhus/type-fest/pull/414 shouldn't be applicable(?)

Would it therefore make sense to revert https://github.com/sindresorhus/type-fest/pull/414 and reimplement using two separate types in a way such that the original "simple" `Simplify<>` definition is used by these internal usages? 

E.g. re-add the original `Simplify<>` implementation under another name and update the internal usages to refer to that (where the new functionality to cope with non-object types is not required), leaving `type-fest` users free to pick from either type as appropriate?

If this sounds good I could provide a PR - I figured to share the issue & check for feedback first.

Thanks!

### `patch-package` patch I have currently used to revert this change:

```diff
diff --git a/node_modules/type-fest/source/simplify.d.ts b/node_modules/type-fest/source/simplify.d.ts
index 7112dc3..cccd870 100644
--- a/node_modules/type-fest/source/simplify.d.ts
+++ b/node_modules/type-fest/source/simplify.d.ts
[...]

PR #441, opened 2022-08-23
Revert `Simplify` to v2.15.1

skarab42:
Revert `Simplify` to v2.15.1 due to breaking change introduced in v2.16.0 #436

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -45,16 +45,16 @@ export {Jsonify} from './source/jsonify';
 export {Schema} from './source/schema';
 export {LiteralToPrimitive} from './source/literal-to-primitive';
 export {
-  PositiveInfinity,
-  NegativeInfinity,
-  Finite,
-  Integer,
-  Float,
-  NegativeFloat,
-  Negative,
-  NonNegative,
-  NegativeInteger,
-  NonNegativeInteger,
+	PositiveInfinity,
+	NegativeInfinity,
+	Finite,
+	Integer,
+	Float,
+	NegativeFloat,
+	Negative,
+	NonNegative,
+	NegativeInteger,
+	NonNegativeInteger,
 } from './source/numeric';
 export {StringKeyOf} from './source/string-key-of';
 export {Exact} from './source/exact';
--- a/index.d.ts
+++ b/index.d.ts
@@ -40,21 +40,21 @@ export {Entry} from './source/entry';
 export {Entries} from './source/entries';
 export {SetReturnType} from './source/set-return-type';
 export {Asyncify} from './source/asyncify';
-export {Simplify, SimplifyOptions} from './source/simplify';
+export {Simplify} from './source/simplify';
 export {Jsonify} from './source/jsonify';
 export {Schema} from './source/schema';
 export {LiteralToPrimitive} from './source/literal-to-primitive';
 export {
-	PositiveInfinity,
-	NegativeInfinity,
-	Finite,
-	Integer,
-	Float,
-	NegativeFloat,
-	Negative,
-	NonNegative,
-	NegativeInteger,
-	NonNegativeInteger,
+  PositiveInfinity,
+  NegativeInfinity,
+  Finite,
+  Integer,
+  Float,
+  NegativeFloat,
+  Negative,
+  NonNegative,
+  NegativeInteger,
+  NonNegativeInteger,
 } from './source/numeric';
 export {StringKeyOf} from './source/string-key-of';
 export {Exact} from './source/exact';
--- a/test-d/simplify.ts
+++ b/test-d/simplify.ts
@@ -42,3 +42,26 @@ expectType<SomeInterfaceAsTypeWrittenByHand>(valueAsInterface);
 expectAssignable<Record<string, unknown>>(valueAsLiteral);
 expectAssignable<Record<string, unknown>>(valueAsSimplifiedInterface);
 expectNotAssignable<Record<string, unknown>>(valueAsInterface); // Index signature is missing in interface
+
+// The following tests should be fixed once we have determined the cause of the bug reported in https://github.com/sindresorhus/type-fest/issues/436
+
+// // Should return the original type if it is not simplifiable, like a function.
+// type SomeFunction = (type: string) => string;
+// expectType<Simplify<SomeFunction>>((type: string) => type);
+
+// class SomeClass {
+// 	id: string;
+
+// 	private readonly code: number;
+
+// 	constructor() {
+// 		this.id = 'some-class';
+// 		this.code = 42;
+// 	}
+
+// 	someMethod() {
+// 		return this.code;
+// 	}
+// }
+
+// expectType<Simplify<SomeClass>>(new SomeClass());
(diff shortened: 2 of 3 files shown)


skarab42:
@sindresorhus once merged I will add `SimplifyDeep` and the conditional version.

index.d.ts
```diff
@@ -40,21 +40,21 @@ export {Entry} from './source/entry';
...
+  PositiveInfinity,
+  NegativeInfinity,
+  Finite,
+  Integer,
+  Float,
+  NegativeFloat,
+  Negative,
+  NonNegative,
+  NegativeInteger,
+  NonNegativeInteger,
```
sindresorhus, line 57: Don't make unrelated changes.

skarab42: Sorry, I'm having trouble getting xo to work properly with my setup.

skarab42: @sindresorhus Done, but should `xo --fix` replace spaces with tabs? I have to change my VSCode config every time I work on type-fest (and I often forget). I have installed and configured the XO plugin and deactivated eslint and prettier. What am I doing wrong?

sindresorhus: > but should xo --fix replace spaces with tabs?

Yes, this project uses tab indentation.

test-d/simplify.ts
```diff
@@ -42,45 +42,3 @@ expectType<SomeInterfaceAsTypeWrittenByHand>(valueAsInterface);
...
-		this.id = 'some-class';
-		this.code = 42;
-	}
-
-	someMethod() {
-		return this.code;
-	}
-}
-
-expectType<Simplify<SomeClass>>(new SomeClass());
```
sindresorhus, line 65: Let's keep this, but commented out, and with a TODO comment to fix it in the future and with a reference to the issue that caused it to be commented out.

index.d.ts
skarab42: > Yes, this project uses tab indentation.

I known, but when I run `xo --fix` in the project it doesn't do the conversion...

sindresorhus:
@skarab42 I noticed we didn't add back the deep ability as `SimplifyDeep`, was that intentional? There's currently no way for users to get back `Simplify<..., {deep: true>`.

skarab42:
@sindresorhus Yes, because obviously the bug occurs as soon as any [condition is introduced](https://github.com/sindresorhus/type-fest/issues/436#issuecomment-1219699899), even a simple `true extends true`, but as I can't reproduce the case I'll have to check with @andyjy. 

I have made an [TS Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAyglgWzAGzgMxAHgCoD4oC8UA3gNoDSEI240cAdlANZUD2aU2AugFycVUakLgF8A3AFgAUNID0szgAs4AZyiqo9VsCgB3VgCcmDAObTQkWIhToQAEQgQwOfETKVqtdYxYh2nXiskVAwHJxwBT2FccWk5BWxFaH0jAEMDVgBXegATdTVgVm9gDJzMgGNoVPoQKHLWXLhgOAbzL3hg2zCwAHVDJnSs3JdCKBLM6AgAD2AIXIKDCagAfhJIoTofNg5uPg6bUMde-sHsnIiPDa4YqD56CAA3CANJGSl5KABab7ipC2hugBGUb7EJYABKEHqBnOKhKpgANLBWAgIAA5Vg5CC4XBiKAfPpGNqWboAJhB1jB3UwkOhsPh9BMSJgKPRmOxuPxCgAolNIOVZnlCnp+mMklAVKlUVBUmoAEYQZCsXQAOjVxIBRwAzBTOocnISBhkzjSoYZ6QZEcjURisTi8QSTsbcnomoo6g0ck0WvRfh9vp9fv8oAAFVgqb0NEMZMBqNxjVhgO6ZBAKl5QZAQNDAZOp55QWJ-dpwABeEGjibjJFdOWAilzabxSTgJkUOc0KbTBdewbDEeaDQAgrl4GXRn3I-QK7GoAAyKxl6cqV7SBizAxoVKVa1srEkaRQKBgdJzdsTgf0Yc5UcQV6H8rKZA5PjEA+Ho8n+hn8OTq83u-dtIIhAA) that illustrates a part of the problem in case you want to report it to the TS team.

andyjy:
> @sindresorhus Yes, because obviously the bug occurs as soon as any [condition is introduced](https://github.com/sindresorhus/type-fest/issues/436#issuecomment-1219699899), even a simple `true extends true`, but as I can't reproduce the case I'll have to check with @andyjy.

It seems reasonable to me to introduce `SimplifyDeep<>` as a separate user-accessible type.  Use of the conditional only caused an error for (my) particularly complex types, I would not expect it to be an issue for simpler cases (which I'd expect to be the majority).

A new user-accessible `SimplifyDeep<>` would enable people to opt-in to utilising it when they specifically wish to - it seems reasonable to expect not to apply it to particularly complex types where it would break. If people do run into issues applying it to overly complex types it would be immediately apparent, and not a breaking change.

(The issue was introducing it within the internals of other types where it was not required and caused existing types to stop working when applied to input types that worked previously.)

andyjy:
> I have made an [TS Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAyglgWzAGzgMxAHgCoD4oC8UA3gNoDSEI240cAdlANZUD2aU2AugFycVUakLgF8A3AFgAUNID0szgAs4AZyiqo9VsCgB3VgCcmDAObTQkWIhToQAEQgQwOfETKVqtdYxYh2nXiskVAwHJxwBT2FccWk5BWxFaH0jAEMDVgBXegATdTVgVm9gDJzMgGNoVPoQKHLWXLhgOAbzL3hg2zCwAHVDJnSs3JdCKBLM6AgAD2AIXIKDCagAfhJIoTofNg5uPg6bUMde-sHsnIiPDa4YqD56CAA3CANJGSl5KABab7ipC2hugBGUb7EJYABKEHqBnOKhKpgANLBWAgIAA5Vg5CC4XBiKAfPpGNqWboAJhB1jB3UwkOhsPh9BMSJgKPRmOxuPxCgAolNIOVZnlCnp+mMklAVKlUVBUmoAEYQZCsXQAOjVxIBRwAzBTOocnISBhkzjSoYZ6QZEcjURisTi8QSTsbcnomoo6g0ck0WvRfh9vp9fv8oAAFVgqb0NEMZMBqNxjVhgO6ZBAKl5QZAQNDAZOp55QWJ-dpwABeEGjibjJFdOWAilzabxSTgJkUOc0KbTBdewbDEeaDQAgrl4GXRn3I-QK7GoAAyKxl6cqV7SBizAxoVKVa1srEkaRQKBgdJzdsTgf0Yc5UcQV6H8rKZA5PjEA+Ho8n+hn8OTq83u-dtIIhAA) that illustrates a part of the problem in case you want to report it to the TS team.

I saw someone posted the following explanation in [this StackOverflow answer](https://stackoverflow.com/a/57683652) - it seems to be TypeScript working as designed.

> The conditional type inference is conceptually a no-op, but it's used to [distribute union types](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types) and to force the compiler to evaluate the "true" branch of the conditional (if you redefine Expand<T> without it, sometimes the compiler will just output the mapped type {[K in keyof RelevantType]: RelevantType[K]}, which is not what you want to see).

Issue #395, opened 2022-05-07
Add `EmptyObject` (the type you'd expect for `{}`)


fregante:
https://stackoverflow.com/a/60112239/288906

```ts
type EmptyObject = Record<string, never>;

const one: EmptyObject = {}; // yes ok
const two: EmptyObject = {a: 1}; // error
```

sindresorhus:
Makes sense for discoverability 👍

HunterKohler:
I like this idea. However there may some issues due to unexpected behavior in niche instances. I have just one of those and an adjustment.

First, this would be more complete:

```ts
type EmptyObject = Record<keyof any, never>;
```

Second, I saw this example from someone on stack overflow, which seems to violate the type safety of this.

```ts
type EmptyObject = Record<keyof any, never>;

type Union = EmptyObject | { id: string }; 

const a: Union = {}; 
const b: string = a.id // no error
const c = a.id; // inferred type: `never`;
```

[A Typescript playground for the above](https://www.typescriptlang.org/play?#code/C4TwDgpgBAogtmUB5ARgKwgY2FAvFAJSwHsAnAEwB4BrCEYgMygEMA7EAGilYgDcJSAPgDcAWABQE0JCgBVVgEtirPLATJ0WHAB8oAbygLyALigBnYKQWsA5lAC+wqBImZlFlqflKV+PY+dxN1YPFFMLK1tVZgA6IygAegTuYigBUjJXdxxMaLjyJyTDVgZ0iHIoaQhTAAMeflIa4SA).

fregante:
Try `Record<never, never>`

https://www.typescriptlang.org/play?#code/C4TwDgpgBAogtmUB5ARgKwgY2FAvFAJSwHsAnAEwB4A7CANwlIBopaHSA+AbgFgAofqEhQAqtQCWxanlgJk6LDgA+UAN5Rx5AFxQAzsFLjqAcygBfLlH79MU-VACGOsZOn5VFq31vV7KHfqGJjIOAHSaUAD0kazEUIykZDZ2OJgh4eSW0RrUAGYJEORQQhA6AAZsjGVcQA

bbrk24:
@fregante `Record<never, never>` doesn't solve the original problem:

```ts
type EmptyObject = Record<never, never>;

const foo: EmptyObject = {a: 1}; // should error, but is allowed
```

In fact, the tooltip suggests `Record<never, never>` is just `{}`.

[...]

PR #447, opened 2022-08-26
Add `EmptyObject` and `IsEmptyObject` types

skarab42:
Related to #395 

// @fregante @bbrk24

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -125,7 +125,7 @@ Click the type names for complete docs.
 
 ### Utilities
 
-- [`EmptyObject`](source/empty-object.d.ts) - Represents a strictly empty plain object. In other words, this type will only accept the `{}` value.
+- [`EmptyObject`](source/empty-object.d.ts) - Represents a strictly empty plain object, the `{}` value.
 - [`IsEmptyObject`](source/empty-object.d.ts) - Returns a `boolean` for whether the type is strictly equal to an empty plain object. In other words, this utility will only match the `{}` value.
 - [`Except`](source/except.d.ts) - Create a type from an object type without certain keys. This is a stricter version of [`Omit`](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys).
 - [`Writable`](source/writable.d.ts) - Create a type that strips `readonly` from all or some of an object's keys. The inverse of `Readonly<T>`. Formerly named `Mutable`.
--- a/readme.md
+++ b/readme.md
@@ -126,7 +126,7 @@ Click the type names for complete docs.
 ### Utilities
 
 - [`EmptyObject`](source/empty-object.d.ts) - Represents a strictly empty plain object, the `{}` value.
-- [`IsEmptyObject`](source/empty-object.d.ts) - Returns a `boolean` for whether the type is strictly equal to an empty plain object. In other words, this utility will only match the `{}` value.
+- [`IsEmptyObject`](source/empty-object.d.ts) - Returns a `boolean` for whether the type is strictly equal to an empty plain object, the `{}` value.
 - [`Except`](source/except.d.ts) - Create a type from an object type without certain keys. This is a stricter version of [`Omit`](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys).
 - [`Writable`](source/writable.d.ts) - Create a type that strips `readonly` from all or some of an object's keys. The inverse of `Readonly<T>`. Formerly named `Mutable`.
 - [`Merge`](source/merge.d.ts) - Merge two types into a new type. Keys of the second type overrides keys of the first type.
--- a/readme.md
+++ b/readme.md
@@ -125,7 +125,8 @@ Click the type names for complete docs.
 
 ### Utilities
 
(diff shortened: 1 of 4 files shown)


readme.md
```diff
@@ -150,6 +151,7 @@ Click the type names for complete docs.
 - [`UnionToIntersection`](source/union-to-intersection.d.ts) - Convert a union type to an intersection type.
 - [`LiteralToPrimitive`](source/literal-to-primitive.d.ts) - Convert a [literal type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types) to the [primitive type](source/primitive.d.ts) it belongs to.
 - [`Stringified`](source/stringified.d.ts) - Create a type with the keys of the given type changed to `string` type.
+- [`IsEmptyObject`](source/empty-object.d.ts) - Returns a `boolean` whether the type is strictly equal to `{}`.
```
sindresorhus, line 154: Just saying `{}` is ambiguous because it's both a JS literal and TS type.

https://stackoverflow.com/a/52193484/64949

readme.md
```diff
@@ -125,6 +125,7 @@ Click the type names for complete docs.
 
 ### Utilities
 
+- [`EmptyObject`](source/empty-object.d.ts) - Represents a strictly empty object. In other words, this type accepts `{}` only.
```
sindresorhus, line 128:
```suggestion
- [`EmptyObject`](source/empty-object.d.ts) - Represents a strictly empty object `{}`.
```

test-d/empty-object.ts
```diff
@@ -0,0 +1,20 @@
+import {expectAssignable, expectError, expectType} from 'tsd';
+import type {EmptyObject, IsEmptyObject} from '../index';
+
+declare let foo: EmptyObject;
+
+expectAssignable<{}>(foo);
+expectAssignable<{}>(foo = {});
+
+expectError(foo = []);
```
sindresorhus, line 9:
```suggestion
expectError(foo = []);
expectError(foo = {x: 1});
```

sindresorhus: And can you add a test for the example in https://github.com/sindresorhus/type-fest/issues/395#issuecomment-1120609111?

source/empty-object.d.ts
```diff
@@ -0,0 +1,39 @@
+declare const emptyObjectSymbol: unique symbol;
+
+/**
+Represents a strictly empty object. In other words, this type accepts `{}` only.
+
```
sindresorhus, line 5: Can you elaborate shortly on why `let foo = {};` is not good enough?

source/empty-object.d.ts
```diff
@@ -0,0 +1,39 @@
...
+let foo: EmptyObject = {};
+
+foo = {}; // Pass
+
+foo = []; // Error
+foo = 42; // Error
+foo = null; // Error
+foo.bar = {}; // Error
+```
+
```
sindresorhus, line 19: Maybe we can also mention at the end here that `Record<string, never>`, `Record<keyof any, never>`, and `Record<never, never>` do not work and link to https://github.com/sindresorhus/type-fest/issues/395 I'm sure some users will ask why not use one of these solutions instead.

sindresorhus (changes requested):

test-d/empty-object.ts
skarab42: > And can you add a test for the example in [#395 (comment)](https://github.com/sindresorhus/type-fest/issues/395#issuecomment-1120609111)?

Isn't that the one? https://github.com/sindresorhus/type-fest/blob/ad34426f02a19845340a598cc77a80028cd0f327/test-d/empty-object.ts#L13

sindresorhus: Are we looking at the same example? I don't see how that is the same as one using a union:

```ts
type Union = EmptyObject | { id: string }; 

const a: Union = {}; 
const b: string = a.id // no error
const c = a.id; // inferred type: `never`;
```

readme.md
```diff
@@ -125,6 +125,8 @@ Click the type names for complete docs.
 
 ### Utilities
 
+- [`EmptyObject`](source/empty-object.d.ts) - Represents a strictly empty plain object. In other words, this type will only accept the `{}` value.
```
sindresorhus, line 128: This feels a bit too verbose.
```suggestion
- [`EmptyObject`](source/empty-object.d.ts) - Represents a strictly empty plain object, the `{}` value.
```

readme.md
```diff
@@ -125,6 +125,8 @@ Click the type names for complete docs.
...
+- [`IsEmptyObject`](source/empty-object.d.ts) - Returns a `boolean` whether the type is strictly equal to an empty plain object. In other words, this utility will only match the `{}` value.
```
sindresorhus, line 129:
```suggestion
- [`IsEmptyObject`](source/empty-object.d.ts) - Returns a `boolean` for whether the type is strictly equal to an empty plain object. In other words, this utility will only match the `{}` value.
```

test-d/empty-object.ts
skarab42: Good point ;)

PR #443, opened 2022-08-25
Add `MergeDeep` type

skarab42:
This is a complete refactoring of #408. @sindresorhus 

The new type supports almost all conventional merging strategies.

```ts
import type {MergeDeep} from 'type-fest';

type Foo = {
	life: number;
	items: string[];
	a: {b: string; c: boolean};
};

interface Bar {
	name: string;
	items: number[];
	a: {b: number};
}

type FooBar = MergeDeep<Foo, Bar>;
// FooBar = {
// 	life: number;
// 	name: string;
// 	items: (string | number)[];
// 	a: {b: number; c: boolean};
// };

type FooBarUnion = MergeDeep<Foo, Bar, {recordMergeMode: 'union'}>;
// FooBarUnion = {
// 	name: string;
// 	life: number;
// 	items: string[] | number[];
// 	a: {b: string; c: boolean} | {b: number};
// };

type FooBarReplace = MergeDeep<Foo, Bar, {recordMergeMode: 'replace'}>;
// FooBarReplace = {
// 	name: string;
// 	life: number;
// 	items: number[];
// 	a: {b: number};
// };

type FooBarMergeUnion = MergeDeep<Foo, Bar, {recordMergeMode: 'merge-or-union'}>;
// FooBarMergeUnion = {
// 	name: string;
// 	life: number;
// 	items: (string | number)[];
// 	a: {b: string | number; c: boolean};
// };

type FooBarMergeReplace = MergeDeep<Foo, Bar, {recordMergeMode: 'merge-or-replace'}>;
// FooBarMergeReplace = {
// 	name: string;
// 	life: number;
// 	items: (string | number)[];
// 	a: {b: number; c: boolean};
// };

// Merge two arrays
type ArrayUnion = MergeDeep<string[], number[], {arrayMergeMode: 'union'}>; // => string[] | number[]
type ArraySpread = MergeDeep<string[], number[], {arrayMergeMode: 'spread'}>; // => (string | number)[]
type ArrayReplace = MergeDeep<string[], number[], {arrayMergeMode: 'replace'}>; // => number[]
type ArrayMergeUnion = MergeDeep<string[], number[], {arrayMergeMode: 'merge-or-union'}>; // => string[] | number[]
type ArrayMergeSpread = MergeDeep<string[], number[], {arrayMergeMode: 'merge-or-spread'}>; // => (string | number)[]
type ArrayMergeReplace = MergeDeep<string[], number[], {arrayMergeMode: 'merge-or-replace'}>; // => number[]

[...]

PR #452, opened 2022-08-30
Add `MergeDeep` type

skarab42:
@sindresorhus Sorry for the extra work, I'm a bit embarrassed :( I need to review the basics of git and especially conflict resolution. Fixes #443

### DIFF
--- a/source/merge-deep.d.ts
+++ b/source/merge-deep.d.ts
@@ -77,11 +77,7 @@ type MergeDeepArrayOrTupleElements<
 	? Destination
 	: Destination extends []
 		? Source
-		: IsMergeable<Destination, Source> extends true
-			? MergeDeepOrReturn<never, Destination, Source, Options>
-			: Options['arrayMergeMode'] extends 'union'
-				? Destination | Source
-				: Source; // 'replace'
+		: MergeDeepOrReturn<Source, Destination, Source, Options>;
 
 /**
 Returns the first tuple element type or the array type.
--- a/source/merge-deep.d.ts
+++ b/source/merge-deep.d.ts
@@ -304,6 +304,8 @@ export type MergeDeepOptions = {
 
 	See {@link MergeDeep} for usages and examples.
 
+	Note: Top-level arrays and tuples are always spread.
+
 	@default 'spread'
 	*/
 	arrayMergeMode?: ArrayMergeMode;
--- a/source/merge-deep.d.ts
+++ b/source/merge-deep.d.ts
@@ -200,13 +200,20 @@ Merge mode for array/tuple elements.
 type ArrayMergeMode = 'spread' | 'replace';
 
 /**
+Test if it sould spread top-level arrays.
+*/
+type ShouldSpread<Options extends MergeDeepOptions> = Options['spreadTopLevelArrays'] extends false
+	? Options['arrayMergeMode'] extends 'spread' ? true : false
+	: true;
+
+/**
 Merge two arrays/tuples according to the chosen {@link MergeDeepOptions.arrayMergeMode arrayMergeMode} option.
 */
 type DoMergeArrayOrTuple<
 	Destination extends UnknownArrayOrTuple,
 	Source extends UnknownArrayOrTuple,
 	Options extends MergeDeepOptions,
-> = Options['arrayMergeMode'] extends 'spread'
+> = ShouldSpread<Options> extends true
 	? Array<Exclude<Destination, undefined>[number] | Exclude<Source, undefined>[number]>
 	: Source; // 'replace'
 
@@ -278,7 +285,7 @@ type MergeDeepOrReturn<
 			: DefaultType
 		: Destination extends UnknownArrayOrTuple
 			? Source extends UnknownArrayOrTuple
-				? MergeDeepArrayOrTuple<Destination, Source, Options>
+				? MergeDeepArrayOrTuple<Destination, Source, Merge<Options, {spreadTopLevelArrays: false}>>
 				: DefaultType
 			: DefaultType>;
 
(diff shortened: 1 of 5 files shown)


sindresorhus:
There are some unresolved comments in the previous pull request:

- https://github.com/sindresorhus/type-fest/pull/443#discussion_r957058900
- https://github.com/sindresorhus/type-fest/pull/443#discussion_r957059558

sindresorhus:
Don't forget to add `Fixes` to the pull request description.

skarab42:
As mentioned here https://github.com/sindresorhus/type-fest/pull/443#discussion_r959225107 I noticed that all the options prefixed with `merge-or-` are the options that allow you to go deep, so maybe they should be called `deep-replace`, `deep-spread`, etc...

sindresorhus:
> I have to do the same thing in the deep version.

Let me know when this is done.

skarab42:
> Let me know when this is done.

@sindresorhus  Done, but I think since the test run on `type-fest@2.19.0` it fail because it does not incorporate the latest changes on which this type depends aka `OmitIndexSignature` and `PickIndexSignature`.

(P.S: sorry I didn't mean to edit your comment)

skarab42:
@sindresorhus `Merge` is broken :( I do a fix!

skarab42:
@sindresorhus  Well, by introducing more and more fixes and options into this type, it becomes too complicated and will quickly become unmaintainable in the long run. I'm thinking of going back to a simpler and more optimised version that doesn't try to cover every edge cases.

sindresorhus:
I agree

skarab42:
@sindresorhus  In blood and sweat I finally got something I like. Unfortunately I can't make it any simpler without sacrificing features that I consider essential:

- Preservation and merging of index signatures.
- Preservation and normalisation of optional properties.
- Merging of variable length tuples.
- Recursion into arrays/tuples.

sindresorhus:
This is looking much better. Thanks for working so hard on it.

What's the reasoning for choosing `spread` as default `arrayMergeMode`? Isn't `replace` more expected behavior as that's how it behaves in native JS when you merge (using object.assign or object spread).

I struggle to see when I would use the `union` `arrayMergeMode`. Do you have any real-world use-cases in mind?

skarab42:
> Isn't `replace` more expected behavior as that's how it behaves in native JS when you merge (using object.assign or object spread).

This is indeed the case when you merge two objects, but when you merge two arrays it's the opposite. So I wonder if we shouldn't export a `MergeDeepArray` type with `spread` by default and `MergeDeepObject` with `replace` by default.

> I struggle to see when I would use the `union` `arrayMergeMode`.

```ts
const initialState = {data: ['a', 'b', 'c']};

// Do the merge
let nextState = {...initialState, data: [1, 2, 3]};

// Error when we try to revert back to inital state, data is typed as number[]
nextState = initialState;
```

sindresorhus:
> This is indeed the case when you merge two objects, but when you merge two arrays it's the opposite. So I wonder if we shouldn't export a MergeDeepArray type with spread by default and MergeDeepObject with replace by default.

Yeah, the important distinction is whether the array is top-level or deep in an object. When it's top-level, it makes to spread, when it's in an object, it makes sense to replace by default.

> In fact I'm not really convinced of the usefulness of union the above example is not realistic as we're going to type the state correctly from the beginning.

Maybe drop it for now and see if there's any requests for it?

skarab42:
@sindresorhus I hope this time it's all right.

skarab42:
> I also don't see when you would want `arrayMergeMode: 'replace'` for top-level arrays. 

Probably when used in another type.

> So maybe it's easier to make it so that top-level arrays are always spread and document that `arrayMergeMode` only applies to nested arrays?

@sindresorhus  Yep right. I added the `spreadTopLevelArrays` option with a default value of `true` which is the easiest solution I found without having to create a new branch and duplicate some code.

sindresorhus:
> Probably when used in another type.

Not sure I understand. Do you have a use-case in mind?

> I added the spreadTopLevelArrays option with a default value of true which is the easiest solution I found without having to create a new branch and duplicate some code.

My question is, do we even need an option for it?

---

Sorry this is taking long. I'm just trying hard to keep the type as simple as possible and ensure we don't add stuff people don't actually need.

skarab42:
> My question is, do we even need an option for it?

We don't need to make it public, the vast majority of users won't need it (and for the moment I can't find any specific use cases). I can make it static but that won't simplify the type itself, only it signature.

skarab42:
@sindresorhus Alternatively, we could also mark the option as `@internal`?

sindresorhus:
> I can make it static but that won't simplify the type itself, only it signature.

👍

skarab42:
@sindresorhus Done ;)

source/merge-deep.d.ts
```diff
@@ -0,0 +1,469 @@
...
+			: never
+>;
+
+/**
+Merge two objects or two arrays/tuples recursively into a new type.
+
+- Properties that only exist in one object are copied into the new object.
+- Properties that exist in both objects are merged if possible or replaced by the one of the source if not.
+- Top-level arrays and tuples are always spread.
+- By default, inner arrays and tuples are replaced. See {@link MergeDeepOptions.arrayMergeMode arrayMergeMode} option to change this behaviour.
```
sindresorhus, line 364: ^ Here it says `replaced`. But here it says `spread`: https://github.com/sindresorhus/type-fest/pull/452/files#diff-5ce63d84bd0497ef6dd72ef89147e7bf80f464a1185b1a34139c6c1cb6c37ee5R307

sindresorhus:
Awesome work! I'm very happy with what we ended up with ❤️

PR #455, opened 2022-09-01
Add index signatures support on `Merge` and improve performances

### LINKED ISSUE #304: Merge type increases type instantiations count drastically
I ran into `error TS2589: Type instantiation is excessively deep and possibly infinite.` and get this result with `--extendedDiagnostics`:

```
Files:                        1527
Lines of Library:            28535
Lines of Definitions:        76407
Lines of TypeScript:         17557
Lines of JavaScript:        195957
Lines of JSON:                   0
Lines of Other:                  0
Nodes of Library:           117553
Nodes of Definitions:       273566
Nodes of TypeScript:         72611
Nodes of JavaScript:        825495
Nodes of JSON:                   0
Nodes of Other:                  0
Identifiers:                431173
Symbols:                    410281
Types:                       90155
Instantiations:            6707979            <========== the instantiations count exceeds the limitation
Memory used:               531849K
Assignability cache size:    53667
Identity cache size:          1231
Subtype cache size:           1904
Strict subtype cache size:    8066
```
tsc has a [5 million limitation on instantiation counts](https://github.com/microsoft/TypeScript/pull/32079#discussion_r297478404)

After I replace all the `Merge` with simple intersection with `Omit`, I get this:

```
Files:                        1527
Lines of Library:            28535
Lines of Definitions:        76407
Lines of TypeScript:         17522
Lines of JavaScript:        195957
Lines of JSON:                   0
Lines of Other:                  0
Nodes of Library:           117553
Nodes of Definitions:       273566
Nodes of TypeScript:         72570
Nodes of JavaScript:        825495
Nodes of JSON:                   0
Nodes of Other:                  0
Identifiers:                431152
Symbols:                    448585
Types:                       80108
Instantiations:             514523         <========== 10x smaller
Memory used:               556046K
Assignability cache size:    77690
Identity cache size:          1259
Subtype cache size:           2096
Strict subtype cache size:    8430
```

Using type-fest@2.3.4

### LINKED ISSUE #382: Merge doesn't is not that useful with indexed types
When using `Merge` with an indexed type on either left or right hand all the well-known properties are lost. This is because `keyof` of an indexed type always returns 'string | number', because every key is possible in that object.

Example:
``` typescript
type Example = {
  foo: string,
  bar: symbol,
  [key: string]: unknown,
}

type Foo = {
  bar: Date,
}

/*
  We would expect something like:
  {
    foo: string,
    bar: Date,
    [key: string]: unknown,
  }
  
  but it actually returns:
  {
    [key: string]: unknown,
    [key: number]: unknown,
  }
*/
type result = Merge<Example, Foo>
```

This is a well-known caveat with Typescript that was discussed in a few issues (https://github.com/microsoft/TypeScript/issues/45367, https://github.com/microsoft/TypeScript/issues/31153), the behavior is actually intended.

And `Omit` indeed loses the well-known keys in the same way:
``` typescript
/*
  Returns:
  {
    [key: string]: unknown,
    [key: number]: unknown,
  }
*/
type omit = Omit<Example, 'foo'>

/*
  Our current implementation of Except on the other hand works fine with indexed types
  This returns:
  {
    bar: Date,
    [key: string]: unknown,
  }
*/
type except = Except<Example, 'foo'>

```

What I propose is to either change `Merge` to support indexed types and return what I'm expecting on the example (similar to except), which could be a breaking change for people expecting to lose typing on these cases (but I doubt is a useful behavior), **or** we can create a new `MergeWellKnownKeys` that treats indexed types. What do you guys think?

I can work on a PR as soon as we decide on an implementation. It should be easy because we already have an internal type doing the same thing on our company's project

skarab42:
This PR adds support for index signatues requested in #382 and by the way it improves performance by reducing the number of instantiations by ~4 times #304.

Fix #304 
Fix #382

### DIFF
--- a/source/merge.d.ts
+++ b/source/merge.d.ts
@@ -1,7 +1,6 @@
-import type {Except} from './except';
+import type {OmitIndexSignature} from './omit-index-signature';
+import type {PickIndexSignature} from './pick-index-signature';
 import type {Simplify} from './simplify';
-
-type Merge_<FirstType, SecondType> = Except<FirstType, Extract<keyof FirstType, keyof SecondType>> & SecondType;
 
 /**
 Merge two types into a new type. Keys of the second type overrides keys of the first type.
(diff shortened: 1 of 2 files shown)


skarab42:
@sindresorhus I have to do the same thing in the deep version.

skarab42:
// @chbdetta @augustobmoura

sindresorhus:
Very nice! 👍

augustobmoura:
Thanks, @skarab42! :+1:

PR #465, opened 2022-09-15
Fix `PackageJson` and `JsonObject` types

### LINKED ISSUE #79: `PackageJson` is not a valid `JsonObject`

### LINKED ISSUE #272: Make `JsonObject` work as expected with TS 4.4 `--exactOptionalPropertyTypes`
#269 brought to my attention that #65 changed the `JsonObject` so that all keys are marked as optional.

Before TypeScript 4.4 a `{ foo?: string}` and a `{ foo?: string|undefined }` was always treated as equal, so #65 made sense then.

But with TS 4.4 the new [`--exactOptionalPropertyTypes`](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#exact-optional-property-types) config was introduced, which (correctly )no longer treats `{ foo?: string}` as equal to `{ foo?: string|undefined }`.

For some reason though, the change introduced in #65 still gets treated in the old way: https://github.com/sindresorhus/type-fest/blob/6b6d81ce894517268391db63b3d6538e971c8126/source/basic.d.ts#L22

I would propose reverting #65 and moving back to:

```typescript
export type JsonObject = {[key: string]: JsonValue};
```

I think it behaves more correctly, as can be seen [in this playground](https://www.typescriptlang.org/play?exactOptionalPropertyTypes=true#code/PQKhCgFkEMBcGMAWBTAzgAmugUgZQPIBy6A9gEYBWy8sAdOACqICWGsAngA7LrzQB26MjwCuqZADMRAG3SwS6ZPwkkATvB6oSAWx7N+nEbDkLhOAoQC08HZzjMy0nmswYsqEd1WWO3E0J5kAA9YJQATZDD0CVUdWnQAERJ+AHJjMR5YFjc3dDDmVWpjQtgRVUFfHmg2FHQM1XQAdxIZKMRoADdMhTCWx2RrauNmWAAudAADCi1+fEoi1zr+AGt+EkbBavQAYTF5bQAlNE5k8Qn4gEl+VFDoMIAadHYW3hbpKODQ-ijnsp29nRHVAna6ZLg8GI6dAjfxKDyFJ4tBqVUj8aTsOriDB4IjWWz2fpycGocYTfShVQSaAaf43QHHU6BELhbEzOZUGjoADe6EAZAToAC+53AAAE+KEAOZqDEAIWqzHg4BAwHAwROqmMKOwbPmnIAvNyANoAaWQGP06Buqn0EoAugB+cba5IANWg0hEyAFAG5VUF1ZrwegAILSCn8ODIZ2zXXGA1cw3LM3jK0221OmZuj1e33gUAQGAIFC5HHEaCqVTQdj0MWRqWqWXyxXKv0Bol+aPBitV9AG6NZz2G225-NQOBINCYfgYjru5hRUvoTjW7QjZhddCz7M18XIeuN1AKpUqtVqQMdmYABRXa43BtT-Al6AAPuh+CJtMIGq+yCQSE4BBfN8ZGkEcwDHItJwEGc5wXCxN3dT0a3EHgJmjZgJHYCZoQkRERDfZBIn8WBK2uFRVG0TB226VFMnaYYckPCUI0JeRJn7RDkGFWtJWldA5UPZsT39M9qJwTNON7cTkmvZhV1gdceFfaN2QWZSZi7St2FzGxrmMaAM2SVT9W5cB0HMoQRAlEklgiCR9EicABXAcBdJuIRxlDcNIxU2MpK5MyLLIKybJEb5JAcsInJctz-2QWhpBICUAApoEeMgAEpvSAA).

skarab42:
This PR fix #272 and fix #79

### DIFF
--- a/source/basic.d.ts
+++ b/source/basic.d.ts
@@ -19,7 +19,7 @@ This type can be useful to enforce some input to be JSON-compatible or as a supe
 
 @category JSON
 */
-export type JsonObject = {[Key in string]?: JsonValue};
+export type JsonObject = {[Key in string]: JsonValue} & {[Key in string]?: JsonValue | undefined};
 
 /**
 Matches a JSON array.
--- a/source/package-json.d.ts
+++ b/source/package-json.d.ts
@@ -1,4 +1,5 @@
 import type {LiteralUnion} from './literal-union';
+import type {JsonObject, JsonValue} from './basic';
 
 declare namespace PackageJson {
 	/**
@@ -27,7 +28,7 @@ declare namespace PackageJson {
 		};
 
 	export type DirectoryLocations = {
-		[directoryType: string]: unknown;
+		[directoryType: string]: JsonValue | undefined;
 
 		/**
 		Location for executable scripts. Sugar to generate entries in the `bin` property by walking the folder.
@@ -483,7 +484,7 @@ declare namespace PackageJson {
 		/**
 		Is used to set configuration parameters used in package scripts that persist across upgrades.
 		*/
-		config?: Record<string, unknown>;
+		config?: JsonObject;
 
 		/**
 		The dependencies of the package.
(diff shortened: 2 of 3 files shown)


skarab42:
// @voxpelli

PR #479, opened 2022-09-29
fix(readonly-deep): make arrays immutable (not just readonly deep)

RebeccaStevens:
This PR essentially makes it so that `ReadonlyDeep<T[]>` returns `Readonly<readonly T[]>` instead of just `readonly T[]` (like how `Set` and `Map` are handled).

However, as TypeScript handles arrays "specially" a small work around is needed to make `Readonly<readonly T[]>` actually work as expected.

### DIFF
--- a/source/readonly-deep.d.ts
+++ b/source/readonly-deep.d.ts
@@ -63,9 +63,11 @@ type ReadonlySetDeep<ItemType> = {} & Readonly<ReadonlySet<ReadonlyDeep<ItemType
 /**
 Same as `ReadonlyDeep`, but accepts only `object`s as inputs. Internal helper for `ReadonlyDeep`.
 */
-type ReadonlyObjectDeep<ObjectType extends object> = {
-	readonly [KeyType in keyof ObjectType]: ReadonlyDeep<ObjectType[KeyType]>
-};
+type ReadonlyObjectDeep<ObjectType extends object> =
+	ObjectType extends readonly unknown[]
+		// Only apply the array fix to arrays to prevent potential issues.
+		? {readonly [KeyType in keyof (ObjectType & {})]: ReadonlyDeep<ObjectType[KeyType]>;}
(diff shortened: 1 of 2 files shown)


sindresorhus:
> This PR essentially makes it so that ReadonlyDeep<T[]> returns Readonly<readonly T[]> instead of just readonly T[] (like how Set and Map are handled).

Why is `readonly T[]` not good enough?

RebeccaStevens:
With `readonly T[]`, the methods of the array are not readonly, and thus they can be reassigned like so:

```ts
const foo: readonly string[] = [];
foo.every = () => true;
```

Currently `Set<T>` (and `Map`) is changed to `Readonly<ReadonlySet<T>>` so arrays should be treated the same.

Maybe it would be best to make a new `Immutable` type that does this sort of thing and the `ReadonlyDeep` type can focus on the more traditional readonlyness transformations?

ulken:
> With `readonly T[]`, the methods of the array are not readonly, and thus they can be reassigned like so:
> 
> ```ts
> const foo: readonly string[] = [];
> foo.every = () => true;
> ```
> 
> Currently `Set<T>` (and `Map`) is changed to `Readonly<ReadonlySet<T>>` so arrays should be treated the same.
> 
> Maybe it would be best to make a new `Immutable` type that does this sort of thing and the `ReadonlyDeep` type can focus on the more traditional readonlyness transformations?

Since @sindresorhus [asked for opinions](https://twitter.com/sindresorhus/status/1586348988000702464) on the matter:

1. I'm in favor of aligning with the other data structures (although, as always, would have preferred for it to be the built-in default case — are there any valid reason for it not to be or has it just been overlooked)?
2. Leaving out any implementation details, I prefer the separation of concerns and composability of leaving `ReadonlyDeep` as is and introducing a `Immutable` (or similar) type.

sindresorhus:
Lets go with a new `Immutable` type.

Issue #468, opened 2022-09-19
Jsonify does not look deeper than one `toJSON` level


MGibson1:
I have a type to Jsonify, similar to this:
```typescript
class Top {
  override: ArrayBuffer;
  inner: Inner;

  toJSON() {
    const stringOverride = "override";
    return {
      override: stringOverride,
      inner: this.inner,
    };
  }
}

class Inner {
  buffer: ArrayBuffer;

  toJSON() {
    const stringBuffer = "buffer";
    return {
      buffer: stringBuffer,
    };
  }
}
```

In this example, `Jsonify<Top>` will return `never`, because `Inner` is not a valid `JsonValue`, however, `Jsonify<Inner>` is a `JsonValue`. The `Jsonify<T>` type will inspect `T` for a `toJSON` method, but mandates that the return type be a valid `JsonValue`.

The relevant section of `Jsonify` is 
```typescript
...
	? T extends {toJSON(): infer J}
		? (() => J) extends (() => JsonValue) // Is J assignable to JsonValue?
			? J // Then T is Jsonable and its Jsonable value is J
			: never // Not Jsonable because its toJSON() method does not return JsonValue
...
```

I would like to expand the final `never` above to mimic the recursive `Jsonify` call below it to do the same deep `Jsonify` process even if a `toJSON` method is defined. This is more similar to how `JSON.stringify` works, which will recursively call `toJSON` on each property that defines it.

sindresorhus:
This looks like an oversight to me. PR welcome.

Issue #466, opened 2022-09-16
I think I found a bug in your Jsonify type?


Evertt:
The title ends with a question mark, because I've literally never installed this package before. I was just reading the source code out of interest.

And while reading the code for `Jsonify`, I saw something that looked like a bug to me. Namely here:

https://github.com/sindresorhus/type-fest/blob/main/source/jsonify.d.ts#L103

At first I was like _"oh really? I never knew that `Map` and `Set` are converted to empty objects by `JSON.stringify()`, that's interesting, let's try that!"_. And indeed:

<img width="253" alt="image" src="https://user-images.githubusercontent.com/1267282/190583200-c5b29aba-4a94-4827-bcb2-dc5625972827.png">

However, that line that I linked to, is still a bug, I believe. Because the `{}` there is not representing `an empty object value`, but instead it represents the `{}` type, which as far as I know means `any non-nullish type`. So instead you should use your own [`EmptyObject` type](https://github.com/sindresorhus/type-fest/blob/main/source/empty-object.d.ts), don't you think?

sindresorhus:
Seems like a bug to me. It was introduced in https://github.com/sindresorhus/type-fest/commit/8ca2959bc8613e6856681dd14abb1ebcbfef6b25.

// @frontsideair

frontsideair:
Yeah, sounds like a bug. It looks like there's already a PR addressing it, otherwise I'd be looking at it. Thanks for taking the initiative!

PR #481, opened 2022-09-30
Fix for #468 and #466

Evertt:
I hope the tests I added are up to your standards.


sindresorhus:
Thanks for working one this. It would be great if you could split it up into two pull requests (one for each fix). And you seem to have some listing issues.

Evertt:
Is it okay if I just fix the linting errors? 😇
I notice I'm not motivated enough to split this into 2 PRs. 😬

sindresorhus:
Combined PRs like this one makes it harder to review, harder to read the Git history later on, and makes it harder to revert changes in case of a regression.

PR #487, opened 2022-10-11
Update tests to use `expectNotAssignable` where appropriate

trevorade:
Specifically, current tests frequently use `expectError` when `expectNotAssignable` would have been more appropriate.

Also updates the `includes.ts` test to have explanatory comments for its `expectError` usage.

Lastly, updates the contributing doc to suggest `expectNotAssignable` first + updates links.

### DIFF
--- a/.github/contributing.md
+++ b/.github/contributing.md
@@ -6,7 +6,7 @@
 - Please help review the other open pull requests. If there are no open pull requests, provide some feedback on some of the open issues.
 - Create a new file in the `test-d` directory and write at least one type test.
 	- See the other tests for inspiration.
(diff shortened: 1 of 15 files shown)


trevorade:
**Context:** While pulling in `type-fest` and trying to get the tests running, there was some confusion as to what `expectError` was doing as written in the tests. It's pretty clear the intent of the test cases updated in this PR is better represented using `expectNotAssignable` so updating.

There were several other usages of `expectError` which could have been rewritten to also use `expectNotAssignable` though that would not have been as simple of an update so probably merits its own PR if desired.

sindresorhus:
Good catch 👍

PR #491, opened 2022-10-13
Add TypeScript canary on cron job that tests on `typescript@next`

voxpelli:
As TS has a tendency to break in one way or another on new releases we should try to catch that early, and one way to do that is to run automated canary tests against the nightly `typescript@next` and the `typescript@latest`

This way we can catch regressions early and notify the TS team about them early.

I have made used of this in some of my modules, which has helped catch some regressions early. See eg: https://github.com/voxpelli/pony-cause/blob/main/.github/workflows/ts.yml

### DIFF
--- a/.github/workflows/ts-canary.yml
+++ b/.github/workflows/ts-canary.yml
@@ -1,0 +1,24 @@
+name: TypeScript Canary
+
+on:
+  schedule:
+    - cron: '15 21 * * 4'
+
+jobs:
+  types:
+    name: TypeScript ${{ matrix.typescript-version }}
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ts-canary.yml
```diff
@@ -0,0 +1,24 @@
+name: TypeScript Canary
+
+on:
+  schedule:
+    - cron: '15 21 * * 4'
```
sindresorhus, line 5: Can you add a comment that describes the interval? Most people cannot read cron syntax.

voxpelli: Done ✔️

Issue #471, opened 2022-09-23
type-fest@3 - How to declare optional properties in JsonObject?


peterjuras:
After the update to version 3.0.0, it is no longer possible to have optional properties on a JsonObject:

```ts
import { JsonObject } from "type-fest";

interface MyObject extends JsonObject {
  optionalProp?: string
}
```

Gives a TypeError:
```
(property) MyObject.optionalProp?: string | undefined
Property 'optionalProp' of type 'string | undefined' is not assignable to 'string' index type '((string | number | boolean | JsonObject | JsonArray) & (string | number | boolean | JsonObject | JsonArray | undefined)) | null'.ts(2411)
```

How should this type be used correctly? The code above worked fine with version 2.x.

[TypeScript Playground](https://codesandbox.io/s/typescript-playground-export-forked-1y15br?file=/index.ts)

sindresorhus:
Probably related to https://github.com/sindresorhus/type-fest/pull/465.

// @skarab42

skarab42:
Unfortunately I can't find (yet) a way to deny the `undefined` value alone and at the same time allow optional properties with `exactOptionalPropertyTypes` set to `false` when extending interface.

A first workaround is to use a type rather than an interface. [Playground](https://www.typescriptlang.org/play?exactOptionalPropertyTypes=false#code/JYWwDg9gTgLgBAbzgKQM4QHYHkBGArAUwGN4BfOAMyghDgCIYBPMAgWgoNRjoG4AoPkxZwAso1yEScALwp02fMXgAyRHzhwIYGMEwBDADYAFamAD8ALjhcowDAHN+pPkA)

```
type MyObject = JsonObject & {
  optionalProp?: string;
}
```

A second workaround is to set `exactOptionalPropertyTypes` to `true` ([Playground](https://www.typescriptlang.org/play?exactOptionalPropertyTypes=true#code/JYWwDg9gTgLgBAbzgKQM4QHYHkBGArAUwGN4BfOAMyghDgCIYBPMAgWgoNRjoG4AoPsAwwCUCgEMiBOAFlGuQiTgEAHiIwATVCnTZ8xeAj5w4EMDGCZxAGwAK1MAH4AXHC5QhAc36k+QA)). 

If none of the above solutions are suitable, a generic type such as `ToJsonObject<MyObject>` can be provided.

peterjuras:
Can you mention in which usecases undefined values are a problem for JsonObjects?

[...]

PR #492, opened 2022-10-16
Add `Jsonifiable` type

### LINKED ISSUE #356: Proposal: Jsonifyable
I'm hitting a problem where `(string | undefined)[]` is incompatible with `JsonValue`. 

```ts
export type BrokenWrapper<TValue extends JsonValue> = {};

export type MyType = string[];

export type BrokenWrappedType = BrokenWrapper<PartialDeep<MyType>>;
```
https://codesandbox.io/s/nice-water-lp2n5

Will produce:

```
Type '(string | undefined)[]' does not satisfy the constraint 'JsonValue'.
  Type '(string | undefined)[]' is not assignable to type 'JsonArray'.
    Type 'string | undefined' is not assignable to type 'JsonValue'.
      Type 'undefined' is not assignable to type 'JsonValue'.  ts(2344)
```

I think this error is correct because `undefined` is certainly not allowed in a JSON value.

What I want is a way to express an object that *can* pass JSON.stringify without error. Ie, Jsonifyable!

It's almost identical, but `undefined` is allowed as a value:

```ts
export type JsonifyableObject = { [Key in string]?: JsonifyableValue };
export type JsonifyableArray = JsonifyableValue[];
export type JsonifyablePrimitive = string | number | boolean | null | undefined;
//                                                                  ^^^ this is the magic
export type JsonifyableValue =
  | JsonifyablePrimitive
  | JsonifyableObject
  | JsonifyableArray;
```

If you'd like me to turn this into a PR, I'd be happy to. I'm using it in my production code right now.

itsjohncs:
Closes #356.

### DIFF
--- a/source/jsonifiable.d.ts
+++ b/source/jsonifiable.d.ts
@@ -6,7 +6,7 @@ type JsonifiableArray = Jsonifiable[];
 /**
 Matches a value that can be losslessly converted to JSON.
 
-`undefined` is allowed in object fields (ex: `{a?: number}`) as a special case even though `JSON.stringify({a: undefined})` is `{}`.
+`undefined` is allowed in object fields (ex: `{a?: number}`) as a special case even though `JSON.stringify({a: undefined})` is `{}` because it makes this class more widely useful and checking for undefined-but-present values is likely an anti-pattern.
 
 @example
 ```
--- a/source/jsonifiable.d.ts
+++ b/source/jsonifiable.d.ts
@@ -1,7 +1,7 @@
 import type {JsonPrimitive} from './basic';
 
 type JsonifiablePrimitive = JsonPrimitive | {toJSON: () => Jsonifiable};
-type JsonifiableObject = {[Key in string]?: Jsonifiable};
+type JsonifiableObject = {[Key in string]: Jsonifiable};
 type JsonifiableArray = Jsonifiable[];
 
 /**
--- a/source/jsonifiable.d.ts
+++ b/source/jsonifiable.d.ts
@@ -1,7 +1,6 @@
 import type {JsonPrimitive, JsonValue} from './basic';
 
-type JsonifiablePrimitive = JsonPrimitive | {toJSON: () => JsonValue};
-type JsonifiableObject = {[Key in string]?: Jsonifiable};
+type JsonifiableObject = {[Key in string]?: Jsonifiable} | {toJSON: () => JsonValue};
 type JsonifiableArray = Jsonifiable[];
 
 /**
@@ -30,4 +29,4 @@ JSON.stringify(good); // {"number": 3, "date": "2022-10-17T22:22:35.920Z"}
 
 @category JSON
 */
-export type Jsonifiable = JsonifiablePrimitive | JsonifiableObject | JsonifiableArray;
+export type Jsonifiable = JsonPrimitive | JsonifiableObject | JsonifiableArray;
--- a/source/jsonifiable.d.ts
+++ b/source/jsonifiable.d.ts
@@ -6,6 +6,8 @@ type JsonifiableArray = readonly Jsonifiable[];
 /**
 Matches a value that can be losslessly converted to JSON.
 
+Can be used to type values that you expect to pass to `JSON.stringify`.
+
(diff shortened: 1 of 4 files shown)


itsjohncs:
Whoops, I somehow missed that `undefined` being allowed is desired! Admittedly I did a pretty cursory reading of #356. I'll update the type to permit `undefined`.

source/jsonifiable.d.ts
```diff
@@ -0,0 +1,32 @@
+import type {JsonPrimitive, JsonValue} from './basic';
+
+type JsonifiableObject = {[Key in string]?: Jsonifiable} | {toJSON: () => JsonValue};
+type JsonifiableArray = Jsonifiable[];
+
+/**
+Matches a value that can be losslessly converted to JSON.
+
+`undefined` is allowed in object fields (ex: `{a?: number}`) as a
+special case even though `JSON.stringify({a: undefined})` is `{}`.
```
sindresorhus, line 10: Don't hard-wrap.

It think the user will ask themselves why `undefined` is allowed, so I think the docs here should quickly answer that.

test-d/jsonifiable.ts
```diff
@@ -0,0 +1,16 @@
+import {expectAssignable, expectNotAssignable} from 'tsd';
+import type {Jsonifiable} from '..';
+
+expectAssignable<Jsonifiable>(1);
+expectAssignable<Jsonifiable>('');
+expectAssignable<Jsonifiable>(null);
+expectAssignable<Jsonifiable>(new Date());
+expectAssignable<Jsonifiable>({a: new Date()});
+expectAssignable<Jsonifiable>([new Date()]);
```
sindresorhus, line 9: Can you add an assertion for readonly array too?

source/jsonifiable.d.ts
```diff
@@ -0,0 +1,32 @@
+import type {JsonPrimitive, JsonValue} from './basic';
+
+type JsonifiableObject = {[Key in string]?: Jsonifiable} | {toJSON: () => JsonValue};
+type JsonifiableArray = Jsonifiable[];
+
+/**
+Matches a value that can be losslessly converted to JSON.
+
```
sindresorhus, line 8: Would you be useful to mention some use-cases for this type. (I realize there is an example, but nothing beats just reading some text quickly)

source/jsonifiable.d.ts
```diff
@@ -0,0 +1,32 @@
+import type {JsonPrimitive, JsonValue} from './basic';
+
+type JsonifiableObject = {[Key in string]?: Jsonifiable} | {toJSON: () => JsonValue};
```
sindresorhus, line 3: Shouldn't this look deeply for `toJSON`?

See https://github.com/sindresorhus/type-fest/commit/e11d30b727fc8b3e9438fbf15ec8619449e5566e and https://github.com/sindresorhus/type-fest/issues/468

sindresorhus (changes requested):

source/jsonifiable.d.ts
itsjohncs: Could you give me a test case that you're worried about? I'm not sure I understand your concern. I added this test case as a guess of what you mean though:

```typescript
expectAssignable<Jsonifiable>({a: {deeply: {nested: {toJsonObject: new Date()}}}});
```

itsjohncs: Disregard my last comment. I must have messed up the test I did which prompted https://github.com/sindresorhus/type-fest/pull/492/commits/d8d78f8c052561b8fefbe431c95cb6f507e2b8e1. I'll roll back that change and add a test.

sindresorhus (approved):

skarab42 (approved):

MichaelDeBoey:
Should we use this type in `Jsonify` as well?

skarab42:
> Should we use this type in `Jsonify` as well?

No
- `Jsonifiable` is used to type values that you expect to pass to `JSON.stringify` where { foo: undefined } and plain {} are functionally equivalent.
- `Jsonify` is used to type parsed JSON value, coming from eg. JSON.parse() where undefined as a value is impossible.

FI https://github.com/sindresorhus/type-fest/issues/471#issuecomment-1285314479

shaharmor:
@skarab42 why no also export `JsonifiableObject` and `JsonifiableArray`?

PR #493, opened 2022-10-18
Update Opaque test to not need @ts-expect-error

trevorade:
When attempting to get these tests running in our environment, we ran into some issues with the `@ts-expect-error` instances in `test-d/opaque.ts` causing issues.

This update simply updates the code to not need them. For example, in the case of unused types or variables, use them.


trevorade:
I'll update my PR in a bit to fix the issues...

trevorade:
I'm going to just go ahead and recreate this :)

Issue #336, opened 2021-12-09
snakeCasedProperties and camelCasedProperties have different behaviours when parsing numbers


olivierbeaulieu:
Given the following example:

```
type F = SnakeCasedPropertiesDeep<
  CamelCasedPropertiesDeep<{
    foo_1: boolean;
  }>
>;
```

I would expect each operation to be the opposite operation of each other - resulting in F being back to its original value `F = { foo_1: boolean }`.

That is not the current behaviour, we as have `F = { foo1: boolean }`.

Is that behaviour correct? If so, how can I get back to my original value when jumping from one case to the other?

kmordan24:
@olivierbeaulieu I ran into this same issue.

I absolutely love this library, but I really do feel like char->number transitions should add underscore as well. It might be reasonable to not make this assumption, especially with something like { "http2": true } where we would NOT want http2 -> http_2. There is no clear standard on this but the most common approach is "column_1" rather than "column1" if you are doing everything in snake case...

Here are some Frankenstein'd types.... That being said this is not nearly as elegant as the type definitions written in this library, so idk.

Snake To Camel Case:
```
export type CamelizeInputType = Record<PropertyKey, any> | Array<any>;

export type SnakeToCamelCase<S extends PropertyKey> = S extends number
  ? S
  : S extends `${infer T}_${infer U}`
  ? `${T}${Capitalize<SnakeToCamelCase<U>>}`
  : S;

export type SnakeToCamelCaseNested<T> = T extends Function | RegExp | Date
  ? T
  : T extends (infer E)[]
  ? SnakeToCamelCaseNested<E>[]
  : T extends CamelizeInputType
  ? {
      [K in keyof T as SnakeToCamelCase<Extract<K, PropertyKey>>]: SnakeToCamelCaseNested<T[K]>;
    }
  : T;
```


Camel To Snake Case
```
export type SnakeCaseInputType = Record<PropertyKey, any> | Array<any>;
type UpperAlphabetic =
  | 'A'
  | 'B'
  | 'C'
  | 'D'
  | 'E'
  | 'F'
  | 'G'
  | 'H'
  | 'I'
  | 'J'
  | 'K'
  | 'L'
  | 'M'
  | 'N'
  | 'O'
  | 'P'
  | 'Q'
  | 'R'
  | 'S'
  | 'T'
  | 'U'
  | 'V'
  | 'W'
  | 'X'
  | 'Y'
  | 'Z';

[...]

PR #501, opened 2022-10-21
Add `preserveConsecutiveUppercase` option to `CamelCase` and add `SplitWords` type

### LINKED ISSUE #224: Feature Request: CamelCase to behave like lodash.camelCase and camelcase
Right now, `CamelCase` (and, obviously, `CamelCasedProperties` and `CamelCasedPropertiesDeep`) behaves weirdly when handling two character strings.

When it's a 2 character string (like `ID`, for example), it will transform the string as expected (`ID` to `id`), but strings like `HostIDPlayer` will leave the `ID` part untouched (`HostIDPlayer` to `hostIDPlayer`).

```ts
const a: CamelCase<'ID'> = 'id';  // Works, no TS error 
const b: CamelCase<'HostIDPlayer'> = 'hostIdPlayer';  // Error, TS expects 'hostIDPlayer'
```

Libraries like `lodash.camelCase` and `camelcase` will transform the strings as expected as can be seen in [this codesandbox](https://codesandbox.io/s/festive-leavitt-byu6y?file=/src/index.js).

I can't say I know enough about TS' type system, but is there a way to make `CamelCase` behave like `lodash.camelCase` and `camelcase`? That was, to me, what I expected the behavior to be.

### LINKED ISSUE #488: `CamelCase` does not properly camelize consecutive characters
Hi!

I'm trying to use the `CamelCase` type in a project. I have the following code:

```ts
type Sample = CamelCase<"GameID">;
```

My expectation according to the docs is `Sample` would now be typed as a string value of `"gameId"`. However, it is instead typed as `"gameID"`, with __ID__ being all uppercase.

This caught me off-guard, as the example in _camel-case.d.ts_ has this code listing:

```ts
interface RawOptions {
	'dry-run': boolean;
	'full_family_name': string;
	foo: number;
	BAR: string;
	QUZ_QUX: number; // <-- here
	'OTHER-FIELD': boolean;
}

const dbResult: CamelCasedProperties<RawOptions> = {
	dryRun: true,
	fullFamilyName: 'bar.js',
	foo: 123,
	bar: 'foo',
	quzQux: 6, // <-- here
	otherField: false
};
```

For my use case, this issue impacts the `CamelCasedProperties` and `CamelCasedPropertiesDeep` types, which leverage `CamelCase`.

skarab42:
Fix #224
Fix #488 

```ts
expectType<CamelCase<'fooBAR'>>('fooBAR');
expectType<CamelCase<'fooBAR', {preserveConsecutiveUppercase: false}>>('fooBar');

expectType<CamelCase<'fooBARBiz'>>('fooBARBiz');
expectType<CamelCase<'fooBARBiz', {preserveConsecutiveUppercase: false}>>('fooBarBiz');

expectType<CamelCase<'foo BAR-Biz_BUZZ'>>('fooBARBizBUZZ');
expectType<CamelCase<'foo BAR-Biz_BUZZ', {preserveConsecutiveUppercase: false}>>('fooBarBizBuzz');
```

In anticipation for #336 this PR also adds the `SplitWords` type (internal for now).

```ts
type Words0 = SplitWords<'helloWorld'>; // ['hello', 'World']
type Words1 = SplitWords<'helloWORLD'>; // ['hello', 'WORLD']
type Words2 = SplitWords<'hello-world'>; // ['hello', 'world']
type Words3 = SplitWords<'--hello the_world'>; // ['hello', 'the', 'world']
type Words4 = SplitWords<'lifeIs42'>; // ['life', 'Is', '42']
```


source/split-words.d.ts
```diff
@@ -0,0 +1,57 @@
+import type {IsLowerCase, IsNumeric, IsUpperCase, WordSeparators} from './internal';
+
+type SkipEmptyWord<Word extends string> = Word extends '' ? [] : [Word];
+
+type RemoveLastChar<Sentence extends string, Char extends string> = Sentence extends `${infer LeftSide}${Char}`
```
sindresorhus, line 5: `Char` => `Character`. Same for the other cases.

sindresorhus:
Nice work 👍

RichiCoder1:
It seems like this PR and `camelcase` have two different behaviors:
* `camelCase("TPSCode", { preserveConsecutiveUppercase: true})` => `TPSCode`
* `CamelCase<'TPSCode', {preserveConsecutiveUppercase: true}>` => `tPSCode`

skarab42:
> Is there anyway to fix or consolidate this behavior?

@RichiCoder1  As a temporary fix you can use [Capitalize](https://www.typescriptlang.org/docs/handbook/utility-types.html#capitalizestringtype)

ulken:
Since `PascalCase` [uses](https://github.com/sindresorhus/type-fest/blob/7294cf0e14883724f899ee2531cfed21fc46656e/source/pascal-case.d.ts#L1) `CamelCase` under the hood, I'd expect the options to propagate. Upgrading broke existing usages of `PascalCase`, without (seemingly) any way to get the old behavior back.

I can work around it, but I still think it should be configurable.

PR #504, opened 2022-10-24
Make TupleToUnion support T[] instead of just tuples

trevorade:
This implementation is no longer recursive which may be more performant. It also works with arrays whereas the previous implementation only worked with tuples.

In addition, it also resolves an issue where infinite tuples were not supported. Previously, `TupleToUnion<[string, ...number[]]>` resolved to `string` and now it correctly resolves to `string | number`.

### DIFF
--- a/source/tuple-to-union.d.ts
+++ b/source/tuple-to-union.d.ts
@@ -48,4 +48,4 @@ type NumberBool = typeof numberBool[number];
 
 @category Array
 */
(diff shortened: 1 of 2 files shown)


sindresorhus:
// @mattmahn

trevorade:
Any clarification you'd like @mattmahn or @sindresorhus?

skarab42 (approved):

buschtoens:
I just wanted to leave this comment to appreciate that this works now and that you've submitted this PR. Thank you! ❤️

PR #511, opened 2022-11-14
Add `GlobalThis` type

trevorade:
This type is for times when you need to access non-standard properties off of ~`window`, `self`, or~ `globalThis`. For example:

```ts
import type {GlobalThis} from 'type-fest';

interface ExtraProps extends GlobalThis {
  readonly GLOBAL_TOKEN: string;
}

const extraProps = globalThis as ExtraProps;
verifyToken(extraProps.GLOBAL_TOKEN);
```

---

Normally, you would define a declaration file to add global variables.

This change permits introducing global symbols in a more targeted manner.

---

Note: In the test, it seems like `expectError` should work here rather than using `@ts-expect-error` but, unfortunately, it does not.

---

I'd be happy to hear other suggestions on supporting this use case!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -236,6 +236,7 @@ Click the type names for complete docs.
 
 ### Miscellaneous
 
+- [`GlobalThis`](source/global-this.d.ts) - Declare locally scoped properties on `globalThis`.
 - [`PackageJson`](source/package-json.d.ts) - Type for [npm's `package.json` file](https://docs.npmjs.com/creating-a-package-json-file). It also includes support for [TypeScript Declaration Files](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) and [Yarn Workspaces](https://classic.yarnpkg.com/lang/en/docs/workspaces/).
 - [`TsConfigJson`](source/tsconfig-json.d.ts) - Type for [TypeScript's `tsconfig.json` file](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html).
 
--- a/source/global-this.d.ts
+++ b/source/global-this.d.ts
@@ -1,0 +1,21 @@
+/**
+Declare locally scoped properties on `globalThis`.
+
+When defining a global variable in a declaration file is inappropriate, it can be helpful to define a `type` or `interface` (say `ExtraGlobals`) with the global variable and then cast `globalThis` via code like `globalThis as unknown as ExtraGlobals`.
+
+Instead of casting through `unknown`, you can update your `type` or `interface` to extend `GlobalThis` and then directly cast `globalThis`.
+
+@example
+```
+import type {GlobalThis} from 'type-fest';
+
+type ExtraGlobals = GlobalThis & {
+	readonly GLOBAL_TOKEN: string;
+};
+
+(globalThis as ExtraGlobals).GLOBAL_TOKEN;
+```
+
+@category Type
+*/
+export type GlobalThis = typeof globalThis;
--- a/test-d/global-this.ts
+++ b/test-d/global-this.ts
@@ -1,0 +1,14 @@
+import {expectType} from 'tsd';
+import type {GlobalThis} from '../index';
+
+type ExtraProps = GlobalThis & {
+	readonly GLOBAL_TOKEN: string;
+};
+
+// Verify `globalThis` can be cast to a type which extends `GlobalThis`.
+expectType<string>((globalThis as ExtraProps).GLOBAL_TOKEN);
+
+// Verify that object literals cannot be cast to a type which extends `GlobalThis`.
+declare function consumeExtraProps(extraProps: ExtraProps): void;
+// @ts-expect-error
+consumeExtraProps(({something: 'value'}) as ExtraProps);
--- a/index.d.ts
+++ b/index.d.ts
@@ -101,5 +101,6 @@ export type {Get} from './source/get';
 export type {LastArrayElement} from './source/last-array-element';
 
 // Miscellaneous
+export type {GlobalThis} from './source/global-this';
 export type {PackageJson} from './source/package-json';
 export type {TsConfigJson} from './source/tsconfig-json';


source/global-this.d.ts
```diff
@@ -0,0 +1,40 @@
...
+// @ts-ignore
+type WindowSansInfinityNorNaN = ExcludeAny<Window & {
+	Infinity: never;
+	NaN: never;
+}>;
+
+/**
+If `T` is `any`, returns unknown. Safe as `SomeType & unknown` is `SomeType`.
+*/
+type ExcludeAny<Type> = Type extends never ? unknown : Type;
```
sindresorhus, line 40: Is it expected that it returns `unknown` and not `never` when the input is `any`?

Alternative name: `AnyToUnknown`

sindresorhus: This should be moved to internal.d.ts

sindresorhus: Maybe: `type AnyToUnknown<T> = unknown extends T ? unknown : T;`

sindresorhus (changes requested):

sindresorhus:
I don't think we should support `window` and `self`. `globalThis` is the correct one to use for new code.

test-d/global-this.ts
```diff
@@ -0,0 +1,16 @@
...
+
+// Verify `globalThis` type objects can be casted with a type which extends `GlobalThis`.
+expectType<string>((globalThis as ExtraProps).GLOBAL_TOKEN);
+expectType<string>((window as ExtraProps).GLOBAL_TOKEN);
+expectType<string>((self as ExtraProps).GLOBAL_TOKEN);
+
+// Verify that non-`globalThis` type objects cannot be casted with a type which extends `GlobalThis`.
+declare function consume(): void;
+// @ts-expect-error
+consume(({something: 'value'}) as ExtraProps);
```
papb, line 16: How does this test work? By looking at it it seems that `@ts-expect-error` would simply catch the fact that you're passing an argument to a function that takes no arguments.

source/global-this.d.ts
trevorade: No longer relevant as `ExcludeAny` is no longer required.

trevorade:
> I don't think we should support `window` and `self`. `globalThis` is the correct one to use for new code.

Fair enough. Just need to guard our codebase from `window` usages :)

test-d/global-this.ts
trevorade: Yeah. Head palm. That's the trouble with `@ts-expect-error`. It's too broad.

Thx for pointing this out. Fixed now.

source/global-this.d.ts
```diff
@@ -0,0 +1,17 @@
+/**
+Aids in defining global symbols in a narrow scope. Permits extending `globalThis`.
```
sindresorhus, line 2: The first line here should be the same as in the readme. Below this first line could be a more in-depth explanation.

readme.md
```diff
@@ -236,6 +236,7 @@ Click the type names for complete docs.
 
 ### Miscellaneous
 
+- [`GlobalThis`](source/global-this.d.ts) - A supertype for custom subtypes of `globalThis`. Useful to declare global symbols that are targeted to a narrow scope without casting through `unknown`.
```
sindresorhus, line 239: Would be great if the description could be a single sentence and a little bit more succinct. The more in-depth description can be in the docs for the type.

source/global-this.d.ts
trevorade: Done.

readme.md
```diff
@@ -236,6 +236,7 @@ Click the type names for complete docs.
 
 ### Miscellaneous
 
+- [`GlobalThis`](source/global-this.d.ts) - Type for `globalThis`. Extend to declare locally scoped global symbols.
```
sindresorhus, line 239:
```suggestion
- [`GlobalThis`](source/global-this.d.ts) - Declare locally scoped properties on `globalThis`.
```

source/global-this.d.ts
```diff
@@ -0,0 +1,21 @@
+/**
+Type for `globalThis`. Extend to declare locally scoped global symbols.
+
+When defining a global symbol in a declaration file is inappropriate, it can be helpful to define an `interface` or `type` (say `Foo`) and then cast `globalThis` via code like `globalThis as unknown as Foo`.
```
sindresorhus, line 4: I would use a different word than "symbol" as symbol has a specific meaning in JS that does not apply to this.

source/global-this.d.ts
```diff
@@ -0,0 +1,21 @@
...
+
+Instead of casting through `unknown`, you can extend `GlobalThis` and then directly cast `globalThis` to your local `interface`.
+
+@example
+```
+import type {GlobalThis} from 'type-fest';
+
+interface ExtraGlobals extends GlobalThis {
```
sindresorhus, line 12: Super nitpick, but I would prefer using a `type` here.

sindresorhus (changes requested):

source/global-this.d.ts
trevorade: Should we show both?

To be honest, I was surprised that this code-base makes interfaces prohibited and enforced that I write this as:

```ts
type ExtraGlobals = GlobalThis & {
	readonly GLOBAL_TOKEN: string;
};
```

At least for my organization, we have made the explicit decision to use `interface` rather than `type` for this sort of thing.

sindresorhus: I prefer to just show type. It's expected that a user can translate it to interface if they prefer that.

readme.md
trevorade: Done

source/global-this.d.ts
trevorade: Done.

source/global-this.d.ts
trevorade: Updated.

sindresorhus:
Nice work 👍

Issue #486, opened 2022-10-10
Problem with `SetNonNullable` in Type Predicate Context


joealden:
[Playground link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAwg9gOwGYEsDmUC8UDeUCGAXFAEZxwA2E+CUAPlAgK4UUDcpxZl1tDzrKAF82AKFEB6AFRZMc+QsVLlK1WvXKoUieJQBbMHABOwXFFCQoAZQjAAcojssK+ElQIBna7YcInrVyoARmEoJCM4PSgAIgsIAFokCA9gaLFRAGNEFKgs5HQACXwPAAUIsBDsAB4AFTK4MCgIAA9gCAQAEy8AawgQOCRYRFQ0AD4AClEoXOH0Ynh8tAAaKagwcuI68tEASmI8kagULxt7R2dAiCCqhZGlqC2G0axnnFWjWyYjWgAhcioaOMDugANrrBoAXR2YhE4jyOTaKQAYghKlAgbM0PNMTsXrhVtMUIMMYsiqVykESXcYvhojtcW9pkymcC0AA6fBiZncpkSCRQAB6AH4CcJREJxNJZBoZbK5fK5FodKI4t4zn4Lm4IAAmKqrH7FCA1cAQFbTADSfQ8NTgpyarXaXSgvX6gwNHiNJqwzr6Ayg7s9kFEz2wjKgIMtICOtBdfoDxsgEOIkftbU6Xkj1tttlFQqgvn8Li1VXjJojfQho1FxFLkHLIAhMPS8NMrLJ9TA2u9tQ7qcdPV9g1u6Amq1Z2MWZrWGweHd2+0xRxOPnOAS1uuHy1n5RDr3en2+-v+vCpoPBYChTbh2VMiOAKK72FPWKGi1xmD3zKJ6LbxQ72ufe5olpel8R5aY4j9VkOS5cC4N5flhVFCUhCAA).

When using`SetNonNullable` in a type predicate context where the predicate contains a generic from the function, I am seeing the following error: `A type predicate's type must be assignable to its parameter's type (2677)`.

When I use the `SetNonNullable` from my own types (see https://github.com/sindresorhus/type-fest/issues/228#issuecomment-1136017869), this is not an issue. Is there something we can do to the existing type to prevent this error, and is there any reason that the type definition provided in https://github.com/sindresorhus/type-fest/issues/228#issuecomment-1136017869 wouldn't suffice?

[...]

PR #515, opened 2022-11-23
Refactor `SetNonNullable` Type

joealden:
Resolves https://github.com/sindresorhus/type-fest/issues/486.


sindresorhus:
Would you be able to add a type test? So we can ensure it doesn't regress in the future.

joealden:
@sindresorhus I've given this a try in e0f5d9d, but I'm not totally sure how to write a test that tests for this case, as we aren't wanting to test the resulting type, we are instead wanting to test that the type declaration itself (within a function declaration) doesn't throw an error. If the current codes doesn't suffice, could you please take a stab at this?

sindresorhus:
Thanks :)

PR #517, opened 2022-11-25
Refactor `Join` Type

Cuppachino:
As of 4.7 we don't need a suppression comment. 

[Typescript 4.7 🔗 extends constraints on infer type variables](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#extends-constraints-on-infer-type-variables)


sindresorhus:
Thanks :)

PR #518, opened 2022-11-25
Add `IndexOf` Type

Cuppachino:


sindresorhus:
Thanks for submitting a new type. Make sure you read and follow: https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md#submitting-a-new-type

sindresorhus:
I think the type should be named `Indices`. `IndexOf` makes it sound like a type for defining `Array#indexOf`.

Cuppachino:
@sindresorhus I think that's a good change.

source/indices.d.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Generates a tuple based on the length of the input array
+ * and fills each index with the index number.
+ *
```
sindresorhus, line 4: Remove the `*`. See the other types.

source/indices.d.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Generates a tuple based on the length of the input array
+ * and fills each index with the index number.
```
sindresorhus, line 3: Don't hard-wrap.

source/indices.d.ts
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * @example type FruitIndices = Indices<['apple', 'banana', 'plum']> // [0, 1, 2]
+ * @example type FruitIndex = Indices<['apple', 'banana', 'plum']>[number] // 0 | 1 | 2
+ *
+ * @category Array
+ * @category Utility
+ */
+export type Indices<
```
sindresorhus, line 11: Check if it's possible to use `unknown` for any of the places you use `any`. I prefer to use `unknown` whenever possible.

source/indices.d.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Generates a tuple based on the length of the input array
+ * and fills each index with the index number.
+ *
+ * @example type FruitIndices = Indices<['apple', 'banana', 'plum']> // [0, 1, 2]
+ * @example type FruitIndex = Indices<['apple', 'banana', 'plum']>[number] // 0 | 1 | 2
```
sindresorhus, line 6: Use the same example formatting as other types.

readme.md
```diff
@@ -215,6 +215,7 @@ Click the type names for complete docs.
 - [`MultidimensionalReadonlyArray`](source/multidimensional-readonly-array.d.ts) - Create a type that represents a multidimensional readonly array of the given type and dimensions.
 - [`ReadonlyTuple`](source/readonly-tuple.d.ts) - Create a type that represents a read-only tuple of the given type and length.
 - [`TupleToUnion`](source/tuple-to-union.d.ts) - Convert a tuple/array into a union type of its elements.
+- [`Indices`](source/index-of.d.ts) - Generate a tuple of an array's indices (`[0, 1, 2...]`). Useful for creating an index union (`0 | 1 | 2 | ...`)
```
sindresorhus, line 218: The second sentence can be in indices.d.ts, but it's too verbose to be in the readme.

sindresorhus (changes requested):

source/indices.d.ts
Cuppachino: that's neat. I like this style of comments,

source/indices.d.ts
Cuppachino: Works without issue 👍

Cuppachino (commented):
Would it be alright to add `IndexOf` to this pr? 

Typescript would struggle if you wanted to do something like 
```ts
function <List extends unknown[], Index extends Indices<T>[number]>
```
But the issue can be avoided if we use an additional generic.
```ts
import type {Indices} from '../index';

export type IndexOf<Tuple extends unknown[] | readonly unknown[]> =
	Indices<Tuple> extends infer IndexTuple extends number[]
		? IndexTuple[number]
		: never;
```
```ts
function <List extends unknown[], Index extends IndexOf<T>
```

Cuppachino:
there's a better way to do it without recursion.

source/indices.d.ts
```diff
@@ -0,0 +1,21 @@
+/**
+Create a tuple type of an array's indices.
+
```
sindresorhus, line 3: There needs to be a text explanation with some use-cases where this type could be useful.

source/indices.d.ts
```diff
@@ -0,0 +1,21 @@
...
+```
+type FruitIndices = Indices<['apple', 'banana', 'plum']> // [0, 1, 2]
+```
+
+Can be used to create an index union (`0 | 1 | 2 | ...`)
+
+@example
+```
+type FruitIndex = Indices<['apple', 'banana', 'plum']>[number] // 0 | 1 | 2
+```
```
sindresorhus, line 14: Would be great to show a practical example of `FruitIndex` in use. I feel the example is not complete without.

sindresorhus:
@Cuppachino Friendly bump :)

cuppachino:
What about something like this?

```ts
export type StringIncludes<
	S extends string,
	Search extends string,
	I extends NumberLike<IndexOf<Chars<S>>> = Assert<
		0,
		NumberLike<IndexOf<Chars<S>>>
	>
> = SplitAt<S, I> extends [infer _A, infer B]
	? B extends `${infer _A}${Search}${infer _B}`
		? true
		: false
	: never
```
```ts
type _StringIncludes_Cases /* ⊆ */ = [
	ExpectTrue<StringIncludes<'abc', 'a'>>,
	ExpectTrue<StringIncludes<'abc', 'bc'>>,
	ExpectTrue<StringIncludes<'abc', 'abc'>>,
	ExpectFalse<StringIncludes<'NOT', 'NOT-INCLUDE'>>,
	//
	ExpectTrue<StringIncludes<'abc', '', 0>>,
	ExpectTrue<StringIncludes<'abc', '', '0'>>,
	ExpectTrue<StringIncludes<'abc', 'a', '0'>>,
	ExpectTrue<StringIncludes<'abc', 'abc', '0'>>,
	ExpectFalse<StringIncludes<'NOT', 'NOT-INCLUDE', '0'>>,
	//
	ExpectTrue<StringIncludes<'abc', 'b', 1>>,
	ExpectTrue<StringIncludes<'abc', 'bc', '1'>>,
	ExpectFalse<StringIncludes<'abc', 'a', '1'>>,
	ExpectFalse<StringIncludes<'abc', 'ab', '1'>>,
	ExpectFalse<StringIncludes<'NOT', 'NOT-INCLUDE', '1'>>,
	//
	ExpectTrue<StringIncludes<'abc', 'c', 2>>,
	ExpectTrue<StringIncludes<'abc', 'c', '2'>>,
	ExpectFalse<StringIncludes<'abc', 'a', '2'>>,
	ExpectFalse<StringIncludes<'abc', 'ab', '2'>>,
	ExpectFalse<StringIncludes<'abc', 'abc', '2'>>,
	ExpectFalse<StringIncludes<'NOT', 'NOT-INCLUDE', '2'>>
]
```
For clarity, the other generics I used can be found [here](https://github.com/cuppachino/type-space/pull/30)

source/indices.d.ts
```diff
@@ -1,24 +1,21 @@
...
+
+@category Array
+@category Utility
+*/
+export type Indices<T extends unknown[] | readonly unknown[]> = {
```
cuppachino, line 19: maybe this would be more proper?

```ts
type Indices<T extends unknown[] | readonly unknown[]> = Exclude<
  keyof T,
  keyof unknown[]
> extends `${infer Index extends number}`
  ? Index
  : never
  ```

Comment (cuppachino) [reply source/indices.d.ts:19]: I thought coercing the index to a number would be useful, but in the places I have used it, I think it is easier to work with when left as a string. I'll make another example highlighting what I'm talking about

Comment (cuppachino) [reply source/indices.d.ts:19]: here's that example
```ts
type Zip<Keys extends Record<Indices<Values>, PropertyKey>, Values extends readonly unknown[]> = {
  [K in Indices<Values> as Keys[K]]: Values[K]
}

type Keys = ['a', 'b']
type Values = ['apple', 'butter']

type Zipped = Zip<Keys, Values>
// { a: "apple"; b: "butter"; }
```

sindresorhus:
> What about something like this (for a real example)?

It's a bit complicated. Maybe something like this:

```ts
// Import the Indices utility type
import { Indices } from './indices';

// A function that gets the value at a specific index in a typed array
function at<T extends unknown[] | readonly unknown[]>(arr: T, index: Indices<T>): T[typeof index] {
  return arr[index as any];
}

// Example usage:
const myArray = ['a', 'b', 'c', 'd', 'e'] as const;

// These calls will compile, as the indices are valid for the array type
const firstElement = at(myArray, "0"); // "a"
const thirdElement = at(myArray, "2"); // "c"

// This call will not compile, as the index is invalid for the array type
const invalidElement = at(myArray, "5"); // Error: Type '"5"' is not assignable to type '"0" | "1" | "2" | "3" | "4"'.
```

source/indices.d.ts
sindresorhus: > maybe this would be more proper?

Why would it be more proper?

sindresorhus: > I thought coercing the index to a number would be useful, but in the places I have used it, I think it is easier to work with when left as a string. I'll make another example highlighting what I'm talking about

It should be a number. We could have a utility type to turn the numbers to strings.

sindresorhus:
Bump

sindresorhus:
Closing for lack of response: https://github.com/sindresorhus/type-fest/issues/634

PR #522, opened 2022-11-26
Add IsEqual type

### LINKED ISSUE #346: Export `IsEqual` type
https://github.com/sindresorhus/type-fest/blob/02849add1fa7e4b0a65c0b8071d3821b12d71933/source/internal.d.ts#L1-L11

I needs better docs and usage example.

Read: https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md#submitting-a-new-type

doz13189:
Close #346

### DIFF
--- a/source/is-equal.d.ts
+++ b/source/is-equal.d.ts
@@ -1,0 +1,21 @@
+/**
+Returns a boolean for whether the two given types are equal.
+
+@link https://github.com/microsoft/TypeScript/issues/27024#issuecomment-421529650
+@link https://stackoverflow.com/questions/68961864/how-does-the-equals-work-in-typescript/68963796#68963796
+
+@example
+```
+import type {IsEqual} from 'type-fest';
+
+type Expect<T extends true> = T
+type Case = Expect<IsEqual<Uppercase<"sample">, "SAMPLE">>
+```
+
+@category Utilities
+*/
+export type IsEqual<X, Y> =
+	(<G>() => G extends X ? 1 : 2) extends
+	(<G>() => G extends Y ? 1 : 2)
+		? true
+		: false;
--- a/source/is-equal.d.ts
+++ b/source/is-equal.d.ts
@@ -8,24 +8,6 @@ Returns a boolean for whether the two given types are equal.
 ```
 import type {IsEqual} from 'type-fest';
 
-// `IsEqual` compares 1 and 1, then gets true as a result.
-type ExpectTrue = IsEqual<1, 1>;
-//=> true
-```
-
-@example
-```
-import type {IsEqual} from 'type-fest';
-
-// `IsEqual` compares 2 and 1, then gets false as a result.
-type ExpectFalse = IsEqual<2, 1>;
-//=> false
-```
-
-@example
-```
-import type {IsEqual} from 'type-fest';
-
 // `Includes` Returns a boolean for whether the given array includes the given item.
 // `IsEqual` compares the given array of 0th and the given item, then return true if they are equal.
 type Includes<Value extends readonly any[], Item> =
--- a/source/is-equal.d.ts
+++ b/source/is-equal.d.ts
@@ -11,8 +11,8 @@ Use-cases:
 ```
 import type {IsEqual} from 'type-fest';
 
-// `Includes` Returns a boolean for whether the given array includes the given item.
-// `IsEqual` compares the given array of 0th and the given item, then return true if they are equal.
+// This type returns a boolean for whether the given array includes the given item.
+// `IsEqual` is used to compare the given array at position 0 and the given item and then return true if they are equal.
 type Includes<Value extends readonly any[], Item> =
 	Value extends readonly [Value[0], ...infer rest]
 		? IsEqual<Value[0], Item> extends true
--- a/source/is-equal.d.ts
+++ b/source/is-equal.d.ts
@@ -4,6 +4,9 @@ Returns a boolean for whether the two given types are equal.
 @link https://github.com/microsoft/TypeScript/issues/27024#issuecomment-421529650
 @link https://stackoverflow.com/questions/68961864/how-does-the-equals-work-in-typescript/68963796#68963796
 
+Use-cases:
+- If you want to make a conditional branch based on the result of a comparison of two types.
+
 @example
 ```
 import type {IsEqual} from 'type-fest';
--- a/source/is-equal.d.ts
+++ b/source/is-equal.d.ts
@@ -6,13 +6,19 @@ Returns a boolean for whether the two given types are equal.
 
 @example
 ```
-type ExpectTrue = IsEqual<1, 1>
+import type {IsEqual} from 'type-fest';
+
+// `IsEqual` compares 1 and 1, then gets true as a result.
+type ExpectTrue = IsEqual<1, 1>;
 //=> true
 ```
 
 @example
 ```
-type ExpectFalse = IsEqual<2, 1>
+import type {IsEqual} from 'type-fest';
+
+// `IsEqual` compares 2 and 1, then gets false as a result.
+type ExpectFalse = IsEqual<2, 1>;
 //=> false
 ```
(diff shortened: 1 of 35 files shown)


source/is-equal.d.ts
```diff
@@ -0,0 +1,21 @@
...
+```
+import type {IsEqual} from 'type-fest';
+
+type Expect<T extends true> = T
+type Case = Expect<IsEqual<Uppercase<"sample">, "SAMPLE">>
+```
+
+@category Utilities
+*/
+export type IsEqual<X, Y> =
```
sindresorhus, line 17: X/Y are commonly used for coordinates. It would be better to use A and B.

source/is-equal.d.ts
```diff
@@ -0,0 +1,21 @@
+/**
+Returns a boolean for whether the two given types are equal.
+
```
sindresorhus, line 3: It would be useful to list some use-cases for this type.

source/is-equal.d.ts
```diff
@@ -0,0 +1,21 @@
...
+@link https://github.com/microsoft/TypeScript/issues/27024#issuecomment-421529650
+@link https://stackoverflow.com/questions/68961864/how-does-the-equals-work-in-typescript/68963796#68963796
+
+@example
+```
+import type {IsEqual} from 'type-fest';
+
+type Expect<T extends true> = T
+type Case = Expect<IsEqual<Uppercase<"sample">, "SAMPLE">>
```
sindresorhus, line 12:
```suggestion
type Expect<T extends true> = T;
type Case = Expect<IsEqual<Uppercase<'sample'>, 'SAMPLE'>>;
```

sindresorhus: I feel like the example is incomplete without also showing how `Case` is used.

sindresorhus (changes requested):

source/is-equal.d.ts
doz13189: Thank you for your review. May I ask your opinion on which one is a good sample ?

A is the example I'm using. It's actually a little more complicated. The purpose is to check if the processed type is as expected.

```ts
// A
type Expect<T extends true> = T;
const ExpectTrue:Expect<IsEqual<Uppercase<'sample'>, 'SAMPLE'>> = true
```

B is a sample for many. I think Equal is generally used in types.

```ts
// B
type Includes<Value extends readonly any[], Item> =
	Value extends readonly [Value[0], ...infer rest]
		? IsEqual<Value[0], Item> extends true
			? true
			: Includes<rest, Item>
		: false;
```

sindresorhus: 👍

sindresorhus:
https://github.com/sindresorhus/type-fest/pull/522/files#r1033002320 still needs to be done

source/is-equal.d.ts
```diff
@@ -0,0 +1,37 @@
+/**
+Returns a boolean for whether the two given types are equal.
+
+@link https://github.com/microsoft/TypeScript/issues/27024#issuecomment-421529650
+@link https://stackoverflow.com/questions/68961864/how-does-the-equals-work-in-typescript/68963796#68963796
+
+@example
+```
+type ExpectTrue = IsEqual<1, 1>
```
sindresorhus, line 9: Each example should have the `import` statement so it can be copy-pasted.

source/is-equal.d.ts
```diff
@@ -0,0 +1,37 @@
...
+```
+type ExpectFalse = IsEqual<2, 1>
+//=> false
+```
+
+@example
+```
+import type {IsEqual} from 'type-fest';
+
+type Includes<Value extends readonly any[], Item> =
```
sindresorhus, line 23: Needs a code comment on what the type does.

source/is-equal.d.ts
```diff
@@ -0,0 +1,45 @@
...
+type ExpectTrue = IsEqual<1, 1>;
+//=> true
+```
+
+@example
+```
+import type {IsEqual} from 'type-fest';
+
+// `IsEqual` compares 2 and 1, then gets false as a result.
+type ExpectFalse = IsEqual<2, 1>;
```
sindresorhus, line 21: I don't really see the point of this and the previous example. No one would actually do this.

doz13189: I thought these examples would help users understand, but you are right, these examples will not be used. Therefore, I have removed the examples.

sindresorhus:
https://github.com/sindresorhus/type-fest/pull/522/files#r1033002320 still needs to be done

doz13189:
> https://github.com/sindresorhus/type-fest/pull/522/files#r1033002320 still needs to be done

Sorry, I misread.
https://github.com/sindresorhus/type-fest/pull/522/commits/401ad60676f4fb8b3c7770eeaaa7dc6b775ff201

trevorade:
FYI

There are two self-identity bugs that can pretty simply be fixed: [TS Playground](https://www.typescriptlang.org/play?#code/PQgEGUAsHsHcGMCGBnAps0qBOXpY7JKgHagCuxAltFcQOah6iXEAu2a8rLDrRoyRAFtUoVgE8ADqgCwAKHkBveaDFTRASWQBRAI5lEAGwA8AQQA0oAEIA+UAF4Vq0AApjAcRsuAlA7vvMAA92YgATDFNQAH5QAEZQAC5QACZfVGCScKdVN08fP1AA9JDw62i4xJTvbPLWLDJZOWdKgDMjNABueSdFNWlMfXaHUC09AxNexCT4gF9QAB9QSenQGctlipm7GacwZoA9KO6mpb7RVEHDDHsRnUvjDdnQADJTqc310HfZ7d3gZtAh3kOwUcjAADFKIExPxFkg0EonBJ+qN7hZrHZHCcch4vL57P4giUIgtCuV4klUkTMsgark8QUihkwhgrKSAjEKVUajE6g0akk2ldUF1QapesjzpdrrcxkYHl8VnNFo9Vp9vqtfic9s4gWLTpKBuMZajxgqNXNXqq1m8lVrVDrVHqQfIIVCYaJXvDGsoTobTfL0bYHHTcfkCWTijTQJFXhyKpS0systjXGH8YSoyyynHyZVUjyxPVGs1Be0RcdxWcjUMbgGJorNqTreq7as-gC9T1qxdjcN6+alS9bR8Rz929r-gcjnIXWD-pDoXxPV8wqTvbS5L7VP67mag5jQ3kM5HkyTc4t41yqVmU816eHM2ec2TL3nE4W+SXnGXhaLu4ava1rK9zWs2jazK2mz2qAjqAjOAHqDWVz9nu8pgVaEFqmOmoTg6U66jOIJAA)

Both of these evaluate to false:
```ts
IsEqual<{a: 1} & {a: 1}, {a: 1}>
IsEqual<{a: 1} | {a: 1}, {a: 1}>
```

The fix is to expand the check a touch:
```ts
type IsEqual<A, B> =
    ? true
    : false;
```

---

There's also this bug but I'm unaware of a fix: [TS Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAkgzgUQI4FcCGAbAPAQQDRQBCAfFALwCwAUFFABRYDixdAlOaY1BAB7AQA7ACZwoOKAH4oARigAuKACZ2vfsLjVaDZmw5QuqwSKKSZ8pa02ngAJxQQrCgGaY4EANzVqAbyihI3Kiu5LCIQdi+aAqyAL5QAGRQvgBGCopQMQSR0QSpShmkMdQA9MW05QB6EtRAA)

This evaluates to false:

```ts
IsEqual<{a: 1} & {b: 2}, {a: 1, b: 2}>
```

sindresorhus:
@trevorade Would you mind moving this to an issue instead? Otherwise it will be forgotten.

trevorade:
> @trevorade Would you mind moving this to an issue instead? Otherwise it will be forgotten.

Created https://github.com/sindresorhus/type-fest/issues/537

PR #523, opened 2022-11-28
Make `PartialDeep` affect `ReturnType` of functions.

daomtthuan:
Add option in `PartialDeepOptions`: `recurseIntoFunctions`
Make `PartialDeep` affect `ReturnType` of functions.

E.g:
image

### DIFF
--- a/source/partial-deep.d.ts
+++ b/source/partial-deep.d.ts
@@ -10,8 +10,7 @@ export type PartialDeepOptions = {
 	@default false
 	*/
 	readonly recurseIntoArrays?: boolean;
-	
-	  
+
 	/**
 	Whether to affect the return type of functions.
 
--- a/source/partial-deep.d.ts
+++ b/source/partial-deep.d.ts
@@ -12,7 +12,7 @@ export type PartialDeepOptions = {
 	readonly recurseIntoArrays?: boolean;
 	
 	  
-  	/**
+	/**
 	Whether to affect the return type of functions.
 
 	@default false
--- a/source/partial-deep.d.ts
+++ b/source/partial-deep.d.ts
@@ -10,6 +10,14 @@ export type PartialDeepOptions = {
 	@default false
 	*/
 	readonly recurseIntoArrays?: boolean;
+	
+	  
+  	/**
+	Whether to affect the return type of functions.
+
+	@default false
+	*/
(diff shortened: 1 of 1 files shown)


papb:
Why just the return type, and not the arguments?

daomtthuan:
@papb 
Is your mean like this?
```typescript
const test: PartialDeepTestType = {
  doSomething: (a: string | undefined, b: number | undefined) => ({
    x: 1,
  }),
};
```
In this case, I just want make Partial ReturnType.
If contributors eccept this solution for Partial ReturnType, I will expend scope for arguments.

papb:
Sorry, my question was too summarized and ended up unclear. Let me rephrase in two new questions.

1. What is the motivation for your situation? Can you show a real example in which you would like to modify a function type so that its return value was made partial?
2. Do you see value in performing PartialDeep on the arguments as well? For example, transforming `({ a: number, b: number }) => void` into `({ a?: number, b?: number }) => void`?

If the option is going to be called `recurseIntoFunctions`, I would expect it to apply for arguments and return types. Another option could be to add two separate options, `recurseIntoFunctionReturnTypes` and `recurseIntoFunctionArguments`.

daomtthuan:
@papb 
In my case.
If 1 one object has property is `function` that has return type is `object`

```ts
type ABC = {
  doSomething: (arg1: number, arg2: string) => { result1: number; result2: string; result3: boolean };
};

type PartialDeepABC = PartialDeep<ABC>;
```

`PartialDeep` just makes type `ABC` can has `doSomething` or not
And if it has `doSomething`, the return type of `doSomething` must be `{ result1: number; result2: string; result3: boolean }`

```ts
const xyz: PartialDeepABC = {
  doSomething: (arg1, arg2) => {
    return { result1: arg1, result2: arg2, result3: true };
  },
};
```

But I want partial deep into that **return object**, `{ result1?: number; result2?: string; result3?: boolean }`
Like this:

```ts
const xyz: PartialDeepABC = {
  doSomething: (arg1, arg2) => {
    if (arg1 === 1) {
      return { result1: 123 };
    }

    if (arg2 === 'bar') {
      return { result2: 'foo' };
    }

    return { result3: false };
  },
};
```

I agree with you to change `recurseIntoFunctions` to `recurseIntoFunctionReturnTypes`

PR #525, opened 2022-11-30
fix(Exact): fix support for opaque type

### LINKED ISSUE #508: `Exact` does not work with `Opaque` after 2.19.0
`Exact` no longer works when there are any object properties which have a type created using `Opaque`. This used to work using version 2.18.0 and lower.

```ts
type SpecialName = Opaque<string, "special name">;

type OnlyAcceptName = {
  name: SpecialName;
};

function onlyAcceptNameImproved<T extends Exact<OnlyAcceptName, T>>(args: T) {}

onlyAcceptNameImproved({
  // Error: Type 'SpecialName' is not assignable to type 'never'
  name: "name" as SpecialName
});
```

Example using type-fest 2.18.0: https://codesandbox.io/s/twilight-shape-bc391p?file=/src/index.ts

Same example using latest version: https://codesandbox.io/s/boring-morse-96egn3?file=/src/index.ts

zorji:
Fixes #508


sindresorhus:
The workaround makes sense to me. Can you fix the lint issues?

zorji:
Hi @sindresorhus 

Sorry for the late response and thanks for reviewing this. I am not very familiar with Opaque type and not sure what is the right use case for it.

I've added a test for string and number and fixed lint issues.

PR #529, opened 2022-12-15
Update TsConfigJson.CompilerOptions to latest TypeScript

Septh:
This PR adds:
- [preserveValueImports](https://www.typescriptlang.org/tsconfig#preserveValueImports) - Added in TS 4.5
- [moduleSuffixes](https://www.typescriptlang.org/tsconfig#moduleSuffixes) - Added in TS 4.7
- [moduleDetection](https://www.typescriptlang.org/tsconfig#moduleDetection) - Added in TS 4.7

### DIFF
--- a/source/tsconfig-json.d.ts
+++ b/source/tsconfig-json.d.ts
@@ -1027,7 +1027,7 @@ declare namespace TsConfigJson {
 
 		Requires TypeScript version 4.5 or later.
 		*/
-		moduleSuffixes?: string[]
+		moduleSuffixes?: string[];
 
 		/**
 		Control what method is used to detect module-format JS files.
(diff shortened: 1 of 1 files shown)


source/tsconfig-json.d.ts
```diff
@@ -997,6 +1012,31 @@ declare namespace TsConfigJson {
...
+		List of file name suffixes to search when resolving a module.
+
+		Requires TypeScript version 4.5 or later.
+		*/
+		moduleSuffixes?: string[];
+
+		/**
+		Control what method is used to detect module-format JS files.
+
+		Requires TypeScript version 4.7 or later.
```
sindresorhus, line 1035: `type-fest` requires TS 4.7, so this is moot.

sindresorhus: Because the only way to get this type is to already be on TS 4.7, so the message of requiring TS 4.7 is not needed.

PR #530, opened 2022-12-15
Tweak documentation/comments/organization to mention npm workspaces

### LINKED ISSUE #510: Standard npm package.json `workspaces` field is missing from PackageJson type
Workspace setup from Yarn have been covered in #77 but now NPM itself has a workspaces field which is not covered in the `PackageJson` type: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#workspaces

bmish:
* Link to both [npm Workspaces](https://docs.npmjs.com/cli/using-npm/workspaces) and [Yarn Workspaces](https://classic.yarnpkg.com/docs/workspaces/).
* Mention that `nohoist` is [supported](https://classic.yarnpkg.com/blog/2018/02/15/nohoist/) by Yarn, but [not supported](https://github.com/npm/rfcs/issues/287) by npm.
* Move `workspaces` type from `YarnConfiguration` to `PackageJsonStandard` since it's supported by both npm and yarn now. There should be no behavior change from the resulting `PackageJson` type. The `YarnConfiguration` type has a potentially breaking change but I'm not sure that's considered a public type?

Fixes #510 (although nothing was actually broken to my knowledge, the resulting `PackageJson` type is the same).

Follow-up to the original change to add `workspaces` in #77.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -237,7 +237,7 @@ Click the type names for complete docs.
 ### Miscellaneous
 
 - [`GlobalThis`](source/global-this.d.ts) - Declare locally scoped properties on `globalThis`.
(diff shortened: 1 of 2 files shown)


klh:
you should update this to explain that npms nohoist === npm i --install-strategy=nested

PR #536, opened 2023-01-02
Add missing ES2022 to CompilerOptions.target

Septh:
Can't believe this was missing.

### DIFF
--- a/test-d/tsconfig-json.ts
+++ b/test-d/tsconfig-json.ts
@@ -1,4 +1,4 @@
-import {expectType, expectAssignable} from 'tsd';
+import {expectType} from 'tsd';
 import type {TsConfigJson} from '../index';
 
 const tsConfig: TsConfigJson = {};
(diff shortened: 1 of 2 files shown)


test-d/tsconfig-json.ts
```diff
@@ -11,3 +11,37 @@ expectType<string[] | undefined>(tsConfig.files);
...
+	es2015: 'es2015',
+	es2016: 'es2016',
+	es2017: 'es2017',
+	es2018: 'es2018',
+	es2019: 'es2019',
+	es2020: 'es2020',
+	es2021: 'es2021',
+	es2022: 'es2022',
+	esnext: 'esnext',
+});
```
sindresorhus, line 47: I don't see the point of this test. You're just duplicating the type. It would not catch any mistakes.

Septh: Indeed. Will update shortly.

PR #538, opened 2023-01-09
Fix bug with split words

bconnorwhite:
SplitWords seems to have a bug when casing changes from uppercase to lowercase after a separator.

For example, these tests both fail - the type resolves to `never`:
```ts
expectType<SplitWords<'hello WORLD lowercase'>>(['hello', 'WORLD', 'lowercase']);
expectType<SplitWords<'hello WORLD-lowercase'>>(['hello', 'WORLD', 'lowercase']);
```

Removing the carry over of `LastCharacter` after a delimiter should fix this

### DIFF
--- a/source/split-words.d.ts
+++ b/source/split-words.d.ts
@@ -33,7 +33,7 @@ export type SplitWords<
 > = Sentence extends `${infer FirstCharacter}${infer RemainingCharacters}`
 	? FirstCharacter extends WordSeparators
 		// Skip word separator
-		? [...SkipEmptyWord<CurrentWord>, ...SplitWords<RemainingCharacters, LastCharacter>]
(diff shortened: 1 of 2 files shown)


skarab42 (approved):

sindresorhus:
@bconnorwhite Does your changes here and in the other pull request affect any of the exposed types? I'm asking as neither of these types are exported publicly. So not sure whether it needs a new release.

bconnorwhite:
@sindresorhus

#538 
- Fixes some cases where `CamelCase` would incorrectly resolve to `never`

#539
- `Trim` removes all whitespace, not just spaces

bconnorwhite:
`SplitString` (and as a result `CamelCase`) still doesn't use the full list of whitespace characters added by #539. I'll open a separate PR for it

PR #539, opened 2023-01-09
Fix issue with IsNumeric when input is whitespace

bconnorwhite:
I'm not actually sure if this is a bug or not, but it at least seems confusing. (Feel free to close if this is expected behavior).

Currently `IsNumeric` passes with whitespace:
```ts
IsNumeric<" ">; // true
IsNumeric<" \t\n\r\f\v">; // true
```
Instead I've just added a check to ensure that if the type extends \`${number}\`, it isn't purely whitespace.


sindresorhus:
I agree it's a bug, but I'm not sure the proposed solution is correct. This kinda feels like a TypeScript bug. Maybe worth opening an issue there first?

A fix for this will also need to come with some tests.

source/internal.d.ts
```diff
@@ -47,6 +47,8 @@ export type WordSeparators = '-' | '_' | ' ';
 
 export type StringDigit = '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9';
 
+type Whitespace = ' ' | '\t' | '\n' | '\r' | '\f' | '\v';
```
sindresorhus, line 50: There are a lot more whitespace characters in Unicode.

bconnorwhite: I added a few more using this page as a reference: https://en.wikipedia.org/wiki/Whitespace_character

bconnorwhite:
As a side note, because `SplitWords` parses one character at a time I don't think this actually causes any issues, but "1e2" also passes `IsNumeric`. There are probably three useful types here:
- IsNumeric - does the string pass `Number()`?
- IsNumber - IsNumeric, but after trimming spaces first
- IsDigit - Only allow 0-9
Right now only `SplitWords` is using `IsNumeric` though, so maybe not worth adding these types

sindresorhus:
> It sounds like this is intended, here is an issue in the TypeScript repo: https://github.com/microsoft/TypeScript/issues/46109

I don't think you read the whole thread because the last comment says it's something they want to fix: https://github.com/microsoft/TypeScript/issues/46109#issuecomment-930307987

source/internal.d.ts
```diff
@@ -109,10 +137,23 @@ Returns a boolean for whether the string is uppercased.
...
+	: T extends `${Whitespace}${infer Rest}`
+		? IsWhitespace<Rest>
+		: false;
+
 /**
 Returns a boolean for whether the string is numeric.
 */
-export type IsNumeric<T extends string> = T extends `${number}` ? true : false;
+export type IsNumeric<T extends string> = T extends `${number}`
+	? IsWhitespace<T> extends true
```
sindresorhus, line 153: Can you add a comment that says this is a workaround for https://github.com/microsoft/TypeScript/issues/46109#issuecomment-930307987 ?

source/internal.d.ts
```diff
@@ -47,6 +47,34 @@ export type WordSeparators = '-' | '_' | ' ';
...
+  | '\u2007'
+  | '\u2008'
+  | '\u2009'
+  | '\u200A'
+  | '\u2028'
+  | '\u2029'
+  | '\u202F'
+  | '\u205F'
+  | '\u3000'
+  | '\uFEFF';
```
sindresorhus: Super nitpicky, but I prefer the modern escape syntax: `\u{7A}`.

https://exploringjs.com/es6/ch_unicode.html#sec_escape-sequences

sindresorhus:
> IsNumeric - does the string pass Number()?

I don't think the `Number` behavior is useful. So we should just fix `isNumeric`.

> IsDigit - Only allow 0-9

Could be a useful type, but beyond the scope of this pull request. Should be called `IsASCIIDigit` as `IsDigit` would be assumed to include all Unicode digits.

bconnorwhite:
Updated the escape formats.

Also realized that we need to use Trim, otherwise something like this passes:
```ts
IsNumeric<" 1 ">
```
Trim isn't currently accounting for all whitespace characters, so I updated Trim to use the new Whitespace type in `internal.d.ts`, which should also more accurately match `String.prototype.trim`.

PR #544, opened 2023-01-14
Add `IsUnion` type

### LINKED ISSUE #236: Proposal: `IsUnion`
Inspired by - https://stackoverflow.com/a/53955431/5039290

```ts
type IsUnion<T> = [T] extends [UnionToIntersection<T>] ? false : true;
```

**Use case**

Create a specific type that guarantees only unions from a generic type that could be mixed-bag.

```ts
type MemberInfo = {
  first_name: string;
  last_name: string;
  date_of_birth: number;
  favourite_fruit: 'apple' | 'mango' | 'banana';
  industry: 'tech' | 'art' | 'medicine';
}

type ChoiceBasedField = {
  [K in keyof MemberInfo]: IsUnion<MemberInfo[K]> extends true ? K : never
}[keyof RespondentFields];

// can accept 'favourite_fruit' or 'industry', but errors on anything else
const renderDropdown = (field: ChoiceBasedField) => (
  <select> ... </select>
);
```

trevorade:
Permits detecting if a given type is a singular type or a union of types. Useful when you only want to accept a union of types or a single type.

This commit uses a pattern where, instead of returning `true` and `false`, the caller can specify the result in the true and false cases. Effectively, you can write:

```ts
IsUnion<T, IfTrue, IfFalse>
```

instead of:

```ts
IsUnion<T> extends true ? IfTrue : IfFalse
```

Fixes #236

### DIFF
--- a/source/is-union.d.ts
+++ b/source/is-union.d.ts
@@ -1,0 +1,24 @@
+import type {UnionToIntersection} from './union-to-intersection';
+
+/**
+Detects whether a given type is a union of types.
+
+While you can default to returning a boolean based on the condition, the resulting typing will be simpler if you instead specify types for True and False.
+
+There are some types that are themselves unions such as `boolean` and enum types.
+
+Use-cases:
+- You want to ensure a given type is not a union but a singular type.
+
+@example
+```
+import type {IsUnion} from 'type-fest';
+
+// If `MaybeSingle` is a union type, resolves to `never`, otherwise MaybeSingle.
+type EnsureSingleType<MaybeSingle> = IsUnion<MaybeSingle, never, MaybeSingle>;
+```
+
+@category Utilities
+*/
+export type IsUnion<MaybeUnion, True = true, False = false> =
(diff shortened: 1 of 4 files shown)


source/is-union.d.ts
```diff
@@ -0,0 +1,24 @@
...
+
+Use-cases:
+- You want to ensure a given type is not a union but a singular type.
+
+@example
+```
+import type {IsUnion} from 'type-fest';
+
+// If `MaybeSingle` is a union type, resolves to `never`, otherwise MaybeSingle.
+type EnsureSingleType<MaybeSingle> = IsUnion<MaybeSingle, never, MaybeSingle>;
```
sindresorhus, line 18: I would like a more real-world example.

sindresorhus (changes requested):

trevorade:
Opting to drop this. I do use this in the codebase I work on but I'm fine keeping local there.

PR #555, opened 2023-02-17
fix(PackageJson): ensure `Imports` and `Exports` reflect Node.js expectations

Xunnamius:
Seems my previous PR was incomplete. While reviewing the Node.js source, I noticed a couple more issues. Namely: the `Imports` and `Exports` types should both accept fallback arrays containing exports conditions objects with optional properties. Also, the `Imports` type should accept everything the `Exports` type accepts (with the added constraint that entry points must start with `"#"`).

This PR addresses these and adds tests.

### DIFF
--- a/source/package-json.d.ts
+++ b/source/package-json.d.ts
@@ -224,7 +224,10 @@ declare namespace PackageJson {
 	string
 	>;
 
-	type ExportConditions = {[condition in ExportCondition]: Exports};
+	/**
+	A mapping of conditions and the paths to which they resolve.
+	*/
+	type ExportConditions = {[condition in ExportCondition]?: Exports};
 
 	/**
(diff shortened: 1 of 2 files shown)


source/package-json.d.ts
```diff
@@ -224,7 +224,10 @@ declare namespace PackageJson {
 	string
 	>;
 
-	type ExportConditions = {[condition in ExportCondition]: Exports};
+	/**
+	A mapping of conditions and the paths to which they resolve.
+	*/
+	type ExportConditions = {[condition in ExportCondition]?: Exports};
```
privatenumber, line 230: The `?` here introduces `undefined` as a valid exports value in the type, which is not possible by JSON.

PR #560, opened 2023-02-28
Enhance `Except` type

### LINKED ISSUE #556: `Except` is not strict
```ts
import { Except } from "type-fest";

type Foo = {
    a: number;
    b: string;
    c: boolean;
};

type FooWithoutA = Except<Foo, 'a'>;

const foo: Foo = {
    a: 1,
    b: 'b',
    c: true
}

const fooWithoutA: FooWithoutA = foo; // not error
```

[playground link](https://www.typescriptlang.org/play?ts=4.9.5&ssl=27&ssc=37&pln=1&pc=1#code/C4TwDgpgBAkgzgUQI4FcCGAbAPAQQDRQBCAfFALwBQU1UAFFgOLG0CU5pDUEAHsBAHYATOFBxQA-FACMUAFxQATGx58hcKjXpNW7KJxUDhRCdLmKWG6pOAAnFBEtmAZpjgQA3BQqhIUAGIAlhh8NlgA0hAgACrgEAQI3ADGGCiCEDGQpGSwiKiY4ZEZcVAJyanpsaQGalC29ib8EABuEDZmtBHRsVy8hiKlKWlFDc2tZp1FLJ7e3aUQYMBYAPIARgBWEInARQSdcMPVRgDWkQD2TlCrG1tFWVAA3o4A2hPdAfxQJyDnl+ub2900CJAsFWgUupBdpF9pUALryK7-IovQqxWGeAC+0x80D8p1O5AejjQ8n4KAAtitWp4aFAVvI4LZ3gBzGk0RLyFb4jAQND8THY7p404AdQCwAAFqcUMAxNk5gssMKCAByNAq4jTRKnfiMqBOfHyYWEx60knSPCOelQFUrFWW2kc2p2BwYrza3XAfX4sWS6Wyo0+8VSmVy72nIA)

Is this intended behavior? if yes, how can i restrict `foo` for being assign to `fooWithoutA`?
Thank you.

orimiles5:
Closes #556.

Hello There!

I would like to introduce a PR that enhances the `Except` type to be more strict. Although the updated type remains non-strict by default, users have the option to use the strict type by setting the `strict` option to `true`. The rationale for this change was in response to the feedback provided in issue #556, which requested a stricter version of the `Except` type.

Additionally, I can make further modifications to the types that use the `Except` type, if you would like.

To assist with the adoption of the updated type, I have included some examples in the JSDoc. If you have any questions or feedback about this change, please do not hesitate to let me know.

Thank you for your time and consideration.


sindresorhus:
// @pheis @tommy-mitchell @rayrw Would any of you be willing to help review this?

readme.md
```diff
@@ -902,4 +902,4 @@ You can find some examples in the [TypeScript docs](https://www.typescriptlang.o
 
 ## License
 
-SPDX-License-Identifier: (MIT OR CC0-1.0)
+SPDX-License-Identifier: (MIT OR CC0-1.0)
```
sindresorhus, line 905: Don't make unrelated changes.

source/except.d.ts
```diff
@@ -36,22 +45,70 @@ This type is a stricter version of [`Omit`](https://www.typescriptlang.org/docs/
...
+*/
+type StrictExcept<ObjectType, KeysType extends keyof ObjectType> = NonStrictExcept<ObjectType, KeysType> & Partial<Record<KeysType, never>>;
+
+/**
+The Except type is the exported type, which determines the appropriate method to use (NonStrictExcept or StrictExcept) based on the options provided as the third argument (which is set to false by default).
+
 @example
 ```
-import type {Except} from 'type-fest';
+import {Except} from 'type-fest';
```
sindresorhus, line 73: Don't make unrelated changes.

sindresorhus:
@orimiles5 Can you think of any reason to want the non-strict version? If not, we could plan to remove the option and default to strict in the next major version.

tommy-mitchell:
@sindresorhus are you suggesting keeping the default as non-strict in this PR, with an option to be strict? Or that the this PR should have an option to be non-strict, and that the next version will remove that option?

source/except.d.ts
```diff
@@ -25,9 +25,18 @@ type Filtered = Filter<'bar', 'foo'>;
...
+type Filter<KeyType, ExcludeType> =
+IsEqual<KeyType, ExcludeType> extends true ? never :
+	KeyType extends ExcludeType ? never : KeyType;
+
+type ExceptOptions = {
+	/**
+	@default false
+	*/
+	strict?: boolean;
+};
```
tommy-mitchell, line 39: Could add a description to the option:

```suggestion
type ExceptOptions = {
	/**
	Disallow assigning non-specified properties.

    Setting this to `false` is not recommended. Included for backwards-compatability.

	@default false
	*/
	strict?: boolean;
};
```

source/except.d.ts
```diff
@@ -25,9 +25,18 @@ type Filtered = Filter<'bar', 'foo'>;
 //=> 'bar'
 ```
 
-@see {Except}
+@see {NonStrictExcept}
 */
-type Filter<KeyType, ExcludeType> = IsEqual<KeyType, ExcludeType> extends true ? never : (KeyType extends ExcludeType ? never : KeyType);
+type Filter<KeyType, ExcludeType> =
+IsEqual<KeyType, ExcludeType> extends true ? never :
+	KeyType extends ExcludeType ? never : KeyType;
```
tommy-mitchell, line 32: Unrelated change.

```suggestion
@see {Except}
*/
type Filter<KeyType, ExcludeType> = IsEqual<KeyType, ExcludeType> extends true ? never : (KeyType extends ExcludeType ? never : KeyType);
```

source/except.d.ts
```diff
@@ -36,22 +45,70 @@ This type is a stricter version of [`Omit`](https://www.typescriptlang.org/docs/
...
+	a: 1,
+	b: 'b',
+	c: true,
 };
+
+const fooWithoutA: FooWithoutA = foo; // Error
+//=> StrictExcept<Foo, 'a'>;
+*/
+export type Except<ObjectType, KeysType extends keyof ObjectType, Options extends ExceptOptions = {strict: false}> =
+Options['strict'] extends false ? NonStrictExcept<ObjectType, KeysType> : StrictExcept<ObjectType, KeysType>;
```
tommy-mitchell, line 114: No need for an extra type, we can just inline the solution:

```suggestion
/**
Create a type from an object type without certain keys.

Use option `strict: true` to disallow assigning additional properties to this type.

This type is a stricter version of [`Omit`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-5.html#the-omit-helper-type). The `Omit` type does not restrict the omitted keys to be keys present on the given type, while `Except` does. The benefits of a stricter type are avoiding typos and allowing the compiler to pick up on rename refactors automatically.

This type was proposed to the TypeScript team, which declined it, saying they prefer that libraries implement stricter versions of the built-in types ([microsoft/TypeScript#30825](https://github.com/microsoft/TypeScript/issues/30825#issuecomment-523668235)).

@example
```
import type {Except} from 'type-fest';

type Foo = {
	a: number;
	b: string;
};

type FooWithoutA = Except<Foo, 'a'>;
//=> {b: string}

const fooWithoutA: FooWithoutA = {a: 1, b: '2'};
//=> errors: 'a' does not exist in type '{ b: string; }'

type FooWithoutB = Except<Foo, 'b', {strict: true}>;
//=> {a: number} & Partial<Record<"b", never>>

const fooWithoutB: FooWithoutB = {a: 1, b: '2'};
//=> errors: Type 'string' is not assignable to type 'undefined'.
```

@category Object
*/
export type Except<ObjectType, KeysType extends keyof ObjectType, Options extends ExceptOptions = {strict: false}> = {
	[KeyType in keyof ObjectType as Filter<KeyType, KeysType>]: ObjectType[KeyType];
} & (Options['strict'] extends true
	? Partial<Record<KeysType, never>>
	: {});
```

tommy-mitchell, line 114: No need for extra types, we can just inline:
```suggestion
@example
```
import type {Except} from 'type-fest';

type Foo = {
	a: number;
	b: string;
};

type FooWithoutA = Except<Foo, 'a'>;
//=> {b: string}

const fooWithoutA: FooWithoutA = {a: 1, b: '2'};
//=> errors: 'a' does not exist in type '{ b: string; }'

type FooWithoutB = Except<Foo, 'b', {strict: true}>;
//=> {a: number} & Partial<Record<"b", never>>

const fooWithoutB: FooWithoutB = {a: 1, b: '2'};
//=> errors: Type 'string' is not assignable to type 'undefined'.
```

@category Object
*/
export type Except<ObjectType, KeysType extends keyof ObjectType, Options extends ExceptOptions = {strict: false}> = {
	[KeyType in keyof ObjectType as Filter<KeyType, KeysType>]: ObjectType[KeyType];
} & (Options['strict'] extends true
	? Partial<Record<KeysType, never>>
	: {});
```

tommy-mitchell: Wouldn't allow me to make a suggestion there, but in the `Except` doc comment:

```diff
/**
Create a type from an object type without certain keys.

+ Use option `strict: true` to disallow assigning additional properties to this type. 

// ...
*/
```

test-d/except.ts
```diff
@@ -11,7 +11,7 @@ type Example = {
 	bar: string;
```
tommy-mitchell, line 11: Add test cases for `strict: true`:

```diff
-import {expectType} from 'tsd';
+import {expectType, expectError} from 'tsd';
import type {Except} from '../index';

declare const except: Except<{a: number; b: string}, 'b'>;
expectType<{a: number}>(except);
+expectError(except.b);
+
+const nonStrict = {
+	a: 1,
+	b: '2',
+};
+
+const nonStrictAssignment: typeof except = nonStrict; // No error
+
+declare const strictExcept: Except<{a: number; b: string}, 'b', {strict: true}>;
+
+expectError(() => {
+	const strictAssignment: typeof strictExcept = nonStrict;
+});

// ...
```

tommy-mitchell: @sindresorhus this will fail because of `ts6133`:

```sh
> npx tsd

  test-d/except.ts:13:6
  ✖  13:6  nonStrictAssignment is declared but its value is never read.  
  ✖  18:7  strictAssignment is declared but its value is never read.     

  2 errors
```

Should we configure `tsd` in `package.json` with

```json
"tsd": {
	"compilerOptions": {
		"noUnusedLocals": true
	}
}
```

Or should `tsd` ignore that error by default?

tommy-mitchell:
> Additionally, I can make further modifications to the types that use the Except type, if you would like.

Which types use `Except`?

tommy-mitchell (changes requested):
I think this PR could be simpler.

sindresorhus:
> are you suggesting keeping the default as non-strict in this PR, with an option to be strict? Or that the this PR should have an option to be non-strict, and that the next version will remove that option?

Making it `{strict: true}` by default is a breaking change, so the option is required for now. I'm asking whether there are any good use-cases for having it as `false`. If not, we can consider switching it to `true` by default in the next major release and then removing the option.

rayrw (commented):
I personally think the definition of `strictness` in this type could be more than one depending on which criteria do we prioritize (check if assignable v.s. accessible).

Yes, using the current `Except` type cannot prevent us from assigning objects with excess properties because of TypeScript's design decision. However, The default behavior actually can check if we can access to that property.

Taking the same `fooWithoutA` variable in the original issue https://github.com/sindresorhus/type-fest/issues/556 as an example. When we try to access `fooWithoutA.a`, we get this error.
```
Property 'a' does not exist on type 'Except<Foo, "a">'.(2339)
```

With the `strict` option in this PR turned on, we cannot assign `foo` to `fooWithoutA` anymore. However, this comes with a trade off. We have to make `a` accessible. `fooWithoutA.a` is now valid and is shown in auto-complete.

Screen Shot 2023-03-11 at 22 20 06

Screen Shot 2023-03-11 at 22 12 30

So far I cannot find a perfect solution that can prevent both, but I wouldn't say either one is wrong. At the time being, maybe an explicit option like `makeExceptPropertiesUndefined` is more accurate? What do you think?

readme.md
tommy-mitchell: I believe the trailing new line was changed.

test-d/except.ts
sindresorhus: I think `tsd` should ignore it.

sindresorhus:
I think the option should have a specific name for what exactly it's enabling. `strict` is confusing as the type is already a strict version of `Omit`.

sindresorhus:
> With the strict option in this PR turned on, we cannot assign foo to fooWithoutA anymore. However, this comes with a trade off. We have to make an accessible. fooWithoutA.a is now valid and is shown in auto-complete.

The trade-off should be documented.

source/except.d.ts
```diff
@@ -29,9 +29,22 @@ type Filtered = Filter<'bar', 'foo'>;
...
+
+	@default false
+	*/
+	requireExactProps?: boolean;
+};
+
 /**
 Create a type from an object type without certain keys.
 
+Use option `requireExactProps: true` to disallow assigning additional properties to this type. Notice that any omitted properties in the resulting type will be present as `undefined`.
```
tommy-mitchell, line 46:
```suggestion
Use option `requireExactProps: true` to disallow assigning additional properties to this type. Note that any omitted properties in the resulting type will be present in suggestions as `undefined`.
```

tommy-mitchell: We could also add a note like this to the readme:

---

- [`Exact`](source/exact.d.ts) - Create a type that does not allow extra properties. Use option `requireExactProps: true` to disallow assigning additional properties to this type. (Note that that any omitted properties in the resulting type will be present in suggestions as `undefined`.)

---

Which means line 46 could probably just be appended to line 44:

```ts
/**
Create a type from an object type without certain keys. Use option `requireExactProps: true` to disallow assigning additional properties to this type. Notice that any omitted properties in the resulting type will be present as `undefined`.

This type is a stricter...

...
*/
```

readme.md
```diff
@@ -129,7 +129,8 @@ Click the type names for complete docs.
 
 - [`EmptyObject`](source/empty-object.d.ts) - Represents a strictly empty plain object, the `{}` value.
 - [`IsEmptyObject`](source/empty-object.d.ts) - Returns a `boolean` for whether the type is strictly equal to an empty plain object, the `{}` value.
-- [`Except`](source/except.d.ts) - Create a type from an object type without certain keys. This is a stricter version of [`Omit`](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys).
+- [`Except`](source/except.d.ts) - Create a type from an object type without certain keys. Use option `requireExactProps: true` to disallow assigning additional properties to this type. (Note that that any omitted properties in the resulting type will be present in suggestions as `undefined`.)
```
sindresorhus, line 132: The description here should not be changed. The option is explained in the type docs. That's enough.

source/except.d.ts
```diff
@@ -29,8 +29,19 @@ type Filtered = Filter<'bar', 'foo'>;
...
+	Setting this to `false` is not recommended. Included for backwards-compatability.
+
+	@default false
+	*/
+	requireExactProps?: boolean;
+};
+
 /**
-Create a type from an object type without certain keys.
+Create a type from an object type without certain keys. Use option `requireExactProps: true` to disallow assigning additional properties to this type. Note that any omitted properties in the resulting type will be present in suggestions as `undefined`.
```
sindresorhus, line 44: This text should be in the docs for `ExceptOptions` instead. The only text we need here is just:

> We recommend setting the `requireExactProps` option to `true`.

readme.md
sindresorhus: @orimiles5 This still needs to be fixed.

sindresorhus (changes requested):

PR #562, opened 2023-03-03
IsStringLiteral: add new util

### LINKED ISSUE #541: Feature request: `IsStringLiteral`
What do you think of adding a type that can say if a string is a string literal (like `'foo/bar'`) and not a regular `string`?

```ts
export type IsStringLiteral<T> = string extends T ? false : true;
```

Ayc0:
Fixes #541


sindresorhus:
Thanks for working on this. I may go for https://github.com/sindresorhus/type-fest/pull/563 as it implements all the utilities.

Issue #454, opened 2022-08-31
Proposal: `readme.md` builder with examples links to TS Playground


skarab42:
Following the hint made by @fregante in #428, I propose a readme generator for the API part. 

@sindresorhus  I did a POC and you can see the generated readme [there](https://github.com/skarab42/type-fest/blob/readme-builder-poc/readme.md#array). 

To generate the code I use the TS compiler so no additional dependencies to get the summaries, tags and examples, the only library that is added is [lz-string](https://github.com/pieroxy/lz-string) which is the same one that is used by the TS Playground to generate the share links.

In addition to generating the links for the examples, this would allow us to not worry about adding new types to the readme and always have a version that matches the code.

skarab42:
@sindresorhus Any thoughts on this?

sindresorhus:
I think it's a great idea 👍

1. I didn't check, but it would be nice if a type could appear in multiple categories if it has multiple categories defined.
2. You need to ensure that all the types sources use the description from the readme.
3. `source | example` should be capitalized
4. `Categories:` => `Categories`
5. Is it really useful to show the type count in the categories section? Feels like noise to me.
6. We should remove the `Template literal` category. It was only there previously because we had to document them as only supporting the new TS version.

tommy-mitchell:
@sindresorhus https://github.com/tommy-mitchell/type-fest/tree/feat-readme-builder

I've been working on expanding this, converting the scripts to TypeScript and using `liquidjs` for the readme template. I've also added `eslint-formatter-pretty` for formatting warnings while generating the readme:

<img width="496" alt="image" src="https://github.com/sindresorhus/type-fest/assets/1403701/dde0ded6-f1d9-4dde-abd7-815817b30a00">

Thoughts? Haven't changed anything with the type categories or added missing examples/summaries, etc.

sindresorhus:
@tommy-mitchell Looks great 👍

Nitpick: I would make it `TS Playground` => `Playground`

sindresorhus:
The code in the inline examples are missing indentation:

<img width="185" alt="Screenshot 2023-08-14 at 15 17 03" src="https://github.com/sindresorhus/type-fest/assets/170270/4c1df1f4-2299-4609-aa23-88128c09831a">

PR #563, opened 2023-03-06
feat: add `IsXLiteral` types

### LINKED ISSUE #339: Proposal: `IsLiteral`
Returns `T` if `T` is a literal type belonging to `U`.

```ts
type IsLiteral<T, U> = T extends U ? (U extends T ? never : T) : never;
```

```ts
type A = IsLiteral<1, number>;
//   ^? 1

type B = IsLiteral<number, number>;
//   ^? never
```

**Playground Link:** [Provided](https://www.typescriptlang.org/play?target=8#code/C4TwDgpgBAkgzgGQJbAgJwIYBsA8AVAGigFUA+KAXijyggA9UA7AEzhKgH536nXrOojCADd0UAFz9JQ0WgDcAKAWhIUAIKVYiFOmw4AjEUYBXALYAjdKUUB6G1AcA9DkpXQAQpvjJUmXCYt0IzNLNGsFOwcoZwUgA)

Not sure if this is useful. Maybe for certain high-level generic types?

### LINKED ISSUE #541: Feature request: `IsStringLiteral`

tommy-mitchell:
Closes #541, closes #339.

Adds types to check if a given type is a `Primitive` literal:

- `IsLiteral`
- `IsStringLiteral`
- `IsNumericLiteral`
- `IsBooleanLiteral`
- `IsSymbolLiteral`
- ~`IsNullLiteral`~
- ~`IsUndefinedLiteral`~

Marking as draft as I still need to add documentation comments to the types. 

<details>
<summary>Info About Removed Types</summary>

I'm not sure if it's possible to check for `null` or `undefined` literals without also allowing their values, e.g:

```ts
import type {IsNullLiteral} from 'type-fest';

declare const _null: null;

type ShouldBeTrue = IsNullLiteral<null>;
//=> true

type ShouldBeFalse = IsNullLiteral<typeof _null>;
//=> true
```

Marking as draft as I still need to figure out the above issue and add documentation comments to the types. Currently the test cases for `_null` and `_undefined` values are commented out.
</details>

### DIFF
--- a/source/is-literal.d.ts
+++ b/source/is-literal.d.ts
@@ -2,7 +2,23 @@ import type {Primitive} from './primitive';
 import type {Numeric} from './numeric';
 import type {IsNever, IsNotFalse} from './internal';
 
-/** @link https://stackoverflow.com/a/52806744/10292952 */
+/**
+Returns a boolean for whether the given type `T` is the specified `LiteralType`.
+
+@link https://stackoverflow.com/a/52806744/10292952
+
+@example
+```
+LiteralCheck<1, number>
+//=> true
+
+LiteralCheck<number, number>
+//=> false
+
+LiteralCheck<1, string>
+//=> false
+```
+*/
 type LiteralCheck<T, LiteralType extends Primitive> = (
 	IsNever<T> extends false // Must be wider than `never`
 		? [T] extends [LiteralType] // Must be narrower than `LiteralType`
@@ -13,8 +29,25 @@ type LiteralCheck<T, LiteralType extends Primitive> = (
 		: false
 );
 
+/**
+Returns a boolean for whether the given type `T` is one of the specified literal types in `LiteralUnionType`.
+
+@example
+```
+LiteralChecks<1, Numeric>
+//=> true
+
+LiteralChecks<1n, Numeric>
+//=> true
+
+LiteralChecks<bigint, Numeric>
+//=> false
+```
+*/
 type LiteralChecks<T, LiteralUnionType> = (
-	// Conditional type to force union distribution
+	// Conditional type to force union distribution.
+	// If `T` is none of the literal types in the union `LiteralUnionType`, then `LiteralCheck<T, LiteralType>` will evaluate to `false` for the whole union.
+	// If `T` is one of the literal types in the union, it will evaluate to `boolean` (i.e. `true | false`)
 	IsNotFalse<LiteralUnionType extends Primitive
 		? LiteralCheck<T, LiteralUnionType>
 		: never
--- a/source/is-literal.d.ts
+++ b/source/is-literal.d.ts
@@ -1,37 +1,29 @@
 import type {Primitive} from './primitive';
-import type {Numeric} from './numeric';
-import type {IsNotFalse} from './internal';
+import type {IsNever, IsNotFalse} from './internal';
 
 /** @link https://stackoverflow.com/a/52806744/10292952 */
 type LiteralCheck<T, LiteralType extends Primitive> = (
-	[T] extends [never] // Must be wider than `never`
-		? false
-		: T extends LiteralType // Must be narrower than `LiteralType`
-			? LiteralType extends T // Cannot be wider than `LiteralType`
+	IsNever<T> extends false // Must be wider than `never`
+		? [T] extends [LiteralType] // Must be narrower than `LiteralType`
+			? [LiteralType] extends [T] // Cannot be wider than `LiteralType`
 				? false
 				: true
 			: false
+		: false
 );
 
-type LiteralChecks<T, Union> = Union extends Primitive ? IsNotFalse<LiteralCheck<T, Union>> : false;
+type LiteralChecks<T, LiteralUnionType> = (
+	// Conditional type to force union distribution
+	LiteralUnionType extends Primitive
+		? IsNotFalse<LiteralCheck<T, LiteralUnionType>>
+		: false
+);
 
 export type IsStringLiteral<T> = LiteralCheck<T, string>;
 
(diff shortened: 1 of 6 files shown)


tommy-mitchell:
Thinking about it some more, I’m not sure if there’s even really a reason to check for `null` or `undefined` literals, beyond completeness sake. And the more I think about it, I don’t think it’s possible to either.

source/is-literal.d.ts
```diff
@@ -0,0 +1,59 @@
...
+type IsLiteralTuple<T> = [
+	IsStringLiteral<T>,
+	IsNumericLiteral<T>,
+	IsBooleanLiteral<T>,
+	IsSymbolLiteral<T>,
+	IsNullLiteral<T>,
+	IsUndefinedLiteral<T>,
+];
+
+export type IsLiteral<T extends Primitive> = Includes<IsLiteralTuple<T>, true>;
```
sindresorhus, line 59: Do we really need to use `Includes` here? Every type slows down type checking and it seems like it could just be `|`.

sindresorhus:
See https://github.com/sindresorhus/type-fest/pull/562/files#diff-e50dff1575d9cd1e723da9f89c4f438d9b5e9c3120d93a41a8139a2f869d05e4R1 for inspiration for docs comments.

sindresorhus:
Lets just drop `null` and `undefined` literal. I cannot think of any use-case for them.

test-d/is-literal.ts
```diff
@@ -0,0 +1,76 @@
...
+declare const _string: string;
+declare const _number: number;
+declare const _bigint: bigint;
+declare const _boolean: boolean;
+declare const _symbol: symbol;
+declare const _null: null;
+declare const _undefined: undefined;
+
+// Literals should be true
+expectType<IsLiteral<typeof stringLiteral>>(true);
```
Ayc0, line 28: You should add a test for the type `string` and expect `false`

tommy-mitchell:
Should the category be `Utilities`?

source/is-literal.d.ts
tommy-mitchell: On second thought, we can be more clever:

```ts
type Intersection = true & false /* &  ... */;
//=> `never` (would be `false` without a `true`)

type IsValid = IfNever<Intersection>;
//=> `true`

// or

type Union = true | false /* | ... */;
//=> `boolean` (would be `false` without a `true`)

type IsValid = Y extends false ? false : true;
//=> `true`
```

In either case, `IsValid` will be `true` only if the type being tested contains `true` in it (e.g. not all `false`). I'd lean towards the latter, as it's simpler and involves unions, which feel more idiomatic.

source/is-literal.d.ts
```diff
@@ -0,0 +1,38 @@
...
+/** @link https://stackoverflow.com/a/52806744/10292952 */
+type LiteralCheck<T, LiteralType extends Primitive> = (
+	IsNever<T> extends false // Must be wider than `never`
+		? [T] extends [LiteralType] // Must be narrower than `LiteralType`
+			? [LiteralType] extends [T] // Cannot be wider than `LiteralType`
+				? false
+				: true
+			: false
+		: false
+);
```
tommy-mitchell, line 14: Once #564 lands, this will need to be updated to `import type {IsNever} from './is-never';`

tommy-mitchell:
Had to force push because I committed on the wrong user.

tommy-mitchell:
Currently, `IsLiteral` has a `Primitive` type constraint:

```ts
type IsLiteral<T extends Primitive> = // ...
```

This means that testing types that aren't assignable to `Primitive` (e.g. `unknown`) is a compiler error:

```ts
type ShouldBeFalse = IsLiteral<unknown>;
//=> Type 'unknown' does not satisfy the constraint 'Primitive'.
```

The other `IsXLiteral` types don't have this constraint. Should this be removed?

tommy-mitchell:
Seems like these aren't being used as type guards either. The following errors at `Capitalize<T>`:

```ts
type CapitalizedString<T> = IsStringLiteral<T> extends true ? Capitalize<T> : string;

// (type parameter) T in type CapitalizedString<T>
// Type 'T' does not satisfy the constraint 'string'. ts(2344)
```

Not sure what I'm doing wrong.

tommy-mitchell:
All types are now documented with examples.

sindresorhus:
> Should the category be Utilities?

👍 And maybe also a new one: `Is` or `Type Guard`?

source/is-literal.d.ts
```diff
@@ -0,0 +1,247 @@
+import type {Primitive} from './primitive';
+import type {Numeric} from './numeric';
+import type {IsNever, IsNotFalse} from './internal';
+
+/**
+Returns a boolean for whether the given type `T` is the specified `LiteralType`.
```
sindresorhus, line 6: Can you make the first line match the readme description for all the types? We'll eventually have some automation to generate the readme: https://github.com/sindresorhus/type-fest/issues/454

tommy-mitchell: `LiteralCheck` is an internal type helper. The public types have matching descriptions, I believe.

I left a comment on the other PR asking how we should word the description: https://github.com/sindresorhus/type-fest/pull/564#issuecomment-1459343557

sindresorhus:
> IsType?

Makes it sound like it's a special type, not a type of type. Let's go with `Type Guard` for now.

sindresorhus:
Looks great :)

PR #564, opened 2023-03-06
feat: add `IsX`/`IfX` types for `any`/`never`/`unknown`

### LINKED ISSUE #129: Proposal: `IfAny`, `IsAny`, `IsUnknown`, etc
https://github.com/sindresorhus/type-fest/pull/127/files#r474306457

Thoughts?

tommy-mitchell:
Closes #129.

Adds `IsAny`, `IfAny`, etc. types:

```ts
import type {IsAny, IfAny} from 'type-fest';

type ShouldBeTrue = IsAny<any>;
//=> true

type ShouldBeFalse = IfAny<''>;
//=> false

type ShouldBeNever = IfAny<any, never, 'not-never'>;
//=> never
```

The `IfX` types are defined using the pattern mentioned in https://github.com/sindresorhus/type-fest/issues/129#issuecomment-1451284276:

```ts
// source/if-any.d.ts

export type IfAny<T, TypeIfAny = true, TypeIfNotAny = false> = (
  IsAny<T> extends true ? TypeIfAny : TypeIfNotAny
);
```

~Marked draft as concrete use cases are needed for documenting the types.~

<details>
<summary>Info about a blocker that has been resolved</summary>

Marked draft as concrete use cases are needed for documenting the types, and a test can't pass without https://github.com/SamVerschueren/tsd/pull/173:

```ts
// test-d/if-any.ts

expectError<IfAny>(anything);

// reports:
// test-d\if-any.ts:25:12
//   ✖  25:0   Expected an error, but found none.
//   ✖  25:12  Generic type IfAny requires between 1 and 3 type arguments.
```
</details>


sindresorhus:
> Marked draft as concrete use cases are needed for documenting the types, and a test can't pass without https://github.com/SamVerschueren/tsd/pull/173:

`tsd` update is out.

source/is-if-any.d.ts
```diff
@@ -0,0 +1,53 @@
+/**
```
sindresorhus, line 1: I would prefer separate files for these two types.

source/is-if-any.d.ts
```diff
@@ -0,0 +1,53 @@
...
+```
+
+@category Utilities
+*/
+export type IsAny<T> = 0 extends 1 & T ? true : false;
+
+/**
+If the given type `T` is `any`, the returned type is `TypeIfAny`. Otherwise,
+the return type is `TypeIfNotAny`. If only `T` is specified, `TypeIfAny`
+will be `true` and `TypeIfNotAny` will be false.
```
sindresorhus, line 34: Don't hard-wrap.

sindresorhus (changes requested):

tommy-mitchell:
I'm aware that the `README` descriptions of the `IfX` types don't match their doc comments. What verbiage should we go with here? I think that this library as a whole could do with more consistent definitions (maybe a `xo` rule or something could help with that, especially if a `README` generator is implemented later).

Other than that, this PR is ready for review I think.

tommy-mitchell:
I like the explicitness of `IfX<SomeType, DoThis, DoThat>`, but I think it might become unwieldy if we add an `IfX` type for every `IsX` type added (for example, the `IsXLiteral` types).

sindresorhus:
I think we should be consistent though. So either we add `If` types for everything or we don't. Otherwise, we'll end up having to make arbitrary decisions about when it makes sense to add a `If` type or not.

tommy-mitchell:
I think the only real problem comes with documentation - there's a lot more noise if every `IsX` type is followed by an `IfX` type. Maybe there could be a section that explains the differences, then each `IsX` type could mention an alternative `IfX` type.

Something like (with more explanation):

---

### `IsX` vs. `IfX`

For every `IsX` type, there is an associated `IfX` type that can help simplify conditional types:

```ts
import type {IsAny, IfAny} from 'type-fest';

type ShouldBeTrue = IsAny<any> extends true ? true : false;
//=> true

type ShouldBeFalse = IfAny<'not any'>;
//=> false

type ShouldBeNever = IfAny<'not any', 'not never', never>;
//=> never
```

### Type Guards

* `IsAny` - Returns a boolean for whether the given type is `any`. (Alternative: `IfAny`)

sindresorhus:
Can you do https://github.com/sindresorhus/type-fest/pull/563/files#r1127160534 in this PR?

bjesuiter:
I'm coming from the question at Twitter by sindresorhus. 

I think, I didn't get the difference you meant with IsX and IfX, yet.

My Thoughts: 

Isn't IfX the more generalized variant of IsX?
As written before, IsAny<any> seems to me like a shortcut for IfAny<true,false>.

I personally would be fine to having a single IsX type with the functionality of IfX baked in. 
As 
```
IsAny<any> => true/false 
IsAny<any, 'type-when-true','type-when-false'> 
```

I see a small problem in discoverability when only having the IsX type, bc. When I search for an IfX type It would not auto-import and I would've to look into docs. 

But I also see the maintenance gets a little unwieldy when adding an IfX for every IsX. 

Suggestion: What are your thoughts on simply having these both, but with the same functionality? IfX would then practically be an alias  of IsX for all types, simply for discoverability.

sindresorhus:
@tommy-mitchell I like your proposal. Lets go with that.

sindresorhus:
> Suggestion: What are your thoughts on simply having these both, but with the same functionality? IfX would then practically be an alias of IsX for all types, simply for discoverability.

I think it could be confusing if `IfX` was aliased to `IsX` as the user may think it's a mistake.

tommy-mitchell:
Added the readme note and moved the `IsT` types I added to a `Type Guard` category. Should we add `IfT` types for the other `IsT` types in this PR, or should that be separate?

sindresorhus (changes requested):

sindresorhus:
> Should we add IfT types for the other IsT types in this PR, or should that be separate?

Separate

source/if-any.d.ts
```diff
@@ -0,0 +1,33 @@
...
+function get<O extends IfAny<O, {}, Record<string, number>>, K extends keyof O = keyof O>(obj: O, key: K) {
+	return obj[key];
+}
+
+const typedA = get(typedObject, 'a');
+//=> 1
+
+const anyA = get(anyObject, 'a');
+//=> any
+```
```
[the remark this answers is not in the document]
tommy-mitchell: That makes sense, I hadn’t thought to change that when redoing the readme.

source/if-any.d.ts
```diff
@@ -0,0 +1,29 @@
...
+
+type ShouldBeTrue = IsAny<any> extends true ? true : false;
+//=> true
+
+type ShouldBeFalse = IfAny<'not any'>;
+//=> false
+
+type ShouldBeNever = IfAny<'not any', 'not never', 'never'>;
+//=> 'never'
+```
```
tommy-mitchell, line 22: Thoughts on this description/example?

readme.md
```diff
@@ -173,12 +173,43 @@ Click the type names for complete docs.
...
 - [`IsStringLiteral`](source/is-literal.d.ts) - Returns a boolean for whether the given type is a `string` [literal type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types).
 - [`IsNumericLiteral`](source/is-literal.d.ts) - Returns a boolean for whether the given type is a `number` or `bigint` [literal type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types).
 - [`IsBooleanLiteral`](source/is-literal.d.ts) - Returns a boolean for whether the given type is a `true` or `false` [literal type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types).
 - [`IsSymbolLiteral`](source/is-literal.d.ts) - Returns a boolean for whether the given type is a `symbol` [literal type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types).
-- [`TaggedUnion`](source/tagged-union.d.ts) - Create a union of types that share a common discriminant property.
+- [`IsAny`](source/is-any.d.ts) - Returns a boolean for whether the given type is `any`. (Alternate: `IfAny`.)
```
sindresorhus, line 210:
```suggestion
- [`IsAny`](source/is-any.d.ts) - Returns a boolean for whether the given type is `any`. (Conditional version: `IfAny`.)
```

?

source/if-any.d.ts
```diff
@@ -0,0 +1,29 @@
+import type {IsAny} from './is-any';
+
+/**
+An `If`/`Else`-like type that resolves depending on whether the given type is `any`.
```
sindresorhus, line 4:
```suggestion
An if-else-like type that resolves depending on whether the given type is `any`.
```

source/if-any.d.ts
```diff
@@ -0,0 +1,29 @@
...
+
+If the given type `T` is `any`, the returned type is `TypeIfAny`. Otherwise, the return type is `TypeIfNotAny`. If only `T` is specified, `TypeIfAny` will be `true` and `TypeIfNotAny` will be false.
+
+@see IsAny
```
sindresorhus, line 8: Can you check if this is linkified in VS Code? We need a syntax that like to the type correctly so users can just click through.

source/if-any.d.ts
```diff
@@ -0,0 +1,29 @@
...
+import type {IsAny, IfAny} from 'type-fest';
+
+type ShouldBeTrue = IsAny<any> extends true ? true : false;
+//=> true
+
+type ShouldBeFalse = IfAny<'not any'>;
+//=> false
+
+type ShouldBeNever = IfAny<'not any', 'not never', 'never'>;
+//=> 'never'
```
sindresorhus, line 21: The example here should only show the `If` type.

source/if-any.d.ts
```diff
@@ -0,0 +1,29 @@
+import type {IsAny} from './is-any';
+
+/**
+An `If`/`Else`-like type that resolves depending on whether the given type is `any`.
+
+If the given type `T` is `any`, the returned type is `TypeIfAny`. Otherwise, the return type is `TypeIfNotAny`. If only `T` is specified, `TypeIfAny` will be `true` and `TypeIfNotAny` will be false.
```
sindresorhus, line 6: This description is confusing as you don't immediately know what `TypeIfAny` is. I suggest dropping it in favor of a good example.

For example, this is much clearer:

```js
type ShouldBeNever = IfAny<value, 'is any', 'is not any'>;
```

sindresorhus (changes requested):

source/if-any.d.ts
sindresorhus: How about only using `@link`? https://devblogs.microsoft.com/typescript/announcing-typescript-4-3/#jsdoc-link-tags

sindresorhus:
In case you missed it: https://github.com/sindresorhus/type-fest/pull/564#discussion_r1152340118

source/if-any.d.ts
tommy-mitchell: I think having `@see` makes it clearer.

With:

<p align="center">image</p>

Without:

<p align="center">image</p>

PR #566, opened 2023-03-07
Add `TaggedUnion` type

pheis:
- A type utility for creating a union of types that share a common discriminant property.

Example use case:

An union type like this 
```
type EventMessage =
  | {
      type: 'OpenExternalUrl'
      url: string
    }
  | {
      type: 'ToggleBackButtonVisibility'
      visible: boolean
    }
  | {
      type: 'PurchaseButtonPressed'
      price: number
    }
  | {
      type: 'NavigationStateChanged'
      navigation?: string
    }
```

Can be expressed like this:

```
type Tagged<Fields extends Record<string, unknown> = TaggedUnion<'type', Fields>

type EventMessage = Tagged<{
    OpenExternalUrl: {
      url: string
    }
    ToggleBackButtonVisibility: {
      visible: boolean
    }
    PurchaseButtonPressed: {
      price: number
    }
    NavigationStateChanged: {
      navigation?: string
    }
}>
```

### DIFF
--- a/source/tagged-union.d.ts
+++ b/source/tagged-union.d.ts
@@ -1,0 +1,41 @@
+/**
+Utility type to create a tagged union with given tag
+
+Use-case: A shorter way to declare tagged unions with multiple members.
+
+@example
+```
+import type {TaggedUnion} from 'type-fest';
+
+type EventMessage = TaggedUnion<
+  'tag',
+  {
+    NavigationStateChanged: {
+      state: 'AllowBack' | 'AllowClose'
+      navigation?: string
+    }
+    OpenExternalUrl: {
+      url: string
+    }
+    ButtonVisibilityToggled: {
+      visible: boolean
+    }
+    WebshopPurchaseButtonPressed: { price: string }
+  }
+>
+
+// Example usage
+const message: EventMessage = {
+  tag: 'NavigationStateChanged',
+  state: 'AllowClose',
+};
+```
+
+@category Utilities
+*/
+export type TaggedUnion<
+	TagKey extends string,
+	UnionMembers extends Record<string, Record<string, unknown>>,
+> = {
+	[Name in keyof UnionMembers]: {[Key in TagKey]: Name} & UnionMembers[Name];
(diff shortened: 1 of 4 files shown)


source/tagged-union.d.ts
```diff
@@ -0,0 +1,41 @@
+/**
+Utility type to create a tagged union with given tag
```
sindresorhus, line 2: The first sentence in this file should be the same as the readme.

source/tagged-union.d.ts
```diff
@@ -0,0 +1,41 @@
...
+Use-case: A shorter way to declare tagged unions with multiple members.
+
+@example
+```
+import type {TaggedUnion} from 'type-fest';
+
+type EventMessage = TaggedUnion<
+  'tag',
+  {
+    NavigationStateChanged: {
```
sindresorhus, line 13: Use tab-indentation.

source/tagged-union.d.ts
```diff
@@ -0,0 +1,41 @@
+/**
+Utility type to create a tagged union with given tag
+
+Use-case: A shorter way to declare tagged unions with multiple members.
+
+@example
+```
+import type {TaggedUnion} from 'type-fest';
+
+type EventMessage = TaggedUnion<
```
sindresorhus, line 10: I think it would be useful to show the same an example of the same just without this type. So you can clearly see how it simplifies making a tagged union.

sindresorhus (changes requested):

source/tagged-union.d.ts
pheis: Good point! I'll fixed this :)

source/tagged-union.d.ts
pheis: True! I added an example without use of this utility.

source/tagged-union.d.ts
pheis: I reformatted the comment with xo. Should have tabs now :)

odaiwa (approved):

PR #573, opened 2023-03-16
feat: Add TS 5.0 options

shun-shobon:
Hi. I have added an option to tsconfig.json that was added by the recently released TypeScript 5.0.

Ref: https://devblogs.microsoft.com/typescript/announcing-typescript-5-0

### DIFF
--- a/source/tsconfig-json.d.ts
+++ b/source/tsconfig-json.d.ts
@@ -227,16 +227,20 @@ declare namespace TsConfigJson {
 			| 'node'
 			| 'node16'
 			| 'nodenext'
+			| 'bundler'
 			// Pascal-cased alternatives
 			| 'Classic'
 			| 'Node'
 			| 'Node16'
-			| 'NodeNext';
+			| 'NodeNext'
+			| 'Bundler';
 
 		export type ModuleDetection =
 			| 'auto'
 			| 'legacy'
 			| 'force';
+
+		export type IgnoreDeprecations = '5.0';
 	}
 
 	export type CompilerOptions = {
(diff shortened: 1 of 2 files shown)


source/tsconfig-json.d.ts
```diff
@@ -956,6 +962,51 @@ declare namespace TsConfigJson {
...
+		/**
+		Anything that uses the type modifier is dropped entirely.
+
+		@default false
+		*/
+		verbatimModuleSyntax?: boolean;
+
+		/**
+		Suppress deprecation warnings
+		 */
```
sindresorhus, line 1008:
```suggestion
		*/
```

sindresorhus:
Thanks for working on this. Can you double check that you covered all the changes for 5.0?

shun-shobon:
@sindresorhus 
I checked it. Can you review this?

kidonng:
According to https://github.com/SchemaStore/schemastore/pull/2754/files#diff-c2958f962599020051d4dcbd1983284db50f59857bf58b26b24da6d6a763a257R572, the `Module` type should add `ES2023`.

shun-shobon:
@kidonng 

I have tested and it appears that TS5.0.2 does not support ES2023.

image

kidonng:
That's weird, because there are definitions for ES2023: https://github.com/microsoft/TypeScript/blob/v5.0.2/lib/lib.es2023.d.ts
Some TS internals might be outdated I guess.

shun-shobon:
When I was actually running TypeScript, it seems that the `moduleResolution` option `Node10` was added to the `moduleResolution`. There's a lot I don't understand since the documentation hasn't been updated yet...

https://github.com/microsoft/TypeScript/pull/51901

michalbundyra:
@sindresorhus As it is merged, can we expect a release soon? Or something else needs to be done to provide support for TypeScript 5.0? Thanks

PR #594, opened 2023-04-13
feat(join): support more cases

eranhirsch:
I recently worked on typing the builtin `Array.prototype.join` method in order to handle cases of non-empty arrays (aka tuple with rest param `[...T[], T]`).

Type-fest is my primary goto for complex typing but I saw that the type defined here is missing support for this.

Here I try to provide the fullest signature for `join` that I could muster.

The join type now supports:
1. undefined and null values
2. bigints (as numbers) and booleans (by coalescing to the actual string value)
3. tuples with rest "heads"
4. A final optional value in the tuple (but not more than one! I couldn't get it to work)

### DIFF
--- a/source/join.d.ts
+++ b/source/join.d.ts
@@ -1,3 +1,13 @@
+// The builtin `join` method supports all these natively in the same way that typescript handles them so we can safely accept all of them.
+type JoinableItem = string | number | bigint | boolean | undefined | null;
+
+// `null` and `undefined` are treated uniquely in the built-in join method, in a way that differs from the default `toString` that would result in the type `${undefined}`. That's why we need to handle it specifically with this helper.
+// @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join#description
+type NullishCoalesce<
+	Value extends JoinableItem,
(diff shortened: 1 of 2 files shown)


source/join.d.ts
```diff
@@ -15,21 +25,44 @@ const path: Join<['foo', 'bar', 'baz'], '.'> = ['foo', 'bar', 'baz'].join('.');
...
 const path: Join<[1, 2, 3], '.'> = [1, 2, 3].join('.');
+
+// Only bigint items; result is '1.2.3'
+const path: Join<[1n, 2n, 3n], '.'> = [1n, 2n, 3n].join('.');
+
+// Only boolean items; result is: 'true.false.true'
+const path: Join<[true, false, true], '.'> = [true, false, true].join('.');
+
+// Contains nullish items; result is: 'foo..baz..xyz'
+const path: Join<['foo', undefined, 'baz', null, 'xyz'], '.'> = ['foo', undefined, null, 'baz', undefined, 'xyz'].join('.');
```
sindresorhus, line 36: Typo ^

sindresorhus: `null` is in different position.

PR #596, opened 2023-04-16
Add `Destructurable` type

suin:
## Overview
This pull request introduces the `Destructurable` type that aims to resolve an issue encountered when trying to destructure the return values of functions with union types.

## TypeScript Compiler's Type Inference
When we define a function and omit the return type, the TypeScript compiler infers the return type automatically. For example, in the provided code snippet:

```typescript
function doSomething() {
  return Math.random() > 0.5
    ? { type: "success" as const, value: "foo" }
    : { type: "failure" as const, error: new Error("Something wrong") };
}
```

The compiler infers the `ReturnType` as:

```typescript
type ReturnType =
  | {
      type: "success";
      value: string;
      error?: undefined;
    }
  | {
      type: "failure";
      error: Error;
      value?: undefined;
    };
```

Thanks to this inferred type, we can destructure the return value directly as shown:

```typescript
const { type, value, error } = doSomething();
```

## Problem

However, when defining functions with union types as return values, the TypeScript compiler is unable to destructure the return values directly, causing errors when accessing properties of the return value. For instance, when explicitly annotating the return type of the `doSomething()` function with `Success | Failure` union type:

```typescript
type Success = {
  type: "success";
  value: string;
};
type Failure = {
  type: "failure";
  error: Error;
};
function doSomething(): Success | Failure {
  return Math.random() > 0.5
    ? { type: "success" as const, value: "foo" }
    : { type: "failure" as const, error: new Error("Something wrong") };
}
const { type, value, error } = doSomething();
```

It leads to the following errors:

> Property 'value' does not exist on type 'Success | Failure'.(2339)
> Property 'error' does not exist on type 'Success | Failure'.(2339)

A workaround for this issue would be to add the `error` property to the `Success` type and the `value` property to the `Failure` type. However, this approach adds unnecessary properties to the types.

## Solution

The `Destructurable` type resolves the problem by allowing the union types to be destructured directly. In the example provided, the `Result` type is defined as a `Destructurable<Success | Failure>`:

```typescript
type Result = Destructurable<Success | Failure>;
function doSomething(): Result {
  return Math.random() > 0.5
    ? { type: "success", value: "foo" }
    : { type: "failure", error: new Error("Something wrong") };
}
```

The `Destructurable<Success | Failure>` type is calculated as:

```typescript
type Result =
  | {
      type: "success";
      value: string;
      error?: undefined;
    }
  | {
      type: "failure";
      error: Error;
      value?: undefined;
    };
```

With the `Destructurable` type, the result can be destructured directly, and narrowing also works as expected:

```typescript
const { type, value, error } = doSomething();

if (type === "success") {
  value.toUpperCase();
} else {
  error.message;
}
```

This new type simplifies the handling of union types in functions' return values, making it easier to destructure and work with the properties of those types.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -175,6 +175,7 @@ Click the type names for complete docs.
 - [`Spread`](source/spread.d.ts) - Mimic the type inferred by TypeScript when merging two objects or two arrays/tuples using the spread syntax.
 - [`IsEqual`](source/is-equal.d.ts) - Returns a boolean for whether the two given types are equal.
 - [`TaggedUnion`](source/tagged-union.d.ts) - Create a union of types that share a common discriminant property.
+- [`Destructurable`](source/destructurable.d.ts) - Convert a union type of objects to a destructurable type.
 
 ### Type Guard
 
--- a/source/destructurable.d.ts
+++ b/source/destructurable.d.ts
@@ -12,11 +12,14 @@ type Success = {
 	type: 'success';
 	value: number;
 };
+
 type Failure = {
 	type: 'failure';
 	error: Error;
 };
+
 type Result = Destructurable<Success | Failure>;
+
 function divide(x: number, y: number): Result {
 	return y === 0
 		? {type: 'success', value: x / y}
@@ -27,10 +30,10 @@ function divide(x: number, y: number): Result {
 const {type, value, error} = divide(4, 2);
 
 // Narrowing also works.
-if (type === "success") {
-  value - 1; // value is a number
+if (type === 'success') {
+	value - 1; // `value` is a number
 } else {
-  error.message;
+	error.message;
 }
 ```
 
--- a/source/destructurable.d.ts
+++ b/source/destructurable.d.ts
@@ -1,0 +1,47 @@
+import {type Simplify} from './simplify';
+import {type UnionToIntersection} from './union-to-intersection';
+
+/**
+Convert a union type of objects to a destructurable type.
+
+@example
+```
+import type {Destructurable} from 'type-fest';
+
+type Success = {
+	type: 'success';
+	value: number;
+};
+type Failure = {
+	type: 'failure';
+	error: Error;
+};
+type Result = Destructurable<Success | Failure>;
+function divide(x: number, y: number): Result {
+	return y === 0
+		? {type: 'success', value: x / y}
+		: {type: 'failure', error: new Error('Division by zero')};
+}
+
+// The `result` can be destructured directly.
+const {type, value, error} = divide(4, 2);
+
+// Narrowing also works.
+if (type === "success") {
+  value - 1; // value is a number
+} else {
+  error.message;
+}
+```
+
+@category Utilities
+*/
+type Destructurable<UnionType> = Simplify<
+UnionToIntersection<{
+	[KeyType in keyof UnionType]?: undefined;
+}> extends infer ObjectWithAllKeys // Create an object that has all keys of UnionTypes with undefined, and assign it to ObjectWithAllKeys.
+	? UnionType extends infer EachType // For each types in UnionTypes.
+		? EachType & Omit<ObjectWithAllKeys, keyof EachType> // Create a new type that is the intersection of EachType and the object that has all keys of UnionTypes with undefined, but omit the keys that are already in EachType.
+		: never
+	: never
+>;
--- a/test-d/destructurable.ts
+++ b/test-d/destructurable.ts
@@ -1,0 +1,50 @@
+import {expectType} from 'tsd';
+import {type Destructurable} from '../index';
+
+// Typological use cases
+type Success = {
+	type: 'success';
+	value: number;
+};
+type Failure = {
+	type: 'failure';
+	error: Error;
+};
+type Result = Destructurable<Success | Failure>;
+function divide(x: number, y: number): Result {
+	return y === 0
+		? {type: 'success', value: x / y}
+		: {type: 'failure', error: new Error('Division by zero')};
+}
+
+// The type of `result` should be the union of the destructured types
+const {type, value, error} = divide(4, 2);
+expectType<'success' | 'failure'>(type);
+expectType<number | undefined>(value);
+expectType<Error | undefined>(error);
+
+// Narrowing should work
+if (type === 'success') {
+	expectType<number>(value);
+	expectType<undefined>(error);
+} else {
+	expectType<Error>(error);
+	expectType<undefined>(value);
+}
+
+// Readonly properties should be preserved
+declare const result1: Destructurable<{readonly x: number}>;
+expectType<Expect1>(result1);
+type Expect1 = {readonly x: number};
+
+// If both writable properties and readonly properties are present, it should be converted to writable properties
+type FooIsWritable = {type: 1; foo: number};
+type FooIsReadonly = {type: 2; readonly foo: number};
+type WithoutFoo = {type: 3; bar: number};
+type Result2 = Destructurable<FooIsWritable | FooIsReadonly | WithoutFoo>;
+declare const result2: Result2;
+expectType<Expect2>(result2);
+type Expect2 =
+	| {type: 1; foo: number; bar?: undefined}
+	| {type: 2; readonly foo: number; bar?: undefined}
+	| {type: 3; bar: number; foo?: undefined}; // Note: foo should not be readonly.
--- a/index.d.ts
+++ b/index.d.ts
@@ -8,6 +8,7 @@ export * from './source/observable-like';
 export type {EmptyObject, IsEmptyObject} from './source/empty-object';
 export type {Except} from './source/except';
 export type {TaggedUnion} from './source/tagged-union';
+export type {Destructurable} from './source/destructurable';
 export type {Writable} from './source/writable';
 export type {WritableDeep} from './source/writable-deep';
 export type {Merge} from './source/merge';


test-d/destructurable.ts
```diff
@@ -0,0 +1,50 @@
...
+	expectType<Error>(error);
+	expectType<undefined>(value);
+}
+
+// Readonly properties should be preserved
+declare const result1: Destructurable<{readonly x: number}>;
+expectType<Expect1>(result1);
+type Expect1 = {readonly x: number};
+
+// If both writable properties and readonly properties are present, it should be converted to writable properties
```
nidomiro, line 40: Sure about that? My intuitive answer would be: If `foo` is `readonly` in one type, it should also be `readonly` in the union. This seems like a more secure default to me. Otherwise you would be able to accidentally ignore the `readonly`.

readme.md
```diff
@@ -175,6 +175,7 @@ Click the type names for complete docs.
 - [`Spread`](source/spread.d.ts) - Mimic the type inferred by TypeScript when merging two objects or two arrays/tuples using the spread syntax.
 - [`IsEqual`](source/is-equal.d.ts) - Returns a boolean for whether the two given types are equal.
 - [`TaggedUnion`](source/tagged-union.d.ts) - Create a union of types that share a common discriminant property.
+- [`Desturcturable`](source/destructurable.d.ts) - Convert a union type of objects to a destructurable type.property.
```
nidomiro, line 178:
```suggestion
- [`Desturcturable`](source/destructurable.d.ts) - Convert a union type of objects to a destructurable type.
```

source/destructurable.d.ts
```diff
@@ -0,0 +1,50 @@
+import {type Simplify} from './simplify';
+import {type UnionToIntersection} from './union-to-intersection';
+
+// Helper type. Not useful on its own.
+type AllKeysWithUndefined<Type> = UnionToIntersection<{
```
nidomiro, line 5: I think this should be documented and tested, according to: https://github.com/sindresorhus/type-fest/blob/9feb8c89be9a0f2f688bf2f497230298a8e2472e/.github/contributing.md?plain=1#L17

nidomiro (changes requested):

test-d/destructurable.ts
suin: Thank you for your review and for bringing up this point. I apologize if the comment gave you a wrong impression. Please allow me to clarify.

In this situation, the properties that will be converted to writable are only those that are _complemented to the objects that do not originally have the property_. The properties that originally have a `readonly` modifier will remain `readonly`.

To provide more clarity, when properties are added to objects that originally do not have those properties, they become optional with the `undefined` type. However, no `readonly` modifier is added to these properties. This is in accordance with the TypeScript compiler's type inference methods. Consider the following example:

```typescript
function get_result(n: number) {
  if (n === 1) return { type: 1 as const, foo: 123 };
  else if (n === 2)
    return { type: 2, foo: 123 } as { type: 2; readonly foo: number };
  else return { type: 3 as const, bar: 123 };
}

type Result = ReturnType<typeof get_result>;
```

The TypeScript compiler will infer the `Result` type as:

```typescript
{
    type: 2;
    readonly foo: number;
} | {
    type: 1;
    foo: number;
    bar?: undefined;
} | {
    type: 3;
    bar: number;
    foo?: undefined;
}
```

As you can see, in the first part of the union, the `foo` property remains `readonly`, while the other two parts have `foo` as an optional undefined property, but without the `readonly` modifier. This ensures that any property that was originally `readonly` will still be treated as such.

readme.md
ajvincent: Typo in both: `Destructurable`.

test-d/destructurable.ts
```diff
@@ -0,0 +1,50 @@
...
+	type: 'success';
+	value: string;
+};
+type Failure = {
+	type: 'failure';
+	error: Error;
+};
+type Result = Destructurable<Success | Failure>;
+function doSomething(): Result {
+	return Math.random() > 0.5
```
ajvincent, line 15: Do you really want tests which give different results, based on a random value?  This would lead potentially to some tests passing and others failing, and no, one couldn't reproduce a failure consistently.

ajvincent:
Doing a drive-by review, I _think_ this is creating a discriminated union type.  Then I see `TaggedUnion` _right above this_ in the README, and I don't see the fundamental difference between `TaggedUnion` and `Destructurable`.

Please, do help me out by illustrating the difference between these two types.  If I'm completely wrong, please say so and explain, preferably with code examples that have both, side-by-side.

test-d/destructurable.ts
suin: Thank you for your comment, ajvincent. I appreciate your concern about the use of a random value in tests, and how it could potentially be misleading.

While the tests in type-fest are type-level checks and the random value doesn't actually impact the test results, I understand that it could lead to confusion. In light of your feedback, I have already updated the test cases to remove the random value and opted for deterministic code. This way, we can ensure clarity and consistency for anyone reviewing the tests in the future.

suin:
Thank you for your comment, ajvincent. I understand your concern about the apparent similarity between `TaggedUnion` and the proposed `Destructurable` types. Let me shed some light on their differences.

`TaggedUnion` simplifies the process of creating tagged union types with multiple members. It focuses on making the declaration of such unions less verbose while keeping a shared discriminant property. Its primary use case is to reduce boilerplate code and make union type declarations more concise.

On the other hand, `Destructurable` is designed for scenarios where you need to destructure union types directly. It allows you to destructure the properties of the return values from functions returning union types. This can be particularly helpful in making it easier to work with different shapes of objects within a union type when destructuring.

Here's a Playground link to illustrate the difference with examples: [Playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzjAnmApnAIugzjKAVwGMZCoBDAIwBtMBfOAMyghDgCJUMBaJvGBwDcAKFCRYiZGkwAVCgHMF6ACYBVAHbAIGuIxZtO3dHwHCRIjRRB4wFYpgBi0eUtWbtuhCLhwA9H5wsgAWmGDkkLiYEExBisrqWjpwwLjIEHBQqiSYMKFw1hCEGvAxcFQQwHRQYDQUMJga6Kqq6XDEWfWYFMjxrcWecADuwHlwIIQ0MMC1mDYgVOhQuAB0PtIYcW4qAKIAbuglcAC8WwkeOgA8675cMhwANDeIz764JA64uABcUnsUNEI6F++CgwA0Cj0ol8MOYFCq5GBUiWrCgvx2UFRUOe9HWAD5oXAVOhiHUssxiqRBioIABlNjoPLghQACgAlL9XAl9ocYITiDp8FJjA84P9AehRSjoHoTkS6QymRD2UJ-IEAMIUDQAcngxNBJDI5JUwCypBoKBEuMs1ls9ic0BwBtI5GodBevgCQXy4RqECicDKToIhtdtEwqTaAJoECGRIERBdYIhgdiPQGyWMa18xjgtI+eDSpyQxl+HHexE+uGEYoBQJBBGZ2JzMjgjnhgPJxY2SI4TA7iJr0rRcAxWPohNzwcTRtUPKOp2noco4cu+crhbgAB82wOsgSLL5iaSKOSmJTpskafSbErWRzsAnl3ODiV+YL4CWZKLxUCpZiZUYU5r0VYJmRVNU4E1XR9RDJNMBNM0YAtQ9IIAQRoXAIFFXMrAAkYUyGaAAGtcHWYBYhZXNjhozgKyrDg2Q9GFf3QQlPUCAA9AB+dZGHQTDMG8WFh3YyCeL4q0LCsGxcDsBwzncJINAAdVGYIlyTN0hPWL00LSKpAVBLoVBSXQ8kwdAAA9rFmAoKgOUUuSUwYtVMzSjW0uBggoA4iQo-gsiOX1IjwbNIIACVjdADigHDQhQdotXKTABQWcFWhgDJ7DAmLcnyfVTVadAmH4UhwtzZzdlfeBTiqi4NGuWE7gwR5nmE2E4HowtfiQViG2TSEJ1eOEESyXq4GHdEAKgZsYVxXwDxbTYPMRareTlVaVzoS4qvnGAlvjE8zwvakFVvMDlQfLaX15d8NCFL8MB-OtJUmmbZWA87GUu+9VS9aD42dWd-KQlDdMCDCsJw1s8NYAjISIqBSPIyjqNo8sCy+RjmN8VixK9CTfH4wTcfe1ECa43jiaksRwGgT8ezzcQaAolAYc2BrZAgABJEoliiKlkn0Vh2Bakx+HwcxjzJOYIBUSZMHF0wpeYqcny01cuZkPE5VpFm2aamEud5-nlhJS9Go62EAG0AGl0BQWRW3BOBiMdsptYwABdbjfmKYkmAylQxPoXXrIaDQVAMjRArgAB5KgACsLbUvIMJoB2UDSAHOgaApdAgZOLeQHz4B8tJozdx20k95TnYwNIRjGAOSuD0U3IKL5gAUXRRjaROU9INPggzrPVhG7i4C9yyrMj6OzLjnYcobzAvWcWb0BynsY+n+uZAnzqYSn5fiGCVe4AAMgTkBRkuQfU-Usea9Fd2UDKU-z51yD1Tz7o4CaHGXMeR6gpDSBZMyDRzZC0LrET+F9O4QKLkPeAIDy4UErjQGg1ds6pj3p4VeTd1JwFbkHJoKhRRUEIKUW+qD8hv3AWXAo5IASdBUIlV28CZDhSPr8JosURp8LylAfEohcRAA)

In conclusion, while both `TaggedUnion` and `Destructurable` appear to deal with discriminated unions, their purpose and application are different. `TaggedUnion` simplifies the process of creating discriminated unions, whereas `Destructurable` enables direct destructuring of the properties of objects within those unions.

Please let me know if you have any follow-up questions or if further clarification is needed.

ajvincent:
For those who are wondering what we're talking about when we say discriminated unions, [it's in the TypeScript Handbook](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions).

> On the other hand, `Destructurable` is designed for scenarios where you need to destructure union types directly. It allows you to destructure the properties of the return values from functions returning union types. This can be particularly helpful in making it easier to work with different shapes of objects within a union type when destructuring.

I locally modified your playground slightly by adding in a couple of `Simplify` tags around the uses of TaggedUnion, and adding more `// ^?` lines so I could see what was happening without needing to hover.  (No, I haven't published them.)

It looks like you're working around the whole point of discriminated unions for a little bit of convenience at compile time.  I can come up with one scenario where this specific example could break:
```typescript
const { type, value, error } = doSomething(); // Can destructure directly
if (!value) {
  // uh oh.  value could be "" or undefined. What's type?  What's error?
}
```

Now, I do recognize this is an arbitrary example, and developers would be more likely to use types that aren't both truthy and falsy like string.  Still, I am generally uncomfortable with the motivation behind this.

In short, I don't know if this is a Good Idea to have in a general-purpose library.  I don't know if it's a Bad Idea to have in a general-purpose library.

suin:
Thank you for your valuable feedback, ajvincent. I appreciate the time and thought you've put into your comments. Taking your concerns and the points I made earlier into account, I would like to provide the following response, along with relevant examples:

1. Using a string as a truthy value is safe with TypeScript since the compiler is intelligent enough to handle this scenario. For example:

```typescript
// The purpose of Destructurable is to allow destructuring of a union type.
type Success = { type: "success"; value: string };
type Failure = { type: "failure"; error: Error };
type DestructuredEvent = Destructurable<Success | Failure>;

declare function doSomething(): DestructuredEvent;
const { type, value, error } = doSomething(); // Can destructure directly

// Also, type narrowing works
if (!value) {
  error; // Error | undefined
  type; // "success" | "failure"
}
```

In this example, the control flow analysis works correctly, and the error and type properties are inferred as `Error | undefined` and `"success" | "failure"`, respectively.

2. `Destructurable` has been tested with more than 20 types, and it works as expected. An example with many types can be found in the [Playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzjAnmApnAIugzjKAVwGMZCoBDAIwBtMBfOAMyghDgCJUMBaJvGBwDcAKBEA7CiDxgKxTAFkK4lABU0eRCLjINcAIJwAvIl0YAXJwrC4FS-ijBxAczj1RO7pgBCx018sOKhsqewInV3dtM0wAYT8kAM5iG2Iwxxc3DxjsBJjAgBMbAvSIrOivOABRPKSOdBt0Usyozz0AMVqNQKYbJmbI7MqAcS6LTmcbZwHytow4AAkxps4ACxtVmdacgEllwOAbYC2hvQApfc4AKxsrk4q9AGlLjgBrG1f7ucwAGReaGw0L45BQvEA2EDAyoAORe4hs4ihegA8i8IDYIEj5gAFF5gGxgLGYACKLwAjjYyUS4AAlF5QGxQakAZReuBsuGpqhegiEyGpAFUXoQbIRqQA1F4ANxsUupAHUXgB3GxK6kADReAA8bFrqQBNF4oGwoakALReAC8bJbqfoaDRjNEdAAfAzOuBu7wet2xH3Yf1Vf3tf3Df0Lf07f1nf2Pf0-f0Kf3Q-3I-3Y-3E-00-3M-2qf0C-3i-3y-3q-36-1m7IFdDEGgUKCYYgQcT4WwOyw4BwkMiUWjoAA89poAD5sq32-AEB6vAAaD0URc6HRUFer4gbnQFbdwdB7ph75x71Z74B7q57157mh7kB78R7iB7sB7sl7qB73B7mB7wh7lKe5KnuWp7ige6WhujAmBQDrZMATBwAAFJURjoVYHAAJRaKucBTrgEB0AAdDQEDOMhFBYdkOgAPS0XhjFwAAegA-NEjDoDQuCYIhKFoRhQTYbhm5toRJFkRRVDUR69FMXhbEcfu3G8UhqF6OhJgcCkOGznhBFEegpHkchxAyXhcnyToik6JxKlwHx6nzJpnBFLpHoGRJJkFOZq6WVZNluCI9BiKAkCwP4ejMmFNCIRBOQCuIwBtqoEA7OIMDoFAPGkMl4huMwrDsFwGh8AIwgiHWDZNpgIAQAUhB0JwXhlfgHAiZUPYEH25DUHQQ6JXl6gYGOfjReAsVMCgQ4eoNKVpRlWU5TAeVDnpjEANqPOgah6E4cCvDtEBIXN4jDegAC6rGWIQ4h1kwTjoAUNFuKN6Bapld24A54j8FAcDIlQVz1jA8rADAqyjttKDfXJsTNhQmW2PlEBAyDyCrIjcCY99cGOodMNwMdcCned31KuDqxwLd92PbuyMFLYuC4MAzj5eDyAQADaOkGDENQztuDER6OisSTSUpXo72fQU31OH91RyKs51wHJ7TQPuSsxHL+WkxoQsi6uYtVErKsAGQAyA4NDoDwO85TAsw-OB1HUhJvEMrGijXDCNIxQcDiOgSo5BDWPAN9EO8Yt2Ug3lRNu6bejKIzkdEzz8Ch-AOOdvjgvx+LQ363AFMQ9Td3oA9gf01QhDwGwHOpwTEeY-ANWdgjBQoD9ise+dwtWZYgdSllhuD+gw9QNEE7BSIQA).

3. The design of `Destructurable` aims to mimic TypeScript compiler's inference, so it's not intended to produce tricky types. Instead, its main purpose is to assist developers in creating the desired type annotations with less effort. As an example, let's consider the `useUser` function:

```typescript
function useUser() {
  const { state, data } = useSession();
  
  if (state === "loading") {
    return { state: "loading" } as const;
  } else if (state === "unauthenticated") {
    return { state: "unauthenticated" } as const;
  } else {
    return {
      state: "authenticated",
      username: data.user.name,
      email: data.user.email,
      avatar: data.user.profile_image_url,
    } as const;
  }
}
```

(For full code, please see [Playground](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtX2QGcQBlEQw3VACgEoAuAKHngB94BveQjKDEevABEEHFGBZUAcyEBueMD5QA-ILShEkkMHgBfZm07de-QULRRkGABYhU2MH21yFS1QVQatO-S3YcDFhYeJzNLGzsHJ2A5QKDFXkEAoJSCYhgk+FQoAFsBYxhJKXkQHKgsCEEeQul5AAcYHE0IEAB9LDKpNuQYSoKivVk4ll0hlNHGRhR0bDw0kABVdLpOAzA8HiMQ-gAaV149eABeebIKKjox+CxEeGpthCOn4VFxIqFaVZS4DB78LgeZleEmkQkOUEI8HWqB4V108BAEGI11u9xMj2e5myVls9iwjn4MU+ySCPz+W3RZgsOMi+OiYPhEKhGwwcIRSIQJJYZJg-2GxlCwnCuKihKEO35RBAvNy+QSUAAdFKYArsnkJakEWUKoJ5Ur0grSuUIBrUlAAG5KDL7RXKhUNJoVNodKBdVo9E1xRmQ6Gwgz6fSMDAATzqCAASiBfrzjvBI9HUAAVUMgAA8IbDTXmS2lAD4hgB6AtBAB6ykYQA))

Using `Destructurable` in this scenario helps maintain only the authenticated type while reducing the manual effort of annotating types. The compiler would infer the return type of the `useUser` function as:

```typescript
{
  readonly state: "loading";
  readonly username?: undefined;
  readonly email?: undefined;
  readonly avatar?: undefined;
} | {
  readonly state: "unauthenticated";
  readonly username?: undefined;
  readonly email?: undefined;
  readonly avatar?: undefined;
} | {
  readonly state: "authenticated";
  readonly username: string;
  readonly email: string;
  readonly avatar: string;
}
```

If developers were to manually write this return type annotation, they would have to copy the complete inferred type, which can be time-consuming and error-prone. Also, any changes to the authenticated fields would require modifications to the loading and unauthenticated types, making maintenance more difficult.

By using `Destructurable`, developers can avoid this manual effort without sacrificing type safety or readability. It provides an easier way to express the desired type annotations, enabling developers to focus on writing clean and maintainable code.

I understand your concerns, and the points you've raised are valid. However, `Destructurable` can be helpful in certain scenarios where developers want to reduce the effort of manually annotating types that mimic compiler-generated types. It may not be a perfect solution for all use cases or everyone's preferences, but it could be useful for some developers who encounter situations where its features align with their requirements.

Thank you once again for sharing your perspective, and I'm grateful for the opportunity to engage in this discussion. Your insights help make the TypeScript community a better place for all.

source/destructurable.d.ts
```diff
@@ -0,0 +1,47 @@
+import {type Simplify} from './simplify';
+import {type UnionToIntersection} from './union-to-intersection';
+
+/**
+Convert a union type of objects to a destructurable type.
+
```
sindresorhus, line 6: There needs to be a more in-depth explanation of the problem space and how it's solved here, and why the type is useful.

sindresorhus:
Closing for lack of response.

https://github.com/sindresorhus/type-fest/issues/662

PR #597, opened 2023-04-17
Add `OverrideProperties` type

nidomiro:
Adds the `OverrideProperties` type.
It is basically a `Merge` with the constraint, that you can only override properties.

Example:
```ts
type Foo = {
	a: string
	b: string
}
type Bar = OverrideProperties<Foo, { b: number }>
//=> type Bar = { a: string, b: number }

type Baz = OverrideProperties<Foo, { c: number }>
// error TS2559: Type '{ c: number; }' has no properties in common with type 'Partial{ a: unknown; b: unknown; }>'.

```

### DIFF
--- a/source/override-properties.d.ts
+++ b/source/override-properties.d.ts
@@ -1,7 +1,7 @@
 import type {Merge} from './merge';
 
 /**
-Override existing properties of the given type. Similar to Merge, but enforces that the original type has the properties you want to override.
+Override existing properties of the given type. Similar to `Merge`, but enforces that the original type has the properties you want to override.
 
 This is useful when you want to override existing properties with a different type and make sure that these properties really exist in the original.
 
@@ -17,6 +17,7 @@ type Bar = OverrideProperties<Foo, {b: number}>
 type Baz = OverrideProperties<Foo, {c: number}>
 // error TS2559: Type '{c: number}' has no properties in common with type 'Partial{a: unknown; b: unknown}>'.
 ```
+
 @category Object
 */
 export type OverrideProperties<
--- a/source/override-properties.d.ts
+++ b/source/override-properties.d.ts
@@ -1,7 +1,7 @@
 import type {Merge} from './merge';
 
 /**
-Override existing property types in TOriginal with the types in TOverride.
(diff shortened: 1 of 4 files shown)


nidomiro:
sure, I thought I did.
Maybe you can point out what I'm missing?

sindresorhus:
> Please help review the other open pull requests. If there are no open pull requests, provide some feedback on some of the open issues.

It takes a lot of effort to review all new pull requests. If people submitting new types could also help review other pull requests, it would lighten my load. 🙏

source/override-properties.d.ts
```diff
@@ -0,0 +1,24 @@
...
+	a: string
+	b: string
+}
+type Bar = OverrideProperties<Foo, {b: number}>
+//=> {a: string, b: number}
+
+type Baz = OverrideProperties<Foo, {c: number}>
+// error TS2559: Type '{c: number}' has no properties in common with type 'Partial{a: unknown; b: unknown}>'.
+```
+*/
```
sindresorhus, line 20: You need to add categories. See other types for example.

source/override-properties.d.ts
```diff
@@ -0,0 +1,24 @@
+import type {Merge} from './merge';
+
+/**
+Override existing property types in TOriginal with the types in TOverride.
```
sindresorhus, line 4: This should match the readme description.

source/override-properties.d.ts
```diff
@@ -0,0 +1,24 @@
...
+
+This is useful when you want to override existing properties with a different type and make sure that these properties really exist in TOriginal.
```
sindresorhus, line 6:
```suggestion
This is useful when you want to override existing properties with a different type and make sure that these properties really exist in the original.
```

sindresorhus (changes requested):

source/override-properties.d.ts
nidomiro: I didn't see that, thanks for pointing out

source/override-properties.d.ts
nidomiro: Done. I also rearranged the entry in the readme to be right after the merge types.

sindresorhus:
Thanks :)

PR #605, opened 2023-04-27
Bump node and TypeScript versions in GitHub Actions

charpeni:
Bumping node and TypeScript versions in `.github/workflows/main.yml`.

Will add more thoughtful comments in a review.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
-          - 19
+          - 20
           - 18
           - 16
           - 14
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -28,14 +28,15 @@ jobs:
...
+          - "latest"
+          - "~4.9.0"
+          - "~4.8.0"
+          - "~4.7.0"
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 16
+          node-version: 18
```
charpeni, line 39: Node 18 is the latest LTS.

.github/workflows/main.yml
```diff
@@ -28,14 +28,15 @@ jobs:
       fail-fast: false
       matrix:
         typescript-version:
-          - 'latest'
-          - '~4.8.0'
-          - '~4.7.0'
+          - "latest"
```
charpeni, line 31: `latest` is now TypeScript 5+.

.github/workflows/main.yml
```diff
@@ -28,14 +28,15 @@ jobs:
...
+          - "~4.9.0"
```
charpeni, line 32: `type-fest` supports TypeScript 4.7+, so we should make sure to add coverage for TypeScript 4.9 too.

.github/workflows/main.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
-          - 19
+          - 20
```
charpeni, line 13: Node 19 is now in maintenance mode, and Node 20 is out as the current version (will be promoted as active LTS on October 24, 2023).

Issue #598, opened 2023-04-17
Add typescript 4.7 as a peer dependency?


patrick-mcdougle:
Would you consider putting typescript 4.7 as a peer dependency so people who install this transitively (via some other library) can get notified that their (older) version of typescript isn't going to work with this library? Could save developers some time troubleshooting.

PR #606, opened 2023-04-27
Add `typescript` as peer dependency

charpeni:
Resolves https://github.com/sindresorhus/type-fest/issues/598.

`type-fest` requires TypeScript >= 4.7, let's make that explicit via a peer dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,9 @@
 		"typescript": "^5.0.2",
 		"xo": "^0.53.1"
 	},
+	"peerDependencies": {
+		"typescript": "^4.7.0 || ^5.0.0"
+	},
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -40,6 +40,9 @@
 		"typescript": "^5.0.2",
 		"xo": "^0.53.1"
 	},
+	"peerDependencies": {
+		"typescript": "^4.7.0 || ^5.0.0"
```
sindresorhus, line 44:
```suggestion
		"typescript": ">=4.7.0"
```

We should allow any version from 4.7.0 and above. I realize a TS v6 is far off, but still.

PR #609, opened 2023-04-28
Fix OverrideProperties doesn't throw an error #608

### LINKED ISSUE #608: OverrideProperties doesn't throw an error if it has one correct property
I tried it in TypeScript 5.0.4.

```
type Foo = {
  a: string
  b: string
}

// It works well.
type Baz = OverrideProperties<Foo, {c: number}>

// It should throw an error, but it doesn't.
type Baz2 = OverrideProperties<Foo, {
  b: number
  c: number
}>
```

jinta4020:
Fixes #608, used Mapping Types instead of Partial.

### DIFF
--- a/source/override-properties.d.ts
+++ b/source/override-properties.d.ts
@@ -25,5 +25,5 @@ type Fizz = OverrideProperties<Foo, {b: number; c: number}>
 */
 export type OverrideProperties<
 	TOriginal,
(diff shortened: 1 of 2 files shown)


sindresorhus:
@nidomiro Can you review?

nidomiro (approved):
@jinta4020 thanks for the catch and fix 👍 

@sindresorhus I can confirm the bug and that the fix works

Issue #273, opened 2021-09-27
Proposal: `RequiredDeep`


osdiab:
similar to PartialDeep but the inverse, where all KV pairs are required.

sindresorhus:
What's the use-case?

osdiab:
In my application I have the concept of an "in progress" form which is stored in one DB table, and then the "completed and finalized" form data which is stored in another DB table. The schemas are expressed as Zod schemas and saved in the database as JSON blobs (since the schema is still in flux and we wanted flexibility).

Some data is nested, some are optional in that in-progress form and some are not, but the thing that's definitely true is that all leaves need to be there in the finalized version of the form data. So it's more ergonomic for me to express that type as a deeply required version of the in-progress form.

There are other ways around it but yeah figured that I'm not the only person who might have some usecase for this kind of type.

sindresorhus:
Makes sense. Accepted.

andersonba:
Up

sindresorhus:
If anyone wants to work on this, see the initial attempt and feedback: https://github.com/sindresorhus/type-fest/pull/401

moltar:
https://github.com/ts-essentials/ts-essentials/tree/master/lib/deep-required

PR #401, opened 2022-05-29
Add `RequiredDeep` type

ilchenkoArtem:
Fixes #273

PR #614, opened 2023-05-11
Add `RequiredDeep` type

jarpoole:
Closes https://github.com/sindresorhus/type-fest/issues/273

This PR is based on a past attempt (see https://github.com/sindresorhus/type-fest/pull/401) with all the feedback integrated. The only exception is that I did not change `args` to `arguments` because as was helpfully noted this is an error inside a strict context like a module.


jarpoole:
CI fails because of https://github.com/xojs/xo/issues/718. Seems like this was fixed in https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2076. I don't really understand what is going wrong here...

source/required-deep.d.ts
```diff
@@ -0,0 +1,62 @@
+import {BuiltIns} from './internal';
+
+type ExcludeUndefined<T> = Exclude<T, undefined>;
+
+/**
+Create a type from another type with all keys and nested keys set to required.
+
```
sindresorhus, line 7: Missing use-cases.

source/required-deep.d.ts
```diff
@@ -0,0 +1,62 @@
...
+type RequiredSettings = RequiredDeep<Settings>;
+Result: {
+	textEditor: {
+		fontSize: number;
+		fontColor: string;
+		fontWeight: number;
+	}
+	autocomplete: boolean;
+	autosave: boolean;
+}
```
sindresorhus, line 31: This should be a proper code comment

sindresorhus:
Is this case handled? https://github.com/sindresorhus/type-fest/pull/401#issuecomment-1169144534

source/required-deep.d.ts
```diff
@@ -0,0 +1,62 @@
...
+export type RequiredDeep<T, E extends ExcludeUndefined<T> = ExcludeUndefined<T>> = E extends BuiltIns
+	? E
+	: E extends Map<infer KeyType, infer ValueType>
+		? Map<RequiredDeep<KeyType>, RequiredDeep<ValueType>>
+		: E extends Set<infer ItemType>
+			? Set<RequiredDeep<ItemType>>
+			: E extends ReadonlyMap<infer KeyType, infer ValueType>
+				? ReadonlyMap<RequiredDeep<KeyType>, RequiredDeep<ValueType>>
+				: E extends ReadonlySet<infer ItemType>
+					? ReadonlySet<RequiredDeep<ItemType>>
```
sindresorhus, line 49: Looks like `WeakSet` and `WeakMap` handling is missing.

sindresorhus:
Should it handle promises? I looked at some different implementations of this kind of type and noticed this one handles promises: https://github.com/ts-essentials/ts-essentials/blob/master/lib/deep-required/index.ts

source/required-deep.d.ts
jarpoole: Done!

source/required-deep.d.ts
jarpoole: Done!

jarpoole:
> Should it handle properties that are promises? I looked at some different implementations of this kind of type and noticed this one handles promises: https://github.com/ts-essentials/ts-essentials/blob/master/lib/deep-required/index.ts

Yes! I have added support.

source/required-deep.d.ts
jarpoole: Good call!

jarpoole:
> Is this case handled? [#401 (comment)](https://github.com/sindresorhus/type-fest/pull/401#issuecomment-1169144534)

I *think* I've handled that case now but I'm not 100% sure.

source/required-deep.d.ts
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+Test if the given function has multiple call signatures.
+
+Needed to handle the case of a single call signature with properties.
+
+Multiple call signatures cannot currently be supported due to a TypeScript limitation.
+@see https://github.com/microsoft/TypeScript/issues/29732
+*/
+type HasMultipleCallSignatures<T extends (...arguments: any[]) => unknown> =
```
sindresorhus, line 88: Use the type in `source/internal.d.ts`.

sindresorhus (changes requested):

sindresorhus:
Thanks :)

Issue #616, opened 2023-05-13
WritableKeysOf 


kkmuffme:
get only keys that do not have `readonly` set

There are already `Writable` and `RequiredKeysOf`. A `WritableKeysOf` would be useful

sindresorhus:
👍

sindresorhus:
Implementation: https://github.com/sindresorhus/type-fest/issues/211#issuecomment-844057983

Issue #211, opened 2021-05-19
Add a `ReadonlyKeys` which finds readonly keys


voxpelli:
I needed to find out what keys that are assignable and remove all keys but those, but found no way in `type-fest` or stock `typescript`, but did stumble upon [this trick](https://stackoverflow.com/a/52473108/20667) from @mattmccutchen on Stack Overflow.

Making an issue here to keep track of this, if I get time later then I may wrap up a PR, else feel free to make one for this.

The example from Stack Overflow is:

```typescript
// https://github.com/Microsoft/TypeScript/issues/27024#issuecomment-421529650
type IfEquals<X, Y, A, B> =
    (<T>() => T extends X ? 1 : 2) extends
    (<T>() => T extends Y ? 1 : 2) ? A : B;

// Alternatively:
/*
type IfEquals<X, Y, A, B> =
    [2] & [0, 1, X] extends [2] & [0, 1, Y] & [0, infer W, unknown]
    ? W extends 1 ? B : A
    : B;
*/

type WritableKeysOf<T> = {
    [P in keyof T]: IfEquals<{ [Q in P]: T[P] }, { -readonly [Q in P]: T[P] }, P, never>
}[keyof T];
```

PR #619, opened 2023-05-15
Add `WritableKeysOf` and `ReadonlyKeysOf` type

kkmuffme:
Fix https://github.com/sindresorhus/type-fest/issues/211 
Fix https://github.com/sindresorhus/type-fest/issues/616 
+ opposite since it's very similar and might be useful

### DIFF
--- a/source/readonly-keys-of.d.ts
+++ b/source/readonly-keys-of.d.ts
@@ -1,0 +1,37 @@
+/**
+Extract all readonly keys from the given type.
+
+This is useful when you want to create a new type that contains readonly keys only.
+
+@example
+```
+import type {ReadonlyKeysOf, Except} from 'type-fest';
+
+interface User {
+	name: string;
+	surname: string;
+	readonly id: number;
+}
+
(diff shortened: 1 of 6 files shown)


source/readonly-keys-of.d.ts
```diff
@@ -0,0 +1,37 @@
...
+    id: 123,
+};
+```
+
+@category Utilities
+*/
+// https://github.com/Microsoft/TypeScript/issues/27024#issuecomment-421529650
+type IfEquals<X, Y, A, B> =
+	(<T>() => T extends X ? 1 : 2) extends
+	(<T>() => T extends Y ? 1 : 2) ? A : B;
```
sindresorhus, line 28: Use https://github.com/sindresorhus/type-fest/blob/bba240fb01ac0b540cb8f163d17854046fa08905/source/is-equal.d.ts#L12 

4 type parameters are pretty unreadable. Better to use extends:

```
IsEqual<Value[0], Item> extends true
			? true
			: Includes<rest, Item>
```

PR #624, opened 2023-05-22
Add `SetReadonly` type

### LINKED ISSUE #620: Add `SetReadonly`
Alongside

- [SetOptional](https://github.com/sindresorhus/type-fest/blob/main/source/set-optional.d.ts) - Create a type that makes the given keys optional.
- [SetRequired](https://github.com/sindresorhus/type-fest/blob/main/source/set-required.d.ts) - Create a type that makes the given keys required.
- [SetNonNullable](https://github.com/sindresorhus/type-fest/blob/main/source/set-non-nullable.d.ts) - Create a type that makes the given keys non-nullable.

it would be helpful for me to have a `SetReadonly` which sets all given keys to `readonly`

Shinigami92:
close #620

### DIFF
--- a/source/set-readonly.d.ts
+++ b/source/set-readonly.d.ts
@@ -1,0 +1,38 @@
+import type {Except} from './except';
+import type {Simplify} from './simplify';
+
+/**
+Create a type that makes the given keys readonly. The remaining keys are kept as is.
+
+Use-case: You want to define a single model where the only thing that changes is whether or not some of the keys are readonly.
+
+@example
+```
+import type {SetReadonly} from 'type-fest';
+
+type Foo = {
+	a: number;
+	readonly b: string;
+	c: boolean;
+}
+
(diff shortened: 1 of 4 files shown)


source/set-readonly.d.ts
```diff
@@ -0,0 +1,33 @@
...
+// }
+```
+
+@category Object
+*/
+export type SetReadonly<BaseType, Keys extends keyof BaseType> =
+	Simplify<
+	Except<BaseType, Keys> &
+	Readonly<Pick<BaseType, Keys>>
+	>;
```
sindresorhus, line 33: Would be great if you could implement the same fix as this: https://github.com/sindresorhus/type-fest/commit/6a82900b45c7c2bcf3f5964dd07bc5e8a3999dac

Shinigami92: done

PR #626, opened 2023-05-29
fix(PackageJson): disallow undefined in exports

privatenumber:
Fixes https://github.com/sindresorhus/type-fest/pull/555/files#r1209049594

### DIFF
--- a/source/package-json.d.ts
+++ b/source/package-json.d.ts
@@ -209,25 +209,11 @@ declare namespace PackageJson {
 	export type Dependency = Partial<Record<string, string>>;
 
 	/**
-	Conditions which provide a way to resolve a package entry point based on the environment.
-	*/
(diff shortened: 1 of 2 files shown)


source/package-json.d.ts
```diff
@@ -208,26 +208,12 @@ declare namespace PackageJson {
...
-	| 'require'
-	| 'node'
-	| 'node-addons'
-	| 'deno'
-	| 'browser'
-	| 'electron'
-	| 'react-native'
-	| 'default',
-	string
-	>;
```
privatenumber, line 225: I can no longer use `ExportCondition` as the key type. However, the only requirement for an export condition is that it should be a string.

Keeping a list of common export conditions will only add to the maintenance workload, and provide little value for autocompletion, so I suggest removing it.

Issue #636, opened 2023-06-14
`LastArrayElement<[...T[], U]>` should be `U`, but it is `T | U`


ootidea:
related: #635

ahrjarrett:
@sindresorhus asked me to cross post a reply I made to a comment by @ootidea on #647 here.

Looking at this issue, I'm not sure that they're directly applicable (I agree that this issue is probably a bug), but since it might lead to other issues being created, I'll go ahead and put them here.

Please don't forget to tag me if you'd like me to discuss further, otherwise I'll probably miss it in my inbox!

@ootidea's comment:
https://github.com/sindresorhus/type-fest/pull/647#issuecomment-1636404282
> LastArrayElement<[]> should be undefined, but it is never.
LastArrayElement<T[]> should be T | undefined, but it is T.
LastArrayElement<[...1[], 2]> should be 2, but it is 1 | 2.
LastArrayElement<[1, 2?]> should be 1 | 2, but it is never.
LastArrayElement<[1?, 2?]> should be 1 | 2 | undefined, but it is never.
LastArrayElement<[0, 1, 2?, ...3[]]> should be 1 | 2 | 3, but it is 0 | 1 | 2 | 3 | undefined.
LastArrayElement<any> should be any, but it is unknown.

@ahrjarrett's response:
https://github.com/sindresorhus/type-fest/pull/647#issuecomment-1636584082
> For instance, the expectation that LastArrayElement<[]> return undefined is useful only as the return type of a term-level JavaScript function that gets the last element of an array, or returns undefined if the array is empty.
> My 2 issues with this:
> Array.prototype.last does not exist (AFAIK). You could write a function called last that behaves the way you describe -- but you could also write a function that returns an error message if the array is empty, is throws, or does any number of things instead. Choosing undefined out of any other option feels somewhat arbitrary.
> This doesn't seem to be in alignment with the design goals of type-fest, since it doesn't export (AFAIK) any term-level API. If someone tried to compose LastArrayElement with another type-level function, getting undefined would be very surprising behavior, because the only way that would happen would be if the value undefined was in the array.

PR #647, opened 2023-07-14
optimization: refactor `LastArrayElement` to avoid recursion

ahrjarrett:
Hello type-fest!

I was poking around the repo and noticed that `LastArrayElement` was using recursion to iterate until reaching the final element, which is unnecessary since TS 4.2:

[Playground](https://www.typescriptlang.org/play?noImplicitAny=false&target=6&jsx=0&ts=4.2.3#code/HYQwtgpgzgDiDGEAEAVArjANsg3gKCSQgA8YB7AJwBckqBPGZTEKKgHgCMojiqJgAJtwoQQAssEx0kaYAGtgZAO7AA2gF0AfAUJIAvEmBowHCBR59B3LqoBE2YAHMqAC1vqkAfiRcL-IUgiYhJSSAAUAJbAAGZmPgCUGl4+SDqEAFyGEABuZmlImb4klgFB4pLSqgB0NSDAdAA0SFGx5hwe3hypugVZuRR4AL54eKCQsAjIKAwQ2LmYAMqM8BHREfAgVBESSPiEY9BwiKjQNHu6JOTUtDNEAI5sIE0cmvpIqk8+HsX+3KocTRAHVoFDQyEy0RAmCgEHyl0oNHojCQUMwjz8VhBYI0rwMIB0wx0SKYLBoBhQpyqqLYqnyFNYVQgD3QWAgVWYrBpAEYGgAmBoAZi0TQFmgadMpTLYLOw7NJbDKIWkYVYFCijiQAB9DMZTBREsKkKr1VqdSYzNpCFohkA)

Since people are paying more attention to type-level performance lately, this change seemed like a quick and reasonable PR to make.

### DIFF
--- a/source/last-array-element.d.ts
+++ b/source/last-array-element.d.ts
@@ -18,11 +18,7 @@ typeof lastOf(array);
 @category Array
 @category Template literal
 */
-export type LastArrayElement<ValueType extends readonly unknown[]> =
-	ValueType extends readonly [infer ElementType]
-		? ElementType
-		: ValueType extends readonly [infer _, ...infer Tail]
-			? LastArrayElement<Tail>
-			: ValueType extends ReadonlyArray<infer ElementType>
-				? ElementType
-				: never;
+export type LastArrayElement<Elements extends readonly unknown[]>
+	= number extends Elements["length"] ? Elements extends readonly (infer Element)[] ? Element : never
+	: Elements extends readonly [...any, infer Target] ? Target
+	: never;
--- a/source/last-array-element.d.ts
+++ b/source/last-array-element.d.ts
@@ -19,6 +19,10 @@ typeof lastOf(array);
 @category Template literal
 */
 export type LastArrayElement<Elements extends readonly unknown[]>
-	= number extends Elements["length"] ? Elements extends readonly (infer Element)[] ? Element : never
-	: Elements extends readonly [...any, infer Target] ? Target
-	: never;
+		= number extends Elements['length']
+			? Elements extends ReadonlyArray<infer Element>
+				? Element
+				: never
+			: Elements extends readonly [...any, infer Target]
+				? Target
+				: never;


ootidea:
These are my opinions.

`LastArrayElement<[]>` should be `undefined`, but it is `never`.
`LastArrayElement<T[]>` should be `T | undefined`, but it is `T`.
`LastArrayElement<[...1[], 2]>` should be `2`, but it is `1 | 2`.
`LastArrayElement<[1, 2?]>` should be `1 | 2`, but it is `never`.
`LastArrayElement<[1?, 2?]>` should be `1 | 2 | undefined`, but it is `never`.
`LastArrayElement<[0, 1, 2?, ...3[]]>` should be `1 | 2 | 3`, but it is `0 | 1 | 2 | 3 | undefined`.

related: #635, #636

ahrjarrett:
Personally I think those should be handled in a separate PR. The purpose of this PR was solely to make the type not recursive.

Unless, do any of these represent regressions from `LastArrayElement`'s current behavior @ootidea ?

ahrjarrett:
Also while I agree with some of your opinions, others seem useful only in certain use cases.

For instance, the expectation that `LastArrayElement<[]>` return `undefined` is useful only as the return type of a term-level JavaScript function that gets the last element of an array, or returns `undefined` if the array is empty.

My 2 issues with this:

1. `Array.prototype.last` does not exist (AFAIK). You could write a function called `last` that behaves the way you describe -- but you could also write a function that returns an error message if the array is empty, is throws, or any number of things.

2. This doesn't seem to be in alignment with the design goals of `type-fest`, since it doesn't export (AFAIK) any term-level API. If someone tried to compose `LastArrayElement` with another type-level function, getting `undefined` would be very surprising behavior, because the only way that would happen would be if the value `undefined` was in the array.

sindresorhus:
I agree with @ahrjarrett. That's out of scope for this pull request.

sindresorhus:
@ahrjarrett Would be great if you could copy-paste your comments to the relevant issues instead: https://github.com/sindresorhus/type-fest/issues/635, https://github.com/sindresorhus/type-fest/issues/636

PR #650, opened 2023-07-23
fix(ReadonlyDeep): lessen instantiation excessively deep errors

### LINKED ISSUE #467: `Type instantiation is excessively deep` in `ReadonlyDeep`
Hi,
I recently updated my declaration of `ReadonlyDeep` to use the new call signature handling from #359, however, I get `error TS2589: Type instantiation is excessively deep and possibly infinite.` :(

See the repro in the [playground](https://www.typescriptlang.org/play?ts=4.8.2#code/JYWwDg9gTgLgBDAnmApnA3nASighgEwgDsAbRAERRTDgF84AzKCEOAciVQFoGUBnGGwDcAKFCRYcAFRxcfOACkAigFcUURI2as2AKwCOajcJEiGKogGMYwYoyIAKFAC5seQqQpUwAHmVHEAD4ASlcANwhgfAwRAEgAEidg0VogA) for details.

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/467">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/467/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

ethanresnick:
I was repeatedly getting "instantiation excessively deep" errors with `ReadonlyDeep`. After more debugging, I realized that the root issue stemmed from using `ReadonlyDeep` on a recursive type similar to `type-fest`'s `JsonValue` type. With that discovery, I was able to add a few more cases to `ReadonlyDeep` that fixed all the errors I was seeing, and should make the type more powerful/robust for everyone. The individual changes are:

1. Previously, `ReadonlyDeep<JsonValue>` [immediately triggered](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgJQKYEMAmEB2AbATwBFVUwAaOAKQGdcA1dPAV1TgF84AzKCEOAEQwCYVAFouqGjAEBuAFDyuzHAGMYwXHHQAKAB4AuFBmz5ipMAB5aDJqwB8ASkTy4buFFQxmUHHD0K7EA) an 'instantiation excessively deep' error. Now, it does not, and it returns a type that's quite close to the type you'd expect if you manually converted `JsonValue` to readonly. The produced type is not identical (which is why the test uses `expectAssignable` rather than `expectType`), because `ReadonlyDeep` doesn't/can't preserve the true recursive structure; instead, after about four or five levels of expansion, the type permits `any`. But it is much better than before. 

    In order to accomplish this, I had to add an [explicit case](https://github.com/sindresorhus/type-fest/pull/650/files#diff-c2b2b242b2574eeb84280c4a411f7e6b4a9a82a063df60eab9f9f5e20dc42e8fR61) for detecting arrays. This is the only change that was required to make `ReadonlyDeep<JsonValue>` not error, I believe. This change should also prevent errors for many other recursive types involving arrays. Previously, arrays hit the `ReadonlyObjectDeep` case, which seemed to contribute to hitting the recursion limit.

2. Adding the array case mentioned above led the tests for tuples to fail (as they started hitting this case, rather than `ReadonlyObjectDeep`), so I had to add cases to detect tuples specifically. I tried a number of approaches, including continuing to pass `T` to `ReadonlyObjectDeep` when it looked like `T` was a tuple, but the approach in this PR is the only one I found that passed all the tests. This change actually improves the returned type when the input type is a tuple with a leading spread. Specifically, `ReadonlyDeep<[...string[], number]>` [used to give](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgJQKYEMAmEB2AbATwBFVUwAaOAeTHQEcBXVOAXzgDMoIQ4AiGAmFQBadqgDOMXgG4AULIFC4AITgBeFBmz5ipMAB4A2gDpTkqMBwBzQwF1KOBiABGqKLYB80oA) `readonly [...(string | number)[], string | number]`; now it correctly produces `readonly [...string[], number]`.

### DIFF
--- a/source/readonly-deep.d.ts
+++ b/source/readonly-deep.d.ts
@@ -48,9 +48,18 @@ export type ReadonlyDeep<T> = T extends BuiltIns
 			? ReadonlyMapDeep<KeyType, ValueType>
 			: T extends Readonly<ReadonlySet<infer ItemType>>
 				? ReadonlySetDeep<ItemType>
-				: T extends object
-					? ReadonlyObjectDeep<T>
-					: unknown;
+				: // Identify tuples to avoid converting them to arrays inadvertently; special case `readonly [...never[]]`, as it emerges undesirably from recursive invocations of ReadonlyDeep below.
+				T extends readonly [] | readonly [...never[]]
+					? readonly []
+					: T extends readonly [infer U, ...infer V]
+						? readonly [ReadonlyDeep<U>, ...ReadonlyDeep<V>]
+						: T extends readonly [...infer U, infer V]
+							? readonly [...ReadonlyDeep<U>, ReadonlyDeep<V>]
+							: T extends ReadonlyArray<infer ItemType>
+								? ReadonlyArray<ReadonlyDeep<ItemType>>
+								: T extends object
+									? ReadonlyObjectDeep<T>
+									: unknown;
 
 /**
 Same as `ReadonlyDeep`, but accepts only `ReadonlyMap`s as inputs. Internal helper for `ReadonlyDeep`.
--- a/test-d/readonly-deep.ts
+++ b/test-d/readonly-deep.ts
@@ -1,6 +1,7 @@
-import {expectType, expectError} from 'tsd';
-import type {ReadonlyDeep} from '../index';
-import type {ReadonlyObjectDeep} from '../source/readonly-deep';
+import {expectType, expectError, expectAssignable} from 'tsd';
+import type {Opaque, tag} from '../source/opaque';
+import type {ReadonlyDeep, ReadonlyObjectDeep} from '../source/readonly-deep';
+import type {JsonValue} from '../source/basic';
 
 type Overloaded = {
 	(foo: number): string;
@@ -15,6 +16,17 @@ type Namespace = {
 type NamespaceWithOverload = Overloaded & {
 	baz: boolean[];
 };
+
+type OpaqueObjectData = {a: number[]} | {b: string};
+type OpaqueObject = Opaque<OpaqueObjectData, {token: unknown}>;
+
+type ReadonlyJsonValue =
+  | {readonly [k: string]: ReadonlyJsonValue}
+  | readonly ReadonlyJsonValue[]
+  | number
+  | string
+  | boolean
+  | null;
 
 const data = {
 	object: {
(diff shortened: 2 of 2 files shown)


ethanresnick:
@sindresorhus Anything I can do to help move this forward, or more info you need from me? I'm running into these 'instantiation excessively deep' errors quite a lot with `ReadonlyDeep` and would love to see this merged.

source/readonly-deep.d.ts
```diff
@@ -55,16 +56,33 @@ export type ReadonlyDeep<T> = T extends BuiltIns
...
+	>;
+}>;
+
+/**
+Utility to reduce depth of TypeScript's internal type instantiation stack, to
+avoid "TS2589: Type instantiation is excessively deep and possibly infinite."
+
+Details: https://github.com/kysely-org/kysely/pull/483
+*/
+type DrainOuterGeneric<T> = [T] extends [unknown] ? T : never;
```
sindresorhus, line 88: Move this to internal.d.ts

source/readonly-deep.d.ts
```diff
@@ -55,16 +56,33 @@ export type ReadonlyDeep<T> = T extends BuiltIns
...
-};
+type ReadonlyObjectDeep<ObjectType extends object> = DrainOuterGeneric<{
+	readonly [KeyType in keyof ObjectType]: DrainOuterGeneric<
+	ReadonlyDeep<ObjectType[KeyType]>
+	>;
+}>;
+
+/**
+Utility to reduce depth of TypeScript's internal type instantiation stack, to
+avoid "TS2589: Type instantiation is excessively deep and possibly infinite."
```
sindresorhus, line 84: Don't hard-wrap.

sindresorhus (changes requested):

sindresorhus:
It would be nice to have some tests that would produce "Type instantiation is excessively deep and possibly infinite" before this change.

source/readonly-deep.d.ts
```diff
@@ -36,7 +37,7 @@ Note that types containing overloaded functions are not made deeply readonly due
 @category Set
 @category Map
 */
-export type ReadonlyDeep<T> = T extends BuiltIns
+export type ReadonlyDeep<T> = T extends BuiltIns | Opaque<unknown>
```
sindresorhus, line 40: The `Opaque` addition could use a short code comment on why it's there.

ethanresnick:
Hi @sindresorhus! Ok, I've done substantially more debugging here and have updated the PR accordingly. I'll update the OP soon to explain each change. Every change has a test, I believe.

ethanresnick:
@sindresorhus How does it look now (if you have a chance to look)? Thanks!

ethanresnick:
Thanks @sindresorhus!

ehoogeveen-medweb:
By the way, does `WritableDeep` suffer from the same issues or is it unaffected? The implementations were very similar before this change.

sindresorhus:
https://github.com/sindresorhus/type-fest/issues/664

PR #660, opened 2023-08-04
Add `UnknownRecord` type

### LINKED ISSUE #659: Expose `UnknownRecord` in the public API
Currently `UnknownRecord` can be imported only from an internal file:
```ts
import { UnknownRecord } from 'type-fest/source/internal.js';
```

However, this type is very handy and I use it frequently in my app. I believe it's fair enough to expose it like the other types:
```ts
import { UnknownRecord } from 'type-fest';
```

MatanYadaev:
Closes #659

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -110,7 +110,7 @@ Click the type names for complete docs.
 
 - [`EmptyObject`](source/empty-object.d.ts) - Represents a strictly empty plain object, the `{}` value.
 - [`IsEmptyObject`](source/empty-object.d.ts) - Returns a `boolean` for whether the type is strictly equal to an empty plain object, the `{}` value.
-- [`UnknownRecord`](source/unknown-record.d.ts) - Matches any unknown record.
+- [`UnknownRecord`](source/unknown-record.d.ts) - Represents an object with unknown value. You probably want this instead of {}.
 - [`Except`](source/except.d.ts) - Create a type from an object type without certain keys. This is a stricter version of [`Omit`](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys).
 - [`Writable`](source/writable.d.ts) - Create a type that strips `readonly` from all or some of an object's keys. The inverse of `Readonly<T>`.
 - [`WritableDeep`](source/writable-deep.d.ts) - Create a deeply mutable version of an `object`/`ReadonlyMap`/`ReadonlySet`/`ReadonlyArray` type. The inverse of `ReadonlyDeep<T>`. Use `Writable<T>` if you only need one level deep.
--- a/readme.md
+++ b/readme.md
@@ -110,7 +110,7 @@ Click the type names for complete docs.
 
 - [`EmptyObject`](source/empty-object.d.ts) - Represents a strictly empty plain object, the `{}` value.
 - [`IsEmptyObject`](source/empty-object.d.ts) - Returns a `boolean` for whether the type is strictly equal to an empty plain object, the `{}` value.
-- [`UnknownRecord`](source/unknown-record.d.ts) - Represents an object with unknown value. You probably want this instead of {}.
+- [`UnknownRecord`](source/unknown-record.d.ts) - Represents an object with `unknown` value. You probably want this instead of `{}`.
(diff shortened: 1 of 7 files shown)


readme.md
```diff
@@ -110,6 +110,7 @@ Click the type names for complete docs.
 
 - [`EmptyObject`](source/empty-object.d.ts) - Represents a strictly empty plain object, the `{}` value.
 - [`IsEmptyObject`](source/empty-object.d.ts) - Returns a `boolean` for whether the type is strictly equal to an empty plain object, the `{}` value.
+- [`UnknownRecord`](source/unknown-record.d.ts) - Matches any unknown record.
```
MatanYadaev, line 113: I copied this description from the internal jsDoc. Is it good enough?

source/internal.d.ts
```diff
@@ -3,6 +3,8 @@ import type {Simplify} from './simplify';
 import type {Trim} from './trim';
 import type {IsAny} from './is-any';
 
+export type {UnknownRecord} from './unknown-record';
```
MatanYadaev, line 6: I exported this type here to prevent breaking changes.

source/unknown-record.d.ts
```diff
@@ -0,0 +1,6 @@
+/**
+Matches any unknown record.
+
+@category Object
+*/
+export type UnknownRecord = Record<PropertyKey, unknown>;
```
MatanYadaev, line 6: Should I add examples and tests?

tommy-mitchell: Yes

readme.md
tommy-mitchell: Maybe:

> Represents an object with any value. You probably want this instead of `{}`.

source/internal.d.ts
sindresorhus: It's better to update the files that use this to use the public type instead.

sindresorhus:
Make sure you adhere to: https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md#submitting-a-new-type

source/internal.d.ts
MatanYadaev: @sindresorhus I did it, for example at `merge-deep.d.ts`. But removing this export from this file will introduce a breaking change.
On my private projects, I have many `import { UnknownRecord } from 'type-fest/source/internal.js';` that will get broken due to removing this export.

MatanYadaev:
@sindresorhus I added examples and tests, let me know if it's good now :)

source/internal.d.ts
sindresorhus: > On my private projects, I have many import { UnknownRecord } from 'type-fest/source/internal.js'; that will get broken due to removing this export.

If you use internal stuff, it's generally on you if something breaks. But I'll allow it. In the next major version of type-fest, it will not be possible to import internal types, so it will not be a problem then.

sindresorhus:
You still have unresolved feedback: https://github.com/sindresorhus/type-fest/pull/660/files#r1284735104

readme.md
MatanYadaev: I'm not sure it's better, the word `any` might confuse here. WDYT about "Represents an object with unknown keys and values"?

sindresorhus: @MatanYadaev Your version can be confusing too as the keys are not `unknown`.

I would go with:

> Represents an object with `unknown` value. You probably want this instead of `{}`.

readme.md
```diff
@@ -110,6 +110,7 @@ Click the type names for complete docs.
 
 - [`EmptyObject`](source/empty-object.d.ts) - Represents a strictly empty plain object, the `{}` value.
 - [`IsEmptyObject`](source/empty-object.d.ts) - Returns a `boolean` for whether the type is strictly equal to an empty plain object, the `{}` value.
+- [`UnknownRecord`](source/unknown-record.d.ts) - Represents an object with unknown value. You probably want this instead of {}.
```
tommy-mitchell, line 113:
```suggestion
- [`UnknownRecord`](source/unknown-record.d.ts) - Represents an object with `unknown` value. You probably want this instead of `{}`.
```

sindresorhus:
Thanks :)

PR #668, opened 2023-08-24
fix(jsonify): support object with any

### LINKED ISSUE #667: Jsonify makes Record<string, any> disappear
Here's a [TS Playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBDAnmApnA3nAUgZwgO2ADNE4BfOIqCEOAIiVQFoiUcY6BuAKG8bQAiAQxhC4AXmx5CJADzpucOGCGIANhCEATAFxwASigDG0LbPZRg+AOYAaOEPyIAfDzKveW42qFQ0J-HY4XThhUR4AvDUUADoNawAKLRiVdU0tAEpuIA):

```ts
import type { Jsonify } from "type-fest";

type Data = Jsonify<{
  payload: Record<string, any>;
}>;

declare const d: Data;
console.log(d.payload) // <-- Property 'payload' does not exist on type 'JsonifyObject<{}>'.
```

This feels like a bug to me. It's causing an issue for me in the upcoming version of Remix: https://github.com/remix-run/remix/issues/7246

My current workaround is to change the type to `unknown` instead of `any` (https://github.com/edmundhung/conform/pull/272 which will hopefully be merged soon). I think is the proper thing to do, but sometimes that's hard because you may not control the type. I think it's more correct to keep the record around on the type.

kentcdodds:
This only adds a test. I'm not sure I have enough TS knowledge to actually solve the issue.

Closes #667

### DIFF
--- a/source/jsonify.d.ts
+++ b/source/jsonify.d.ts
@@ -7,7 +7,7 @@ import type {NegativeInfinity, PositiveInfinity} from './numeric';
 import type {TypedArray} from './typed-array';
 
 type HasAnyProperty<T extends object> = {
-    [K in keyof T]: IsAny<T[K]> extends true ? K : never;
+	[K in keyof T]: IsAny<T[K]> extends true ? K : never;
 }[keyof T] extends never ? false : true;
 
 // Note: The return value has to be `any` and not `unknown` so it can match `void`.
--- a/source/jsonify.d.ts
+++ b/source/jsonify.d.ts
@@ -6,6 +6,10 @@ import type {IsNever} from './is-never';
 import type {NegativeInfinity, PositiveInfinity} from './numeric';
 import type {TypedArray} from './typed-array';
 
+type HasAnyProperty<T extends object> = {
+    [K in keyof T]: IsAny<T[K]> extends true ? K : never;
+}[keyof T] extends never ? false : true;
+
 // Note: The return value has to be `any` and not `unknown` so it can match `void`.
 type NotJsonable = ((...arguments_: any[]) => any) | undefined | symbol;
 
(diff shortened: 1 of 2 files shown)


source/jsonify.d.ts
```diff
@@ -6,6 +6,10 @@ import type {IsNever} from './is-never';
 import type {NegativeInfinity, PositiveInfinity} from './numeric';
 import type {TypedArray} from './typed-array';
 
+type HasAnyProperty<T extends object> = {
+    [K in keyof T]: IsAny<T[K]> extends true ? K : never;
```
kentcdodds, line 10: Whoops, sorry about that.

```suggestion
	[K in keyof T]: IsAny<T[K]> extends true ? K : never;
```

kentcdodds:
Whelp, I did my best 😅 This is outside my wheelhouse.

kentcdodds:
Ok, I'm working on this more and I've discovered some *very* odd behavior that I think may be causing this issue...

Here's the code I'm playing with:

```ts
type UndefinedExtendsHi = undefined extends 'hi' ? true : false
```

Of course, `UndefinedExtendsHi` should be `false` ([playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAqgdgEwgMwJZwggogD2BRAZwAlUoBeKAV0RXUygjwIUKgHIALVdqAfijAATlWgAuKMgCGAG0IQAUAGMA9nELAoQiISozgE+EjQZszIqQqTZ8hUA)).

However, if I place this code in [`source/internal.d.ts`](https://github.com/sindresorhus/type-fest/blob/main/source/internal.d.ts), I get `true`!? 🧐

image

It's also incorrect if I put it in `jsonify.d.ts`:

image

Interestingly, if I put it in the jsonify test it behaves as expected:

image

So something is *really* whacky in the `source` directory. I would love for someone with more knowledge to help me understand what's going on here.

kentcdodds:
Oh dear... It looks like there's an inconsistency between `.ts` and `.d.ts` files and it's a reported bug in TypeScript: https://github.com/microsoft/TypeScript/issues/33174 🙈

Adding `strictNullChecks: true` to the tsconfig seems to fix the issue.

sindresorhus:
Closing as this is not a problem with type-fest. We already recommend strict TS config in the readme.

PR #672, opened 2023-08-25
feat: add Tagged for composable tagged types

### LINKED ISSUE #665: Support composable tagged types
Type-fest's current implementation of opaque types is great for attaching _one_ tag to a type. However, it does not support attaching _more than one_ tag to the same type, which is often useful to represent multiple validations that have been performed on a value, or multiple "aspects" of how it can be interpreted.

For example, the Typescript discussions about tagged types [use the example](https://github.com/microsoft/TypeScript/pull/33038) of a string that can be an absolute path, a normalized path, _or both_. Similarly, in my code, I have a tagged type representing URLs and a tagged type representing the ids used by a particular sub-system, but these subsystem ids happen to also be URLs. Therefore, I'd like to have:

```ts
type Url = Opaque<string, "Url">;
type SubsystemId = Opaque<Url, "SubsystemId">;

// NB: SubsystemId above is ultimately the same as:
// string & Tagged<"Url"> & Tagged<"SubsystemId">
```

That way, I can pass a `SubsystemId` to any utility function I have that takes a `Url`, too.

If you try this sort of thing in type-fest today, though, [you run into problems](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgeTAQwI4FcCmcC+cAZlBCHAOQwCeYOAtETgM4wUDcAsAFA11wBBAEbMIAGywwcABTQwAFnAC8KdNhwAeVlGAA7AOYAaSsNESpshRQB8XXrTwA5aCDRjgALxwATS4pWomLhaMDoGxhTOUK7uXr5y8jZ2fHjS8hC6OI5YIEI4UMqqQZraekaUaRlZOXlQSTw83jgAxmJoUHhEWLrNMMAZcMDMpuKSMgkaACpwOAAeUrrezHClBtYAFMAwAFxwkwCUu1uDy9MAZIIioxYJdk2t7Z3dvf26cB3YwB3MlZnZuflNjs4L9qgCoIc4AA3CDAbx2HgAekRKywYEgzDwAHc8G4OmhvNQ4PI0FDcdC3Lg4Ao5MS0MsqtS0Po4HIYGhmvIfNSIIMYMsUgA6RotNodODNDKsOC6FxuTw+Py7KIxBXxBQI7jImUQLHGHFwLFoXTwGC8qHy7xyPDHLqwLlQfXyYCclbpLBibys9FiIloFYtDJe9ks43eJEosQ4eBYZboZgCrkUiR4M3EZ59KXvHCfMIsmmm5k8YBETbDK7mcYKday6LyuJ+fb7RA8OBwbUKIYStxiN0QD1eohoYBiYx5ZpoWOppPoBQnOBNIh6LY4X1wAD6spg69ZcDA6UyMpq+Tsbe1elYOAJxmOzCwzWaOB8y3Hk8x1KTKTgECIcAABrWqoNgkf50i+T5vK27Yoh03j3tyaYATgZJQH+xhQAk+QfsacAqvWioJHAFwjJWfhQR8WBfCwoL-LUNZyrEBEKPsdj4EAA):

```ts
type AbsolutePath = Opaque<string, 'AbsolutePath'>;
type NormalizedPath = Opaque<string, 'NormalizedPath'>;
type PhoneNumber = Opaque<string, 'PhoneNumber'>;

declare function isAbsolutePath<T extends string>(it: T): it is T & AbsolutePath;
declare function requiresPhoneNumber(it: PhoneNumber): void;

// suppose we already have a value that has one tag attached to its type.
declare const normalizedPath: NormalizedPath;

// now, we want to validate it further, which should apply a second tag and
// let us pass the value to functions requiring that tag
if(isAbsolutePath(normalizedPath)) {
  // this call should fail, because the path is definitely _not_ a phone number;
  // instead, it succeeds because the type of `normalizedPath` has been 
  // reduced to `never`, rather than `NormalizedPath & AbsolutePath`
  requiresPhoneNumber(normalizedPath);
}
```

The root of the issue is that:

1. the Typescript compiler has some built-in hacks that tell it not to reduce `string & Tagged<"NormalizedPath">` to `never`, even though there's no runtime value that can actually be both a string and an object w/ a symbol-named property, which is what that type requires; 

2. however, as soon as the type becomes `string & Tagged<"NormalizedPath"> & Tagged<"AbsolutePath">`, those hacks are insufficient and the type is immediately reduced to `never`, as shown [here](https://www.typescriptlang.org/play?#code/PTAEGEHsAcEsFMAmoDOAXATgQ1gcwBZqgBmGkAtqGgJ7TwC0x86AsAFCLwDGANlhvFBdIAO3RUsuAFygAriNgBHWYJTVyAI0g8A3O3YgIMBMnTY8hEmUo06jZmnade-QbcEAVSbiQAeD5AA1vAiAHygALygAN7sAJACWIiiPNSgANpokgC6MgHBInpsAL5FBmAetMxcGLDQRMLkcDzwGKD4WCioFIIasrA8aPSwIu1YXIFdAO74sFz4oFg8PJBToLBE5VSQoAD6IpBou6ACiLJcbrNdaDsABiLwAG6ttwA0oE8hVPiQsgTfrXgAHIugcTvI0LByPB2O5QAANSKoTAjXCgABkoC8uB8iF8ACIAHKQDDkJawABeSAAClg0Ph8eFMdjcQSAIIaFDaWRoeC0+mMopAA). The compiler  sees `{ [tag]: "NormalizedPath" } & { [tag]: "AbsolutePath" }` and determines (correctly) that the strings in the  `[tag]` property cannot both be satisfied.

The way to work around this, as described [here](https://github.com/microsoft/TypeScript/pull/33290), is to use an object type to store all the tags, and use _one key per tag_ in that object. I.e., instead of `SubsystemId` ultimately being represented like:

```ts
string & { [tag]: "Url" } & { [tag]: "SubsystemId" }
```

which reduces to `never`, you represent it as:

```ts
string & { [tag]: { Url: void } } & { [tag]: { SubsystemId: void } }
```

which is equivalent to:

```ts
string & { [tag]: { Url: void, SubsystemId: void } }
```

This ends up working very well with the structural nature of Typescript, in that a function that needs its argument to only have one of the tags can still be called with an argument that has more tags (by the normal TS rules for allowing excess object keys); meanwhile, the structure of the type prevents reduction to `never`, because the `{ [tag]: { Url: void, SubsystemId: void } }` part actually _is satisfiable_, which is enough for the TS compiler hack to preserve the full type when intersecting with `string`.

All of this is to say: **supporting types with multiple tags is definitely possible, and I'd be happy to put up a PR. `UnwrapOpaque` can be made to work too. However, representing tagged types like this would be a breaking change, in two ways:**

1. The `Token` parameter on `Opaque` is currently unconstrained; with this implementation, it would have to extend `string | number | symbol`. I don't actually think that limitation is a big problem, because `Token` is usually a literal string in my experience, and, for people who want even stronger guarantees about the uniqueness/non-assignability of different opaque types, the ability to use a `unique symbol` type as the `Token` is sufficient. However, this example from the current documentation would no longer be supported:

    ```ts 
    type Person = {
      id: Opaque<number, Person>;
      name: string;
    }; 
    ```

     One would instead need to do:

    ```ts
    type PersonId = Opaque<number, "PersonId">; 

    type Person = {
      id: PersonId;
      name: string;
    }; 
    ```

    or, for the super-paranoid:
    
    ```ts
    declare const personId: unique symbol;
    type PersonId = Opaque<number, typeof personId>; 
    ```

2. Second, the underlying representation is different enough that type-fest users who've built their own utility types that inspect the current `Opaque` types might need to update their logic.

So, my questions for you @sindresorhus are: Do you think this multiple tags functionality is valuable? If so, do you think it's acceptable to break the current API in the (imo, relatively small) ways mentioned above?

If those breaks are not acceptable, another option would be to introduce a new type that's basically the same as `Opaque`, except that it supports multiple tags. I think that `UnwrapOpaque` could be made to work on both the current `Opaque` type and the new version. The only potential downsides to that approach are: 

1. We'd need a new name for this new type, and (for a while) there'd be two types in type-fest with similar functionality. Personally, I think having a new name could be a good thing, because the `Opaque` name is actually kinda misleading: the type is not opaque at all, in that the type of the runtime value is very plainly exposed and usable — i.e., an `Opaque<string, 'X'>` can be passed to any function that takes a string, which means its "string-ness" is not hidden and can be depended on in code outside the module where the "opaque" type was defined. Given that, I'd be tempted to call the new type "Tagged" and rename the existing `Tagged` type (which isn't exported so should be fine to rename) to something like `TagContainer`.

2. Existing `Opaque` types that use a `Token` which isn't a string, number, or symbol couldn't interoperate with the new tagged types.

3. All the types would be a bit more complicated, to support interoperability between the old and new form of tagged types.

ethanresnick:
This adds two new types, `Tagged` and `UnwrapTagged`, per https://github.com/sindresorhus/type-fest/issues/665.

It also tweaks `UnwrapOpaque` so that it works on types produced by `Tagged` (as well as still working on types produced by `Opaque`, of course). This is a small gesture at interoperability between `Opaque` and `Tagged`, to make it easier for codebases to migrate from `Opaque` to `Tagged`.

Fixes #665

### DIFF
--- a/source/opaque.d.ts
+++ b/source/opaque.d.ts
@@ -1,17 +1,25 @@
 declare const tag: unique symbol;
 
-declare type Tagged<Token> = {
+declare type TagContainer<Token> = {
 	readonly [tag]: Token;
 };
 
+type MultiTagContainer<Token extends PropertyKey> = {
+	readonly [tag]: {[K in Token]: void};
+};
+
 /**
-Create an opaque type, which hides its internal details from the public, and can only be created by being used explicitly.
+Attach a "tag" to an arbitrary type. This allows you to create distinct types, that aren't assignable to one another, for runtime values that would otherwise have the same type. (See examples.)
 
-The generic type parameter can be anything. It doesn't have to be an object.
+The generic type parameters can be anything.
 
-[Read more about opaque types.](https://codemix.com/opaque-types-in-javascript/)
+Note that `Opaque` is somewhat of a misnomer here, in that, unlike [some alternative implementations](https://github.com/microsoft/TypeScript/issues/4895#issuecomment-425132582), the original, untagged type is not actually hidden. (E.g., functions that accept the untagged type can still be called with the "opaque" version -- but not vice-versa.)
 
-There have been several discussions about adding this feature to TypeScript via the `opaque type` operator, similar to how Flow does it. Unfortunately, nothing has (yet) moved forward:
+Also note that this implementation is limited to a single tag. If you want to allow multiple tags, use `Tagged` instead.
+
+[Read more about tagged types.](https://medium.com/@KevinBGreene/surviving-the-typescript-ecosystem-branding-and-type-tagging-6cf6e516523d)
+
+There have been several discussions about adding similar features to TypeScript. Unfortunately, nothing has (yet) moved forward:
 	- [Microsoft/TypeScript#202](https://github.com/microsoft/TypeScript/issues/202)
 	- [Microsoft/TypeScript#15408](https://github.com/Microsoft/TypeScript/issues/15408)
 	- [Microsoft/TypeScript#15807](https://github.com/Microsoft/TypeScript/issues/15807)
@@ -59,7 +67,7 @@ getMoneyForAccount(2);
 // You can use opaque values like they aren't opaque too.
 const accountNumber = createAccountNumber();
 
-// This will not compile successfully.
+// This will compile successfully.
 const newAccountNumber = accountNumber + 2;
 
 // As a side note, you can (and should) use recursive types for your opaque types to make them stronger and hopefully easier to type.
@@ -71,10 +79,10 @@ type Person = {
 
 @category Type
 */
-export type Opaque<Type, Token = unknown> = Type & Tagged<Token>;
+export type Opaque<Type, Token = unknown> = Type & TagContainer<Token>;
 
 /**
-Revert an opaque type back to its original type by removing the readonly `[tag]`.
+Revert an opaque or tagged type back to its original type by removing the readonly `[tag]`.
 
 Why is this necessary?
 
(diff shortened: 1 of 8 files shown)


source/opaque.d.ts
```diff
@@ -1,17 +1,25 @@
 declare const tag: unique symbol;
 
-declare type Tagged<Token> = {
+declare type TagContainer<Token> = {
 	readonly [tag]: Token;
 };
 
+type MultiTagContainer<Token extends string | number | symbol> = {
```
sindresorhus, line 7:
```suggestion
type MultiTagContainer<Token extends PropertyKey> = {
```

sindresorhus (changes requested):

sindresorhus:
This looks very good 👍

sindresorhus:
You need to also add them to the readme.

sindresorhus:
Some types like https://github.com/sindresorhus/type-fest/blob/main/source/exact.d.ts, https://github.com/sindresorhus/type-fest/blob/main/source/invariant-of.d.ts, and https://github.com/sindresorhus/type-fest/blob/main/source/conditional-pick-deep.d.ts have special handling of `Opaque`. That should probably extend to `Tagged` too.

ethanresnick:
Ok, I think I addressed all the comments. For the three types you mentioned that use `Opaque` (`Exact`, `InvariantOf`, and `ConditionalPickDeep`), I did something different in each case:

- `Exact`: I extended the current approach to work with `Tagged` types too. As I mentioned in the inline code comment, I'm not sure the current approach is exactly correct/ideal, but that can be a topic for a separate PR.

- `InvariantOf`: I kept the same basic logic, but inlined it to remove the dependency on `Opaque`, in case we want to deprecate `Opaque` at some point. I also put the tag that's enforcing the invariance in a differently-named symbol, so that it won't conflict with tags added by `Tagged`/`Opaque`. (Before, it was impossible to do `Opaque<InvariantOf<X>>` or `InvariantOf<Opaque<X>>`, as both `InvariantOf` and `Opaque` would've tried to put conflicting types at the same phantom `[tag]` key.)

> `ConditionalPickDeep` didn't seem like it should be using opaque/tagged types at all; from my reading, it looks like it really just wanted a `unique symbol` type. So, I updated that accordingly, and all the tests still pass.

sindresorhus:
Thanks ;)

PR #673, opened 2023-08-26
fix(jsonify): fix array jsonify handler

### LINKED ISSUE #629: Jsonify: `readonly` messes with Jsonified tuple types
### Actual

```ts
type A = Jsonify<[1, 2, 3]>; // type A = [1, 2, 3]
type B = Jsonify<readonly [1, 2, 3]>; // type B = (1 | 2 | 3)[]
```

### Expected

```ts
type A = Jsonify<[1, 2, 3]>; // type A = [1, 2, 3]
type B = Jsonify<readonly [1, 2, 3]>; // type B = readonly [1, 2, 3]
```

Emiyaaaaa:
hey！this is my first contribution

1. fix #629
2. fix another jsonfity array bug
  ```typescript
  // main branch
  declare const mixTupleJson: Jsonify<['1', typeof fn, 2]>; // type: ['1', 2]
  // this pr
  declare const mixTupleJson: Jsonify<['1', typeof fn, 2]>; // type: ['1', null, 2]

  // because `FilterNonNever` filter all NotJsonable value. but in array, NotJsonable value should be `null`
```

### DIFF
--- a/test-d/jsonify.ts
+++ b/test-d/jsonify.ts
@@ -222,7 +222,7 @@ expectType<[string, string]>(tupleJson);
 declare const tupleRestJson: Jsonify<[string, ...Date[]]>;
 expectType<[string, ...string[]]>(tupleRestJson);
 
-declare const mixTupleJson: Jsonify<['1', typeof fn, 2]>;
+declare const mixTupleJson: Jsonify<['1', (_: any) => void, 2]>;
 expectType<['1', null, 2]>(mixTupleJson);
 
 declare const tupleStringJson: Jsonify<string[] & ['some value']>;
@@ -330,6 +330,6 @@ expectType<{a: any}>(objectWithAnyProperty);
 declare const objectWithAnyProperties: Jsonify<Record<string, any>>;
 expectType<Record<string, any>>(objectWithAnyProperties);
 
-/// #629
+// Regression test for https://github.com/sindresorhus/type-fest/issues/629
 declare const readonlyTuple: Jsonify<readonly [1, 2, 3]>;
 expectType<[1, 2, 3]>(readonlyTuple);
--- a/test-d/jsonify.ts
+++ b/test-d/jsonify.ts
@@ -222,6 +222,9 @@ expectType<[string, string]>(tupleJson);
 declare const tupleRestJson: Jsonify<[string, ...Date[]]>;
 expectType<[string, ...string[]]>(tupleRestJson);
 
+declare const mixTupleJson: Jsonify<['1', typeof fn, 2]>;
+expectType<['1', null, 2]>(mixTupleJson);
+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by sindresorhus** [comment]
> fix another jsonfity array bug

Would you be able to split this out into a separate pull request? That would make it easier to review.

**Comment by sindresorhus** [comment]
// @MichaelDeBoey @sachinraja

**Comment by Emiyaaaaa** [comment]
> > fix another jsonfity array bug
> 
> Would you be able to split this out into a separate pull request? That would make it easier to review.

1. fix #629 bug just make array and readonly array use same handler.
2. but previous array handler `JsonifyList` is wrong, this handler filter all NotJsonable value, that will make `JsonifyList<Array<number>>` type to `never[]` but not `null[]`, it's obviously not right.
3. so, how previous jsonify array test passed ?
https://github.com/sindresorhus/type-fest/blob/main/source/jsonify.d.ts#L119C1-L119C1 this line is wrong, not all array match this rule `T extends [unknown, ...unknown[]]` like `Jsonify<Array<number>>` but match next rule `T extends ReadonlyArray<infer U>`, this handler just right for Array<...>, wrong for readonly [1,2,3]


so i fell split this is a little difficult, sorry about that

**Comment by MichaelDeBoey** [comment]
CC/ @pcattori

File: test-d/jsonify.ts [hunk 1]
```diff
@@ -222,6 +222,9 @@ expectType<[string, string]>(tupleJson);
 declare const tupleRestJson: Jsonify<[string, ...Date[]]>;
 expectType<[string, ...string[]]>(tupleRestJson);
 
+declare const mixTupleJson: Jsonify<['1', typeof fn, 2]>;
```
MichaelDeBoey, line 225: Maybe we should just use `null` or an explicit function declaration instead of a variable that's declared somewhere totally different?

```suggestion
declare const mixTupleJson: Jsonify<['1', null, 2]>;
```
```suggestion
declare const mixTupleJson: Jsonify<['1', (_: any) => void, 2]>;
```

test-d/jsonify.ts
```diff
@@ -328,5 +331,5 @@ declare const objectWithAnyProperties: Jsonify<Record<string, any>>;
 expectType<Record<string, any>>(objectWithAnyProperties);
 
 /// #629
```
MichaelDeBoey, line 333: To keep this in line with other comments in the file

```suggestion
// Regression test for https://github.com/sindresorhus/type-fest/issues/629
```

MichaelDeBoey (changes requested):

sindresorhus:
Can you fix the merge conflict?

MichaelDeBoey (approved):

Emiyaaaaa:
> Can you fix the merge conflict?

fixed

sindresorhus:
Thanks :)

PR #682, opened 2023-09-08
feat(union-to-intersection): allow indexing by the resulting type

niieani:
Previously, using `UnionToIntersection` in the key position would result in the error: 
> Type 'unknown' cannot be used as an index type.

```ts
type Objects = { a: string } | { b: string } | { c: string }
type X = Objects[UnionToIntersection<keyof Objects>]
```

After the change the error goes away, and it doesn't affect how the type works otherwise.
The use case is when you use generics and want to ensure the type is `never` for unions of objects.

### DIFF
--- a/source/union-to-intersection.d.ts
+++ b/source/union-to-intersection.d.ts
@@ -56,5 +56,6 @@ export type UnionToIntersection<Union> = (
 		// Infer the `Intersection` type since TypeScript represents the positional
 		// arguments of unions of functions as an intersection of the union.
 ) extends ((mergedIntersection: infer Intersection) => void)
-	? Intersection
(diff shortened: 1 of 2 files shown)


sindresorhus:
Can you add a test?

niieani:
Hi @sindresorhus, sorry I'm an OSS maintainer too, didn't yet have time to follow up. It's on my list.

niieani:
Hi @sindresorhus! Added a test.

sindresorhus:
@JGJP It's generally better to open a new issue than to comment on an old pull request.

som-sm:
@JGJP Yup, you're right! Fixed the test case in #1260.

PR #684, opened 2023-09-08
fix: `IsLiteral` not restrict generics type

### LINKED ISSUE #680: IsLiteral should not require the type to be Primitive but check it internally
Not sure why the IsLiteral type has a restriction, while other literal types don't have (neither do most other types)

```
export type IsLiteral<T extends Primitive> = IsNotFalse<IsLiteralUnion<T>>;
```

should be:

```
export type IsLiteral<T> = T extends Primitive ? IsNotFalse<IsLiteralUnion<T>> : false;
```

Emiyaaaaa:
fix #680

### DIFF
--- a/source/is-literal.d.ts
+++ b/source/is-literal.d.ts
@@ -1,6 +1,6 @@
 import type {Primitive} from './primitive';
 import type {Numeric} from './numeric';
-import type {IsNotFalse} from './internal';
+import type {IsNotFalse, IsPrimitive} from './internal';
 import type {IsNever} from './is-never';
 
 /**
@@ -250,4 +250,7 @@ stripLeading(str, 'abc');
 @category Type Guard
 @category Utilities
 */
-export type IsLiteral<T extends Primitive> = IsNotFalse<IsLiteralUnion<T>>;
+export type IsLiteral<T> =
+	IsPrimitive<T> extends true
+		? IsNotFalse<IsLiteralUnion<T>>
+		: false;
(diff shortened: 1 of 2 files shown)


source/is-literal.d.ts
```diff
@@ -3,6 +3,22 @@ import type {Numeric} from './numeric';
...
+//=> true
+
+IsPrimitive<string>
+//=> true
+
+IsPrimitive<Object>
+//=> false
+```
+ */
+type IsPrimitive<T> = [T] extends [Primitive] ? true : false;
```
sindresorhus, line 20: Can you move this to internal.d.ts? It makes it easier to reuse types and easier to see what should be exported publicly later on.

source/is-literal.d.ts
```diff
@@ -3,6 +3,22 @@ import type {Numeric} from './numeric';
...
+IsPrimitive<'string'>
+//=> true
+
+IsPrimitive<string>
+//=> true
+
+IsPrimitive<Object>
+//=> false
+```
+ */
```
sindresorhus, line 19:
```suggestion
*/
```

source/is-literal.d.ts
```diff
@@ -3,6 +3,22 @@ import type {Numeric} from './numeric';
 import type {IsNotFalse} from './internal';
 import type {IsNever} from './is-never';
 
+/**
+Returns a boolean for whether the given type `T` is Primitive Value or Primitive Type.
```
sindresorhus, line 7:
```suggestion
Returns a boolean for whether the given type is primitive value or primitive type.
```

sindresorhus (changes requested):

PR #686, opened 2023-09-13
Add `UnionToUnorderedTuple` type

kopach:

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -86,6 +86,7 @@ export type {WritableKeysOf} from './source/writable-keys-of';
 export type {HasWritableKeys} from './source/has-writable-keys';
 export type {Spread} from './source/spread';
 export type {TupleToUnion} from './source/tuple-to-union';
+export type {UnionToUnsortedTuple} from './source/union-to-unsorted-tuple';
 export type {IntRange} from './source/int-range';
 export type {IsEqual} from './source/is-equal';
 export type {
--- a/readme.md
+++ b/readme.md
@@ -247,6 +247,7 @@ type ShouldBeNever = IfAny<'not any', 'not never', 'never'>;
 - [`MultidimensionalReadonlyArray`](source/multidimensional-readonly-array.d.ts) - Create a type that represents a multidimensional readonly array of the given type and dimensions.
 - [`ReadonlyTuple`](source/readonly-tuple.d.ts) - Create a type that represents a read-only tuple of the given type and length.
 - [`TupleToUnion`](source/tuple-to-union.d.ts) - Convert a tuple/array into a union type of its elements.
+- [`UnionToUnsortedTuple`](source/union-to-unsorted-tuple.d.ts) - Convert a union type into a tuple type of its elements.
 
 ### Numeric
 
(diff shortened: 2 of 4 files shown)


sindresorhus:
Thanks for contributing.

It would be great if you could check out the previous attempt at adding this type and address the feedback given there (if relevant): https://github.com/sindresorhus/type-fest/pull/167

kopach:
hi @sindresorhus , thanks for pointing this out. Yes indeed this implementation has exactly the same issues as during previous attempt. The problem is in TypeScript itself. It doesn't guarantee that tuple will always preserve order of it's elements. Closing this for now

kopach:
hi @sindresorhus, just found this https://github.com/microsoft/TypeScript/issues/44116#issuecomment-849833071
> Unions are unordered and any code that attempts to observe a union's order is asking for undefined behavior.

So, maybe we shouldn't expect tuple to be ordered in the same way as union? Personally, I would have this `UnionToTuple` type as it is (possibly with additional comment in docs, that order could change). I use this type in my codebase and still find it useful.

kopach:
Maybe just renaming this to `UnionToUnsortedTuple` to indicate this is not sorted?

sindresorhus:
Yeah. I would go with `UnionToUnorderedTuple`.

sindresorhus:
Bump :)

mkovel:
:+1:

kopach:
hi @sindresorhus, @mkovel. I started refactoring this and discovered, that there are other conceptual problems with this type. Not sure what should be the correct result in this case
```ts
type T = UnionToUnsortedTuple<boolean | 'a'>; // [false, true, 'a']
```
There are many more issues & attempts to solve them here:  https://github.com/microsoft/TypeScript/issues/13298
Not sure if we should have such a type in the library. TS should resolve this.
What do you think?

mkovel:
@kopach  Hi, honestly I didn't dive into it deeply. I just reused this code to solve the issue. Maybe it will be helpful.
```ts
export type UnionToIntersection<U> = (
  U extends never ? never : (arg: U) => never
) extends (arg: infer I) => void
  ? I
  : never;

export type UnionToTuple<T> = UnionToIntersection<
  T extends never ? never : (t: T) => T
> extends (_: never) => infer W
  ? [...UnionToTuple<Exclude<T, W>>, W]
  : [];
```

jtenner:
> @kopach Hi, honestly I didn't dive into it deeply. I just reused this code to solve the issue. Maybe it will be helpful.
> 
> ```ts
> export type UnionToIntersection<U> = (
>   U extends never ? never : (arg: U) => never
> ) extends (arg: infer I) => void
>   ? I
>   : never;
> 
> export type UnionToTuple<T> = UnionToIntersection<
>   T extends never ? never : (t: T) => T
> > extends (_: never) => infer W
>   ? [...UnionToTuple<Exclude<T, W>>, W]
>   : [];
> ```

This is by far one of the most amazing types I've ever seen. It's exactly what I needed, and I don't really care much about the undefined behavior aspect. It's just for type checking and cli configuration.

Thank you so much.

sindresorhus:
I think it's still worth adding this type, but this PR is not moving forward, so closing for now: https://github.com/sindresorhus/type-fest/issues/819

Issue #657, opened 2023-07-29
Jsonify fails on nested object with name property. 


RalphSleigh:
Code here:
https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgKQM4QHbAGYE84C+c2UEIcARDLmAKYC02tqMFA3AFAfV1wQBGAKwByzGLQAmwgIYhacALyIOcVcWBQWALmVq9cDLNo6WUYBgDmKtQQ63uNeQJFjJwiDLmLda7Bu0++gYQMJ7GcKbmVnq29jzygugYAOrAMAAWYd5omDi4ADzOoixuRgB8DryJmADyQsXiUh5G2Ul5hfWuTWFlQA

```
import { Jsonify } from "type-fest";

type objNestedName = {
    first: {
        name: string
    }
}

type objNestedNoName = {
    first: {
        notName: string
    }
}

type jsonWithName = Jsonify<objNestedName>
type jsonObjNestedNoName = Jsonify<objNestedNoName>
```

Note if you mouseover the second one you can see inside the object, but the first one is an empty object. Might be wrong but I am fairly sure name is not a reserved word in typescript or ES6 so would expect this to work.

sindresorhus:
When I mouse over them in the playground, they both show the same thing.

RalphSleigh:
### Playground
Yeah, output has changed since I posted the issue. Not sure why there, maybe due to the release? The mouseover now displays both as:

```type objNestedNoName = {
    first: {
        notName: string
    }
}```

Which is not right either, but the structured types plugin gets it right if you enable/run that.

### Locally

I updated to the latest version of type-fest locally and can still see the issue in visual studio code: 

![Screenshot 2023-08-01 115753](https://github.com/sindresorhus/type-fest/assets/4304184/09dbec06-0edd-486a-9ed1-a593d587a43a)
![Screenshot 2023-08-01 115747](https://github.com/sindresorhus/type-fest/assets/4304184/7f7a8796-c2a5-42a1-9640-9b1fc2b35a2d)

RalphSleigh:
Trying to narrow this down, I made a new folder and added that example, Node is: 17.8

`npm install typescript type-fest`

This gets me typescript 5.1.6 and I can see the issue

`npm install type-fest@2.19.0`

No issue

`npm install type-fest@3.0.0`

Its back.

donovanhiland:
[...]

PR #689, opened 2023-09-21
fix(jsonify): removing nested object with name property

duvet86:
Closes https://github.com/sindresorhus/type-fest/issues/657


sindresorhus:
Can you add a test? It should fail before this change and succeed after.

duvet86:
I closed the PR the issue is more complex than what I thought. And yep, I'll add tests, if I find a solution.

PR #691, opened 2023-09-25
Fixes Jsonify on nested objects with only a name property

### LINKED ISSUE #657: Jsonify fails on nested object with name property.

donovanhiland:
Fixes #657

### DIFF
--- a/source/internal.d.ts
+++ b/source/internal.d.ts
@@ -189,11 +189,11 @@ type BaseKeyFilter<Type, Key extends keyof Type> = Key extends symbol
 	? never
 	: Type[Key] extends symbol
 		? never
-		/**
-		 * To prevent a problem where an object with only a 'name' property is incorrectly treated as assignable to a function, we first check if the property is a record.
-		 * This check is necessary because without it, if we don't verify whether the property is a record, an object with a type of `{ name: any }` would return 'never' due to its potential assignability to a function.
-		 * See: https://github.com/sindresorhus/type-fest/issues/657
-		 */
+		/*
+		To prevent a problem where an object with only a `name` property is incorrectly treated as assignable to a function, we first check if the property is a record.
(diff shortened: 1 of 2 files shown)


source/internal.d.ts
```diff
@@ -189,9 +189,11 @@ type BaseKeyFilter<Type, Key extends keyof Type> = Key extends symbol
...
 	: Type[Key] extends symbol
 		? never
-		: [(...arguments_: any[]) => any] extends [Type[Key]]
-			? never
-			: Key;
+		: Type[Key] extends Record<string, unknown>
+			? Key
+			: [(...arguments_: any[]) => any] extends [Type[Key]]
+				? never
+				: Key;
```
sindresorhus, line 196: It would be good with a short code comment on the code that handles the name thing for future reference.

sindresorhus:
Thanks :)

PR #698, opened 2023-10-04
Fix Jsonify for objects with `Record<string, any>` properties

pcattori:
Pushing just the test case up first, since I can't replicate the failing test from #668 locally


pcattori:
I can reproduce the error in a Typescript playground (as described here: https://github.com/sindresorhus/type-fest/issues/667#issuecomment-1745348930), but could not replicate this error within the `type-fest` test suite. Seems to be related to TSConfig settings

PR #700, opened 2023-10-08
feat: `undefinedOnPartialDeep` type

### LINKED ISSUE #681: `UndefinedOnPartialDeep` (opposite of `PartialOnUndefinedDeep`)
It’d be great have a type similar to [`PartialOnUndefinedDeep`](https://github.com/sindresorhus/type-fest/blob/main/source/partial-on-undefined-deep.d.ts), which does the opposite:

```ts
import type { UndefinedOnPartialDeep } from 'type-fest';

interface Settings {
  optionA: string;
  optionB?: number;
  subOption: {
    subOptionA: boolean;
    subOptionB?: boolean;
  };
}

const testSettingsA: Settings = {
  optionA: 'foo',
  optionB: undefined, // TS error if exactOptionalPropertyTypes is true
  subOption: {
    subOptionA: true,
    subOptionB: undefined, // TS error if exactOptionalPropertyTypes is true
  },
};

const testSettingsB: UndefinedOnPartialDeep<Settings> = {
  optionA: 'foo',
  optionB: undefined, // 👉 optionB can be set to undefined now
  subOption: {
    subOptionA: true,
    subOptionB: undefined, // 👉 subOptionB can be set to undefined now
  },
};
```

See https://github.com/sindresorhus/type-fest/pull/426 and [`exactOptionalPropertyTypes` docs](https://www.typescriptlang.org/tsconfig#exactOptionalPropertyTypes) for context.

Emiyaaaaa:
Fixes #681 

Hello Everyone~

I have a question about testing and want some suggest or help:

- This type is only useful when `exactOptionalPropertyTypes` enabled, but this option is disable in type-fest, so tests for this type must pass regardless of bugs.

Then I have an idea:

1. replace [this line](https://github.com/sindresorhus/type-fest/pull/700/files#diff-33d1f8bc49426a80bb2c00b2fa091aada9bb33105abdfc2122bc5da9c3a43019R66) `undefined` with another any flag type like `type FlagType = 1` by generic type. example `undefinedOnPartialDeep<T, UnionType = FlagType>`.
2. test if `undefinedOnPartialDeep<{ a?: string }, FlagType>` is assignable to type `{ a?: string | FlagType }` or not.
3. export a stable new type like `type NewStableType<T> = undefinedOnPartialDeep<T, undefined>`, because 2nd generic type is just for testing, don't hope to export for user. (ignore type name, maybe is `type UndefinedOnPartialDeep<T> = someOtherName<T, undefined>`)


sindresorhus:
// @kachkaev

sindresorhus:
> This type is only useful when exactOptionalPropertyTypes enabled, but this option is disable in type-fest, so tests for this type must pass regardless of bugs.

You could create a separate test file in a subfolder, give it a tsconfig file that extends the existing ones, but changes that flag, and then just run `tsc --noEmit` on that file.

Emiyaaaaa:
> > This type is only useful when exactOptionalPropertyTypes enabled, but this option is disable in type-fest, so tests for this type must pass regardless of bugs.
> 
> You could create a separate test file in a subfolder, give it a tsconfig file that extends the existing ones, but changes that flag, and then just run `tsc --noEmit` on that file.

thanks! fixed in the lastest commit. but `tsc` seems just could test assignable, `expectType` is not working.

sindresorhus:
@Emiyaaaaa Thank you 🙏

It would be amazing if you would be willing to help review some of the open pull requests too: https://github.com/sindresorhus/type-fest/pulls

Emiyaaaaa:
> @Emiyaaaaa Thank you 🙏
> 
> It would be amazing if you would be willing to help review some of the open pull requests too: https://github.com/sindresorhus/type-fest/pulls

Thanks for the invite, I'll try my best.

Issue #132, opened 2020-10-05
Proposal: A version of Pick that distributes over union types


novemberborn:
See https://github.com/microsoft/TypeScript/issues/28791#issuecomment-443520161:

> The issue is that `Pick` doesn't distribute over union types, and therefore your `Omit` type doesn't either. Here's a simplified version of your example:
> 
> ```ts
> type A = { name?: string, variant: 'a' };
> type B = { name?: string, variant: 'b' };
> type C = A | B;
> type X = Omit<C, 'name'>;  // { variant: 'a' | 'b' }
> ```
> Because `Pick` isn't distributive over union types, when the input type is a union it is effectively treated as a single object type with only the common properties, each having a union of the respective property types. So, `Pick` sees `C` as equivalent to `{ name?: string, variant: 'a' | 'b' }` and `X` therefore is `{ variant: 'a' | 'b' }`, which isn't assignable back to `C`.
> 
> There are two ways you can fix the issue. The simplest is to declare `C` as a single object type (and this would be my recommendation):
> 
> ```ts
> type C = { name?: string, variant: 'a' | 'b' };
> ```
> The other is to make `Omit` be distributive over union types:
> 
> ```ts
> export type AllKeys<T> = T extends T ? keyof T : never;
> export type Omit<T, K extends AllKeys<T>> = T extends T ? Pick<T, Exclude<keyof T, K>> : never;
> ```
> This may look a bit odd, but the `T extends T ? XXX : never` pattern simply means "when `T` is a union type, distribute `XXX` over each constituent of `T` and union together the results".
> 
> When you use the distributive form of Omit you now get
> 
> ```ts
> type A = { name?: string, variant: 'a' };
> type B = { name?: string, variant: 'b' };
> type C = A | B;
> type X = Omit<C, 'name'>;  // { variant: 'a' } | { variant: 'b' }
> ```
> and `X` now is assignable back to `C`.

The solution I came up with is:

```ts
type AllKeys<T> = T extends T ? keyof T : never;
type Select<T, K extends AllKeys<T>> = T extends T
  ? Pick<T, Exclude<keyof T, Exclude<keyof T, K>>>
  : never;
```

Note that here I want a `Pick` that is distributive, and the above was about `Omit`. I suppose we'd want both?

Not sure what to name this though.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/sindresorhus) so you can upvote and help fund this issue.
- The funding will be given to active contributors.
[...]

PR #709, opened 2023-10-15
Expose KeysOfUnion type

henriqueinonhe:
Adds the `AllKeys` type, that is similiar to `keyof` but it distributes over unions, that is, it returns the keys of ALL members of a union type.

Although this is a first step to add the types in #132, it is also useful by itself.

Additionally, I only used this name (`AllKeys`) because it's what was suggested in the issue I mentioned above, but there are other possible names we might want to consider, like `KeyOfDistributeUnions` or `DistributedKeyOf`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -5,6 +5,7 @@ export * from './source/basic';
 export * from './source/observable-like';
 
 // Utilities
+export type {KeysOfUnion} from './source/keys-of-union';
 export type {EmptyObject, IsEmptyObject} from './source/empty-object';
 export type {NonEmptyObject} from './source/non-empty-object';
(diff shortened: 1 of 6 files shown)


sindresorhus:
Some more naming alternatives:

- `AllMemberKeys`
- `EveryKeyOf`

sindresorhus:
Hmm, we actually have a type for this already: https://github.com/sindresorhus/type-fest/blob/47626cf470853b1828b44637114292a84bf58817/source/internal.d.ts#L40-L47

I think that name is the best.

henriqueinonhe:
Interesting... Would it be okay to expose this internal type then?

sindresorhus:
You can just replace the use of that type with yours.

PR #712, opened 2023-10-15
Implement autocomplete for OverrideProperties

### LINKED ISSUE #711: Proposal: Implement autocompletion for `OverrideProperties`
Currently, while `OverrideProperties` restricts the keys you can override to the keys that are present in the type you pass (`TOriginal`), it doesn't offer autocomplete:

image

image

But with a simple tweak we can enable autocompletion:

image

Here's a demo showcasing the implementation:

https://stackblitz.com/edit/typescript-zyn86c?file=index.ts

This implementation keeps keys "strict", in the sense that you cannot override/add keys that are not present in the original type.

If maintainers agree that this is a good improvement, I can create a PR.

henriqueinonhe:
Fixes #711

### DIFF
--- a/source/override-properties.d.ts
+++ b/source/override-properties.d.ts
@@ -25,5 +25,9 @@ type Fizz = OverrideProperties<Foo, {b: number; c: number}>
 */
 export type OverrideProperties<
 	TOriginal,
-	TOverride extends {[Key in keyof TOverride]: Key extends keyof TOriginal ? TOverride[Key] : never},
+	TOverride extends Partial<Record<keyof TOriginal, unknown>> & {
+		[Key in keyof TOverride]: Key extends keyof TOriginal
+			? TOverride[Key]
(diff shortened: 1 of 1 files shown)


source/override-properties.d.ts
```diff
@@ -25,5 +25,9 @@ type Fizz = OverrideProperties<Foo, {b: number; c: number}>
 */
 export type OverrideProperties<
 	TOriginal,
-	TOverride extends {[Key in keyof TOverride]: Key extends keyof TOriginal ? TOverride[Key] : never},
+	TOverride extends Partial<Record<keyof TOriginal, unknown>> & {
```
sindresorhus, line 28: Can you add a short code comment about why we do this (to get autocomplete).

henriqueinonhe: Absolutely! Good call!

PR #723, opened 2023-10-23
feat: add metadata support to Tagged type

ethanresnick:
Addition to support the use case described here https://github.com/sindresorhus/type-fest/issues/665#issuecomment-1769133997

### DIFF
--- a/source/opaque.d.ts
+++ b/source/opaque.d.ts
@@ -4,9 +4,7 @@ export type TagContainer<Token> = {
 	readonly [tag]: Token;
 };
 
-type MultiTagContainer<Token extends PropertyKey> = {
-	readonly [tag]: {[K in Token]: void};
-};
+type Tag<Token extends PropertyKey, TagMetadata> = TagContainer<{[K in Token]: TagMetadata}>;
 
 /**
 Attach a "tag" to an arbitrary type. This allows you to create distinct types, that aren't assignable to one another, for runtime values that would otherwise have the same type. (See examples.)
@@ -113,18 +111,25 @@ type WillWork = UnwrapOpaque<Tagged<number, 'AccountNumber'>>; // number
 @category Type
 */
 export type UnwrapOpaque<OpaqueType extends TagContainer<unknown>> =
-	OpaqueType extends MultiTagContainer<string | number | symbol>
+	OpaqueType extends Tag<PropertyKey, any>
 		? RemoveAllTags<OpaqueType>
 		: OpaqueType extends Opaque<infer Type, OpaqueType[typeof tag]>
 			? Type
 			: OpaqueType;
 
 /**
-Attach a "tag" to an arbitrary type. This allows you to create distinct types, that aren't assignable to one another, for runtime values that would otherwise have the same type. (See examples.)
+Attach a "tag" to an arbitrary type. This allows you to create distinct types, that aren't assignable to one another, for distinct concepts in your program that should not be interchangeable, even if their runtime values have the same type. (See examples.)
 
 A type returned by `Tagged` can be passed to `Tagged` again, to create a type with multiple tags.
 
 [Read more about tagged types.](https://medium.com/@KevinBGreene/surviving-the-typescript-ecosystem-branding-and-type-tagging-6cf6e516523d)
+
+A tag's name is usually a string (and must be a string, number, or symbol), but each application of a tag can also contain an arbitrary type as its "metadata". See {@link GetTagMetadata} for examples and explanation.
+
+A type `A` returned by `Tagged` is assignable to another type `B` returned by `Tagged` if and only if:
+  - the underlying (untagged) type of `A` is assignable to the underlying type of `B`;
+	- `A` contains at least all the tags `B` has;
+	- and the metadata type for each of `A`'s tags is assignable to the metadata type of `B`'s corresponding tag.
 
 There have been several discussions about adding similar features to TypeScript. Unfortunately, nothing has (yet) moved forward:
 	- [Microsoft/TypeScript#202](https://github.com/microsoft/TypeScript/issues/202)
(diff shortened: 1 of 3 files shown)


source/opaque.d.ts
```diff
@@ -126,6 +126,8 @@ A type returned by `Tagged` can be passed to `Tagged` again, to create a type wi
 
 [Read more about tagged types.](https://medium.com/@KevinBGreene/surviving-the-typescript-ecosystem-branding-and-type-tagging-6cf6e516523d)
 
+A tag is usually a string (and must be a string, number, or symbol), but each application of a tag can also contain an arbitrary type as its metadata. See {@link GetTagMetadata}.
```
sindresorhus, line 129: The docs needs to mention how it affects the type, like assignability.

sindresorhus: I think the text here should also mention what adding metadata can be useful for.

sindresorhus: > and people who want invariant behavior can get that using the InvariantOf helper type (e.g., type JsonOf<T> = Tagged<string, "JSON", InvariantOf<T>>) 

Maybe also worth documenting how to get invariant behavior.

source/opaque.d.ts
```diff
@@ -162,7 +164,32 @@ const newAccountNumber = accountNumber + 2;
...
 @category Type
 */
-export type Tagged<Type, Tag extends PropertyKey> = Type & MultiTagContainer<Tag>;
+export type Tagged<Type, Tag extends PropertyKey, TagMetadata = void> = Type & MultiTagContainer<Tag, TagMetadata>;
+
+/**
+Given a type and a tag, returns the metadata associated with that tag on that type.
+
+@example
+```
```
sindresorhus, line 173: The example is great, but I don't think it's immediately clear to all users what problem metadata solves. Maybe explain the code example in a couple of succinct sentences?

sindresorhus (changes requested):

ethanresnick:
@sindresorhus thanks! I think I can tackle this this week :)

ethanresnick:
@sindresorhus Ok, I've updated this PR to rebase it on top of main and to add more docs that better explain the use cases for metadata and cover some of the details around assignability!

source/opaque.d.ts
```diff
@@ -113,19 +113,24 @@ type WillWork = UnwrapOpaque<Tagged<number, 'AccountNumber'>>; // number
 @category Type
 */
 export type UnwrapOpaque<OpaqueType extends TagContainer<unknown>> =
```
ethanresnick, line 115: FYI, `UnwrapOpaque` still works on `Tagged` types, and there's a test for that. This new implementation is just simpler, by exploiting the fact that `TagContainer` and `MultiTagContainer` share the same basic structure. It lets us remove `RemoveAllTags`, which was causing some circular type errors when used with tag metadata.

source/opaque.d.ts
```diff
@@ -113,19 +113,24 @@ type WillWork = UnwrapOpaque<Tagged<number, 'AccountNumber'>>; // number
...
-	OpaqueType extends MultiTagContainer<string | number | symbol>
-		? RemoveAllTags<OpaqueType>
-		: OpaqueType extends Opaque<infer Type, OpaqueType[typeof tag]>
-			? Type
-			: OpaqueType;
+	OpaqueType extends Opaque<infer Type, OpaqueType[typeof tag]>
+		? Type
+		: OpaqueType;
```
ethanresnick, line 118: FYI, this updated implementation of `UnwrapOpaque` still works on `Tagged` types, and there's a test for that. The new implementation is just simpler, by exploiting the fact that `TagContainer` and `MultiTagContainer` share the same basic structure. It lets us remove `RemoveAllTags`, which was causing some circular type errors when used with tag metadata.

source/opaque.d.ts
```diff
@@ -151,18 +156,59 @@ function getMoneyForAccount(accountNumber: AccountNumber): AccountBalance {
...
+type Url = Tagged<string, 'URL'>;
+type SpecialCacheKey = Tagged<Url, 'SpecialCacheKey'>;
+
+// You can also pass a union of tag names, so this is equivalent to the above, although it doesn't give you the ability to assign distinct metadata to each tag.
+type SpecialCacheKey2 = Tagged<string, 'URL' | 'SpecialCacheKey'>;
+```
+
+@category Type
+*/
+export type Tagged<Type, Tag extends PropertyKey, TagMetadata = never> = Type & MultiTagContainer<Tag, TagMetadata>;
```
ethanresnick, line 182: I ended up making a tag's default metadata `never` instead of `void`, as `never` has more consistent/sensible assignability semantics

source/opaque.d.ts
```diff
@@ -4,8 +4,8 @@ export type TagContainer<Token> = {
 	readonly [tag]: Token;
 };
 
-type MultiTagContainer<Token extends PropertyKey> = {
-	readonly [tag]: {[K in Token]: void};
+type Tag<Token extends PropertyKey, TagMetadata> = {
```
ethanresnick, line 7: I decided to rename `MultiTagContainer` to `Tag` because it makes the IDE popups much clearer for tagged types. E.g. in these types:

```ts
type AbsolutePath = Tagged<string, 'AbsolutePath'>;
type NormalizedPath = Tagged<string, 'NormalizedPath'>;
type NormalizedAbsolutePath = AbsolutePath & NormalizedPath;
```

Hovering over `NormalizedAbsolutePath` will now show 

```ts
string & Tag<'AbsolutePath', never> & Tag<'NormalizedPath', never>
```

instead of 

```ts
string & MultiTagContainer<'AbsolutePath', never> & MultiTagContainer<'NormalizedPath', never>
```

source/opaque.d.ts
```diff
@@ -4,9 +4,7 @@ export type TagContainer<Token> = {
 	readonly [tag]: Token;
 };
 
-type MultiTagContainer<Token extends PropertyKey> = {
-	readonly [tag]: {[K in Token]: void};
-};
+type Tag<Token extends PropertyKey, TagMetadata> = TagContainer<{[K in Token]: TagMetadata}>;
```
ethanresnick, line 7: I decided to rename `MultiTagContainer` to `Tag` because it makes the IDE popups much clearer for tagged types. E.g. in these types:

```ts
type AbsolutePath = Tagged<string, 'AbsolutePath'>;
type NormalizedPath = Tagged<string, 'NormalizedPath'>;
type NormalizedAbsolutePath = AbsolutePath & NormalizedPath;
```

Hovering over `NormalizedAbsolutePath` will now show 

```ts
string & Tag<'AbsolutePath', never> & Tag<'NormalizedPath', never>
```

instead of 

```ts
string & MultiTagContainer<'AbsolutePath', never> & MultiTagContainer<'NormalizedPath', never>
```

test-d/opaque.ts
```diff
@@ -113,6 +113,34 @@ const unwrapped2 = 123 as PlainValueUnwrapTagged;
 expectType<number>(unwrapped1);
 expectType<number>(unwrapped2);
 
+// UnwrapTagged/UnwrapOpaque should work on types with multiple tags.
+const unwrapped3 = '' as UnwrapTagged<NormalizedAbsolutePath>;
+const unwrapped4 = '' as UnwrapOpaque<NormalizedAbsolutePath>;
+expectType<string>(unwrapped3);
+expectType<string>(unwrapped4);
```
ethanresnick, line 120: these are new tests that i added because i realized that i accidentally broke `UnwrapTagged` without them in an earlier version of this PR

sindresorhus:
Thanks for writing this. Looks good 👍

ethanresnick:
Thanks for merging! 💯🎉

PR #726, opened 2023-10-23
fix(Writable): support array

### LINKED ISSUE #725: `Writable` is not suppert array
`Writable` Sounds like can make `readonly [a, b, c]` to be writable as `[a, b, c]` but it can not actually, although the documentation indicates that it is only used to handle object.

And the behavior of `Writable` is more like 'WritableObject'

Can we make `Writable` support array? @sindresorhus 

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/725">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/725/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

Emiyaaaaa:
Fixed #725 

Have a question: 
 - Should we suppert `Map` and `Set`? https://github.com/sindresorhus/type-fest/issues/725#issuecomment-1774686086


sindresorhus:
> Should we suppert Map and Set?

Yes, just make sure you document that this differs from `Readonly` and link to this TS issue: https://github.com/microsoft/TypeScript/issues/29655

Issue #592, opened 2023-04-11
A Constructor within a ReadonlyDeep type is not callable


galmalka:
Given the following code:
```typescript
class A {}

const x = {
    constructor: A
}

const readonlyX: ReadonlyDeep<typeof x> = x;

const y = new readonlyX.constructor();
```

The last line gives the following error:
`This expression is not constructable.
  Type 'ReadonlyObjectDeep<typeof A>' has no construct signatures.ts(2351)`

This should probably be legitimate because using the constructor doesn't change the variable itself.

PR #728, opened 2023-10-24
fix(ReadonlyDeep): skip constructor

Emiyaaaaa:
- Fixed: #592


sindresorhus:
Can you check if we need to do the same to any of the other "deep" types?

Issue #759, opened 2023-11-17
Audit all types for whether they need `readonly` modifier


sindresorhus:
Similar to https://github.com/sindresorhus/type-fest/issues/758

// @Emiyaaaaa

Emiyaaaaa:
need to fix (search `extends [` ):
- `JsonifyList`
- `PropertyOf`
- `CamelCaseFromArray`
- `StringArrayToDelimiterCase`
- `DelimiterCasedPropertiesArrayDeep`
- `GetWithPath`
- `Join`
- `MergeDeep` and other help type in merge-deep.d.ts

uncertain:
- `Class` `Constructor` `AbstractClass` `AbstractConstructor`

Issue #758, opened 2023-11-17
[4.8.0] type error on jsonify


AviVahl:
with the new version, I get:
```
ode_modules/type-fest/source/jsonify.d.ts:122:28 - error TS2344: Type 'WritableDeep<T>' does not satisfy the constraint 'unknown[]'.
  Type 'T | (T extends (...arguments_: any[]) => unknown ? {} extends WritableObjectDeep<((...arguments_: any[]) => unknown) & T> ? ((...arguments_: any[]) => unknown) & T : HasMultipleCallSignatures<((...arguments_: any[]) => unknown) & T> extends true ? ((...arguments_: any[]) => unknown) & T : ((...arguments_: Parameters...' is not assignable to type 'unknown[]'.
    Type 'readonly unknown[] & T' is not assignable to type 'unknown[]'.
      The type 'readonly unknown[]' is 'readonly' and cannot be assigned to the mutable type 'unknown[]'.

122              ? JsonifyList<WritableDeep<T>>
                               ~~~~~~~~~~~~~~~
```

changing:
`type JsonifyList<T extends unknown[]>`
to:
`type JsonifyList<T extends readonly unknown[]>`

fixes the issue.

saw this in https://github.com/wixplosives/file-services when I've upgraded to `type-fest@4.8.0`

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/sindresorhus) so you can upvote and help fund this issue.
- The funding will be given to active contributors.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/sindresorhus/type-fest/issues/758">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/758/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/sindresorhus/type-fest/issues/758/pledge.svg">
</picture>
</a>

AviVahl:
can also be seen in a dependabot PR: https://github.com/wixplosives/resolve-directory-context/pull/288

sindresorhus:
Fixed: https://github.com/sindresorhus/type-fest/releases/tag/v4.8.1

sindresorhus:
@Emiyaaaaa Would be great if you could add a test for this.

Emiyaaaaa:
> @Emiyaaaaa Would be great if you could add a test for this.

[...]

Issue #757, opened 2023-11-17
[4.8.0] fails when there are no DOM types


AviVahl:
got this failure with the new version:
```
node_modules/type-fest/source/internal.d.ts:72:52 - error TS2304: Cannot find name 'Element'.

72 export type BuiltIns = Primitive | Date | RegExp | Element;
```

my tsconfig says: `"lib": ["es2021"],` without DOM types.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/sindresorhus) so you can upvote and help fund this issue.
- The funding will be given to active contributors.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/sindresorhus/type-fest/issues/757">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/757/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/sindresorhus/type-fest/issues/757/pledge.svg">
</picture>
</a>

sindresorhus:
Fixed: https://github.com/sindresorhus/type-fest/releases/tag/v4.8.1

PR #762, opened 2023-11-17
`Jsonify`: Remove redundant conditions

Emiyaaaaa:
Related #759 

I think `WritableDeep<T>` is not necessary here.

### DIFF
--- a/source/jsonify.d.ts
+++ b/source/jsonify.d.ts
@@ -6,7 +6,7 @@ import type {IsNever} from './is-never';
 import type {IsUnknown} from './is-unknown';
 import type {NegativeInfinity, PositiveInfinity} from './numeric';
 import type {TypedArray} from './typed-array';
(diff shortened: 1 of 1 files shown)


sindresorhus:
Would you be able to add a test to ensure the case in https://github.com/sindresorhus/type-fest/issues/759 still works?

Emiyaaaaa:
> Would you be able to add a test to ensure the case in #759 still works?

I think #758 actually caused by #757

I clone his project, than use type-fest@4.8.0, throw two errors: #757 and #758 ,than I just remove `Element` in `BuiltIn`, the tests are passed.

So I think the reason for this is that `WritableDeep` uses `BuiltIn`, but there is some error in `BuiltIn` that causes typescript to give an incorrect `WritableDeep` type inference.

Emiyaaaaa:
This PR is just for more robust and remove some useless conditions

PR #764, opened 2023-11-19
fix: `PartialOnUndefinedDeep` do not make unknown partial / only use LiteralKeys to reconstruct type

### LINKED ISSUE #756: PartialOnUndefinedDeep removes non-optional properties when the input type contains an index signature
Example:
```ts
import {PartialOnUndefinedDeep} from "type-fest"

type WithoutSignature = {
  foo: number
  bar: number | undefined
}
// Output1 contains the `foo` property
type Output1 = PartialOnUndefinedDeep<WithoutSignature>

type WithSignature = {
  foo: number
  bar: number | undefined
  [k: string]: unknown
}
// Output2 does not contain the `foo` property
type Output2 = PartialOnUndefinedDeep<WithSignature>
```
[TS Playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwAoENbBQGwPIDsCquAJgKYBmwuJRAIiSWAL5xlQQhwBEMAnmCQFoyJAM4xOAKAm9+cAOrAYACwgBXGAGVgAc1woYqqCTgBeRBLgsIEAFxxcqkACMSUC3Cdo7D567gAfOFVickpqCUYJAHoouGx1MHUARjgAYwhcGBRKEThlYwADMmsCuDA2flgeaT5jeJhEmBSzVHQsPEJSCipaejAAHgVlNU0dPQMjAD5o2MsAPQB+KRljIaUtXX1DYzMEd2Lbe0cXN0tPKG9jv0DgrrCidwBtAGs7MShKbQBdO2Dn3AgAHdcBEZnEEuoAExwIgQUT2CDwdKZbK4PJKQoHUrlCCVXg1WT1RrQlpoGAYHAEELdah0BiDRTrMZbKZg+ZLIA)

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/756">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/756/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

clemclx:
fix #756

- Do not treat unknown like undefined (the condition was `undefined extends T[KeyType]` that is true when `T[KeyType]` is `unknown`)
- The bug reported in #756 was due to the condition when reconstructing type (with `Merge`) to check if the key is present in the partial derived type (condition was always true if the index key type is string `KeyType extends keyof U` is always true). The new condition only check the literal keys of the partial derived type.

Note : this `LiteralKeyOf` type could be added to type-fest, it can be useful I think

### DIFF
--- a/source/partial-on-undefined-deep.d.ts
+++ b/source/partial-on-undefined-deep.d.ts
@@ -71,6 +71,6 @@ type PartialOnUndefinedDeepValue<T, Options extends PartialOnUndefinedDeepOption
 			: unknown;
 
 /**
-Utility type to retrieve only literal keys from type 
+Utility type to retrieve only literal keys from type.
 */
 type LiteralKeyOf<T> = keyof {[K in keyof T as IsLiteral<K> extends true ? K : never]-?: never};
--- a/source/partial-on-undefined-deep.d.ts
+++ b/source/partial-on-undefined-deep.d.ts
@@ -1,5 +1,5 @@
 import type {IfUnknown} from './if-unknown';
-import type {BuiltIns} from './internal';
+import type {BuiltIns, LiteralKeyOf} from './internal';
 import type {IsLiteral} from './is-literal';
 import type {Merge} from './merge';
 
@@ -70,8 +70,3 @@ type PartialOnUndefinedDeepValue<T, Options extends PartialOnUndefinedDeepOption
 		: T extends Record<any, any> | undefined
 			? PartialOnUndefinedDeep<T, Options>
 			: unknown;
-
-/**
-Utility type to retrieve only literal keys from type.
-*/
-type LiteralKeyOf<T> = keyof {[K in keyof T as IsLiteral<K> extends true ? K : never]-?: never};
--- a/source/partial-on-undefined-deep.d.ts
+++ b/source/partial-on-undefined-deep.d.ts
@@ -1,3 +1,4 @@
+import type {IfUnknown} from './if-unknown';
 import type {BuiltIns} from './internal';
 import type {Merge} from './merge';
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
@Emiyaaaaa Looks good to you?

source/partial-on-undefined-deep.d.ts
```diff
@@ -68,3 +70,8 @@ type PartialOnUndefinedDeepValue<T, Options extends PartialOnUndefinedDeepOption
 		: T extends Record<any, any> | undefined
 			? PartialOnUndefinedDeep<T, Options>
 			: unknown;
+
+/**
+Utility type to retrieve only literal keys from type.
+*/
+type LiteralKeyOf<T> = keyof {[K in keyof T as IsLiteral<K> extends true ? K : never]-?: never};
```
sindresorhus, line 77: Can you move this to internal.d.ts? It makes it easier for us to keep track of what should probably be exported publicly.

test-d/partial-on-undefined-deep.ts
```diff
@@ -49,6 +51,8 @@ expectAssignable<{
 	readonly1?: TestingType['readonly1'];
 	readonly2?: TestingType['readonly2'];
 	tuple?: TestingType['tuple'];
+	indexType: Partial<Pick<TestingType['indexType'], 'propertyB'>> & Omit<TestingType['indexType'], 'propertyB'>;
```
Emiyaaaaa, line 54: It would be better to use expectType

```ts
declare const indexType: PartialOnUndefinedDeep<{[k: string]: string | undefined; propertyA: string; propertyB: string | undefined};>
expectType<{[k: string]?: string | undefined; propertyA: string; propertyB?: string | undefined}>(indexType)
```

test-d/partial-on-undefined-deep.ts
```diff
@@ -49,6 +51,8 @@ expectAssignable<{
...
+	indexTypeUnknown: Partial<Pick<TestingType['indexTypeUnknown'], 'propertyB'>> & Omit<TestingType['indexTypeUnknown'], 'propertyB'>;
```
Emiyaaaaa, line 55: Same as above

test-d/partial-on-undefined-deep.ts
```diff
@@ -1,4 +1,4 @@
-import {expectAssignable} from 'tsd';
+import {expectAssignable, expectType} from 'tsd';
 import type {PartialOnUndefinedDeep} from '../index';
 
 type TestingType = {
```
Emiyaaaaa, line 4: because changed the key type of object in this PR, so it would be better to check the property keyword like `readonly`

```ts
readonly readonlyProperty: string | undefined;
```

test-d/partial-on-undefined-deep.ts
```diff
@@ -24,6 +24,8 @@ type TestingType = {
 	readonly2: ReadonlyArray<{propertyA: string; propertyB: number | undefined}> | undefined;
 	tuple: ['test1', {propertyA: string; propertyB: number | undefined}] | undefined;
 };
+declare const indexType: {[k: string]: string | undefined; propertyA: string; propertyB: string | undefined};
+declare const indexTypeUnknown: {[k: string]: unknown; propertyA: string; propertyB: number | undefined};
 
 // Default behavior, without recursion into arrays/tuples
 declare const foo: PartialOnUndefinedDeep<TestingType>;
```
Emiyaaaaa, line 31: add 
```ts
readonly readonlyProperty?: TestingType['readonlyProperty'];
```

sindresorhus:
Bump :)

clemclx:
Hey, sorry for the delay. Let me know if there is something else to fix

russelldavis:
@clemclx @sindresorhus it looks like this is ready to merge? (other than merge conflicts)

PR #776, opened 2023-12-02
`ConditionalPickDeep`: supports interface

### LINKED ISSUE #773: ConditionalPickDeep does not work with deep interfaces
Hi,

If the deep property is an interface, `ConditionalPickDeep` does not pick the nested attribute.

Example:

```ts
type S = "SOME STRING";

interface B {
  b: S;
}

interface C {
  b: B;
  c: S;
}

type Test = ConditionalPickDeep<C, S>;

// Expected: type Test = { b: B, c: S }
// Actual: type Test = { c: S }
```

However the code below works as expected:

```ts
type S = "SOME STRING";

type B = {
  b: S;
}

interface C {
  b: B;
  c: S;
}

type Test = ConditionalPickDeep<C, S>;

// Expected: type Test = { b: B, c: S }
// Actual: type Test = { b: B, c: S }
```

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/773">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/773/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

Emiyaaaaa:
- Fixed #773

### DIFF
--- a/source/internal.d.ts
+++ b/source/internal.d.ts
@@ -78,9 +78,9 @@ Matches non-recursive types.
 export type NonRecursiveType = BuiltIns | Function | (new (...args: any[]) => unknown);
 
 /**
-Returns a boolean for whether the given type is a normal key-value object.
+Returns a boolean for whether the given type is a plain key-value object.
 */
-export type IsRecordObject<T> =
+export type IsPlainObject<T> =
 	T extends NonRecursiveType | UnknownArray | ReadonlyMap<unknown, unknown> | ReadonlySet<unknown>
 		? false
 		: T extends object
--- a/source/internal.d.ts
+++ b/source/internal.d.ts
@@ -3,6 +3,7 @@ import type {Simplify} from './simplify';
 import type {Trim} from './trim';
 import type {IsAny} from './is-any';
 import type {UnknownRecord} from './unknown-record';
+import type {UnknownArray} from './unknown-array';
 
 // TODO: Remove for v5.
 export type {UnknownRecord} from './unknown-record';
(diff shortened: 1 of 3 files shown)


source/internal.d.ts
```diff
@@ -76,6 +77,16 @@ Matches non-recursive types.
 */
 export type NonRecursiveType = BuiltIns | Function | (new (...args: any[]) => unknown);
 
+/**
+Returns a boolean for whether the given type is a normal key-value object.
+*/
+export type IsRecordObject<T> =
```
sindresorhus, line 83: `Record` is a type, so the expectation is that this would detect such type, which I don't think it does exactly. Maybe a better name would be `IsPlainObject`?

source/internal.d.ts
```diff
@@ -76,6 +77,16 @@ Matches non-recursive types.
 */
 export type NonRecursiveType = BuiltIns | Function | (new (...args: any[]) => unknown);
 
+/**
+Returns a boolean for whether the given type is a normal key-value object.
```
sindresorhus, line 81: It has to be more specific what a "normal key-value object" is exactly. Plain object?

source/internal.d.ts
```diff
@@ -76,6 +77,16 @@ Matches non-recursive types.
...
+*/
+export type IsRecordObject<T> =
+	T extends NonRecursiveType | UnknownArray | ReadonlyMap<unknown, unknown> | ReadonlySet<unknown>
```
sindresorhus, line 84: What about WeakMap and other types? I feel like it would be better to come up with a check that didn't need an "exclude list" of types.

sindresorhus (changes requested):

PR #777, opened 2023-12-02
`MergeDeep`: fixed optional when value type is `any` or `never`

### LINKED ISSUE #775: `MergeDeep` removes optionals from `any` and `never` fields
Example:

```ts
interface A{
  propNever?:never
  propAny?:any
  propString?:string
  propNull?:null
  propUndefined?:undefined
  propNumber?:number
  propLiteralUnion?:'A'|'B'
}

type B = MergeDeep<A, {additionalAny?:any, additionalString?:string}>
```
### Expectation:

That `B` keeps all the undefined fields as they are and adds two more.

### Actual:

`B` sets:
- all `undefined | any` to `any`
- all `undefined | never` to `undefined`

It doesn't matter if the types come from the left or the right interface.

image

The merging seems to be conform to expectations for other types (the ones I tried at least).

### My use case:

I am using Supabase and their excellent typescript generation.

Supbase sets the `id` field of operations `update` and `insert` as `type?:never`, to make sure an `id` is never passed to an SQL operation that doesn't need it. So a table definition looks like so:

```ts
interface Tables{
  users:{
    Row:{
      id: string,
      email: string 
    },
    Insert:{
      id?: never,
      email: string
    }
    Update:{
      id?: never,
      email?: string
    }
  }
}
```

It's a neat and elegant system that makes all operations type-sound(-ish).

I use MergeDeep to refine the Supabase type (specifically, strictly typing the json fields). But because of the above bug, this won't work:
```ts
supabase.from("users").update({ email }).eq("id", id)
```
The linter won't let it pass,  it complains that `id` is required.

I am now forced to use
```ts
supabase.from("users").update({ email, id: undefined }).eq("id", id)
```
Or to leave a `//@ts-ignore`.

Both are suitable workaround at the moment, but I would much rather no do that.

Thanks for Typefest, it's one of my favorite libraries! Amazing work.

### Prior literature

- https://github.com/sindresorhus/type-fest/issues/506
- https://github.com/sindresorhus/type-fest/issues/579

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/775">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/775/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

Emiyaaaaa:
- Fixed #775 

I think @skarab42 uses `EnforceOptional` to restore the '?' by check `undefined`, because he dosen't used `keyof` in the implementation.

Now I used `keyof` to resolve the issue #775 , thus `EnforceOptional` is useless, so I seem need remove it, but it will cause some case changed like: 
```ts
MergeDeep<{a: string | undefined}, {}>
// Before: => {a?: string | undefined}
// After:  => {a: string | undefined}
```

But I still think I should remove `EnforceOptional`, what do you think? @sindresorhus

### DIFF
--- a/source/merge-deep.d.ts
+++ b/source/merge-deep.d.ts
@@ -1,7 +1,6 @@
 import type {ConditionalSimplifyDeep} from './conditional-simplify';
 import type {OmitIndexSignature} from './omit-index-signature';
 import type {PickIndexSignature} from './pick-index-signature';
-import type {EnforceOptional} from './enforce-optional';
 import type {Merge} from './merge';
 import type {
 	ArrayTail,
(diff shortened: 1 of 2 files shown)


Emiyaaaaa:
> > thus EnforceOptional is useless, so I seem need remove it, but it will cause some case changed like:
> 
> 
> 
> That is a bug fix. The new behavior is the correct one.
> 
> 
> 
> > But I still think I should remove EnforceOptional, what do you think?
> 
> 
> 
> 👍 

got it, it's ready for review

Issue #784, opened 2023-12-15
Canary: `MergeDeep` failures in `typescript@next` not detected by canary tests


voxpelli:
I'm seeing a lot of these style of errors in the canary tests in my own modules:

```
Error: ../node_modules/read-pkg/node_modules/type-fest/source/merge-deep.d.ts(144,5): error TS2321: Excessive stack depth comparing types 'PickRestType<ArrayTail<ArrayTail<Destination>>>[number]' and 'UnknownArrayOrTuple'.
Error: ../node_modules/read-pkg/node_modules/type-fest/source/merge-deep.d.ts(144,5): error TS2321: Excessive stack depth comparing types 'PickRestType<ArrayTail<ArrayTail<Destination>>>[number]' and 'UnknownRecord'.
Error: ../node_modules/read-pkg/node_modules/type-fest/source/merge-deep.d.ts(144,5): error TS2321: Excessive stack depth comparing types 'PickRestType<ArrayTail<ArrayTail<Source>>>[number]' and 'UnknownArrayOrTuple'.
Error: ../node_modules/read-pkg/node_modules/type-fest/source/merge-deep.d.ts(170,5): error TS2321: Excessive stack depth comparing types 'PickRestType<ArrayTail<ArrayTail<Destination>>>[number]' and 'UnknownArrayOrTuple'.
Error: ../node_modules/read-pkg/node_modules/type-fest/source/merge-deep.d.ts(170,5): error TS2321: Excessive stack depth comparing types 'PickRestType<ArrayTail<ArrayTail<Destination>>>[number]' and 'UnknownRecord'.
Error: ../node_modules/read-pkg/node_modules/type-fest/source/merge-deep.d.ts(170,5): error TS2321: Excessive stack depth comparing types 'PickRestType<ArrayTail<ArrayTail<Source>>>[number]' and 'UnknownArrayOrTuple'.
Error: ../node_modules/read-pkg/node_modules/type-fest/source/merge-deep.d.ts(196,5): error TS2321: Excessive stack depth comparing types 'PickRestType<ArrayTail<ArrayTail<Destination>>>[number]' and 'UnknownArrayOrTuple'.
Error: ../node_modules/read-pkg/node_modules/type-fest/source/merge-deep.d.ts(196,5): error TS2321: Excessive stack depth comparing types 'PickRestType<ArrayTail<ArrayTail<Destination>>>[number]' and 'UnknownRecord'.
[...]

PR #785, opened 2023-12-15
Tests: Ensure that `.d.ts` files are checked by setting `skipLibCheck: false`

voxpelli:
For background discussion, see https://github.com/sindresorhus/tsconfig/issues/15 and my personal decision to use `skipLibCheck: false` in my projects: https://github.com/voxpelli/tsconfig/issues/1

This enables the errors that are shown in https://github.com/sindresorhus/type-fest/issues/784 to be surfaced in here, and as such the canary tests to actually be a canary of something

### DIFF
--- a/.github/workflows/ts-canary.yml
+++ b/.github/workflows/ts-canary.yml
@@ -21,4 +21,6 @@ jobs:
           node-version: lts/*
       - run: npm install
       - run: npm install typescript@${{ matrix.typescript-version }}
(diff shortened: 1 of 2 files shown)


voxpelli:
Regression now shows correctly: https://github.com/sindresorhus/type-fest/actions/runs/7225971119/job/19690545983

Skärmavbild 2023-12-15 kl  20 01 12

PR #810, opened 2024-02-01
Make Simplify works for deep nesting

linbudu599:
This PR makes `Simplify` type works for computed types with deep nesting.

Example:

```typescript
type Foo = {
  foo: string;
  bar: number;
  baz: boolean;
}

type Result = {
  [K in keyof Foo]: Pick<Foo, K>
}

// {
//     foo: Pick<Foo, "foo">;
//     bar: Pick<Foo, "bar">;
//     baz: Pick<Foo, "baz">;
// }
type SimplifiedResult = Simplify<Result>;
```

With deep version:

```typescript
type DeepSimplify<T> = { [KeyType in keyof T]: T[KeyType] extends object ? DeepSimplify<T[KeyType]> : T[KeyType] } & {};

type Result = {
  [K in keyof Foo]: Pick<Foo, K>
}

// {
//     foo: {
//         foo: string;
//     };
//     bar: {
//         bar: number;
//     };
//     baz: {
//         baz: boolean;
//     };
// }
type DeepSimplifiedResult = DeepSimplify<Result>;
```

### DIFF
--- a/source/simplify.d.ts
+++ b/source/simplify.d.ts
@@ -55,4 +55,8 @@ fn(someInterface as Simplify<SomeInterface>); // Good: transform an `interface` 
 
 @category Object
 */
-export type Simplify<T> = {[KeyType in keyof T]: T[KeyType]} & {};
+export type Simplify<T> = {
+	[KeyType in keyof T]: T[KeyType] extends object
+		? Simplify<T[KeyType]>
+		: T[KeyType];
+} & {};


linbudu599:
Seems like this change breaks the other type behaviour in test cases, like `PartialOnUndefinedDeep`, maybe we should bring a new type called `SimplifyDeep` ? @sindresorhus

Emiyaaaaa:
> Seems like this change breaks the other type behaviour in test cases, like `PartialOnUndefinedDeep`, maybe we should bring a new type called `SimplifyDeep` ? @sindresorhus

`SimplifyDeep` is existing but it's in `type-fest/source/merge-deep`, make it to be a stable type will be better

sindresorhus:
Yes. It should be a separate type. That internal type should be properly exposed.

sindresorhus:
Bump :)

PR #823, opened 2024-03-02
Remove Opaque/UnwrapOpaque

ethanresnick:
This PR may be premature, so I'm happy to let it sit for a bit. But, the basic idea is that:

- Now that tag metadata has landed, `Tagged` types are strictly more powerful than `Opaque` ones, thanks to their ability to more naturally support multiple tags.
- `Opaque` has always been misnamed (since the underlying, untagged type isn't hidden at all, which is a critical property of opaque types in other languages).
- So it feels like we're at/approaching a point where `Opaque` and `UnwrapOpaque` should be removed.


source/opaque.d.ts
nplusp, line 1: I would suggest to rename the file then as well.

voxpelli:
We can do a `@deprecated` first, that will help notify people that it will go away and ease people's upgrade path?

ethanresnick:
@nplusp @voxpelli Ok, I've renamed the files from opaque -> tagged and opened #867 to mark these tags as deprecated before removing them.

KinderGouello:
Can you also update the README section ["Alternative type names"](https://github.com/sindresorhus/type-fest?tab=readme-ov-file#alternative-type-names) where Branded redirects to Opaque instead of Tagged?

ethanresnick:
@KinderGouello Done, here and in #867.

sindresorhus:
Closing for now as we don't plan to do this for a while. https://github.com/sindresorhus/type-fest/issues/1070

Issue #831, opened 2024-03-13
TS 5.4.2 breaks `ConditionalKeys` and `SetParameterType`


trevorade:
For `ConditionalKeys`, the root issue is: https://github.com/microsoft/TypeScript/issues/56644 which sound slike it will not be fixed. Basically, the `NonNullable` call in the `ConditionalKeys` type is ignored and does not run. Based on that issue, it's possible this type can be rewritten to be compatible with 5.4.2 but I'm unclear exactly on how...

I haven't debugged `SetParameterType` very closely but I just noticed that it too has issues in TS 5.4.2

voxpelli:
How come the canary is not picking up on this? https://github.com/sindresorhus/type-fest/actions/workflows/ts-canary.yml

voxpelli:
@trevorade Extending the tests to highlight this failure would be a good first step I think 👍 

This is the second breakage in the 5.4-line, after https://github.com/sindresorhus/type-fest/issues/784

trevorade:
Hey @voxpelli, we were just updating our code repository in my company to 5.4.2 and this test started breaking so I figured I would send an early signal. I'm not really looking for contribution opportunities at the moment.

[...]

PR #835, opened 2024-03-18
`SetParameterType`: compatible with break change about in typescript 5.4

Emiyaaaaa:
1. Compatible with break change about the rest index in array in typescript 5.4

~~2. Replace `npm test` with `npx tsc` in github workflow to make testing more comprehensive~~

related #831

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,4 +37,4 @@ jobs:
           node-version: 18
       - run: npm install
       - run: npm install typescript@${{ matrix.typescript-version }}
-      - run: npx tsc
+      - run: npx test
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,4 +37,4 @@ jobs:
           node-version: 18
       - run: npm install
       - run: npm install typescript@${{ matrix.typescript-version }}
-      - run: npm test
+      - run: npx tsc
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,4 +37,4 @@ jobs:
           node-version: 18
       - run: npm install
       - run: npm install typescript@${{ matrix.typescript-version }}
-      - run: npx test
+      - run: npm test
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -37,4 +37,4 @@ jobs:
           node-version: 18
       - run: npm install
       - run: npm install typescript@${{ matrix.typescript-version }}
-      - run: npx tsc
+      - run: npm test
```
voxpelli, line 40: This makes it test more than just typescript and may fail for unrelated reasons

voxpelli: As was mentioned in the issue, `tsd` ships with its own copy of typescript and as such isn't affected by any changes to the `typescript` dependency: https://github.com/sindresorhus/type-fest/issues/831#issuecomment-1995912221

One would instead have to do what I believe @shicks and @trevorade are doing: Use eg. [`expect-type`](https://www.npmjs.com/package/expect-type) instead: https://github.com/sindresorhus/type-fest/pull/499

And since `expect-type` is validated as part of running `tsc` itself, no changes to the GitHub Actions would be done then either.

That said: Replacing `tsd` with `expect-type` is a much broader topic, as highlighted by eg https://github.com/sindresorhus/type-fest/pull/499

Though: Since `expect-type` is already a dev dependency  of `type-fest` as far as I remember it could be good to add a test using that and ensure that breaks the canary tests correctly, at least that way we won't see a regression again.

Emiyaaaaa: Got it

Emiyaaaaa: > Though I'm a bit uncertain if those are actually run using `tsc` rather than `tsd`

Unfortunately it's using `tsd`.

Directory `test-d/` is excluded in tsconfig. https://github.com/sindresorhus/type-fest/blob/4f14bff7321b9a9876dca0719945423abd6b4da1/tsconfig.json#L14C2-L14C16

PR #838, opened 2024-03-20
Update `xo`

voxpelli:
This is a straight up update and auto-fix

Only manual fix was the addition of this in one file:

```ts
// eslint-disable-next-line unicorn/prevent-abbreviations
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,7 +1,13 @@
 name: CI
 on:
-  - push
-  - pull_request
+  push:
+    branches:
+      - main
+    tags:
+      - '*'
+  pull_request:
+    branches:
+      - main
 jobs:
   test:
     name: Node.js ${{ matrix.node-version }}
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
 		"expect-type": "^0.15.0",
 		"tsd": "^0.28.1",
 		"typescript": "^5.2.2",
-		"xo": "^0.56.0"
+		"xo": "^0.58.0"
 	},
 	"xo": {
 		"rules": {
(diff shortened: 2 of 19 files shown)


voxpelli:
* Added `npm-run-all2` as then I got feedback as to which of the five scripts if was that was failing (and it runs the tests in parallel which means quicker execution)
* Stopped running CI tests on pushes to non-`main` branches, they will be the source of PR:s anyway and as such will get CI run as part of the PR as well, which previously resulted in duplicated tests

source/subtract.d.ts
```diff
@@ -1,4 +1,6 @@
-import type {NumberAbsolute, BuildTuple, And, Or} from './internal';
+import type {
+	NumberAbsolute, BuildTuple, And, Or,
```
sindresorhus, line 2: These should be on separate lines.

sindresorhus (changes requested):

source/subtract.d.ts
voxpelli: Right, any idea why `xo` isn't picking up on that?

sindresorhus: No idea. First time I noticed it.

voxpelli: Caused by https://github.com/xojs/eslint-config-xo/pull/83, which requires a newline in beginning and end when there are so many parameters, but doesn't seems to enforce it in-between them 🤔

voxpelli: Fix is to modify it and remove the `minProperties`, one can then add `object-property-newline` to force newlines, but that will force it inbetween all properties and as such be equal to `minProperties: 1`, and its `allowAllPropertiesOnSameLine` allows all properties on a single line even when there's a curly newline, so will add this here as a workaround for now and then we can look at upstreaming to `xo`:

```json
	"object-curly-newline": ["error", {
		"multiline": true,
		"consistent": true
	}]
```

sindresorhus: We still want to only enforce the object to be on newlines when there are minimum 4 properties though. Not sure what the correct config would be for that.

voxpelli: Now it only enforces it if there's a newline somewhere within the object

PR #843, opened 2024-03-22
Create grouped dependabot.yml

### LINKED ISSUE #840: Set up dependabot / renovate
I think it would be great if PR:s like #838 and #839 was opened automatically. Especially something like the `tsd` one. Then it would be visible right away if a new version either breaks something or highlights something that was already broken but wasn't showing before, like #839 / #833.

Thoughts @sindresorhus @skarab42 and others?

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/840">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/840/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

voxpelli:
Fixes #840

I know @sindresorhus thinks this can get noisy, but I tried to make it as non-noisy as possible.

I would really love to get notified when eg. a new `tsd`, `xo` or similar dependency needs to be updated, and same for GitHub Actions.

This setup makes it so that at max one PR for npm and one PR for GitHub Actions will be created, and it will only be created/updated weekly, and for npm it will only create it if the version range _needs_ to be increased, so no needless bumping of the lower range of a version.

Can we give this a test and revisit it in a couple of weeks and see if it did get too noisy?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,18 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      github-actions:
+        patterns: "*"
+
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    versioning-strategy: "increase-if-necessary"
+    groups:
+      development-dependencies:
+        dependency-type: "development"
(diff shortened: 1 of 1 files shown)


sindresorhus (approved):

sindresorhus:
I'm willing to try it out.

PR #849, opened 2024-03-30
Add SingleKey and IfEmptyObject

### LINKED ISSUE #813: Require exactly one key of type `string`
Is there was a way to type an object to require that it has exactly one key? The key itself can be any `string` (or `number` or `symbol` for that matter), but I'd like to ensure there are at least _and_ at most one non-prototype/inherited property.

The use case is for typing the JavaScript implementation of [JsonLogic](http://jsonlogic.com/) ([npm](https://www.npmjs.com/package/json-logic-js) / [DT types](https://www.npmjs.com/package/@types/json-logic-js)). In the runtime code, the "logic" object itself and each nested object must satisfy the condition `Object.keys(logic).length === 1` ([see code](https://github.com/jwadhams/json-logic-js/blob/2.0.2/logic.js#L180)).

I tried `RequireExactlyOne<Record<string, any>, string>`, but that seems to be effectively the same as `Record<string, any>`.

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/813">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/813/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

ash-zzz:
Closes #813 

I also added IfEmptyObject since that was a missing guard type that I needed to use for this


sindresorhus:
I think it should be named `SingleKeyObject`. The current name is a bit ambiguous and could make you single it fetches a single key from an object, for example.

source/single-key.d.ts
```diff
@@ -0,0 +1,26 @@
...
+processOperation({ operation: { name: 'add' } });
+
+processOperation({ operation: { name: 'add' }, 'values': [1, 2] }); // Compliation error
+```
+
+@category Object
+ */
+
+// export type SingleKey<ObjectType> =
+// 	IsUnion<keyof ObjectType> extends true ? never : {} extends ObjectType ? never : ObjectType;
```
sindresorhus, line 32: Leftover

source/single-key.d.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const processOperation = <T>(operation: SingleKey<T>) => {};
+
+processOperation({ operation: { name: 'add' } });
+
+processOperation({ operation: { name: 'add' }, 'values': [1, 2] }); // Compliation error
+```
+
+@category Object
+ */
```
sindresorhus, line 29:
```suggestion
*/
```

source/single-key.d.ts
```diff
@@ -0,0 +1,36 @@
...
+	}
+});
+
+processOperation({
+	operation: {
+		name: 'add'
+	},
+	values: [1, 2]
+});
+// Compilation error
```
sindresorhus, line 25: Document the actual compilation here.

source/single-key.d.ts
```diff
@@ -0,0 +1,36 @@
...
+processOperation({
+	operation: {
+		name: 'add'
+	}
+});
+
+processOperation({
+	operation: {
+		name: 'add'
+	},
```
sindresorhus, line 22: Keep the example simple by not using a nested property here.

sindresorhus (changes requested):

source/single-key.d.ts
ash-zzz: Oops I thought I removed them all

ash-zzz:
> I think it should be named `SingleKeyObject`. The current name is a bit ambiguous and could make you single it fetches a single key from an object, for example.

Agreed, I just renamed it and fixed the documentation

sindresorhus:
Thanks :)

jakeboone02:
Thanks for doing this, @ash-zzz!

Should this work for types where the keys are unrestricted? If not, I'm not sure it satisfies my original request from #813, which was to declare an object type that requires exactly one key without restricting the key itself to any specific string or union of strings.

I tried `SingleKeyObject<Record<string, any>>` and `SingleKeyObject<{ [k: string]: any }>` but they both resolved to `never`. Something like `SingleKeyObject<Record<'union' | 'of' | 'strings', any>>` works fine.

Issue #136, opened 2020-10-22
Can I just install with devDependencies instead of dependencies?


lvdang:
Can I just install with devDependencies instead of dependencies? 

e.g.
 npm install type-fest --save-dev

papb:
If you're not authoring a module that ships its own typescript typings, then yes. Are you? How are you using `type-fest`?

lvdang:
I am using in development, but using webpack bundle 'ts-loader' to convert to .js files on production.

papb:
Since you're not creating a library for other developers to use, you can use it as a dev dependency :+1:

lvdang:
thanks

PR #851, opened 2024-03-30
Update readme.md to include a note about dev mode install

jpshackelford:
Quick update to the README for the question posted at https://github.com/sindresorhus/type-fest/issues/136 since I am least the second to newb to ask.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -75,6 +75,13 @@ PR welcome for additional commonly needed types and docs improvements. Read the 
 npm install type-fest
 ```
 
+<p>Or, for projects that do not export these or dependent types, install as a dev only dependency with: </p>
+
+```sh
+npm install type-fest --save-dev
+```
+
+
 *Requires TypeScript >=5.1*
 
 *Works best with [`{strict: true}`](https://www.typescriptlang.org/tsconfig#strict) in your tsconfig.*


readme.md
```diff
@@ -75,6 +75,13 @@ PR welcome for additional commonly needed types and docs improvements. Read the
 npm install type-fest
 ```
 
+<p>Or, for projects that do not export these or dependent types, install as a dev only dependency with: </p>
```
sindresorhus, line 78: This is not entirely accurate. The types don't need to be exported for it to need to be a dependency, and not dev dependency. Even if the types are used internally in a index.d.ts file (and not exported), `type-fest` needs to a dependency. Basically, if types included in a package use types from `type-fest`, `type-fest` needs to be a dependency.

voxpelli: Unless one explicitly add an ignore to the imports, then they become optional imports, see eg this helper of mine: https://github.com/voxpelli/ts-ignore-import

ShlomoCode: @sindresorhus type dependencies are defined as "development dependencies" because they are only checked by `tsc`, not by the runtime -
If I run the TS code with vite/bun/nodejs v23.6.0+ they doing "stupid" type stripping so i don't need the types dependencies to run the code in production

Only during development or during CICD am I expected to install a full installation, in production it's just a waste of time and space

Issue #637, opened 2023-06-14
`Integer<1e+100>` should be `1e+100`, but it is `never`


ootidea:
TypeScript handles large integers of the `number` type in exponential notation. It seems that `bigint` does not do the same.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/sindresorhus) so you can upvote and help fund this issue.
- The funding will be given to active contributors.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/sindresorhus/type-fest/issues/637">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/637/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/sindresorhus/type-fest/issues/637/pledge.svg">
</picture>
</a>

PR #857, opened 2024-04-07
Add `IsInteger` and  `IsFloat`; Fix `Integer` and `Float` handing with edge case

Emiyaaaaa:
fixed: #637 

Breaking but I think it's the corrent changes:
```ts
type Change1 = Integer<1 | 1.5>
// latest version => never // auther commented: This may be undesired behavior
// this PR        => 1

type Change2 = Integer<1e+100>
// latest version => never
// this PR        => 1e+100 // Number.IsInteger(1e+100) is true

type Change3 = Float<PositiveInfinity>
// latest version => PositiveInfinity
// this PR        => never // Infinity isn't a float.
```

And clarifies how to deal with some special number type like '1.0', '0x10' and negative number

### DIFF
--- a/source/numeric.d.ts
+++ b/source/numeric.d.ts
@@ -1,50 +1,9 @@
-import type {Not} from './internal';
+import type {IsFloat} from './is-float';
+import type {IsInteger} from './is-integer';
 
 export type Numeric = number | bigint;
 
 type Zero = 0 | 0n;
-
-/**
-Returns the given number if it is a float, like `1.5` or `-1.5`.
-*/
-type IsFloat<T extends number> =
-`${T}` extends `${infer _Sign extends '' | '-'}${number}.${infer Decimal extends number}`
-	? Decimal extends Zero
-		? false
-		: true
-	: false;
-
-/**
-Returns the given number if it is an integer, like `-5`, `1` or `100`.
-
-Like [`Number#IsInteger()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/IsInteger) but for types.
-
-@example
-```
-type Integer = IsInteger<1>;
-//=> true
-type IntegerWithDecimal = IsInteger<1.0>;
-//=> true
-type NegativeInteger = IsInteger<-1>;
-//=> true
-type Float = IsInteger<1.5>;
-//=> false
-
-//=> supported non-decimal numbers
-type OctalInteger: IsInteger<0o10>;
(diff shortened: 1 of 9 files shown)


source/numeric.d.ts
```diff
@@ -1,7 +1,51 @@
+import type {Not} from './internal';
+
 export type Numeric = number | bigint;
 
 type Zero = 0 | 0n;
 
+/**
+Returns the given number if it is a float, like `1.5` or `-1.5`.
```
sindresorhus, line 8: It doesn't return the number.

```suggestion
Returns whether the given number is a float, like `1.5` or `-1.5`.
```

source/numeric.d.ts
```diff
@@ -1,7 +1,51 @@
...
+*/
+type IsFloat<T extends number> =
+`${T}` extends `${infer _Sign extends '' | '-'}${number}.${infer Decimal extends number}`
+	? Decimal extends Zero
+		? false
+		: true
+	: false;
+
+/**
+Returns the given number if it is an integer, like `-5`, `1` or `100`.
```
sindresorhus, line 18:
```suggestion
Returns whether the given number is an integer, like `-5`, `1` or `100`.
```

source/numeric.d.ts
```diff
@@ -67,7 +131,10 @@ declare function setYear<T extends number>(length: Integer<T>): void;
...
 // `${bigint}` is a type that matches a valid bigint literal without the `n` (ex. 1, 0b1, 0o1, 0x1)
 // Because T is a number and not a string we can effectively use this to filter out any numbers containing decimal points
-export type Integer<T extends number> = `${T}` extends `${bigint}` ? T : never;
+export type Integer<T extends number> =
+	T extends unknown // To distributive type
+		? IsInteger<T> extends true ? T : never
+		: never; // Never happens
 
 /**
 A `number` that is not an integer.
```
sindresorhus, line 152: It should document the behavior for `Infinity`.

source/numeric.d.ts
sindresorhus: It should document the behavior for `Infinity`.

sindresorhus (changes requested):

Emiyaaaaa:
Should we expose `IsFloat` and `IsInteger`? I think these are useful types. @sindresorhus

sindresorhus:
Yes. I think they are useful too. Up to you whether it's done now or in a later PR.

PR #870, opened 2024-04-24
Fix: "Zero" is not exported

Zhang-Wei-666:
fix:
image


Emiyaaaaa:
Related https://github.com/sindresorhus/type-fest/pull/824

PR #874, opened 2024-04-29
TSConfig JSON: Add `preserve` module type and `ES2022` lib types

blakewilson:
This PR adds:
- The [`preserve`](https://www.typescriptlang.org/tsconfig/#preserve) module type that was added in TypeScript 5.4
- The [`ES2022`](https://www.typescriptlang.org/tsconfig/#lib) libs ([full list from source as of 5.5](https://github.com/microsoft/TypeScript/blob/release-5.5/src/lib/libs.json#L64-L70))


Emiyaaaaa (approved):

blakewilson:
Thanks for the quick review and merge 🙌

Issue #708, opened 2023-10-15
Unable to use `Opaque<Record<a, b>, c>`


fregante:
I think this is a limitation of TypeScript rather than the type's, but I wanted to document it here and see if anyone can come up with a solution. I think the issue is specifically because the `Record` only accepts `string` and is unable to accept the `[tag]` symbol in the union.

```ts
import {Opaque} from "type-fest";

type CustomString = Opaque<string, 'custom'>;
const string: CustomString = 'hello' as CustomString;
// ✅

type CustomObject = Opaque<{foo: 'bar'}, 'custom'>;
const object: CustomObject = {foo: 'bar'} as CustomObject;
// ✅

type CustomRecord = Opaque<Record<string, string>, 'custom'>;
const record: CustomRecord = {foo: 'bar'} as CustomRecord;
// ❌ Conversion of type '{ foo: string; }' to type 'CustomRecord' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
//  Property '[tag]' is missing in type '{ foo: string; }' but required in type 'TagContainer<"custom">'.(2352)

const emptyRecord: CustomRecord = {} as CustomRecord;
// ✅ 🤷‍♂️ 
```

[...]

PR #875, opened 2024-04-29
Don't make `Tagged` type compatibility dependent on type fest version

ethanresnick:
Since removing `Opaque`/`UnwrapOpaque` is obviously gonna require a new major version, I wanted to put this PR up now, in case it makes sense to land both changes in the same major version.

This PR switches the `tag` key used by the `Tagged` type to be a string rather than a symbol, undoing [this commit](https://github.com/causaly/type-fest/commit/7f63a6e7e0884cebd11b645c42a17041969add94). The rationale for that, which I've summarized from [this comment](https://github.com/sindresorhus/type-fest/issues/708#issuecomment-1769357978) is:

1. It's easy for two different copies of `type-fest` to end up installed in a project. Consider a dependency tree like:

    ```
    @my-company/app
    ├─┬ @my-company/types
    │ └── type-fest
    ├─┬ @my-company/some-internal-library
    │ └─┬ @my-company/types
    │   └── type-fest
    └── type-fest
    ```

   Depending on the version of `type-fest` installed at the root (i.e., by the `@my-company/app` package) and the version of the `@my-company/types` package used the chosen version of `@my-company/some-internal-library`, npm may be forced to install multiple `type-fest` copies.

2. When multiple type-fest copies are present, a `Tagged` type produced by one copy is not assignable to a `Tagged` type produced by another copy _even if the types have the same tag names, tag metadata, and base/untagged type_, because Typescript sees the `tag` symbol from each copy of `type-fest` as distinct.

   This is weird behavior, as it ends up making assignability dependent on exactly how/when npm chooses to dedupe packages, which isn't even necessarily stable as new packages are installed or as npm is updated or as installed packages are updated.

   In the example dependency tree above, it means that: if `@my-company/some-internal-library` returns a tagged type from its copy of the `@my-company/types` package, that type may not be assignable to the copy of that same type exported by the copy of `@my-company/types` that `@my-company/app` is depending on directly — even if the tagged type hasn't changed (in terms of its tag names/tag metadata/base type).

This also happens to make `Record<string, T>` taggable (solving https://github.com/sindresorhus/type-fest/issues/708), at the expense of `Record<symbol, T>` not being taggable (which seems like a good tradeoff).

Given that type-fest used a string-based key for many years, I don't think this poses any big issues, but there would be some small user-facing changes. For example, `keyof Tagged<{ foo: string }, 'SomeTag'>` would become `'foo' | '__type-fest-tag__'` instead of `'foo' | typeof tag`; that's not fundamentally different, but  it's maybe a tad less clear. It would probably be helpful too, at that point, for `type-fest` to export a string literal type with the name of the magic tag string (i.e., `"__type-fest-tag__"`), so that users can do `Exclude<keyof Tagged, TypeFestTagKey>`, in much the same way they might've had to do `Exclude<keyof Tagged, symbol>` before.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,5 +96,8 @@
 		"compilerOptions": {
 			"noUnusedLocals": false
 		}
+	},
+	"dependencies": {
+		"tagged-tag": "^0.1.0"
 	}
 }
--- a/package.json
+++ b/package.json
@@ -49,6 +49,9 @@
 		"json",
 		"generics"
 	],
+	"dependencies": {
+		"tagged-tag": "^0.1.0"
+	},
 	"devDependencies": {
 		"@sindresorhus/tsconfig": "^7.0.0",
 		"expect-type": "^1.2.1",
@@ -96,8 +99,5 @@
 		"compilerOptions": {
 			"noUnusedLocals": false
 		}
-	},
-	"dependencies": {
-		"tagged-tag": "^0.1.0"
 	}
 }
--- a/source/tagged.d.ts
+++ b/source/tagged.d.ts
@@ -1,4 +1,4 @@
-declare const tag: unique symbol;
+import type tag from 'tagged-tag';
 
 export type TagContainer<Token> = {
 	readonly [tag]: Token;
@@ -254,3 +254,5 @@ export type UnwrapOpaque<OpaqueType extends TagContainer<unknown>> =
 		: OpaqueType extends Opaque<infer Type, OpaqueType[typeof tag]>
 			? Type
 			: OpaqueType;
+
+export { type default as tag } from 'tagged-tag';
--- a/source/tagged.d.ts
+++ b/source/tagged.d.ts
@@ -255,4 +255,4 @@ export type UnwrapOpaque<OpaqueType extends TagContainer<unknown>> =
 			? Type
 			: OpaqueType;
 
-export { type default as tag } from 'tagged-tag';
+export {type default as tag} from 'tagged-tag';


ethanresnick:
Perhaps a better, less-disruptive approach to the same ends: publish a new package that _just_ has the tag symbol, and commit to never updating the version of that package; that way, it will only be installed once, and every copy of `type-fest` will reference the same symbol.

sindresorhus:
> Perhaps a better, less-disruptive approach to the same ends: publish a new package that just has the tag symbol, and commit to never updating the version of that package; that way, it will only be installed once, and every copy of type-fest will reference the same symbol.

Good idea. Done here: https://github.com/sindresorhus/tagged-tag

The package is a ESM module, so it will not work until type-fest targets ESM, which it will for v5. I have published a 0.1.0 in the meantime to provide feedback. If it's good, I will do a 1.0.0 before merging this PR, and at that point, that package is locked forever.

belgattitude:
I like the approach, but I’m wondering how safe it is with a locked external package ?

 In the past I’ve seen bugs in package managers in the way they hoist a same exact version at different paths. Happened mostly in monorepos with yarn or pnpm. In general a dedupe would fix it but sometimes it goes further (older versions of pnpm or some specific settings). In general it’s safe when using recent pm, but it still comes back from time to time. 

Just an open remark. That said I like the approach

ethanresnick:
I think you're right @belgattitude that using an external package may not solve this in some edge cases, or may occasionally require the user to do a manual `npm dedupe` (or other package manager equivalent). 

At the very least, we have to lock _both the minor + patch versions_ of the external package, not just its major version, because of npm's behavior [described here](https://docs.npmjs.com/cli/v10/commands/npm-dedupe):

> During the installation process, the c@1.0.3 dependency for b was placed in the root of the tree. Though d's dependency on c@1.x could have been satisfied by c@1.0.3, the newer c@1.9.0 dependency was used, because npm favors updates by default, even when doing so causes duplication.

All that said, I think an external package will solve this in the vast majority of cases, and we know that applying this approach in type-fest v5 is totally non-breaking (as a user installing type-fest v5 would've gotten a new copy with a new symbol anyway). So I think we should do this non-breaking, minimally-invasive approach first, and then only switch to a string key if it proves inadequate.

ethanresnick:
@sindresorhus I've updated this to use the new package.

ethanresnick:
Updated!

Issue #878, opened 2024-05-09
ConditionalKeys does not correctly filter out `never` type


roryabraham:
I'm seeing some unexpected behavior in this minimal example:

```ts
type ConditionalKeys<Base, Condition> =
{
	// Map through all the keys of the given base type.
	[Key in keyof Base]-?:
	// Pick only keys with types extending the given `Condition` type.
	Base[Key] extends Condition
	// Retain this key since the condition passes.
		? Key
	// Discard this key since the condition fails.
		: never;

	// Convert the produced object into a union type of the keys which passed the conditional test.
}[keyof Base];

type MyMap = {
	a: {
		b: string;
	};
	c: {
		html: string;
	};
	d: never;
};

type MyFilteredMap = ConditionalKeys<MyMap, {html: string}>;
```

The root cause is that `never` can extend anything. I'm not sure if this is expected behavior or not, but it was a gotcha for me. It can be fixed like so:

```ts
type ConditionalKeys<Base, Condition> = {
    [Key in keyof Base]-?:
        // The key is retained if its value extends the given `Condition` and is not `never`
        Base[Key] extends Condition
            ? (Base[Key] extends never ? never : Key)
            : never;
}[keyof Base];
```

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/sindresorhus) so you can upvote and help fund this issue.
- The funding will be given to active contributors.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/sindresorhus/type-fest/issues/878">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/878/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/sindresorhus/type-fest/issues/878/pledge.svg">
</picture>
</a>

roryabraham:
Happy to open a PR if we agree that this is unexpected behavior.

sindresorhus:
That is indeed unexpected behavior. A pull request would be amazing 🙏

roryabraham:
Opened https://github.com/sindresorhus/type-fest/pull/881. Thanks in advance for reviewing 🙇🏼

PR #881, opened 2024-05-23
Fix ConditionalKeys - prevent never from being included unintentionally

roryabraham:
Fixes https://github.com/sindresorhus/type-fest/issues/878

The following minimal example doesn't work as expected with the minimal example, because it includes `d` in the result, even though the value is `never`

```ts
type ConditionalKeys<Base, Condition> =
{
	// Map through all the keys of the given base type.
	[Key in keyof Base]-?:
	// Pick only keys with types extending the given `Condition` type.
	Base[Key] extends Condition
	// Retain this key since the condition passes.
		? Key
	// Discard this key since the condition fails.
		: never;

	// Convert the produced object into a union type of the keys which passed the conditional test.
}[keyof Base];

type MyMap = {
	a: {
		b: string;
	};
	c: {
		html: string;
	};
	d: never;
};

type MyFilteredMap = ConditionalKeys<MyMap, {html: string}>;
```

### DIFF
--- a/source/conditional-keys.d.ts
+++ b/source/conditional-keys.d.ts
@@ -1,3 +1,5 @@
+import type {IfNever} from './if-never';
+
 /**
 Extract the keys from a type where the value type of the key extends the given `Condition`.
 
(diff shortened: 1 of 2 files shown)


roryabraham:
Ok, updated it to work for targeting `never` as well. I might have use-cases for that actually 🤔

source/conditional-keys.d.ts
```diff
@@ -36,10 +36,16 @@ export type ConditionalKeys<Base, Condition> =
 	[Key in keyof Base]-?:
 	// Pick only keys with types extending the given `Condition` type.
 	Base[Key] extends Condition
-	// Retain this key since the condition passes.
-		? Key
-	// Discard this key since the condition fails.
+		// Retain this key since the condition passes
+		?
+			// If the condition extends never, only include the key if the value for that key is also never
```
roryabraham, line 41: let me know if you want to format this differently. I chose to progressively increase indentation for each level of ternary, and to keep comments indented to the same level as the line they're describing.

roryabraham:
I can't explain this difference - might be a TS bug?

roryabraham:
@sindresorhus let me know how you'd like to proceed. I can remove or comment out the failing test, or revert to https://github.com/sindresorhus/type-fest/pull/881/commits/5ce8679ffe3a3c6c16fbe25f2fca9526fd2447f4 (i.e: don't try to make it work for targeting never). Either way works for me.

roryabraham:
Found a relevant SO: https://stackoverflow.com/questions/58977876/generic-conditional-type-resolves-to-never-when-the-generic-type-is-set-to-never

Issue #879, opened 2024-05-15
Paths<T> should handle recursive types


Xriuk:
Currently `Paths<T>` does not handle types like this (translation of some backend DB entities):
```js
type MyEntity ={
   myOtherEntity: MyOtherEntity;
    ...
}

type MyOtherEntity ={
    myEntity: MyEntity;
    ...
}

type MyEntityPaths = Paths<MyEntity>;
```
TS gives an error like this:
`The type of property ... circularly references itself in mapped type ...`

The main issue here would be determining if any given type recurses, which would require comparing already entered types (which afaik is unfeasible).
A possible solution would be adding an optional "hard limit" of depth parameter, which won't go beyond N levels deep. In this case also types using `Paths<T>` like `PickDeep` should include this parameter and forward it to `Paths<T>`.

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/sindresorhus) so you can upvote and help fund this issue.
- The funding will be given to active contributors.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/sindresorhus/type-fest/issues/879">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/879/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/sindresorhus/type-fest/issues/879/pledge.svg">
</picture>
</a>

PR #891, opened 2024-06-11
`Paths`: prevent infinite recursion

Emiyaaaaa:
fixed: #879

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
 	},
 	"xo": {
 		"rules": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -47,6 +47,7 @@
 	},
 	"xo": {
 		"rules": {
+			"@typescript-eslint/no-extraneous-class": "off",
```
Emiyaaaaa, line 50: This change is not related to this PR, but some dependencies of xo updated automatically caused existing code checks to fail

PR #894, opened 2024-06-20
Improve documentation of `IterableElement` type & add tests for using with `Set`

xeho91:
Added a new type `SetKeys` which is a helper to quickly infer the type of keys from a strongly typed `Set` or `ReadonlySet`.


xeho91:
> I think you can already use the `IterableElement` type for this.

Oh! I feel dumb missing this one. You're right, `IterableElement` gives same result.

Least I can do is update examples of `IterableElement` and mention keywords of the common Iterable instances _(`Set`, `Map`)_ in the `readme.md` next to this type.

Especially this mention: https://github.com/sindresorhus/type-fest/blob/e75dbe1fee378d1aab81091c30da7b29be9e9a5f/source/iterable-element.d.ts#L36

Because I think is too hidden, at least for people like me. Hopefully not too many 😅

sindresorhus:
👍

And also here: https://github.com/sindresorhus/type-fest?tab=readme-ov-file#alternative-type-names Should probably add `SetKeys` there too, as that's a more correct term for it.

PR #897, opened 2024-06-22
Add `StructuredCloneable` type

### LINKED ISSUE #855: Type for serializable objects?
Would it make sense to create a type for [serializable objects](https://developer.mozilla.org/en-US/docs/Glossary/Serializable_object)? 

Similar to Jsonify et al., this would match values that could losslessly be deep-copied via `structuredClone` or `postMessage`.

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/855">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/855/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

thecodewarrior:
This adds two types for values losslessly cloneable using `structuredClone`.

Notes:
- The web and non-web types are separate so people who aren't writing browser code don't pull in web types
- Custom error classes do match, despite the fact that `structuredClone` isn't able to losslessly clone them (they're cloned as base `Error` instances). After doing some research I couldn't figure out a way to include the root Error class but exclude subclasses, so I just accepted the limitation and documented the discrepancy.

Fixes #855


source/structured-cloneable.d.ts
```diff
@@ -0,0 +1,66 @@
...
+	| Date
+	| Error
+	| RegExp
+	| TypedArray;
+
+type StructuredCloneableCollection =
+	| StructuredCloneable[]
+	| {[k: string]: StructuredCloneable; [k: number]: StructuredCloneable}
+	| Map<StructuredCloneable, StructuredCloneable>
+	| Set<StructuredCloneable>;
```
sindresorhus, line 26: I would use the read-only type versions here so it can support read-only collections too.

Emiyaaaaa:
I think this PR needs to wait for #898 to merge.

sindresorhus:
Tests are failing now. Not sure how we can handle DOM types conditionally.

sindresorhus:
Relevant: https://github.com/microsoft/TypeScript/issues/31894

sindresorhus:
I think we could do it by adding a `dom` sub-export in package.json: https://nodejs.org/api/packages.html#subpath-exports

thecodewarrior:
Adding the sub-export would probably involve changes to the testing/build infrastructure (running twice for non-dom/dom builds), which I'm not familiar with, but if you want to implement that I'd be happy to add back the web type. If we're doing a sub-export I think it wouldn't even need to have a separate name, just `import {StructuredCloneable} from 'type-fest/dom'` instead of `import {StructuredCloneable} from 'type-fest'`

sindresorhus:
Thanks :)

PR #901, opened 2024-06-29
`Jsonify`: Convert `undefined` to `null` in union element of array

m-shaka:
`Jsonify` doesn't handle correctly union type with undefined in array.
You get `[1, null]` from `JSON.parse(JSON.stringify([1, undefined]))`.
Like tuple type, `Array<undefined | T>` should be converted to `Array<null | T>`

### DIFF
--- a/test-d/jsonify.ts
+++ b/test-d/jsonify.ts
@@ -192,6 +192,9 @@ expectType<never>(plainSymbol);
 declare const arrayMemberUndefined: Jsonify<Array<typeof undefined>>;
 expectType<null[]>(arrayMemberUndefined);
 
+declare const arrayMemberUnionWithUndefined: Jsonify<Array<typeof undefined | typeof number>>;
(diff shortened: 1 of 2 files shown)


test-d/jsonify.ts
```diff
@@ -192,6 +192,9 @@ expectType<never>(plainSymbol);
 declare const arrayMemberUndefined: Jsonify<Array<typeof undefined>>;
 expectType<null[]>(arrayMemberUndefined);
 
+declare const arrayMemberUnionWithUndefined: Jsonify<Array<typeof undefined | typeof number>>;
+expectType<Array<null | number>>(arrayMemberUnionWithUndefined);
```
sindresorhus, line 196: Can you also do a test where `Array<typeof undefined | typeof number>` is not at the top-level?

m-shaka: Added!

sindresorhus (approved):

PR #908, opened 2024-07-08
Conditionally add node/DOM types to StructuredCloneable

### LINKED ISSUE #899: Web-specific version of `StructuredCloneable` type
See: https://github.com/sindresorhus/type-fest/pull/897#issuecomment-2189741111

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/899">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/899/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

thecodewarrior:
While investigating #899, I discovered a solution that lets us only include DOM/Node types if they actually exist. Based on [this code](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3263bd44/types/node/globals.d.ts#L406-L411) from the node type declarations, I came up with this solution:

```ts
/** Grab the specified type from its corresponding global constructor, if it's declared */
type ExtractFromGlobalConstructor<Name extends string> =
	typeof globalThis extends Record<Name, new (...args: any[]) => infer T> ? T : never
```

This only works with types that have a public constructor in `globalThis`, but `Blob`, `File`, and all the DOM types appear to have constructors and work like this. In order to verify that all the DOM types have public constructors I added `@webgpu/types`, `@types/dom-webcodecs`, and `@types/webxr`, which got me most of the types I was missing, then I created this file to manually check them:
```ts
type DOMKeys =
	| 'AudioData'
	| 'CropTarget'
	| 'CryptoKey'
	| 'DOMException'
	| 'DOMMatrix'
	| 'DOMMatrixReadOnly'
	| 'DOMPoint'
	| 'DOMPointReadOnly'
	| 'DOMQuad'
	| 'DOMRect'
	| 'DOMRectReadOnly'
	| 'FileList'
	| 'FileSystemDirectoryHandle'
	| 'FileSystemFileHandle'
	| 'FileSystemHandle'
	| 'GPUCompilationInfo'
	| 'GPUCompilationMessage'
	| 'ImageBitmap'
	| 'ImageData'
	| 'RTCCertificate'
	| 'VideoFrame'

/**
- `Name` = found global with valid constructor
- `~Name` = found global, but no constructor found
- `!Name` = global not found
*/
type FindGlobals<Name extends string> =
	Name extends string
		? typeof globalThis extends Record<Name, new (...args: any[]) => any> ? Name
			: typeof globalThis extends Record<Name, any> ? `~${Name}`
				: `!${Name}`
		: never;

import "@webgpu/types"

const found: FindGlobals<DOMKeys>[] = [
	'AudioData',
	'!CropTarget', // no typescript declarations exist for this yet
	'CryptoKey',
	'DOMException',
	'DOMMatrix',
	'DOMMatrixReadOnly',
	'DOMPoint',
	'DOMPointReadOnly',
	'DOMQuad',
	'DOMRect',
	'DOMRectReadOnly',
	'FileList',
	'FileSystemDirectoryHandle',
	'FileSystemFileHandle',
	'FileSystemHandle',
	'GPUCompilationInfo',
	'GPUCompilationMessage',
	'ImageBitmap',
	'ImageData',
	'RTCCertificate',
	'VideoFrame',
]
```

Fixes #899

### DIFF
--- a/source/find-global-type.d.ts
+++ b/source/find-global-type.d.ts
@@ -1,8 +1,7 @@
 /**
 Tries to find the type of a global with the given name.
 
-Limitations: Due to peculiarities with the behavior of `globalThis`, "globally defined" only includes `var`
-declarations in `declare global` blocks, not `let` or `const` declarations.
+Limitations: Due to peculiarities with the behavior of `globalThis`, "globally defined" only includes `var` declarations in `declare global` blocks, not `let` or `const` declarations.
 
 @example
 ```
@@ -27,9 +26,7 @@ Tries to find one or more types from their globally-defined constructors.
 
 Use-case: Conditionally referencing DOM types only when the DOM library present.
 
-*Limitations:* Due to peculiarities with the behavior of `globalThis`, "globally defined" has a narrow definition in
-this case. Declaring a class in a `declare global` block won't work, instead you must declare its type using an
-interface and declare its constructor as a `var` (*not* `let`/`const`) inside the `declare global` block.
+*Limitations:* Due to peculiarities with the behavior of `globalThis`, "globally defined" has a narrow definition in this case. Declaring a class in a `declare global` block won't work, instead you must declare its type using an interface and declare its constructor as a `var` (*not* `let`/`const`) inside the `declare global` block.
 
 @example
 ```
--- a/source/find-global-type.d.ts
+++ b/source/find-global-type.d.ts
@@ -1,0 +1,52 @@
+/**
+Tries to find one or more types from their globally-defined constructors.
+
+Use-case: Conditionally referencing DOM types only when the DOM library present.
+
+*Note:* Globally-defined has a narrow definition in this case due to peculiarities with the behavior of `globalThis`.
+Merely declaring a class in a `declare global` block won't work, instead you must declare its constructor as a `var`
+(not `let`/`const`) inside the `declare global` block. This typically is done using a combination of interface +
+constructor style, rather than the es6 class syntax.
+
+@example
+```
+import type {FindGlobalType} from 'type-fest';
+
+class Point {
+	constructor(public x: number, public y: number) {
+	}
+}
+
+type PointLike = Point | FindGlobalType<'DOMPoint'>;
+
+function doSomething(point: PointLike): number {
+	// ...
+}
+```
+
+@example
+```
+import type {FindGlobalType} from 'type-fest';
+
+declare global {
+	// ES6 class syntax won't add the key to `globalThis`
+	class Foo {}
+
+	// interface + constructor style works
+	interface Bar {}
+	interface BarConstructor {
+		new (): Bar;
+	}
+	var Bar: BarConstructor; // not let or const
+}
+
+type FindFoo = FindGlobalType<'Foo'>; // doesn't work
+type FindBar = FindGlobalType<'Bar'>; // works
+```
+
+@category Utilities
+*/
+export type FindGlobalType<Name extends string> =
+	Name extends string
+		? typeof globalThis extends Record<Name, new (...arguments: any[]) => infer T> ? T : never
+		: never;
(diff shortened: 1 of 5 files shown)


Emiyaaaaa (approved):

source/structured-cloneable.d.ts
```diff
@@ -11,37 +11,46 @@ type StructuredCloneablePrimitive =
 	| Number
 	| String;
 
+/** Grab the specified types from their global constructors, if they're declared */
+type ExtractFromGlobalConstructors<Name extends string> =
```
sindresorhus, line 15: This would be useful as an exported type. If you're not interested in doing that, at least put it in internal.d.ts, so we don't forget to eventually export it.

voxpelli:
Created a related PR, #912, that makes the type checks properly fail for the current scenario.

Would be good to get work on that in tandem with this.

I also think we should find a model where we test the types in contexts where eg. both the `dom` lib and `@types/node` is available, to ensure it works as expected in all supported context

source/find-global-type.d.ts
```diff
@@ -0,0 +1,48 @@
...
+	var Bar: BarConstructor; // Not let or const
+}
+
+type FindFoo = FindGlobalType<'Foo'>; // Doesn't work
+type FindBar = FindGlobalType<'Bar'>; // Works
+```
+
+@category Utilities
+*/
+export type FindGlobalType<Name extends string> =
```
voxpelli, line 45: Should we name this `FindGlobalInstanceType` or something instead? As it only looks for global constructors, if used with eg `FindGlobalType<'fetch'>` it won't give the type?

voxpelli (changes requested):

source/find-global-type.d.ts
voxpelli: [`InstanceType`](https://www.typescriptlang.org/docs/handbook/utility-types.html#instancetypetype) being the name of the built in utility type for finding the instance type of a constructor

voxpelli: Maybe something like this?

```ts
type FindGlobalType<Name extends string> =
  typeof globalThis extends Record<Name, infer T>
    ? T
    : never;

type FindGlobalInstanceType<Name extends string> =
  FindGlobalType<Name> extends new (...args: any[]) => infer T
    ? T
    : never;
```

thecodewarrior: That name would probably be better, I wasn't aware of that builtin, and splitting it up like that would actually be nice.

If we wanted to be a bit more precise, we could also use `FindInstanceTypeFromGlobal` or `FindTypeFromGlobalConstructor`. I would love to have a more versatile type for this, since it won't even work for a `declare global { class Foo { } }`, but I don't know any way to do that.

thecodewarrior: Just pushed a rename and split to `FindGlobalType`/`FindGlobalInstanceType`. For reasons I wasn't able to determine, `FindGlobalInstanceType` would return `unknown` for non-existent keys when delegating to `FindGlobalType`, so it still directly references `globalThis`

source/find-global-type.d.ts
```diff
@@ -0,0 +1,67 @@
...
+@category Utilities
+*/
+export type FindGlobalType<Name extends string> = typeof globalThis extends Record<Name, infer T> ? T : never;
+
+/**
+Tries to find one or more types from their globally-defined constructors.
+
+Use-case: Conditionally referencing DOM types only when the DOM library present.
+
+*Limitations:* Due to peculiarities with the behavior of `globalThis`, "globally defined" has a narrow definition in
```
sindresorhus: I recommend opening a TypeScript issue about this limitation.

sindresorhus (approved):

voxpelli (approved):

sindresorhus:
Thanks for contributing 🙏

source/find-global-type.d.ts
thecodewarrior: The odd behavior itself is correct (microsoft/TypeScript#39504), so there's nothing to be done on that front. In terms of avoiding this hack in the first place, there's already an issue regarding that: microsoft/TypeScript#31894

Issue #909, opened 2024-07-08
Exact not function correctly with Dates since last commit


nlynzaad:
Hi,

First off, I'm a new user of the project and just want to say thanks for your efforts in maintaining this exceptionally handy project. 

Was struggling to get the Exact type to work and it seems to be related to a change that was made in commit #902, specifically the change from object to JsonObject. When having a date field, the validation does not throw an error as expected. The error however does appear in the version prior to the commit and in the new version if the Date field is removed.

```
import type {Exact, IsEqual, JsonObject} from "type-fest";
import type {ArrayElement} from "type-fest/source/internal";
import type {TagContainer} from "type-fest/source/opaque";
import {ExactObject} from "type-fest/source/exact";

type UserType = {
	id: string,
	name: string,
	createdAt: Date,
	email?: string,
};

type UserTypeWithoutDate = {
	id: string,
	name: string,
	email?: string | undefined,
};

const testValues = {
	id: 'asd',
	name: 'John',
	createdAt: new Date(),
	email: undefined,
	surname: 'Doe',
};

const testValuesWithoutDate = {
	id: 'asd',
	name: 'John',
	email: undefined,
	surname: 'Doe',
};

export type Exact_OLD<ParameterType, InputType> =
	IsEqual<ParameterType, InputType> extends true ? ParameterType
		// Convert union of array to array of union: A[] & B[] => (A & B)[]
		: ParameterType extends unknown[] ? Array<Exact<ArrayElement<ParameterType>, ArrayElement<InputType>>>
			// In TypeScript, Array is a subtype of ReadonlyArray, so always test Array before ReadonlyArray.
			: ParameterType extends readonly unknown[] ? ReadonlyArray<Exact<ArrayElement<ParameterType>, ArrayElement<InputType>>>
				// Leave tagged types as-is. We could try to make the untagged part Exact, and just leave the tag as-is, but that seems to create instanitation excessively deep errors.
				: ParameterType extends TagContainer<unknown> ? ParameterType
					: ParameterType extends object ? ExactObject<ParameterType, InputType>
						: ParameterType;

export type Exact_NEW<ParameterType, InputType> =
	IsEqual<ParameterType, InputType> extends true ? ParameterType
		// Convert union of array to array of union: A[] & B[] => (A & B)[]
		: ParameterType extends unknown[] ? Array<Exact<ArrayElement<ParameterType>, ArrayElement<InputType>>>
[...]

PR #911, opened 2024-07-09
fix(Exact): fix type when class present

zorji:
This is an attempt to fix the issue reported in https://github.com/sindresorhus/type-fest/issues/909

I previously attempted to fix the issue in https://github.com/sindresorhus/type-fest/pull/902

My original approach was use JsonObject to determine object with primitive type and assume all types not match are classes. However, I found that JsonObject also rejects normal object containing any class type.

I also can't find a way to reliable determine a class, I used `Class<any>` but it seems can't detect `Date` because `Date` is a built-in type and only interface is available.

# What's changed
- updated the type to leave `Function` check as is
- updated the type to return primitive type early to avoid the generated type becomes too complex.

### DIFF
--- a/source/exact.d.ts
+++ b/source/exact.d.ts
@@ -1,8 +1,8 @@
 import type {ArrayElement, ObjectValue} from './internal';
 import type {IsEqual} from './is-equal';
 import type {KeysOfUnion} from './keys-of-union';
-import type {JsonPrimitive} from './basic';
 import type {IsUnknown} from './is-unknown';
+import type {Primitive} from './primitive';
 
 /**
 Create a type from `ParameterType` and `InputType` and change keys exclusive to `InputType` to `never`.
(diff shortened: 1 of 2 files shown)


source/exact.d.ts
```diff
@@ -52,11 +53,15 @@ onlyAcceptNameImproved(invalidInput); // Compilation error
...
-	IsEqual<ParameterType, InputType> extends true ? ParameterType
-		// Convert union of array to array of union: A[] & B[] => (A & B)[]
-		: ParameterType extends unknown[] ? Array<Exact<ArrayElement<ParameterType>, ArrayElement<InputType>>>
-			// In TypeScript, Array is a subtype of ReadonlyArray, so always test Array before ReadonlyArray.
-			: ParameterType extends readonly unknown[] ? ReadonlyArray<Exact<ArrayElement<ParameterType>, ArrayElement<InputType>>>
-				// Only apply Exact for pure object types. For types from a class, leave it unchanged to TypeScript to handle.
-				: ParameterType extends JsonObject ? ExactObject<ParameterType, InputType>
-					: ParameterType;
+	// If the parameter is a primitive, return it as is immediately to avoid it being converted to a complex type
+	ParameterType extends JsonPrimitive ? ParameterType
```
sindresorhus, line 57: Why not the `Primitive` type?

zorji: My mistake, I'll update it later.

PR #912, opened 2024-07-12
Ensure no implicit `@types/` includes

voxpelli:
To avoid the errors seen in #899 as a result of  #897 this is a companion follow up to the #898 that @Emiyaaaaa made that also removes any implicitly includes `@types/` package.

When no [`types`](https://www.typescriptlang.org/tsconfig/#types) is specified in the `tsconfig.json`, then typescript will pull in all types in `./node_modules/@types/`, no matter if those appeared there due to us including it or because it's been hoisted from one of our dependencies.

Since we specify no `@types/` dependencies all such types are hoisted from our dependencies and seems to currently amount to:

Skärmavbild 2024-07-12 kl  01 56 14

Most importantly this means that we have been pulling in a random `@types/node` and implicitly relied on it (seemingly by relying on `xo` which relies on `eslint-import-resolver-webpack`  ➡️ `webpack` ➡️ `terser-webpack-plugin` ➡️ `jest-worker` ➡️ `@types/node@*`)

This PR sets `"types": []` to ensure no such implicit types are included, which uncovers the issue mentioned in #899 which it solves the same way as the dom specific types were solved.

Related: #908

### DIFF
--- a/source/structured-cloneable.d.ts
+++ b/source/structured-cloneable.d.ts
@@ -17,9 +17,10 @@ type StructuredCloneableData =
 	| Date
 	| Error
 	| RegExp
-	| TypedArray
-	| Blob
-	| File;
+	| TypedArray;
+// Requires DOM or @types/node
+//	| Blob
+//	| File
 // DOM exclusive types
 //	| AudioData
 //	| CropTarget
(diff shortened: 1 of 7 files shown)


sindresorhus:
CI is failing

voxpelli:
@sindresorhus Fixed it now by adding an ignore comment in that specific place

Not sure how to best handle tests of stuff that's only available in some contexts. Similar issue as #908 has.

For now I think ignoring is okay as this PR is mostly to ensure no new mistakes are being made

Issue #676, opened 2023-09-01
`ArrayElement` only importable from 'type-fest/source/internal'


bombillazo:
Hello,

For use, the `ArrayElement` type is super convenient, yet we can only import it from the `type-fest/source/internal` path. In Deno this causes some import issues so I was wondering if the library could export `ArrayElement` directly from the main project?

sindresorhus:
We are happy to expose more of the internal types. Someone needs to do the work write better docs, tests, etc for that type.

https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md#submitting-a-new-type

alex-touza:
I'm currently in need for UnionMax and UnionMin types, but I cannot import it, so I have to copy paste the implementation.

PR #906, opened 2024-07-06
Split `internal.d.ts` into multiple files

tommy-mitchell:
Splits the types in `internal.d.ts` into multiple category-based files to improve maintainability, with the following structure:

```
\_ source/
   \_ internal/
      \_ array.d.ts
      \_ characters.d.ts
      \_ keys.d.ts
      \_ numeric.d.ts
      \_ object.d.ts
      \_ string.d.ts
      \_ tuple.d.ts
      \_ type.d.ts (misc types)
      \_ index.d.ts (so imports don't need updated)
   \_ ...
\_ test-d/
   \_ internal/
      \_ [category]/
         \_ [type].ts
   \_ ...
```

Other miscellaneous changes:

- Updates test script to detect source files in subfolders.
- Removes unused type imports.
- Renames `ArrayMax`/`ArrayMin` to `TupleMax`/`TupleMin`. This can be reverted if desired.

I tried maintaining the order of types and their formatting when copying them over, but I could've missed something.

PR #914, opened 2024-07-14
Expose `ArrayElement` type

tommy-mitchell:
Related: #676. (maybe closes?)

Exposes internal `ArrayElement` type:

> Extracts the element type of an array or union of arrays.

```ts
import type {ArrayElement} from 'type-fest';

declare const getMostCommonElement: <T>(array: T[]) => ArrayElement<typeof array>;

getMostCommonElement([1, 2, 3]);
//=> 1 | 2 | 3

getMostCommonElement(['foo', 'bar', 'baz'] as const);
//=> 'foo' | 'bar' | 'baz'
```

### DIFF
--- a/source/array-element.d.ts
+++ b/source/array-element.d.ts
@@ -22,4 +22,7 @@ getMostCommonElement(['foo', 'bar', 'baz'] as const);
 
 @category Array
 */
-export type ArrayElement<ArrayType> = ArrayType extends readonly unknown[] ? ArrayType[0] : never;
+export type ArrayElement<ArrayType extends readonly unknown[]> =
+	ArrayType extends ReadonlyArray<infer ElementType>
+		? ElementType
+		: never;
--- a/source/array-element.d.ts
+++ b/source/array-element.d.ts
@@ -1,0 +1,25 @@
+/**
+Extracts the element type of an array or union of arrays.
+
+Returns `never` if T is not an array.
+
+It creates a type-safe way to access the element type of `unknown` type.
+
+@example
+```
+import type {ArrayElement} from 'type-fest';
+
+declare const getMostCommonElement: <T>(array: T[]) => ArrayElement<typeof array>;
+
+getMostCommonElement([1, 2, 3]);
+//=> 1 | 2 | 3
+
+getMostCommonElement(['foo', 'bar', 'baz'] as const);
+//=> 'foo' | 'bar' | 'baz'
+```
+
+@see {@link ArrayValues} for when you know that the input is an array.
+
+@category Array
+*/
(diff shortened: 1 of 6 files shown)


source/array-element.d.ts
```diff
@@ -0,0 +1,28 @@
...
+```
+
+@see {@link ArrayValues} for when you know that the input is an array.
+
+@category Array
+*/
+export type ArrayElement<ArrayType extends readonly unknown[]> =
+	ArrayType extends ReadonlyArray<infer ElementType>
+		? ElementType
+		: never;
```
tommy-mitchell, line 28: I'm not sure this works @strongpauly

> Should this be more?:
> 
> ```ts
> type ArrayElement<ArrayType extends readonly unknown[]> = 
>   ArrayType extends readonly (infer ElementType)[] ? ElementType : never;
> ```
> 
> Copy paste from [stackoverflow.com/questions/41253310/typescript-retrieve-element-type-information-from-array-type](https://stackoverflow.com/questions/41253310/typescript-retrieve-element-type-information-from-array-type)

// https://github.com/sindresorhus/type-fest/pull/906#discussion_r1673712673

sindresorhus:
Isn't this type already covered by https://github.com/sindresorhus/type-fest/blob/main/source/iterable-element.d.ts ?

source/array-element.d.ts
strongpauly: Well, as long as it returns a union of the types of the array elements, its fit for purpose, IMO.

tommy-mitchell:
> Isn't this type already covered by [`main`/source/iterable-element.d.ts](https://github.com/sindresorhus/type-fest/blob/main/source/iterable-element.d.ts?rgh-link-date=2024-07-14T10%3A56%3A22Z) ?

Hm, I'd assume so, but the tests for `Exact` fail if swapped to use `IterableElement`:

image

fregante:
Is it really worth to import and use a custom type instead of just using a literal type offered by TypeScript?

``` diff
- array[number]
+ ArrayItem<array>
```

fregante:
In your example it does not look like an advantage, it's hard to follow and does not match the order of the JS code.


```ts
              0             1            2        3 
              complexObject .property_1  [0].     .subprop_1

              0             1            2        3
type MyType = MyComplexType['property_1'][number]['subprop_1'] | null
// vs

              2           0             1              3
type MyType = ArrayElement<MyComplexType['property_1']>['subprop_1'] | null

```

source/array-element.d.ts
giao66: good

PR #919, opened 2024-07-19
`Paths`: Add custom delimiter support

imhoffd:
Useful when `.` isn't the de facto delimiter, e.g. with [PostgreSQL json/jsonb operators](https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-JSON-PROCESSING).

### DIFF
--- a/source/paths.d.ts
+++ b/source/paths.d.ts
@@ -11,6 +11,8 @@ Generate a union of all possible paths to properties in the given object.
 
 It also works with arrays.
 
+The delimiter defaults to `.` but can be customized.
+
 Use-case: You want a type-safe way to access deeply nested properties in an object.
 
 @example
--- a/source/paths.d.ts
+++ b/source/paths.d.ts
@@ -47,9 +47,9 @@ open('listB.1'); // TypeError. Because listB only has one element.
 @category Object
 @category Array
 */
-export type Paths<T> = Paths_<T>;
+export type Paths<T, Delimiter extends string = '.'> = Paths_<T, Delimiter>;
 
-type Paths_<T, Depth extends number = 0> =
+type Paths_<T, Delimiter extends string = '.', Depth extends number = 0> =
 	T extends NonRecursiveType | ReadonlyMap<unknown, unknown> | ReadonlySet<unknown>
 		? never
 		: IsAny<T> extends true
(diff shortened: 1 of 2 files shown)


voxpelli:
Can you explain the use case a bit more?

Also, it would already be possible to emulate by wrapping in a `Replace<value, '.', ':'>` or whatever it is that one would want to replace it with

Even if providing a shortcut, I wonder if it could be better to use `Replace<>` internally then rather than extending all the interfaces here.

imhoffd:
Admittedly I'm new to `type-fest` as of yesterday so I'm still learning the API. `Replace<Paths<Obj>, '.', ',', { all: true }>` appears to be suitable for my use case.

However, unfortunately, and possibly as a separate issue, `Paths<T>` doesn't behave well with generic types, as this simple example shows:

```ts
const fn = <T extends object>(path: Replace<Paths<T>, '.', ':'>) => {
  return path
}
```

```
Type 'Paths<T>' does not satisfy the constraint 'string'.
  Type 'number | `${number}`' is not assignable to type 'string'.
    Type 'number' is not assignable to type 'string'. (tsserver 2344)
```

---

> Can you explain the use case a bit more?

Sure, with [PostgreSQL json/jsonb operators](https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-JSON-PROCESSING) the delimiter is a comma, not a dot. I wrote this function for Drizzle using the `Paths` type from this PR. It provides a type-safe way of specifying the path to a field within a jsonb column with commas as the delimiter. Obviously I could use a dot delimiter and use `.replaceAll()`, but I wanted to be more idiomatic towards SQL, not JavaScript. I could also use an array (typed as tuples of string literals) to separate path parts instead (not sure if that utility type exists?).

```ts
/**
 * Equivalent to the `#>>` operator.
 *
 * @see https://www.postgresql.org/docs/current/functions-json.html#FUNCTIONS-JSON-PROCESSING
 */
const jsonbGetText = <T extends ColumnBaseConfig<'json', 'PgJsonb'>>(
  col: PgColumn<T>,
  path: Paths<T['data'], ','>,
): SQL<string> => {
  return sql`${col}#>>'{${sql.raw(path)}}'`
}
```

sindresorhus:
I would prefer to find a way to make using `Replace` work with `Paths`. Every option we add to this type adds implementation complexity and slows down type checking.

imhoffd:
Agreed.

Also, I was able to get around the compile error by using `ToString`:

```ts
const jsonbGetText = <T extends ColumnBaseConfig<'json', 'PgJsonb'>>(
  col: PgColumn<T>,
  path: Replace<ToString<Paths<T['data']>>, '.', ',', { all: true }>,
): SQL<string> => {
  return sql`${col}#>>'{${sql.raw(String(path))}}'`
}
```

PR #920, opened 2024-07-19
Add options to `Paths`

### LINKED ISSUE #917: Expose Max Recursion Depth of Paths
I have a custom type that makes use of `Paths`. Whenever I use this type with a sufficiently complex input `T` I do get an error `Type instantiation is excessively deep and possibly infinite.ts(2589)`.

[The recursion limit of Paths is currently locked at 15.](https://github.com/Emiyaaaaa/type-fest/blob/60712f1d667a88dc2e3ae846281dd852f1303ac0/source/paths.d.ts#L78) When I slightly lower the amount of allowed recursions, e.g. from 15 to 10, my error goes away. 

Is there a reason why the `Depth` parameter is hidden to the outside world? I would like it to be configurable to handle those errors. What do you think?

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/917">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/917/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

fardolieri:
Fixes #917

Introduce `PathsOptions` with `maxRecursionDepth`.

## Things up for debate
- By introducing the new options parameter, which adds a bit more complexity, the previous test-case
https://github.com/sindresorhus/type-fest/blob/8a45ba048767aaffcebc7d190172d814a739feb0/test-d/paths.ts#L109
worked fine with the old recursion limit of 15 but now throws an error _"Type instantiation is excessively deep and possibly infinite"_. Setting the limit to 14 fixes it. Technically, this introduces a small breaking change.
In my experience, most paths that are more than 4 levels deep do not really add value most of the time. So I would suggest having a smaller default like 8 or 10. This would speed up type checking times for users and slightly reduce the bloated wall of intellisense suggestions. If users need more, they can set `maxRecursionDepth` manually.
- I'm not sure why `InternalPaths` is currently exported. Perhaps so that users have a way to get their hands on the `Depth` parameter. If that is the only reason, we could consider not exporting it now that `Paths` supports it.

### DIFF
--- a/source/paths.d.ts
+++ b/source/paths.d.ts
@@ -13,7 +13,7 @@ Paths options.
 */
 export type PathsOptions = {
 	/**
-	Maximum recursions before stopping to search deeper paths.
+	The maximum depth to recurse when searching for paths.
 
 	@default 10
 	*/
--- a/source/paths.d.ts
+++ b/source/paths.d.ts
@@ -3,8 +3,22 @@ import type {EmptyObject} from './empty-object';
 import type {IsAny} from './is-any';
 import type {IsNever} from './is-never';
 import type {UnknownArray} from './unknown-array';
-import type {Sum} from './sum';
-import type {LessThan} from './less-than';
+import type {Subtract} from './subtract';
+import type {GreaterThan} from './greater-than';
+
+/**
+Paths options.
+
+@see {@link Paths}
+*/
+export type PathsOptions = {
+	/**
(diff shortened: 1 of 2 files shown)


voxpelli:
Related: #919

voxpelli:
Might have caused a regression: https://github.com/sindresorhus/type-fest/issues/922

PR #926, opened 2024-08-01
`Path`: add `bracketNotation` option

### LINKED ISSUE #921: Issues with `Paths` type
I'm not sure why this is, but the paths produced by `Paths` ends up being slightly different from `Get`. Specifically, it would be nice if `Paths` produced a string array tuple type like:
```
[
  'arr[0].field'
  'arr.0.field' // both valid
  // ... other paths
]
```
... or with a config option, just:
```
[
  'arr[0].field'
  // ... other paths
]
```
Instead, it produces only the `{dot}{number}{dot}` form, like:
```
[
  'arr.0.field'
  // ... other paths
]
```

The other problem with `Paths` is that it can't handle a type extended from `Record`. For example, imagine this very simple use case:
```ts
import get from 'lodash/get'
import type { Paths, Get } from 'type-fest'

export function getter<const T extends Record<string, any>, P extends string = Paths<T>>(
  obj: T,
  path: P
): Get<T, P> {
  return get(obj, path)
}
```
This produces the TypeScript error:
```
Type 'Paths<T>' does not satisfy the constraint 'string'.
  Type 'number | `${number}`' is not assignable to type 'string'.
    Type 'number' is not assignable to type 'string'.ts(2344)
```
Even though we've specified that the record will only have string keys, `Paths` produces something other than a union of strings.

It looks like a workaround is:
```ts
export function getter<const T extends Record<string, any>, P extends string = Exclude<Paths<T>, number>>(
  obj: T,
  path: P
): Get<T, P> {
  return get(obj, path)
}
```
... but I'm not sure why this should be the case

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/921">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/921/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

Emiyaaaaa:
closed #921 

we can change default style on V5 if need.

### DIFF
--- a/source/internal/numeric.d.ts
+++ b/source/internal/numeric.d.ts
@@ -20,7 +20,7 @@ NumberAbsolute<NegativeInfinity>
 export type NumberAbsolute<N extends number> = `${N}` extends `-${infer StringPositiveN}` ? StringToNumber<StringPositiveN> : N;
 
 /**
-Return a type if it is a number or a number string.
+Check whether the given type is a number or a number string.
 
 Supports floating-point as a string.
 
--- a/source/internal/numeric.d.ts
+++ b/source/internal/numeric.d.ts
@@ -20,6 +20,25 @@ NumberAbsolute<NegativeInfinity>
 export type NumberAbsolute<N extends number> = `${N}` extends `-${infer StringPositiveN}` ? StringToNumber<StringPositiveN> : N;
 
 /**
+Return a type if it is a number or a number string
+
+@example
+```
+type A = IsNumberLike<'1'>;
+//=> true
+
+type B = IsNumberLike<1>;
+//=> true
+
+type C = IsNumberLike<'a'>;
+//=> false
+*/
+export type IsNumberLike<N> =
+	N extends number ? true
+		:	N extends `${number}` ? true
+			: false;
+
+/**
 Returns the minimum number in the given union of numbers.
 
 Note: Just supports numbers from 0 to 999.
--- a/source/internal/numeric.d.ts
+++ b/source/internal/numeric.d.ts
@@ -22,23 +22,29 @@ export type NumberAbsolute<N extends number> = `${N}` extends `-${infer StringPo
 /**
 Return a type if it is a number or a number string.
 
+supports floating-point as a string.
+
 @example
 ```
 type A = IsNumberLike<'1'>;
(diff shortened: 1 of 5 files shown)


source/paths.d.ts
```diff
@@ -18,6 +18,18 @@ export type PathsOptions = {
 	@default 10
 	*/
 	maxRecursionDepth?: number;
+
+	/**
+	Style of the path.
+
+  @default 'a.0.b'
+	*/
+	style?: 'a[0].b' | 'a.0.b';
```
Emiyaaaaa, line 27: There should be a better naming, but I can't think of one right now

Emiyaaaaa:
I notice that Get can accept a 'mixed' path like 0[0].0 in addition to these two styles. so we should add a new style more, or just make Paths output all style?

source/internal/numeric.d.ts
```diff
@@ -19,6 +19,25 @@ NumberAbsolute<NegativeInfinity>
...
+type A = IsNumberLike<'1'>;
+//=> true
+
+type B = IsNumberLike<1>;
+//=> true
+
+type C = IsNumberLike<'a'>;
+//=> false
+*/
+export type IsNumberLike<N> =
```
sindresorhus, line 36: Does this support floating-point as a string? If it does, it should be documented, if not, it should be named `IsNumberOrIntegerString`.

It could also use a couple of tests.

source/paths.d.ts
sindresorhus, line 27:
```suggestion
	/**
	Use bracket notation for array indices.

	@default false

	@example
	```
	type Example = {
		array: string[];
	};

	type A = Paths<Example, {bracketNotation: false}>;
	//=> 'array' | 'array.0'

	type B = Paths<Example, {bracketNotation: true}>;
	//=> 'array' | 'array[0]'
	```
	*/
	bracketNotation?: boolean;
```

Emiyaaaaa: Should we need to leave some buffer to expand this option for the future? like "a[0]['b']" style?

sindresorhus: No

Issue #223, opened 2021-06-26
SnakeCaseProperties doesn't handle PascalCase and UPPERCASE


chriszirkel:
I have a type with PascalCase and a few uppercase properties, e.g.

```
Identifiers: {
    MarketplaceASIN: {
	MarketplaceId: string;
	ASIN: string;
    },
}
```

SnakeCasedPropertiesDeep turns this into:

```
_identifiers: {
    _marketplace_a_s_i_n: {
        _marketplace_id: string;
        _a_s_i_n: string;
    };
};
```

It doesn't handle the cases where the property starts with an uppercase and where the property contains multiple uppercase letters in a row.

I could overcome the first issue by using CamelCasedPropertiesDeep first. 

What I would expect is:
```
identifiers: {
    marketplace_asin: {
        marketplace_id: string;
        asin: string;
    };
};
```

During runtime I use lodash's https://lodash.com/docs/4.17.15#snakeCase method. Is there a way to use the same method for the type?

threehams:
This looks like it might be fairly simple to fix. I've hacked in a userland fix with this:

```
type IsScreamingSnakeCase<TString extends string> = Split<TString, ""> extends (
  | UpperCaseCharacters
  | WordSeparators
  | StringDigit
)[]
  ? true
  : false

type SnakeCase = <TString extends string>(
  str: TString,
) => IsScreamingSnakeCase<TString> extends true
  ? SnakeCase<Lowercase<TString>>
  : SnakeCase<TString>;
```

I'll submit a PR when I have time, hopefully it's as simple as it seems.

voxpelli:
Bets solution would be the one described here I think: https://github.com/sindresorhus/type-fest/issues/224#issuecomment-925357006

PR #930, opened 2024-08-07
Improve `DelimiterCase`

filipw01:
I made some changes that pass more test cases than the current solution. I was heavily inspired by `CamelCase` implementation

Let me know if this is something worth investing more time into or if I'm doing some basic mistake and butchering performance for example

Also, what should be the behavior of `string1` conversion 
* `string1` - this is the current behavior
* `string_1` - another valid behavior
* either one customizable with an option object like `camelCase` - in that case there's still a question of what should be the default (most likely `string1` for backwards compatibility)

Related issues:
https://github.com/sindresorhus/type-fest/issues/223
https://github.com/sindresorhus/type-fest/issues/336

### DIFF
--- a/source/delimiter-case.d.ts
+++ b/source/delimiter-case.d.ts
@@ -1,54 +1,21 @@
-import type {UpperCaseCharacters, WordSeparators} from './internal';
+import {SplitWords} from "./split-words";
 
-// Transforms a string that is fully uppercase into a fully lowercase version. Needed to add support for SCREAMING_SNAKE_CASE, see https://github.com/sindresorhus/type-fest/issues/385
-type UpperCaseToLowerCase<T extends string> = T extends Uppercase<T> ? Lowercase<T> : T;
-
-// This implementation does not support SCREAMING_SNAKE_CASE, it is used internally by `SplitIncludingDelimiters`.
-type SplitIncludingDelimiters_<Source extends string, Delimiter extends string> =
-	Source extends '' ? [] :
-		Source extends `${infer FirstPart}${Delimiter}${infer SecondPart}` ?
-			(
-				Source extends `${FirstPart}${infer UsedDelimiter}${SecondPart}`
-					? UsedDelimiter extends Delimiter
-						? Source extends `${infer FirstPart}${UsedDelimiter}${infer SecondPart}`
-							? [...SplitIncludingDelimiters<FirstPart, Delimiter>, UsedDelimiter, ...SplitIncludingDelimiters<SecondPart, Delimiter>]
-							: never
-						: never
-					: never
-			) :
-			[Source];
 
 /**
-Unlike a simpler split, this one includes the delimiter splitted on in the resulting array literal. This is to enable splitting on, for example, upper-case characters.
-
-@category Template literal
+Convert an array of words to delimiter case starting with a delimiter with input capitalization.
 */
-export type SplitIncludingDelimiters<Source extends string, Delimiter extends string> = SplitIncludingDelimiters_<UpperCaseToLowerCase<Source>, Delimiter>;
+type DelimiterCaseFromArray<
+	Words extends string[],
+	Delimiter extends string,
+	OutputString extends string = ''
+> = Words extends [
+		infer FirstWord extends string,
+		...infer RemainingWords extends string[]
+	]
+	? `${Delimiter}${FirstWord}${DelimiterCaseFromArray<RemainingWords, Delimiter>}`
+	: OutputString;
 
-/**
-Format a specific part of the splitted string literal that `StringArrayToDelimiterCase<>` fuses together, ensuring desired casing.
-
-@see StringArrayToDelimiterCase
-*/
-type StringPartToDelimiterCase<StringPart extends string, Start extends boolean, UsedWordSeparators extends string, UsedUpperCaseCharacters extends string, Delimiter extends string> =
-	StringPart extends UsedWordSeparators ? Delimiter :
-		Start extends true ? Lowercase<StringPart> :
-			StringPart extends UsedUpperCaseCharacters ? `${Delimiter}${Lowercase<StringPart>}` :
-				StringPart;
-
-/**
-Takes the result of a splitted string literal and recursively concatenates it together into the desired casing.
-
-It receives `UsedWordSeparators` and `UsedUpperCaseCharacters` as input to ensure it's fully encapsulated.
-
-@see SplitIncludingDelimiters
-*/
-type StringArrayToDelimiterCase<Parts extends readonly any[], Start extends boolean, UsedWordSeparators extends string, UsedUpperCaseCharacters extends string, Delimiter extends string> =
-	Parts extends [`${infer FirstPart}`, ...infer RemainingParts]
-		? `${StringPartToDelimiterCase<FirstPart, Start, UsedWordSeparators, UsedUpperCaseCharacters, Delimiter>}${StringArrayToDelimiterCase<RemainingParts, false, UsedWordSeparators, UsedUpperCaseCharacters, Delimiter>}`
-		: Parts extends [string]
-			? string
-			: '';
+type RemoveFirstLetter<S extends string> = S extends `${infer _}${infer Rest}` ? Rest : '';
 
 /**
 Convert a string literal to a custom string delimiter casing.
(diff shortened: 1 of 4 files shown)


filipw01:
@voxpelli Could you take a look? At the moment I just want to know if this is something I should dig deeper into

voxpelli:
@filipw01 Its not immediately obvious to me what the improvements are, and my experience with these helpers has been that they easily turn into a game of whackamole – any improvement reveals a couple of regressions.

I do see that you ultimate refer to my wish to unify the two: https://github.com/sindresorhus/type-fest/issues/224#issuecomment-925357006

> To support this use case I think we essentially would have to merge the functionality of the two. I can't remember why I didn't do that initially.

But that was ultimately solved through a new option instead: https://github.com/sindresorhus/type-fest/pull/501

I guess my main feedback would be to split this PR into more commits and try to make it as clear what and why the changes made are made, so that one can as easily as possible follow the reasoning in a review

filipw01:
For the no split on number I decided to align with change-case which seems to be the best at handling different edge cases

source/split-words.d.ts
```diff
@@ -36,22 +39,30 @@ export type SplitWords<
...
 			// Fist char of word
-			? SplitWords<RemainingCharacters, FirstCharacter, FirstCharacter>
-			// Case change: non-numeric to numeric, push word
+			? SplitWords<RemainingCharacters, Options, FirstCharacter, FirstCharacter>
+			// Case change: non-numeric to numeric
 			: [false, true] extends [IsNumeric<LastCharacter>, IsNumeric<FirstCharacter>]
-				? [...SkipEmptyWord<CurrentWord>, ...SplitWords<RemainingCharacters, FirstCharacter, FirstCharacter>]
-				// Case change: numeric to non-numeric, push word
+				// Split on number: push word
+				? Options['splitOnNumber'] extends true
```
filipw01, line 46: When there's a numeric/non-numeric case change I added and used a new option `splitOnNumber` which depending on how it is set will either
1. When set to true - Like before split words on number
2. When set to false - concat to current word

source/screaming-snake-case.d.ts
```diff
@@ -21,13 +11,13 @@ This can be useful when, for example, converting a camel-cased object property t
...
 const someVariable: ScreamingSnakeCase<'fooBar'> = 'FOO_BAR';
+const someVariableNoSplitOnNumber: ScreamingSnakeCase<'p2pNetwork', {splitOnNumber: false}> = 'P2P_NETWORK';
+
 ```
 
 @category Change case
 @category Template literal
-*/
-export type ScreamingSnakeCase<Value> = Value extends string
-	? IsScreamingSnakeCase<Value> extends true
```
filipw01, line 30: I can bring this back, but I believe it shouldn't be necessary, especially that we don't use it for other things like snake case or kebab case

test-d/kebab-case.ts
```diff
@@ -15,3 +18,60 @@ expectType<'foo-bar'>(kebabFromSnake);
...
+expectType<'foo2bar'>(kebabFromNumberInTheMiddleNoSplitOnNumber);
+
+const kebabFromNumberInTheMiddleNoSplitOnNumberEdgeCase: KebabCase<'foo2Bar', {splitOnNumber: false}> = 'foo2-bar';
+expectType<'foo2-bar'>(kebabFromNumberInTheMiddleNoSplitOnNumberEdgeCase);
+
+const kebabFromNumberInTheMiddleNoSplitOnNumberEdgeCase2: KebabCase<'foO2bar', {splitOnNumber: false}> = 'fo-o2bar';
+expectType<'fo-o2bar'>(kebabFromNumberInTheMiddleNoSplitOnNumberEdgeCase2);
+
+const kebabFromNumberInTheMiddleNoSplitOnNumberEdgeCase3: KebabCase<'FOO22Bar', {splitOnNumber: false}> = 'foo22-bar';
+expectType<'foo22-bar'>(kebabFromNumberInTheMiddleNoSplitOnNumberEdgeCase3);
```
filipw01, line 77: I added the same tests to all case types as well as some edge cases visible here. If you have any doubts about the implementation I can add more tests

source/kebab-case.d.ts
```diff
@@ -12,6 +13,7 @@ import type {KebabCase} from 'type-fest';
 // Simple
 
 const someVariable: KebabCase<'fooBar'> = 'foo-bar';
+const someVariableNoSplitOnNumber: KebabCase<'p2pNetwork', {splitOnNumber: false}> = 'p2p-network';
```
filipw01, line 16: I added this new option on case changing types

source/delimiter-case.d.ts
```diff
@@ -1,54 +1,20 @@
...
-
-@see SplitIncludingDelimiters
-*/
-type StringArrayToDelimiterCase<Parts extends readonly any[], Start extends boolean, UsedWordSeparators extends string, UsedUpperCaseCharacters extends string, Delimiter extends string> =
-	Parts extends [`${infer FirstPart}`, ...infer RemainingParts]
-		? `${StringPartToDelimiterCase<FirstPart, Start, UsedWordSeparators, UsedUpperCaseCharacters, Delimiter>}${StringArrayToDelimiterCase<RemainingParts, false, UsedWordSeparators, UsedUpperCaseCharacters, Delimiter>}`
-		: Parts extends [string]
-			? string
-			: '';
+type DelimiterCaseFromArray<
```
filipw01, line 6: This gets the words split and inserts the delimited at the beginning and after every word

Like `['here', 'We', 'Go']` -> `'#here#We#Go'`

source/delimiter-case.d.ts
```diff
@@ -87,13 +54,9 @@ const rawCliOptions: OddlyCasedProperties<SomeOptions> = {
...
-export type DelimiterCase<Value, Delimiter extends string> = string extends Value ? Value : Value extends string
-	? StringArrayToDelimiterCase<
-	SplitIncludingDelimiters<Value, WordSeparators | UpperCaseCharacters>,
-	true,
-	WordSeparators,
-	UpperCaseCharacters,
-	Delimiter
-	>
+ */
+export type DelimiterCase<Value, Delimiter extends string, Options extends SplitWordsOptions = {splitOnNumber: true}> = Value extends string
```
filipw01, line 58: Here we
1. `'hereWeGo'` -> SplitWords -> `['here', 'We', 'Go']`
2. `['here', 'We', 'Go']` -> DelimiterCaseFromArray ->  `'#here#We#Go'`
3. `'#here#We#Go'` -> RemoveFirstLetter ->  `'here#We#Go'`
4. `'here#We#Go'` -> Lowercase -> `'here#we#go'`

filipw01:
@voxpelli At this point where all the commits are made I'd prefer not to do any git wizardry splitting into smaller commits. I left some comments for easier review. I hope this will be enough. If something is unclear let me know. 

I tried to add more edge cases to avoid the whack a mole game

voxpelli:
#975 is also touching some of these files. I will not have time to review either, just noting it

filipw01:
@sindresorhus any chance to moce this forward?

source/delimiter-case.d.ts
sindresorhus, line 58: `SplitWordsOptions = {splitOnNumber: true}` will not allow us to add more options in the future, as if you specify one, you override the existing defaults.

Example: https://github.com/sindresorhus/type-fest/blob/7dfb307e3ecc0e7717de6880395f1de8715c82c1/source/paths.d.ts

sindresorhus (changes requested):

sindresorhus:
You need to resolve the merge conflict.

sindresorhus:
And maybe `splitOnNumber` => `splitOnNumbers`?

source/delimiter-case.d.ts
```diff
@@ -87,13 +54,9 @@ const rawCliOptions: OddlyCasedProperties<SomeOptions> = {
...
+	? string extends Value
```
som-sm, line 59: Replace the literal check with the `IsStringLiteral` type.

Right now, infinite string types (other than `string`) don't work properly, so it's best to skip them.
```ts
type T = DelimiterCase<`on${string}`, '#'>;
//   ^? type T = "on"
```

som-sm (changes requested):

source/delimiter-case.d.ts
filipw01: Added

source/delimiter-case.d.ts
```diff
@@ -1,54 +1,20 @@
...
-			: '';
+type DelimiterCaseFromArray<
+	Words extends string[],
+	Delimiter extends string,
+	OutputString extends string = '',
+> = Words extends [
+	infer FirstWord extends string,
+	...infer RemainingWords extends string[],
+]
+	? `${Delimiter}${FirstWord}${DelimiterCaseFromArray<RemainingWords, Delimiter>}`
```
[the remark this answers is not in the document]
filipw01: Nice, catch. Changed

source/delimiter-case.d.ts
filipw01: Added a test case and fixed it

filipw01:
Looks good 😉

source/words.d.ts
```diff
@@ -1,11 +1,46 @@
...
+
+	type Example2 = Words<'p2pNetwork', {splitOnNumbers: false}>;
+	//=> ["p2p", "Network"]
+	```
+	*/
+	splitOnNumbers?: boolean;
+};
+
+type DefaultOptions = {
+	splitOnNumbers: true;
```
dasa, line 41: Should this default to `false`? 4.36.0 seems to be a breaking change since I have to set this to false to avoid errors using `KebabCase`.

PR #932, opened 2024-08-09
`Get`: update parameter `Path` generic

### LINKED ISSUE #863: Unify `Get` and `Paths` / `PickDeep` etc
Its a bit odd that we have [`Get`](https://github.com/sindresorhus/type-fest/blob/main/source/get.d.ts) which accepts three styles of specifying paths, but doesn't use [`Paths`](https://github.com/sindresorhus/type-fest/blob/main/source/paths.d.ts), and also have [`PickDeep`](https://github.com/sindresorhus/type-fest/blob/main/source/pick-deep.d.ts) etc which use `Paths` and only accepts a single style of specifying paths.

* `Paths` is used to validate that the specified paths are valid options + used to give intellisense, both which are handy.
* `Paths` only supports the `foo.0.bar` style of paths whereas `Get` also supports `foo[0].bar` and `['foo', '0', 'bar']`, is that maybe overkill for `Get`? Or, if useful, should `Paths` etc be extended with the same functionality?

### Desired outcome:

Ideally `Get` and `Paths` / `PickDeep` etc should behave the same when it comes to paths and use the same fundamentals.

### Random notes

`PickDeep` and `Get` feels fairly similar in other ways as well – maybe one could adapt the internals of `PickDeep` to also serve `Get`?

I gave it a look and a try but got a headache and needs to bang my head against something right now.

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/863">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/863/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

Emiyaaaaa:
closed #863

### DIFF
--- a/source/get.d.ts
+++ b/source/get.d.ts
@@ -1,5 +1,5 @@
 import type {StringDigit, ToString} from './internal';
-import type {LiteralUnion} from './literal-union';
+import type {LiteralStringUnion} from './literal-union';
 import type {Paths} from './paths';
 import type {Split} from './split';
 import type {StringKeyOf} from './string-key-of';
@@ -193,8 +193,6 @@ export type Get<
 	BaseType,
 	Path extends
 	| readonly string[]
-	| LiteralUnion<ToString<
-	Paths<BaseType, {bracketNotation: false}> | Paths<BaseType, {bracketNotation: true}>>
-	, string>,
+	| LiteralStringUnion<ToString<Paths<BaseType, {bracketNotation: false}> | Paths<BaseType, {bracketNotation: true}>>>,
 	Options extends GetOptions = {}> =
 		GetWithPath<BaseType, Path extends string ? ToPath<Path> : Path, Options>;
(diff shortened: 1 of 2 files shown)


source/get.d.ts
```diff
@@ -187,5 +189,12 @@ Get<Record<string, string>, 'foo', {strict: true}> // => string
...
 */
-export type Get<BaseType, Path extends string | readonly string[], Options extends GetOptions = {}> =
-	GetWithPath<BaseType, Path extends string ? ToPath<Path> : Path, Options>;
+export type Get<
+	BaseType,
+	Path extends
+	| readonly string[]
+	| LiteralUnion<ToString<
+	Paths<BaseType, {bracketNotation: false}> | Paths<BaseType, {bracketNotation: true}>>
+	, string>,
```
sindresorhus, line 198: Kinda weird code formatting here

sindresorhus:
Does this add any new capabilities or changes? If so, some new tests for would good.

Emiyaaaaa:
> Does this add any new capabilities or changes? If so, some new tests for would good.

no, just a enhancement of autocompletion in IDE

PR #935, opened 2024-08-11
Add `Arrayable` type #270

### LINKED ISSUE #270: Like `Promisable<X>` but for arrays: `X | X[]`
I remember seeing this type somewhere, but I can't find it. Does it exist? Is it native?

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/270">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/270/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

Max10240:
fix #270

### DIFF
--- a/source/arrayable.d.ts
+++ b/source/arrayable.d.ts
@@ -1,0 +1,25 @@
+/**
+Create a type that represents either the value or an array of the value.
+
+@see Promisable
+
+@example
+```
+import type {Arrayable} from 'type-fest';
+
+function bundle(input: string, output: Arrayable<string>) {
+	const outputList = Array.isArray(output) ? output : [output];
+	// ...
+
+	for (const output of outputList) {
+	  console.log(`write to: ${output}`);
+	}
+}
+
+bundle('src/index.js', 'dist/index.js');
+bundle('src/index.js', ['dist/index.cjs', 'dist/index.mjs']);
+```
+
+@category Array
+*/
+export type Arrayable<T> = T | readonly T[];
--- a/source/arrayable.d.ts
+++ b/source/arrayable.d.ts
@@ -9,7 +9,8 @@ import type {Arrayable} from 'type-fest';
 
 function bundle(input: string, output: Arrayable<string>) {
 	const outputList = Array.isArray(output) ? output : [output];
-	// ...
+
+	// …
 
 	for (const output of outputList) {
 	  console.log(`write to: ${output}`);
--- a/index.d.ts
+++ b/index.d.ts
@@ -36,6 +36,7 @@ export type {UndefinedOnPartialDeep} from './source/undefined-on-partial-deep';
 export type {ReadonlyDeep} from './source/readonly-deep';
 export type {LiteralUnion} from './source/literal-union';
 export type {Promisable} from './source/promisable';
+export type {Arrayable} from './source/arrayable';
 export type {Opaque, UnwrapOpaque, Tagged, GetTagMetadata, UnwrapTagged} from './source/opaque';
 export type {InvariantOf} from './source/invariant-of';
 export type {SetOptional} from './source/set-optional';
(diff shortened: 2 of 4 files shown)


source/arrayable.d.ts
```diff
@@ -0,0 +1,25 @@
+/**
+Create a type that represents either the value or array of the value.
```
sindresorhus, line 2:
```suggestion
Create a type that represents either the value or an array of the value.
```

source/arrayable.d.ts
```diff
@@ -0,0 +1,25 @@
...
+
+@see Promisable
+
+@example
+```
+import type {Arrayable} from 'type-fest';
+
+function bundle(input: string, output: Arrayable<string>) {
+  const outputList = Array.isArray(output) ? output : [output];
```
sindresorhus, line 11: Tab-indentation

source/arrayable.d.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+}
+
+bundle('src/index.js', 'dist/index.js');
+bundle('src/index.js', ['dist/index.cjs', 'dist/index.mjs']);
+```
+
+@category Array
+*/
+export type Arrayable<T> = T | T[];
```
sindresorhus, line 25:
```suggestion
export type Arrayable<T> = T | readonly T[];
```

sindresorhus (changes requested):

Emiyaaaaa (approved):

fregante:
@sindresorhus why was `readonly []` chosen here? I'm unable to create a `castArray` function that is compatible with this.

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbwIJSgQwJ5oEYBsCmAvnAGZQQhwDkMGY+AtCfgM4xUDcAUFyQK4A7AMYxgEAXCFo2KdBgCMAHgCSMfCAB8ACgBuaXH3wAuOLMw4CKtZoCUJ1eoDaAXURcAkFHww+UCWYwAOmAWAN19Qxs4AH44PQN8OBNHeMNnbkIefmFRcUlpGACAJit1bVTjU1RzPHxS2xMvNAATcVwMOAcQFzdPb19-aqCQsIqo2IqkuBSI-HSuTKA

Max10240:
@fregante  `Item` cannot be an array type, because `Arrayable<unknown[]>` is equivalent to `unknown[] | readonly unknown[][]`, NOT `unknown[]` or `readonly unknown[]`. If `Item` is an Array type, `Array.isArray` will always return `true`, meaning that the actual return type of `castArray1 / castArray2` is actually `Item` itself, not `Item[]`.
I'm guessing what you're looking for is something like:
```tsx
import { Arrayable } from 'type-fest';

type Primitive =
	| bigint
	| boolean
	| null
	| number
	| string
	| symbol
	| undefined;

function castArray<Item extends Primitive | Record<PropertyKey, unknown>>(value: Arrayable<Item>): Item[] {
	return Array.isArray(value) ? value : [value];
}

```

fregante:
My previous code worked correctly if `readonly` is dropped, so that lengthy type information is a consequence of this keyword. 

You can see it working here: https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbwIJSgQwJ5oEYBsCmAvnAGZQQhwDkMGY+AtCfgM4xUDcAUF7fXAAVcaYADsU6LHnwAeABoA+OAF44cuAB81AbQC63HiQCuogMYxgEUXFNo2EzAEYZASRj4QCgBQA3NLiN8AC5BYTEHKQJXd08AShC3Dz1ELgBIKHwYIyhrCIA6YBYI339A2LgAfjg-APw4EO0awP0uQkMTc0trW3tUTAAmaI9vJuC4CJwoxLiEmOSENIysnPG+jAKitZLa8qrR+rhG0vwWtqA

Issue #796, opened 2024-01-16
SetRequiredDeep - is it possible?


Hatgor:
Hello there! 

Could anyone please advise if I can combine existing methods to achieve the following: 

`type Example = {
  a: string;
  b?: number;
  c: {
    ca: number;
    cb?: string;
  };
};`

`type ExampleWithRequired = SetRequiredDeep<Example, 'b' | 'c.cb'>;`

I've tried to slightly modify SetRequired on my own, but it didn't work as expected: 

<img width="691" alt="image" src="https://github.com/sindresorhus/type-fest/assets/30573650/7cca4caa-bdef-4701-a859-a2b6119947a8">
  
<img width="452" alt="image" src="https://github.com/sindresorhus/type-fest/assets/30573650/3041e241-52d9-42c6-a724-ee6153db2f48">

Is it possible at all? 

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/sindresorhus) so you can upvote and help fund this issue.
- The funding will be given to active contributors.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/sindresorhus/type-fest/issues/796">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/796/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/sindresorhus/type-fest/issues/796/pledge.svg">
</picture>
</a>

Emiyaaaaa:
I think this can be a good feature.

PR #939, opened 2024-08-12
Add `SetRequiredDeep` type

hugomartinet:
Closes https://github.com/sindresorhus/type-fest/issues/796

This pull request adds a new `SetRequiredDeep` which, like `SetRequired`, allows to make one or several keys in a model required. It adds the possibility to select nested keys by specifying their path.

**Example**

```typescript
import type {SetRequiredDeep} from 'type-fest';

type Foo = {
	a?: number;
	b?: string;
	c?: {
		d?: number
	}[]
}

type SomeRequiredDeep = SetRequiredDeep<Foo, 'a' | `c.${number}.d`>;
// type SomeRequiredDeep = {
// 	a: number; // Is now required
// 	b?: string;
// 	c: {
//    d: number // Is now required
//  }[]
// }
```


source/set-required-deep.d.ts
```diff
@@ -0,0 +1,49 @@
...
+@example
+```
+import type {SetRequiredDeep} from 'type-fest';
+
+type Foo = {
+	a?: number;
+	b?: string;
+	c?: {
+    ca?: number
+    cb?: boolean
```
sindresorhus, line 21: Tab-indentation

Emiyaaaaa:
Can this feature be added on `SetRequired` instead of a new type?

sindresorhus:
The problem with doing both shallow and deep in a single method is that the deep logic affects the performance even if you don't need deep. This was the case for `Simplify`, where we had to split it out into a `SimplifyDeep` version.

source/set-required-deep.d.ts
```diff
@@ -0,0 +1,47 @@
...
+
+Use-case: You want to define as required only a some nested keys inside a model.
+
+@example
+```
+import type {SetRequiredDeep} from 'type-fest';
+
+type Foo = {
+	a?: number;
+	b?: string;
```
Emiyaaaaa, line 18: add a array example is better.
```suggestion
     	b?: string;
	array?: {
	  a?: number;
	}[];
```

test-d/set-required-deep.ts
```diff
@@ -0,0 +1,18 @@
...
+declare const variation2: SetRequiredDeep<{a?: number; b?: {c?: string}}, 'a'>;
+expectType<{a: number; b?: {c?: string}}>(variation2);
+
+// Keeps required key as required
+declare const variation3: SetRequiredDeep<{a: number; b?: {c?: string}}, 'a'>;
+expectType<{a: number; b?: {c?: string}}>(variation3);
+
+// Update optional keys to required in a union.
+declare const variation4: SetRequiredDeep<{a?: '1'; b?: {c?: boolean}} | {a?: '2'; b?: {c?: boolean}}, 'a' | 'b.c'>;
+expectType<{a: '1'; b: {c: boolean}} | {a: '2'; b: {c: boolean}}>(variation4);
```
Emiyaaaaa, line 18: need add some tests about array

Emiyaaaaa (changes requested):

test-d/set-required-deep.ts
```diff
@@ -0,0 +1,22 @@
+import {expectType} from 'tsd';
+import type {SetRequiredDeep} from '../index';
+
+// Update an optional nested key to required.
+declare const variation1: SetRequiredDeep<{a?: number; b?: {c?: string}}, 'b.c'>;
+expectType<{a?: number; b: {c: string}}>(variation1);
+
```
Emiyaaaaa, line 7:
```
declare const variation1: SetRequiredDeep<{a?: number; b?: {c?: string}}, 'b'>;
```

test-d/set-required-deep.ts
```diff
@@ -0,0 +1,22 @@
...
+expectType<{a: number; b?: {c?: string}}>(variation2);
+
+// Keeps required key as required
+declare const variation3: SetRequiredDeep<{a: number; b?: {c?: string}}, 'a'>;
+expectType<{a: number; b?: {c?: string}}>(variation3);
+
+// Update optional keys to required in a union.
+declare const variation4: SetRequiredDeep<{a?: '1'; b?: {c?: boolean}} | {a?: '2'; b?: {c?: boolean}}, 'a' | 'b.c'>;
+expectType<{a: '1'; b: {c: boolean}} | {a: '2'; b: {c: boolean}}>(variation4);
+
```
Emiyaaaaa, line 19: add tests:
```
SetRequiredDeep<{a?: number; array?: Array<{b?: number}>}, 'array.${number}.b'>;

SetRequiredDeep<{a?: number; array?: Array<{b?: number}>}, 'array[0].b'>;

SetRequiredDeep<{a?: number; array?: [{b?: number}, {c?: number}]}, 'array[1].c'>;
```

source/set-required-deep.d.ts
```diff
@@ -0,0 +1,45 @@
...
+	array?: {
+		c?: number
+	}[]
+}
+
+type SomeRequiredDeep = SetRequiredDeep<Foo, 'a' | 'array.c'>;
```
Emiyaaaaa, line 24:
```suggestion
type SomeRequiredDeep = SetRequiredDeep<Foo, 'a' | 'array.${number}.c'>;
```

I kinda forgot how it's should be. You can add the example in test to ensure it is worked.

Emiyaaaaa (changes requested):

test-d/set-required-deep.ts
hugomartinet: It's not working for array nested types, because the combination of the required and not required keys do not merge the array back together after setting keys to required.

So the test shows this error, which is legit:
Screenshot 2024-08-19 at 11 53 41
Screenshot 2024-08-19 at 11 54 38

Is there a type in **type-fest** that I could use to handle the merging of nested array types properly ? I searched but couldn't find one.

Thanks for your help!

Emiyaaaaa: `&` is not effect the final result, I think this is a little bug on `SimplifyDeep`, you can test by `expectAssignable`:
```
expectAssignable<SetRequiredDeep<{a?: number; array?: Array<{b?: number; c?: string}>}, `array.${number}.b`>>({array: [{b: 1}]});
```

hugomartinet: Actually, the problem with my current type definition of `SetRequiredDeep` is that : 
If I write 
```typescript
type Result = SetRequiredDeep<{ array?: {b?: number; c?: string}[] }, `array.${number}.b`>
```

I want the result to be :
```typescript
type Result = { array?: {b: number; c?: string}[] }
```

But for now the result will be :
```typescript
type Result = { array: {b: number}[] &  {c?: string}[] }
```
which is incorrect...

hugomartinet: I guess I need to define the `SetRequiredDeep` type recursively right? So I can't use `RequiredDeep` util in my definition

Emiyaaaaa: > Actually, the problem with my current type definition of `SetRequiredDeep` is that : If I write
> 
> ```ts
> type Result = SetRequiredDeep<{ array?: {b?: number; c?: string}[] }, `array.${number}.b`>
> ```

If your query is about the optionality of the "array", I think `array` is correct, not `array?`.

If about `{b: number; c?: string}[]` and `{b: number}[] &  {c?: string}[]`, [this answer](https://stackoverflow.com/questions/58088420/any-behavior-difference-between-a-number-b-string-and-a-number-b) can help you. (I think this should be handled by `SimplifyDeep` in your code, but it's not being done now)

hugomartinet: Thanks for the explanation!

I added the 3 tests you asked for but since SimplifyDeep does not handle the merging of array types, the tests fail (as you can see in the CI). Should I comment them for now?

Emiyaaaaa: > `&` is not effect the final result, `SimplifyDeep` only make result more beautiful and not change the result. I think this is a little bug on `SimplifyDeep`, you can test by `expectAssignable`:
> 
> ```ts
> expectAssignable<SetRequiredDeep<{a?: number; array?: Array<{b?: number; c?: string}>}, `array.${number}.b`>>({array: [{b: 1}]});
> ```

use this instead 👆🏻

hugomartinet: 🙏

Emiyaaaaa:
Need to resolve all conversation

hugomartinet:
> Need to resolve all conversation

Done

test-d/set-required-deep.ts
```diff
@@ -0,0 +1,35 @@
+import {expectAssignable, expectNotAssignable, expectType} from 'tsd';
+import type {MergeDeep, OverrideProperties, SetRequired, SetRequiredDeep, Simplify, SimplifyDeep} from '../index';
+
+// Update an optional nested key to required.
+declare const variation1: SetRequiredDeep<{a?: number; b?: {c?: string}}, 'b.c'>;
+expectType<{a?: number; b: {c: string}}>(variation1);
+
+// Update a root key to required.
+declare const variation2: SetRequiredDeep<{a?: number; b?: {c?: string}}, 'a'>;
+expectType<{a: number; b?: {c?: string}}>(variation2);
```
Beraliv, line 10: Is it worth adding a test case with multiple paths?

```ts
declare const variation2: SetRequiredDeep<{a?: number; b?: {c?: string}}, 'a' | 'b'>;
expectType<{a: number; b: {c?: string}}>(variation2);
```

source/set-required-deep.d.ts
```diff
@@ -0,0 +1,45 @@
...
+// 	b?: string;
+// 	array: {
+//    c: number // Is now required
+//  }[]
+// }
+```
+
+@category Object
+*/
+export type SetRequiredDeep<BaseType, KeyPaths extends Paths<BaseType>> =
```
Beraliv, line 36: I'm playing with this solution using the following example:

```ts
type Teacher = { address?: { postcode?: string; city?: string } };
type P1 = SetRequiredDeep<Teacher, 'address' | 'address.city'>;
//      ^?
```

This gives me:

```ts
type P1 = {
    address: {
        postcode: string;
        city: string;
    };
} | {
    address: {
        postcode: string;
        city: string;
    };
}
```

Is it expected?

Emiyaaaaa: I agree

source/set-required-deep.d.ts
Emiyaaaaa: > I kinda forgot how it's should be. You can add the example in test to ensure it is worked.

This doesn't seem to be resolved @hugomartinet

source/set-required-deep.d.ts
hugomartinet: Yes I think my current implementation could not work for this, since I use `RequiredDeep` and `RequiredDeep` makes all child keys of an object required as well.

sindresorhus (approved):

Beraliv:
@sindresorhus @hugomartinet I'll have a look at your PR later today or tomorrow

source/set-required-deep.d.ts
```diff
@@ -0,0 +1,72 @@
+import type {NonRecursiveType} from './internal';
+import type {Paths} from './paths';
+import type {SimplifyDeep} from './simplify-deep';
+
+/**
+Create a type that makes the given keys required. You can specifiy deeply nested key paths. The remaining keys are kept as is.
```
Beraliv, line 6: There is a typo in the word "specify", it's written as "specifiy"

test-d/set-required-deep.ts
```diff
@@ -0,0 +1,34 @@
+import {expectType} from 'tsd';
+import type {SetRequiredDeep} from '../index';
+
+// Set nested key to required
+declare const variation1: SetRequiredDeep<{a?: number; b?: {c?: string}}, 'b.c'>;
+expectType<{a?: number; b: {c: string}}>(variation1);
```
Beraliv, line 6: you haven't specified `'b' | 'b.c'`, therefore the expected behaviour should be `expectType<{a?: number; b?: {c: string}}>(variation1);` (b is still optional)

test-d/set-required-deep.ts
```diff
@@ -0,0 +1,34 @@
...
+declare const variation5: SetRequiredDeep<{a?: '1'; b?: {c?: boolean}} | {a?: '2'; b?: {c?: boolean}}, 'a' >;
+expectType<{a: '1'; b?: {c?: boolean}} | {a: '2'; b?: {c?: boolean}}>(variation5);
+
+// Set array key to required
+declare const variation6: SetRequiredDeep<{a?: Array<{b?: number}>}, 'a'>;
+expectType<{a: Array<{b?: number}>}>(variation6);
+
+// Set key inside array to required
+declare const variation7: SetRequiredDeep<{a?: Array<{b?: number}>}, `a.${number}.b`>;
+expectType<{a: Array<{b: number}>}>(variation7);
```
Beraliv, line 30: Similar here, the only required path `` `a.${number}.b` `` is specified, not `` 'a' | `a.${number}.b` ``, therefore the expected behaviour should be `expectType<{a?: Array<{b: number}>}>(variation7);`

Beraliv:
@hugomartinet I left my thoughts, I don't think it's ready because of a bug in the implementation, but I will leave it up to you (cc @sindresorhus)

test-d/set-required-deep.ts
hugomartinet: Thanks for this remark. I actually did this on purpose, in my mind I thought that this is what we want.

Since `SetRequired<..., 'b.c'>` would mean that we want `b.c` to always be defined, hence we also need `b` to always be defined. Do you disagree ? 

The reason is that I thought it would be fastidious when having nested objects with many levels. Writing `SetRequired<..., 'a.b.${number}.c'>` is lighter than writing `SetRequired<..., 'a' | 
'a.b' | 'a.b.${number}' | 'a.b.${number}.c'>`.

What do you think ? If you're confident on making only the nested key required I can work on that again 😀

hugomartinet:
> @hugomartinet I left my thoughts, I don't think it's ready because of a bug in the implementation (specified in my 2 comments where exactly), but I will leave it up to you (cc @sindresorhus)

@Beraliv I pushed a correction, I think I managed to take into account the expected behaviour you mentioned, thanks for pointing it out!

hugomartinet:
@sindresorhus @Beraliv How should we proceed on this?
Do you still have comments and suggestions?

Beraliv:
@hugomartinet thank you for your patience! If it waits, I'll leave my comments by the end of the year (next 2 weeks). If not, proceed with what you have and I'll raise bugs later if I find anything unexpected

som-sm (commented):
@hugomartinet The current implementation seems to fail in the following cases:

1. ```ts
    type T1 = SetRequiredDeep<{ a: 1; b: { c?: 1 } }, 'b.c'>;
    //   ^? type T1 = { a: 1; b?: { c: 1 } }
    ```
    Shouldn't make `b` optional.
    <br>

2. ```ts
    type T1 = SetRequiredDeep<{ a: 1; readonly b: { c?: 1 } }, 'b.c'>;
    //   ^? type T1 = { a: 1; b?: { c: 1 } }
    ```
    Doesn't preserve the `readonly` modifier on `b`.
    <br>

3. ```ts
    type T1 = SetRequiredDeep<{ a: 1; readonly b: { c?: 1 } | number }, 'b.c'>;
    //   ^? type T1 = { a: 1; b?: { c: 1 } }
    ```
    Doesn't preserve the `| number` on `b`.
    <br>

4. ```ts
    type T1 = SetRequiredDeep<{ 0: 1; 1: { 2?: string } }, '1.2'>;
    //   ^? type T1 = {0: 1; 1: {2?: string}}
    ```
    Doesn't work with number keys.

Beraliv:
@som-sm thank you for your summary of issues, this is exactly what I experienced when I was testing changes in October. @hugomartinet can you add unit tests for all scenarios I and @som-sm mentioned and fix those? Thank you

hugomartinet:
@som-sm @Beraliv 

Thanks for the update, I added these test cases and fixed them 👍

sindresorhus (approved):

som-sm:
@hugomartinet Looks like it's not working properly with index signatures:
```ts
type Test = SetRequiredDeep<{[x: string]: unknown; a?: number}, 'a'>;
//   ^? {[x: string]: unknown; [x: number]: unknown;}
```
Should have returned `{[x: string]: unknown; a: number;}` instead.

<br>

And the implementation seems to have gotten really complicated, here's a much simpler implementation that passes all the existing tests and also works with index signatures.

```ts
export type SetRequiredDeep<Type, KeyPaths extends Paths<Type>> =
Type extends NonRecursiveType
	? Type
	: SimplifyDeep<(
		{[K in keyof Type as K extends (KeyPaths | StringToNumber<KeyPaths & string>) ? K : never]-?: Type[K]} extends infer RequiredPart
			? keyof RequiredPart extends never ? unknown : RequiredPart
			: never) & {
		[K in keyof Type]: Extract<KeyPaths, `${K & (string | number)}.${string}`> extends never
			? Type[K]
			: SetRequiredDeep<Type[K], KeyPaths extends `${K & (string | number)}.${infer Rest extends Paths<Type[K]>}` ? Rest : never>
	}>;

type StringToNumber<Keys extends string> = Keys extends `${infer N extends number}` ? N : never;
```

hugomartinet:
> @hugomartinet Looks like it's not working properly with index signatures:
> 
> ```ts
> type Test = SetRequiredDeep<{[x: string]: unknown; a?: number}, 'a'>;
> //   ^? {[x: string]: unknown; [x: number]: unknown;}
> ```
> 
> Should have returned `{[x: string]: unknown; a: number;}` instead.
> 
> 
> And the implementation seems to have gotten really complicated, here's a much simpler implementation that passes all the existing tests and also works with index signatures.
> 
> ```ts
> export type SetRequiredDeep<Type, KeyPaths extends Paths<Type>> =
> Type extends NonRecursiveType
> 	? Type
> 	: SimplifyDeep<(
> 		{[K in keyof Type as K extends (KeyPaths | StringToNumber<KeyPaths & string>) ? K : never]-?: Type[K]} extends infer RequiredPart
> 			? keyof RequiredPart extends never ? unknown : RequiredPart
> 			: never) & {
> 		[K in keyof Type]: Extract<KeyPaths, `${K & (string | number)}.${string}`> extends never
> 			? Type[K]
> 			: SetRequiredDeep<Type[K], KeyPaths extends `${K & (string | number)}.${infer Rest extends Paths<Type[K]>}` ? Rest : never>
> 	}>;
> 
> type StringToNumber<Keys extends string> = Keys extends `${infer N extends number}` ? N : never;
> ```
> 
> @hugomartinet WDYT? Do you see any cases where the above implementation might fail?

Looks good to me, thanks for the simplification

som-sm:
> Looks good to me, thanks for the simplification

Great! I'll go ahead and update the PR then.

Although, most likely the following bit can be simplified further to just be the mapped type:
```ts
{[K in keyof Type as K extends (KeyPaths | StringToNumber<KeyPaths & string>) ? K : never]-?: Type[K]} extends infer RequiredPart
	? keyof RequiredPart extends never ? unknown : RequiredPart
	: never
```

Currently, if I replace the above bit with just the mapped type:
```
{[K in keyof Type as K extends (KeyPaths | StringToNumber<KeyPaths & string>) ? K : never]-?: Type[K]}
```

It fails to preserve the structure of the array for some reason. So, `SetReadonlyDeep<{b: Array<{c?: 1}>}>, `b.${number}.c`>` doesn't return `{b: Array<{c: 1}>}`, instead it unwraps the array. Ideally this shouldn't have happened because this case evaluates to: `{} & {b: {} & Array<{c: 1} & {c?: 1}>` which should simplify down to `{b: Array<{c: 1}>}`, but for some reason it doesn't.

Will have to spend more time to understand what's going wrong, but I guess it's fine for now.

som-sm (approved):

hugomartinet:
Thanks @som-sm!

What are the next steps for this PR?

Beraliv:
@som-sm @hugomartinet I'm having a last look at the PR, please bear with me

Beraliv:
From test perspective, everything works as expected, great work @hugomartinet & @som-sm!

Beraliv:
> It fails to preserve the structure of arrays for some reason. So, SetReadonlyDeep<{b: Array<{c?: 1}>}>, `b.${number}.c`> doesn't return {b: Array<{c: 1}>}, instead it unwraps the array. Ideally this shouldn't have happened because this case evaluates to: {} & {b: {} & Array<{c: 1} & {c?: 1}> which should simplify down to {b: Array<{c: 1}>}, but for some reason it doesn't. For now, I'm replacing the empty objects with unknown to make it work.

Very interesting investigation, @som-sm! I tried to follow the path around `SimplifyDeep`: when I remove `SimplifyDeep` and your workaround with `unknown`,

`SetRequiredDeep<{ b: Array<{ c?: 1 }> }, `b.${number}.c`>` is evaluated as:

```ts
{} & {
    b: {} & ({
        c: 1;
    } & {
        c?: 1;
    })[];
}
```

All the next statements evaluated correctly:

```ts
// { c: 1}[]
type B1 = SimplifyDeep<{} & Array<{ c: 1 }>>;

// {b: {c: 1}[]}
type B2 = SimplifyDeep<{
	b: ({
		c: 1;
	} & {
		c?: 1;
	})[];
}>;

// {b: {c: 1}[]}
type B3 = SimplifyDeep<
	{} & {
		b: {} & ({
			c: 1;
		} & {
			c?: 1;
		})[];
	}
>;
```

But when I call `SimplifyDeep<A>`, it again unwraps arrays. I believe it can be a known quirk of TypeScript or simply a bug. I searced issues in TypeScript repo but I cannot find anything.

Beraliv (approved):
@hugomartinet @som-sm amazing work, thanks both of you for addressing the issues, it LGTM now!

som-sm:
@Beraliv Thanks for the detailed breakdown.

<br>

> I believe it can be a known quirk of TypeScript or simply a bug.

I think I've managed to find a simpler reproduction of this issue:
```ts
import {Simplify} from "type-fest";

type T1 = Simplify<{} & {c: 1}[]>; // This works as expected
//   ^? type T1 = {c: 1}[]

type T2 = Simplify<Pick<{}, never> & {c: 1}[]>; // This unwraps the array
//   ^? type T2 = {[x: number]: {c: 1}; length:...
```

`Pick<{}, never>` evaluates down to `{}`, but the outcome is different from when it's a literal `{}`. This also means my workaround doesn't actually need an `unknown`, it works even with a literal `{}`.

som-sm:
Also, simplified down the following conditional:
```ts
{[K in keyof Type as K extends (KeyPaths | StringToNumber<KeyPaths & string>) ? K : never]-?: Type[K]} extends infer RequiredPart
	? keyof RequiredPart extends never ? unknown : RequiredPart
	: never
```
to this:
```ts
BaseType extends UnknownArray
	? {}
	: {[K in keyof BaseType as K extends (KeyPaths | StringToNumber<KeyPaths & string>) ? K : never]-?: BaseType[K]}
```
Refer [this commit](https://github.com/sindresorhus/type-fest/pull/939/commits/a938634cd972028d481a4488ce256a6d12f9b70d).

<br>

**NOTE:**
None of the solutions discussed so far works when the requirement is remove the optional modifier from a tuple, like:
```ts
type T = SetRequiredDeep<{a: [string?]}, "a.0"> // doesn't remove the `?` modifier
//  actual: {a: [string?]} // returns the same thing
//  expected: {a: [string]}
```

But, I guess that shouldn't be blocker.

som-sm:
> What are the next steps for this PR?

@hugomartinet PR LGTM! 

@sindresorhus Shall we merge?

Beraliv:
@som-sm great, this looks even simpler now!

sindresorhus:
> None of the solutions discussed so far work when the requirement is to remove an optional modifier from a tuple, like:

> type T = SetRequiredDeep<{a: [string?]}, "a.0"> // doesn't remove the `?` modifier
//  actual: {a: [string?]} // returns the same thing
//  expected: {a: [string]}
And, it would need more thought in cases like these:

> type T = SetRequiredDeep<{a: [string?, string?]}, "a.1"> // can't make `1` required, w/o making `0` required

It would be great if you could open an issue about this.

sindresorhus:
Looks good. Nice work, everyone 🙏

elijaholmos:
Hi friends, thanks for this PR! I was anticipating the release of this type 🙂 I installed the latest version of type-fest and ran into the following behavior, which seemed unintentional, but I wasn't sure:

```ts
import type { PartialDeep, SetRequiredDeep } from 'type-fest';

type Foo = {
  a: number;
  b?: string;
  c: {
    d: boolean[];
  };
};

type Bar = SetRequiredDeep<PartialDeep<Foo>, 'c.d'>;

function test(bar: Bar) {
  bar.a;
  bar.b;
  bar.c.d;	// error -> 'bar.c' is possibly 'undefined'. ts(18048)
}
```

Is the deep property accessor (`'c.d'`) supposed to make its parent property (`c`) required as well? Logically, that would follow, since there'd be no way to access `c.d` if `c` is inaccessible (where accessibility refers to lack of partiality). Please correct me if I'm wrong or if there's a better way to approach this!

som-sm:
@elijaholmos 
> Is the deep property accessor ('c.d') supposed to make its parent property (c) required as well?

No, only the specified paths are made required, not the entire chain.

This design makes sense because requiring a specific path doesn’t necessarily mean all its parent paths should also be required. Consider the following example:
```ts
import type {SetRequiredDeep} from 'type-fest';

type Destination = SetRequiredDeep<{
	name: string;
	address: string;
	coordinates?: {
		lat?: number; // This was incorrectly marked optional
		lon: number;
	},
}, "coordinates.lat">

declare const d1: Destination;

if (d1.coordinates) {
	let coor = d1.coordinates;
	//  ^? let coor: {
	//         lat: number;
	//         lon: number;
	//     }
}
```
Here, `coordinates` is optional, but if it exists, it guarantees the presence of `lat` and `lon`.

<br>

However, we can introduce an option like `c.*`, which would make all paths starting with `c` required.

elijaholmos:
Thank you for the explanation and example, @som-sm. That makes sense!

PR #949, opened 2024-08-28
`Exact`: Fix usage with recursive types and unions

### LINKED ISSUE #948: `Exact` no longer works with recursive types + unions
Code that worked in v4.21.0 no longer compiles in v4.22.0 after the change to `Exact`. Simplified code below:

```tsx
import { Exact } from 'type-fest';

export type A = { a: Expected; }
export type B = { b: string }
export type Expected = A | B;

const _errors: Exact<Expected, Expected> = {} as Expected;
```

[Playground Link](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgUQB4EMDG8C+cBmUEIcA5DAJ5gCmAtPtQM4ykDcAUO9apLHJTTgBBOAF5EcdAC4UPatmoATVnBxce0eAOpwAQmIkAjGcyjAAdgHNV63lqo60NBYoMiAPno7tMEc8zgAfWooIihGGTQsGAAeJ3kYJQAaWWdExQA+AwQ8dEZUhKVWIA)

We believe that the change which caused `IsEqual` to not be first [here](https://github.com/sindresorhus/type-fest/pull/911/files#diff-fb970dfdcdd46dfa104c565ac5a63e3b706a028781fda85b027b362e19f421cfL55) causes the type to be distributed early, and the `IsEqual` check suddenly fails because it distributes to `Exact<A, Expected> | Exact<B, Expected>` which doesn't make any sense.

`isEqual` _used_ to be first, which is why this worked previously, so I think moving it first again could fix this issue.

cc @zorji 

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/948">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/948/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

abrenneke:
Fixes #948

This moves `IsEqual` to be the first check, before any distribution, restoring the behavior before #911.

Wrote a test for it, and all tests are still passing, including those written for #911!

### DIFF
--- a/test-d/exact.ts
+++ b/test-d/exact.ts
@@ -537,3 +537,19 @@ import type {Exact, Opaque} from '../index';
 	// @ts-expect-error
 	function_(withExcessSurname);
 }
+
+// Spec - recursive type with union
+// @see https://github.com/sindresorhus/type-fest/issues/948
+{
+	type A = {
+		a: Expected;
+	};
+	type B = {
+		b: string;
+	};
+	type Expected = A | B;
+
+	const function_ = <T extends Exact<Expected, T>>(arguments_: T) => arguments_;
+
+	function_({} as Expected);
+}
--- a/source/exact.d.ts
+++ b/source/exact.d.ts
@@ -53,13 +53,14 @@ onlyAcceptNameImproved(invalidInput); // Compilation error
 @category Utilities
 */
 export type Exact<ParameterType, InputType> =
-	// If the parameter is a primitive, return it as is immediately to avoid it being converted to a complex type
-	ParameterType extends Primitive ? ParameterType
-		// If the parameter is an unknown, return it as is immediately to avoid it being converted to a complex type
-		: IsUnknown<ParameterType> extends true ? unknown
-			// If the parameter is a Function, return it as is because this type is not capable of handling function, leave it to TypeScript
-			: ParameterType extends Function ? ParameterType
-				: IsEqual<ParameterType, InputType> extends true ? ParameterType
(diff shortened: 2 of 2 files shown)


abrenneke:
@zorji I'm assuming because the tests pass, that this doesn't revert any of the behavior you fixed in #911 but it's possible I suppose?

test-d/exact.ts
```diff
@@ -537,3 +537,19 @@ import type {Exact, Opaque} from '../index';
 	// @ts-expect-error
 	function_(withExcessSurname);
 }
+
+// Spec - recursive type with union
+// @see https://github.com/sindresorhus/type-fest/issues/948
+{
+	type A = {
+		a: A;
```
zorji, line 545: Sorry, I don't understand this type. Looks like this is an infinite recursive type.

Do you actually mean the following?

```
type A = {
  a: Expected;
}
```

zorji:
@abrenneke Sorry for the late reply. I think the implementation is OK but I have a question for the test case.

test-d/exact.ts
abrenneke: Ah, yes I think that's what I meant

abrenneke: Fixed. I think it would also test the same case with `a?: A`. As long as there's _some_ sort of recursion in either of the types. TS doesn't even really care about the infinite recursion unless you try to actually instantiate or resolve the type 🤷

zorji (approved):

PR #887, opened 2024-06-06
`Schema`: Fix handling of arrays

grigorischristainas:
I opened this pull request based on #883. It includes a refactoring regarding the array handling. Although the issue concerns arrays of objects, I think it makes sense to generalize it to all arrays.

Given the following example: 
```ts
import type {Schema} from 'type-fest';

interface Users {
	names: Array<string>;
        fullnames: Array<{
		firstname: string;
		lastname: string;
	}>;
}

type UserMask = Schema<Users, 'mask' | 'hide' | 'show'>;
```

**Previous implementation:**

```ts
const userMaskSettings: UserMask = {
	names: 'mask',
        fullnames: 'mask'
}
```

**New implementation:**

```ts
const userMaskSettings: UserMask = {
	names: ['mask'],
	fullnames: [
		{
			firstname: 'show',
			lastname: 'hide'
		}
	]
}
```
I would be happy to make any further adjustments as needed.

PR #960, opened 2024-09-27
`Schema`: Add `recurseIntoArrays` option

iamandrewluca:
Allow `Schema` behavior before this PR https://github.com/sindresorhus/type-fest/pull/887 via Options. Please see PR comments for more context.


sindresorhus:
I think it makes more sense to add an option to the `Scheme` type. The behavior introduced here does not really make it "loose" and "loose" is also a ambiguous term for this.

grigorischristainas:
@iamandrewluca @sindresorhus if you agree, I would suggest adding the optional config  `{ recurseIntoArrays: boolean }` (defaulting to `true`). This way, the type would eventually be used as follows:
```
type FooOption = 'A' | 'B';
type FooSchema = Schema<typeof foo, FooOption, {recurseIntoArrays: false}>;
```

iamandrewluca:
I will try to refactor it, to be able to use recurseIntoArrays

grigorischristainas:
Ok, let me know if you need any assistance, I'd be glad to help.

iamandrewluca:
@grigorischristainas I tried the approach, but I think something I didn't do right.

I modified the current tests just to check. In the end, I will write separate type tests for this use case.

iamandrewluca:
> I think I figured it out. Checked other types.

Figured it out with the options. Can't figure out how to make it work ))
I feel I'm missing some TypeScript gotcha.

grigorischristainas:
Thanks @iamandrewluca for the update, I will be able to take a look by the end of day tomorrow, or at the latest on Monday. I could checkout at your branch and commit any changes there.

grigorischristainas:
@iamandrewluca I cloned your branch (`revert-schema-as-schema-loose`), but I can't push my commits, could you please give me that permission? I think you can enable that from `Settings` -> `Collaborators`.

grigorischristainas:
Unfortunately, I cannot fork it because I already have a fork of the original type-fest repository. Maybe I could create a new branch in your forked repo and if you agree with the changes then merge it into yours.

iamandrewluca:
I invited you as a collaborator in my fork.

grigorischristainas:
Thanks, I opened a PR to your forked repo with the changes. Please let me know if you need any additional information.

iamandrewluca:
@grigorischristainas, thanks for the help, I merged your PR. All the changes are visible here now.

grigorischristainas:
@sindresorhus, just following up to see if there’s any update or feedback on this PR. Please take a look whenever you have some time, thank you!

source/schema.d.ts
```diff
@@ -32,12 +33,34 @@ const userMaskSettings: UserMask = {
...
+	active: 'show',
+	passwordHash: 'hide',
+	attributes: 'hide'
 }
 ```
 
 @category Object
 */
-export type Schema<ObjectType, ValueType> = ObjectType extends string
+export type Schema<ObjectType, ValueType, Options extends SchemaOptions = {recurseIntoArrays: true}> = ObjectType extends string
```
sindresorhus, line 63: It should not have `= {recurseIntoArrays: true}` because it will only work correctly as long as there is a single option. Which could cause bugs in the future if we add another option.

source/schema.d.ts
```diff
@@ -32,12 +33,34 @@ const userMaskSettings: UserMask = {
...
+	id: 'show',
+	name: {
+		firstname: 'show',
+		lastname: 'mask',
+	},
+	created: 'show',
+	active: 'show',
+	passwordHash: 'hide',
+	attributes: 'hide'
 }
```
sindresorhus, line 58: I think it would be better to put this in a doc comment on the `recurseIntoArrays` option.

source/schema.d.ts
```diff
@@ -58,14 +83,32 @@ export type Schema<ObjectType, ValueType> = ObjectType extends string
...
+		? Options['recurseIntoArrays'] extends true
+			? Schema<ObjectType[KeyType], K, Options>
+			: K
+		: Schema<ObjectType[KeyType], K, Options> | K;
+};
+
+/**
+@see Schema
+*/
+export type SchemaOptions = {
```
sindresorhus, line 109: This must be re-exported in the index file.

sindresorhus (changes requested):

source/schema.d.ts
```diff
@@ -32,12 +33,13 @@ const userMaskSettings: UserMask = {
...
 	active: 'show',
 	passwordHash: 'hide',
+	attributes: ['mask', 'show']
 }
 ```
 
 @category Object
 */
-export type Schema<ObjectType, ValueType> = ObjectType extends string
+export type Schema<ObjectType, ValueType, Options extends SchemaOptions = {}> = ObjectType extends string
```
iamandrewluca, line 42: Because the `@default` of `recurseIntoArrays` is `true` when missing it probably creates the problems, I see only two use cases in the library that has `@default true` when missing.

- `CamelCaseOptions`, this has the generic default to `{preserveConsecutiveUppercase: true}`
- `GetOptions`, this has the generic default to `{}`

sindresorhus: You can still make it true by default without using the default parameter. Just invert  the check.

iamandrewluca: Fixed

source/schema.d.ts
```diff
@@ -48,7 +50,9 @@ export type Schema<ObjectType, ValueType> = ObjectType extends string
 				: ObjectType extends ReadonlySet<unknown>
 					? ValueType
 					: ObjectType extends Array<infer U>
-						? Array<Schema<U, ValueType>>
+						? Options['recurseIntoArrays'] extends false
```
iamandrewluca, line 53: Question: Because `recurseIntoArrays` is optional, shouldn't we check here also for undefined? 🤔 

```suggestion
						? Options['recurseIntoArrays'] extends false | undefined
```

iamandrewluca: If the user has TS Enabled `exactOptionalPropertyTypes`, this should not be a problem.
If the user does not have this enabled, they can pass undefined for recurseIntoArrays

iamandrewluca: Probably we also need a test case for when `undefined` is passed. Because I see that type-fest has enabled `exactOptionalPropertyTypes`

iamandrewluca: I included the `undefined` in the tests when using `recurseIntoArrays`

PR #965, opened 2024-10-12
Update readme.md to instruct to to install as Dev Dependency

muchisx:
Hi!

Noticed that the proposed install was set as default (save).

It doesn't occur to me a case in which this wouldn't be installed as a Dev Dependency, let me know if there's one!


sindresorhus:
Duplicate of #851

PR #975, opened 2024-11-03
export (Split)Words

eranhirsch:
The SplitWords utility is not under the `/internal/` directory, but isn't exported.

Would it be OK to export it? I want to use it for some types in https://github.com/remeda/remeda


eranhirsch:
Would [`words`](https://lodash.com/docs/4.17.15#words) be enough?

voxpelli:
Noting that #930 is also touching / changing this

PR #977, opened 2024-11-07
`ArrayTail`: Fix support for optional parameters

### LINKED ISSUE #978: ``ArrayTail`` doesn't work when mixed with optional elements
### Bug description

This type:

``
type Test = ArrayTail<[foo?: string | undefined, bar?: number | undefined]>;
``

Is resolved as ``[]``

### Repro

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBDAnmApnA3nAglKBDRAFT2ABs4BfOAMyghDgHIlUBaalAZxkYG4BYAFBDqAVwB2AYxjAI4hFxgAKahAgB+AFxxuUYOIDmAGjgAjPFC1xxokKZRQAlBiEBIKChiio4gYIpCQgD0QXCEABbAnHAenBCkAG5cCBBwANqqGtq6+gZwAD5wEgAmKNT6KMUm5pbaNnYOBUXipeXilQC6QixoAGIS0rLiAAoWeCDRALxwo-ggng6cADw9ENQK3AB8fsGhEVE64RCipMUxXPFJKek1VvX2UE0lZRXFXYI9YcgovYrEZHBpjh8EQSKQlv0pDI5LNxpxtoFBCEwpFop9ygAPZIwcJoMAeBKyUTRKKcUQobrfOC9YBY4r-cEAGU2gLgjLgKAxMBQLWiHjwxTkpEQ6Q6cHUbLg2nZnO5vPOAqFIrSEgA1uIIAB3cTqEwAOgN+g4j0ZDLFEtNYKlbJ2SNCACULolknhomlbnVbA8ni0Xu03nBJNAPNJhZTUHAAOrAHE0ukM1lxyoMiEDaEjMYTbZAA

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/978">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/978/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

ferferga:
Fixes #978

### DIFF
--- a/test-d/array-tail.ts
+++ b/test-d/array-tail.ts
@@ -10,3 +10,17 @@ expectType<[]>(getArrayTail(['a', 'b', 'c']));
 expectType<[]>(getArrayTail([] as const));
 expectType<[]>(getArrayTail(['a'] as const));
 expectType<['b', 'c']>(getArrayTail(['a', 'b', 'c'] as const));
+
+// Optional elements tests
+expectType<[undefined, 'c']>(getArrayTail(['a', undefined, 'c'] as const));
+
+// Mixed optional/required
+type MixedArray = [string, undefined?, number?];
+expectType<[undefined?, number?]>(getArrayTail(['hello'] as MixedArray));
+
+// Optional numbers
+expectType<[undefined, 3]>(getArrayTail([1, undefined, 3] as const));
+
+// Complex mixed case
+type ComplexArray = [string, boolean, number?, string?];
+expectType<[boolean, number?, string?]>(getArrayTail(['test', false] as ComplexArray));
--- a/source/array-tail.d.ts
+++ b/source/array-tail.d.ts
@@ -22,4 +22,8 @@ add3(4);
 
 @category Array
 */
-export type ArrayTail<TArray extends UnknownArrayOrTuple> = TArray extends readonly [unknown, ...infer Tail] ? Tail : [];
+export type ArrayTail<TArray extends UnknownArrayOrTuple> = TArray extends readonly [unknown?, ...infer Tail]
+	? keyof TArray & `${number}` extends never
+		? []
+		: Tail
+	: [];


ferferga:
Not sure how this can be approachable in a backwards-compatible way so tests pass.

som-sm:
@ferferga Following solution addresses the issue mentioned in #978 while ensuring all existing tests continue to pass.

```ts
export type ArrayTail<TArray extends UnknownArrayOrTuple> = keyof TArray & `${number}` extends never
	? []
	: TArray extends readonly [unknown?, ...infer Tail]
		? Tail
		: [];
```

The conditional <code>keyof TArray & \`${number}\` extends never</code> filters out empty arrays and non-tuple arrays like `string[]`, `number[]` etc.
<br>
However, the updated implementation fails when the type is instantiated with unions. For example, `TArray<[1, 2] | []>` now incorrectly resolves to `[]` instead of the expected `[2] | []` (although currently there’s no test covering this behaviour). To ensure the type continues to work correctly with unions, we can simply reorder the conditionals.

```ts
export type ArrayTail<TArray extends UnknownArrayOrTuple> = TArray extends readonly [unknown?, ...infer Tail]
	? keyof TArray & `${number}` extends never
		? []
		: Tail
	: [];
```

ferferga:
@som-sm Thank you very much for your input, works flawlessly! Took the chance to also add some tests.

test-d/array-tail.ts
```diff
@@ -10,3 +10,17 @@ expectType<[]>(getArrayTail(['a', 'b', 'c']));
...
+// Mixed optional/required
+type MixedArray = [string, undefined?, number?];
+expectType<[undefined?, number?]>(getArrayTail(['hello'] as MixedArray));
+
+// Optional numbers
+expectType<[undefined, 3]>(getArrayTail([1, undefined, 3] as const));
+
+// Complex mixed case
+type ComplexArray = [string, boolean, number?, string?];
+expectType<[boolean, number?, string?]>(getArrayTail(['test', false] as ComplexArray));
```
som-sm, line 26: @ferferga These test cases would have passed with the previous implementation as well, the problem was only when the first element was optional, like `["a"?, "b"?]`.

This test would have failed previously but works with updated implementation:
```ts
expectType<ArrayTail<['a'?, 'b'?]>>({} as ['b'?]);
```

<br>

Also, could add a test for unions:
```ts
expectType<ArrayTail<[] | ['a', 'b']>>({} as [] | ['b']);
```

som-sm: Added tests [here](https://github.com/sindresorhus/type-fest/pull/999).

PR #984, opened 2024-11-18
`SharedUnionFieldsDeep`: Fix propagation for non union root types

ocadoret:
`SharedUnionFieldsDeep` is not propagating deeply when root type is not a union type.

This PR aims to fix this.

### DIFF
--- a/source/shared-union-fields-deep.d.ts
+++ b/source/shared-union-fields-deep.d.ts
@@ -83,7 +83,15 @@ function displayPetInfo(petInfo: SharedUnionFieldsDeep<Cat | Dog>['info']) {
 export type SharedUnionFieldsDeep<Union, Options extends SharedUnionFieldsDeepOptions = {recurseIntoArrays: false}> =
 // If `Union` is not a union type, return `Union` directly.
 IsUnion<Union> extends false
-	? Union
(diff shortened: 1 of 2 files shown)


source/shared-union-fields-deep.d.ts
```diff
@@ -83,7 +83,15 @@ function displayPetInfo(petInfo: SharedUnionFieldsDeep<Cat | Dog>['info']) {
 export type SharedUnionFieldsDeep<Union, Options extends SharedUnionFieldsDeepOptions = {recurseIntoArrays: false}> =
 // If `Union` is not a union type, return `Union` directly.
```
Emiyaaaaa, line 84: I kind of forgot why I limited the input to `union`...

If I understand your question correctly, just need to remove that limit 
```typescript
IsUnion<Union> extends false ? Union
```

Emiyaaaaa (changes requested):

source/shared-union-fields-deep.d.ts
ocadoret: You're right, removing this conditional gives the same results. I updated my PR with this simplification 👍

Emiyaaaaa (approved):

PR #997, opened 2024-12-03
Add `AllUnionFields` type

### LINKED ISSUE #996: New Type: AllUnionFields (similar to SharedUnionFields)

Expand on `SharedUnionFields` with a type that keeps all the union fields.
This is a simplified version of #386.

~This new type will simply change non-shared fields to optional with value `unknown`. Or with an option (proposed name `exact`), will be changed to be optional with a value of the union of all values for that property within the union.~

Edit: This new type will change each field to be optional with a value of the union of all values for that property within the union.

### Simple example by modifying the `SharedUnionFields`'s example (probably not that great an example of why you'd actually want to use this type)

```ts
import type {AllUnionFields} from 'type-fest';

type Cat = {
	name: string;
	type: 'cat';
	catType: string;
};

type Dog = {
	name: string;
	type: 'dog';
	dogType: string;
};

function displayPetInfo(petInfo: Cat | Dog) {
	// typeof petInfo =>
	// {
	// 	name: string;
	// 	type: 'cat';
	// 	catType: string;
	// } | {
	// 	name: string;
	// 	type: 'dog';
	// 	dogType: string;
	// }

	console.log('name: ', petInfo.name);
	console.log('type: ', petInfo.type);

	// TypeScript complains about `catType` and `dogType` not existing on type `Cat | Dog`.
	console.log('animal type: ', petInfo.catType ?? petInfo.dogType);
}

function displayPetInfo(petInfo: AllUnionFields<Cat | Dog>) {
	// typeof petInfo =>
	// {
	// 	name: string;
	// 	type: 'cat' | 'dog';
	//  catType?: string;
	//  dogType?: string;
	// }

	console.log('name: ', petInfo.name);
	console.log('type: ', petInfo.type);

	// No TypeScript error.
	console.log('animal type: ', petInfo.catType ?? petInfo.dogType);
}
```

### Type source

_No response_

- [X] I tried my best to look for it

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/996">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/996/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

RebeccaStevens:
fix #996

TODO: Create a deep version of the type.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -130,6 +130,7 @@ export type {ArraySplice} from './source/array-splice';
 export type {ArrayTail} from './source/array-tail';
 export type {SetFieldType} from './source/set-field-type';
 export type {Paths} from './source/paths';
+export type {AllUnionFields} from './source/all-union-fields';
 export type {SharedUnionFields} from './source/shared-union-fields';
 export type {SharedUnionFieldsDeep} from './source/shared-union-fields-deep';
 export type {IsNull} from './source/is-null';
--- a/readme.md
+++ b/readme.md
@@ -203,6 +203,7 @@ Click the type names for complete docs.
 - [`Paths`](source/paths.d.ts) - Generate a union of all possible paths to properties in the given object.
 - [`SharedUnionFields`](source/shared-union-fields.d.ts) - Create a type with shared fields from a union of object types.
 - [`SharedUnionFieldsDeep`](source/shared-union-fields-deep.d.ts) - Create a type with shared fields from a union of object types, deeply traversing nested structures.
+- [`AllUnionFields`](source/all-union-fields.d.ts) - Create a type with all fields from a union of object types.
 - [`DistributedOmit`](source/distributed-omit.d.ts) - Omits keys from a type, distributing the operation over a union.
 - [`DistributedPick`](source/distributed-pick.d.ts) - Picks keys from a type, distributing the operation over a union.
 - [`And`](source/and.d.ts) - Returns a boolean for whether two given types are both true.
--- a/source/all-union-fields.d.ts
+++ b/source/all-union-fields.d.ts
@@ -73,13 +73,15 @@ Extract<Union, NonRecursiveType | ReadonlyMap<unknown, unknown> | ReadonlySet<un
 		?
 		| SkippedMembers
 		| Simplify<
+		// Include fields that are common in all union members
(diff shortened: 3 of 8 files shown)


Emiyaaaaa:
Is there a better name for it?

// @sindresorhus

som-sm (commented):
@RebeccaStevens IMO, the following implementation is simpler than the one proposed in this PR:

```ts
export type AllUnionFields<Union> = [Union] extends [ NonRecursiveType | ReadonlyMap<unknown, unknown> | ReadonlySet<unknown> | UnknownArray]
	? Union
	: Simplify<
	SharedUnionFields<Union> &
	(ReadonlyKeysOfUnion<Union> extends infer ReadonlyKeys extends PropertyKey
		? {
			readonly [P in ReadonlyKeys]?: ValueOfUnion<Union, P>;
		}
		: {}) & {
		[P in Exclude<KeysOfUnion<Union>, ReadonlyKeysOfUnion<Union> | keyof Union>]?: ValueOfUnion<Union, P>;
	}
	>;

type ValueOfUnion<Union, Key> = Union extends unknown ? Key extends keyof Union ? Union[Key] : never : never;

type ReadonlyKeysOfUnion<T> = T extends unknown ? keyof {
	[P in keyof T as IsEqual<{[Q in P]: T[P]}, {readonly [Q in P]: T[P]}> extends true ? P : never]: never
} : never;
```

This passes all the existing test cases and also uses utilities like `ValueOfUnion` and `ReadonlyKeysOfUnion`, which can be exposed as standalone types.

WDYT?

RebeccaStevens:
@som-sm I agree, that implementation is nicer. I pushed an update with that one.

som-sm:
@RebeccaStevens I have updated this PR with the following changes:
1. Moved the helper types to the `internal` folder. Added documentation and tests for them, would be helpful when we expose these types.
2. Improved handling for non-recursive types as mentioned [here](https://github.com/sindresorhus/type-fest/pull/997#issuecomment-2528945947) and added additional test cases accordingly.

I’d appreciate if you could review these updates once.

sindresorhus:
This one looks good to me 👍

PR #999, opened 2024-12-05
Add test cases for `ArrayTail` type

som-sm:
The behaviour of `ArrayTail` with tuples containing all optional elements was recently fixed in [this PR](https://github.com/sindresorhus/type-fest/pull/977). However, as mentioned [here](https://github.com/sindresorhus/type-fest/pull/977#discussion_r1866189549), there are no tests currently covering this scenario.


sindresorhus:
Thanks :)

PR #1006, opened 2024-12-12
Add ArrayTailDeep type

naremloa:
Add a new type `ArrayTailDeep`.It is useful when you need to rewrap some function and partially modify its parameters.

Here is a localStorage rewrap example:
```ts
import type {ArrayTailDeep} from 'type-fest';

type LocalStorageSchema = { id: string, name: string, token: string };
class LocalStorage {
	getItem<Key extends keyof LocalStorageSchema>(
		key: Key,
		...rest: ArrayTailDeep<Parameters<typeof localStorage.getItem>, 1>
	) {
		return localStorage.getItem(key, ...rest);
	}

	setItem<Key extends keyof LocalStorageSchema>(
		key: Key,
		value: LocalStorageSchema[Key],
		...rest: ArrayTailDeep<Parameters<typeof localStorage.setItem>, 2>
	) {
		return localStorage.setItem(key, value, ...rest);
	}
}
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -126,6 +126,7 @@ export type {ArrayValues} from './source/array-values';
 export type {ArraySlice} from './source/array-slice';
 export type {ArraySplice} from './source/array-splice';
 export type {ArrayTail} from './source/array-tail';
+export type {ArrayTailDeep} from './source/array-tail-deep';
 export type {SetFieldType} from './source/set-field-type';
 export type {Paths} from './source/paths';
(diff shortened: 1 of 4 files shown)


naremloa:
> @naremloa Is there really a need for a separate type? Similar behaviour be achieved using `ArraySlice`, like:
> 
> ```ts
> type T = ArraySlice<[1, 2, 3, 4], 1>; // => [2, 3, 4]
> ```

Without considering currying, ArraySlice is sufficient.

PR #1009, opened 2024-12-16
`KeysOfUnion`: Fix assignability with `keyof`

### LINKED ISSUE #1008: Assignability b/w generic `keyof T` and `KeysOfUnion<T>`
Ideally, `keyof T` _should_ be assignable to `KeysOfUnion<T>` and `KeysOfUnion<T>` _should not_ be assignable to `keyof T`.

However, currently the exactly opposite happens.

```ts
import { KeysOfUnion } from "type-fest";

// KeysOfUnion<T> should not be assignable to keyof T
type Assignability1<T, K extends keyof T> = unknown;
type Test1<T> = Assignability1<T, KeysOfUnion<T>>; // should error but doesn't

// keyof T should be assignable to KeysOfUnion<T>
type Assignability2<T, K extends KeysOfUnion<T>> = unknown;
type Test2<T> = Assignability2<T, keyof T>; // shouldn't error but does
```

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/1008">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/1008/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

som-sm:
Fixes #1008

To make `keyof T` assignable to `KeysOfUnion<T>`, we can simply add an union with `keyof ObjectType` to the existing type, as shown below. 

```ts
type KeysOfUnion<ObjectType> =
 | keyof ObjectType
 | ObjectType extends unknown ? keyof ObjectType : never;
```

However, I couldn’t find a straightforward way to prevent `KeysOfUnion<T>` from being assignable to `keyof T`. So, I reworked the implementation to use `UnionToIntersection`, which resolves both issues.

<hr>

There's another hack I figured out while experimenting, this also solves both the issues, but it's very hacky:
```ts
type KeysOfUnion<ObjectType> =
  | keyof ObjectType // Convinces TS that keyof T is part of KeysOfUnion<T>
  | ObjectType extends unknown ? keyof ObjectType: never
  | Extract<Uppercase<string>, Lowercase<ObjectType & string>>; // Convinces TS that KeysOfUnion<T> is broader than keyof T

// keyof T should be assignable to KeysOfUnion<T>
type Assignability2<T, K extends KeysOfUnion<T>> = unknown;
type Test2<T> = Assignability2<T, keyof T>; // doesn't error now

// KeysOfUnion<T> should not be assignable to keyof T
type Assignability1<T, K extends keyof T> = unknown;
type Test1<T> = Assignability1<T, KeysOfUnion<T>>; // does error now
```
[Playground](https://tsplay.dev/mZRl4m)

### DIFF
--- a/source/keys-of-union.d.ts
+++ b/source/keys-of-union.d.ts
@@ -1,3 +1,5 @@
+import type {UnionToIntersection} from './union-to-intersection';
+
 /**
 Create a union of all keys from a given type, even those exclusive to specific union members.
 
@@ -35,6 +37,6 @@ type AllKeys = KeysOfUnion<Union>;
 
 @category Object
 */
-export type KeysOfUnion<ObjectType> = ObjectType extends unknown
-	? keyof ObjectType
-	: never;
+export type KeysOfUnion<ObjectType> =
+  // Hack to fix https://github.com/sindresorhus/type-fest/issues/1008
+  keyof UnionToIntersection<ObjectType extends unknown ? Record<keyof ObjectType, never> : never>;
--- a/test-d/keys-of-union.ts
+++ b/test-d/keys-of-union.ts
@@ -35,3 +35,25 @@ type Expected2 = 'common' | 'a' | 'b' | 'c';
 declare const actual2: KeysOfUnion<Example2>;
 
 expectType<Expected2>(actual2);
+
+// With property modifiers
+declare const actual3: KeysOfUnion<{a?: string; readonly b: number} | {a: number; b: string}>;
+expectType<'a' | 'b'>(actual3);
+
+// KeysOfUnion<T> should NOT be assignable to keyof T
+type Assignability1<T, _K extends keyof T> = unknown;
+// @ts-expect-error
+type Test1<T> = Assignability1<T, KeysOfUnion<T>>;
+
+// Keyof T should be assignable to KeysOfUnion<T>
+type Assignability2<T, _K extends KeysOfUnion<T>> = unknown;
+type Test2<T> = Assignability2<T, keyof T>;
+
+// KeysOfUnion<T> should be assignable to PropertyKey
+type Assignability3<_T, _K extends PropertyKey> = unknown;
+type Test3<T> = Assignability3<T, KeysOfUnion<T>>;
+
+// PropertyKey should NOT be assignable to KeysOfUnion<T>
+type Assignability4<T, _K extends KeysOfUnion<T>> = unknown;
+// @ts-expect-error
+type Test4<T> = Assignability4<T, PropertyKey>;
--- a/test-d/keys-of-union.ts
+++ b/test-d/keys-of-union.ts
@@ -40,20 +40,20 @@ expectType<Expected2>(actual2);
 declare const actual3: KeysOfUnion<{a?: string; readonly b: number} | {a: number; b: string}>;
 expectType<'a' | 'b'>(actual3);
 
-// KeysOfUnion<T> should NOT be assignable to keyof T
+// `KeysOfUnion<T>` should NOT be assignable to `keyof T`
 type Assignability1<T, _K extends keyof T> = unknown;
 // @ts-expect-error
 type Test1<T> = Assignability1<T, KeysOfUnion<T>>;
 
-// Keyof T should be assignable to KeysOfUnion<T>
+// `keyof T` should be assignable to `KeysOfUnion<T>`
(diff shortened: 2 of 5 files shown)


source/keys-of-union.d.ts
```diff
@@ -35,6 +37,6 @@ type AllKeys = KeysOfUnion<Union>;
 
 @category Object
 */
-export type KeysOfUnion<ObjectType> = ObjectType extends unknown
-	? keyof ObjectType
-	: never;
+export type KeysOfUnion<ObjectType> =
+  // Hack to fix https://github.com/sindresorhus/type-fest/issues/1008
```
Emiyaaaaa, line 41: Is there a way not hack?

test-d/keys-of-union.ts
```diff
@@ -35,3 +35,25 @@ type Expected2 = 'common' | 'a' | 'b' | 'c';
 declare const actual2: KeysOfUnion<Example2>;
 
 expectType<Expected2>(actual2);
+
+// With property modifiers
+declare const actual3: KeysOfUnion<{a?: string; readonly b: number} | {a: number; b: string}>;
+expectType<'a' | 'b'>(actual3);
+
+// KeysOfUnion<T> should NOT be assignable to keyof T
+type Assignability1<T, _K extends keyof T> = unknown;
```
Emiyaaaaa, line 44: add test:
```ts
type Assignability1<T extends UnknownRecord, K extends keyof T> = unknown;
type Test1<T extends UnknownRecord> = Assignability1<T, KeysOfUnion<T>>; // should error
```

Emiyaaaaa (changes requested):

test-d/keys-of-union.ts
som-sm: @Emiyaaaaa The test you suggested doesn’t seem to produce an error:

```ts
type Assignability5<T extends UnknownRecord, _K extends keyof T> = unknown;
type Test5<T extends UnknownRecord> = Assignability5<T, KeysOfUnion<T>>; // doesn't error
```

<br>

However, adding slightly different constraints does produce the expected error:
```ts
type Assignability5<T extends Record<string, unknown>, _K extends keyof T> = unknown;
// @ts-expect-error
type Test5<T extends Record<string, unknown>> = Assignability5<T, KeysOfUnion<T>>; // Does error
```

```ts
type Assignability5<T extends object, _K extends keyof T> = unknown;
// @ts-expect-error
type Test5<T extends object> = Assignability5<T, KeysOfUnion<T>>; // Does error
```

Emiyaaaaa: I don't have idea right now, maybe it's a bug of Typescript I'm not sure. I'll try to look into this interesting issue later

source/keys-of-union.d.ts
sindresorhus: I think it's fine to do this for now until we come up with a better solution.

test-d/keys-of-union.ts
som-sm: @sindresorhus These tests don't use either `tsd` or `expect-type`, they are just simple generics, like this:

```ts
type Assignability5<T extends Record<string, unknown>, _K extends keyof T> = unknown;
// @ts-expect-error
type Test5<T extends Record<string, unknown>> = Assignability5<T, KeysOfUnion<T>>; // Does error
```

source/keys-of-union.d.ts
som-sm: Yeah, I agree! 
And this is probably a bug, refer [this comment](https://github.com/microsoft/TypeScript/issues/60815#issuecomment-2553605416).

test-d/keys-of-union.ts
sindresorhus: Ok, maybe just add your tests, and then add `type Assignability5<T extends UnknownRecord, _K extends keyof T> = unknown;
type Test5<T extends UnknownRecord> = Assignability5<T, KeysOfUnion<T>>; // Doesn't error` commented out with a comment that says that they don't produce error even though they should.

som-sm: Yup makes sense, added all the above mentioned cases and some more, refer [2629884](https://github.com/sindresorhus/type-fest/pull/1009/commits/26298846cd99341d4772a665daa6a031d0746cac).

Issue #450, opened 2022-08-29
v5


sindresorhus:
- [x] Target TypeScript 5.9
- [x] https://github.com/sindresorhus/type-fest/issues/1143
- [x] `ArrayTail`: Make `preserveReadonly` option true by default.
- [x] Move to ESM
- [x] Target Node.js 20
- [x] Make the `preserveConsecutiveUppercase` option in `CamelCase` `false` by default
- [x] Remove this: https://github.com/sindresorhus/type-fest/commit/32d94ddf02dba559655a9722bf8c481d4045e766#diff-9589c7ca28e78a91ac6a3982dce6a6be80734e70af0bcea4092ecaebc04190ffR1-R3
- [x] Make `PartialDeep` `allowUndefinedInNonTupleArrays` option `false` by default.
- [x] Make `Split` `strictLiteralChecks` option `true` by default.
- [x] https://github.com/sindresorhus/type-fest/pull/1029
- [x] https://github.com/sindresorhus/type-fest/pull/875
- [x] `Schema`: Remove `undefined` from `recurseIntoArrays` option.
- [x] Add a constraint of `object` to `T` for `WritableKeysOf` and `ReadonlyKeysOf`
- [x] Remove `strict: false` workaround: https://github.com/sindresorhus/type-fest/pull/1096
- [x] Move `ObservableLike` type to separate export since it mutates global: https://github.com/sindresorhus/type-fest/issues/1122#issuecomment-2858728080
- [x] Decide on If/Is types.

voxpelli:
Possibly: https://github.com/sindresorhus/type-fest/issues/863

sindresorhus:
@som-sm @Emiyaaaaa I'm planning v5 for right after TypeScript 5.9 is out. Probably 2-3 months out.

Please discuss here if there are any other breaking changes we should do that are not mentioned in the issue body.

som-sm:
@sindresorhus Do we want to update anything with the `If*` types, as suggested in https://github.com/sindresorhus/type-fest/issues/788#issuecomment-2408420468?

<br>

> I would deprecate the Is* and refer people to the If* (https://github.com/sindresorhus/type-fest/issues/788#issuecomment-2411093782)

I feel more inclined towards this, but `Is*` variants are likely much more common in codebases, so deprecating them could be annoying.

filipw01:
[...]

PR #1014, opened 2024-12-19
`SetOptional`/`SetRequired`/`SetReadonly`: Fix instantiations with index signatures

### LINKED ISSUE #1013: `SetOptional` remove properties of types with index signatures
### Bug description

After #1007, `SetOptional` remove properties of types with index signatures.

v4.30.1 works fine.

image

### Repro

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBDAnmApnA3nAyimB5MGYCAOwEMAbOAXzgDMoIQ4ByJVAWjpQGcYWA3ACghwEjBRQ6ZAMZoAYhAgYhcNfSUAuOHyhiA5sPVwA2gA9tugwF1tAVxIBrEhADuJYdRHs0BOAF5sXAIiUkoAHkUIABpWOiUWAD4jOAB6VLgAPQB+ETV0uABRM1QZCQATbR84P0DMeIhsyxg9En04AB84B3KUOjEUcoFTCx0Wm3snF3caISA

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/1013">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/1013/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

som-sm:
Fixes #1013

- Tweaked the implementation of `DistributedPick` so that it handles `any` properly
- `Set-*` types now use `DistributedPick` instead of `Except`/`Exclude` combination

### DIFF
--- a/source/distributed-pick.d.ts
+++ b/source/distributed-pick.d.ts
@@ -80,6 +80,4 @@ if (pickedUnion.discriminant === 'A') {
 @category Object
 */
 export type DistributedPick<ObjectType, KeyType extends KeysOfUnion<ObjectType>> =
-	ObjectType extends unknown
-		? Pick<ObjectType, Extract<KeyType, keyof ObjectType>>
-		: never;
+	{[Key in keyof ObjectType as Extract<Key, KeyType>]: ObjectType[Key]};
--- a/source/distributed-pick.d.ts
+++ b/source/distributed-pick.d.ts
@@ -80,4 +80,6 @@ if (pickedUnion.discriminant === 'A') {
 @category Object
 */
 export type DistributedPick<ObjectType, KeyType extends KeysOfUnion<ObjectType>> =
-	{[Key in keyof ObjectType as Extract<Key, KeyType>]: ObjectType[Key]};
+	ObjectType extends unknown
+		? Pick<ObjectType, Extract<KeyType, keyof ObjectType>>
+		: never;
--- a/source/set-optional.d.ts
+++ b/source/set-optional.d.ts
@@ -1,4 +1,6 @@
+import type {DistributedPick} from './distributed-pick';
 import type {Except} from './except';
+import type {KeysOfUnion} from './keys-of-union';
 import type {Simplify} from './simplify';
 
 /**
@@ -32,6 +34,6 @@ export type SetOptional<BaseType, Keys extends keyof BaseType> =
 		// Pick just the keys that are readonly from the base type.
 		Except<BaseType, Keys> &
 		// Pick the keys that should be mutable from the base type and make them mutable.
-		Partial<Except<BaseType, Exclude<keyof BaseType, Keys>>>
+		Partial<DistributedPick<BaseType, Keys & KeysOfUnion<BaseType>>>
 		>
 		: never;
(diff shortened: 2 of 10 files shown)


source/distributed-pick.d.ts
```diff
@@ -80,6 +80,4 @@ if (pickedUnion.discriminant === 'A') {
 @category Object
 */
 export type DistributedPick<ObjectType, KeyType extends KeysOfUnion<ObjectType>> =
-	ObjectType extends unknown
```
som-sm, line 83: This conditional is no longer required, because the implementation now uses a homomorphic mapped types, which already distributes over union types.

source/set-optional.d.ts
```diff
@@ -32,6 +34,6 @@ export type SetOptional<BaseType, Keys extends keyof BaseType> =
 		// Pick just the keys that are readonly from the base type.
 		Except<BaseType, Keys> &
 		// Pick the keys that should be mutable from the base type and make them mutable.
-		Partial<Except<BaseType, Exclude<keyof BaseType, Keys>>>
+		Partial<DistributedPick<BaseType, Keys & KeysOfUnion<BaseType>>>
```
som-sm, line 37: Would not require intersection with `KeysOfUnion<BaseType>` once #1009 is finalised.

cc: @Emiyaaaaa

som-sm:
**Note:** The updated implementation of `DistributedPick` introduces subtle changes in it's behaviour, which might be problematic if the goal is to keep `DistributedPick` fully aligned with the built-in `Pick` for non-union types.

#### 1. Behaviour with primitives
```ts
// Previous behaviour
type Test = Pick<string, 'toUpperCase'>;
//   ^? type Test = { toUpperCase: () => string }
```
```ts
// Updated behaviour
type Test = Pick<string, 'toUpperCase'>;
//   ^? type Test = string
```
And this happens because homomorphic mapped types pass through primitives unchanged.

#### 2. Picking `number` from `string` indexor
```ts
// Previous behaviour
type Test = Pick<{ [k: string]: unknown }, number>;
//   ^? type Test = { [x: number]: unknown; }
```

```ts
// Updated behaviour
type Test = Pick<{ [k: string]: number }, number>;
//   ^? type Test = {}
```

<br>

@sindresorhus If you think this might create issues, then I'd create a separate internal `Pick` type instead of modifying `DistributedPick`.

sindresorhus:
I think it would be better to create an internal pick type for now.

Do you think the new behavior is the correct one / better? If so, we could consider changing it in v5: https://github.com/sindresorhus/type-fest/issues/450

Alternatively, if both the old and new behavior makes sense in different situations, we could consider adding an option to the type.

som-sm:
It’s hard to determine which behaviour is better, as these are very rare scenarios.

While we could add an option for this, I doubt it'd be very useful—it might just create confusion. Currently, `DistributedPick` does what it says, it distributes `Pick`, but introducing additional changes might not be that obvious.

For now, I'll create an internal `Pick`.

sindresorhus:
@som-sm Nice work! You have done a lot of great work here lately. Would you be interested in being added as a maintainer? No worries if not :)

som-sm:
@sindresorhus Yeah sure, I’d be happy to help as a maintainer and contribute further.

sindresorhus:
Awesome 🎉

Issue #1016, opened 2024-12-21
SetFieldType bug


salisbury-espinosa:
### Bug description

the optionality of fields is not used in any way as a result

```typescript
import type {SetFieldType} from 'type-fest';

type MyModel = {
	id: number;
	updatedAt?: Date;
};

type MyModelApi = SetFieldType<MyModel,  'updatedAt', string>;
// {
// 	id: number;
// 	updatedAt: string;
// }
```

`updatedAt**?**: Date` => `updatedAt: string;`

PR #1017, opened 2024-12-21
`SetFieldType`: Add option to preserve property modifiers

som-sm:
Fixes: #1016 

Adds a new `preservePropertyModifiers` option.

I've set the default value of this option to `true`, which means it'll regress from the existing behaviour, but that should be fine if the current behaviour is considered a bug.

### DIFF
--- a/source/set-field-type.d.ts
+++ b/source/set-field-type.d.ts
@@ -1,5 +1,15 @@
-import type {Except} from './except';
 import type {Simplify} from './simplify';
+
+type SetFieldTypeOptions = {
+	/**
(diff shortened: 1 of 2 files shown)


som-sm:
@sindresorhus Noticed there's some room for improvement with index signatures as well. I'll create a separate PR for that because that'll require more complicated changes, so it's better to keep those changes separate.

som-sm:
@sindresorhus Noticed there's some room for improvement with index signatures as well. I'll create a separate PR for that because that'll require more complicated changes, so it's better to keep those changes separate.

PR #1019, opened 2024-12-23
`PartialDeep`: Add `allowUndefinedInArrays` option

PicoI2:
Add option `allowUndefinedInArrays` to `PartialDeep`, default to true to respect backward compatibility.

PartialDeep with recurseIntoArrays option works like this
```
import type {PartialDeep} from 'type-fest';

interface Settings {
	languages: string[];
}

const partialSettings: PartialDeep<Settings, {recurseIntoArrays: true}> = {
	languages: [undefined]
};
```
But it can be annoying that `undefined` type could be allowed in an array of string.
So with this new option `allowUndefinedInArrays: false`, the `languages` array is still optional, but type `undefined` is not allowed anymore into the `languages` array.

### DIFF
--- a/source/partial-deep.d.ts
+++ b/source/partial-deep.d.ts
@@ -12,11 +12,14 @@ export type PartialDeepOptions = {
 	readonly recurseIntoArrays?: boolean;
 
 	/**
-	Allows undefined values in non-tuple arrays. When set to `true`, elements of non-tuple arrays can be `undefined`. When set to `false`, only explicitly defined elements are allowed in non-tuple arrays, ensuring stricter type checking. Defaults to `true` if not specified.
+	Allows `undefined` values in non-tuple arrays.
+
+	- When set to `true`, elements of non-tuple arrays can be `undefined`.
+	- When set to `false`, only explicitly defined elements are allowed in non-tuple arrays, ensuring stricter type checking.
 
 	@default true
 	@example
-	You can prevent undefined values in non tuple arrays by passing `{recurseIntoArrays: true; allowUndefinedInNonTupleArrays: false}` as the second type argument:
+	You can prevent `undefined` values in non tuple arrays by passing `{recurseIntoArrays: true; allowUndefinedInNonTupleArrays: false}` as the second type argument:
 
 	```
 	interface Settings {
--- a/source/partial-deep.d.ts
+++ b/source/partial-deep.d.ts
@@ -18,10 +18,13 @@ export type PartialDeepOptions = {
 	- When set to `false`, only explicitly defined elements are allowed in non-tuple arrays, ensuring stricter type checking.
 
 	@default true
+
 	@example
-	You can prevent `undefined` values in non tuple arrays by passing `{recurseIntoArrays: true; allowUndefinedInNonTupleArrays: false}` as the second type argument:
+	You can prevent `undefined` values in non-tuple arrays by passing `{recurseIntoArrays: true; allowUndefinedInNonTupleArrays: false}` as the second type argument:
 
 	```
+	import type {PartialDeep} from 'type-fest';
+
 	type Settings = {
 		languages: string[];
 	};
@@ -48,12 +51,12 @@ import type {PartialDeep} from 'type-fest';
 
 const settings: Settings = {
 	textEditor: {
-		fontSize: 14;
-		fontColor: '#000000';
-		fontWeight: 400;
-	}
-	autocomplete: false;
-	autosave: true;
+		fontSize: 14,
+		fontColor: '#000000',
+		fontWeight: 400
+	},
+	autocomplete: false,
+	autosave: true
 };
 
 const applySavedSettings = (savedSettings: PartialDeep<Settings>) => {
@@ -68,7 +71,7 @@ By default, this does not affect elements in array and tuple types. You can chan
 ```
 import type {PartialDeep} from 'type-fest';
 
-interface Settings {
+type Settings = {
 	languages: string[];
 }
 
--- a/source/partial-deep.d.ts
+++ b/source/partial-deep.d.ts
@@ -1,7 +1,7 @@
 import type {BuiltIns} from './internal';
 
 /**
-@see PartialDeep
+@see {@link PartialDeep}
 */
 export type PartialDeepOptions = {
 	/**
@@ -73,6 +73,8 @@ const partialSettings: PartialDeep<Settings, {recurseIntoArrays: true}> = {
 	languages: [undefined]
 };
 ```
+
+@see {@link PartialDeepOptions}
 
 @category Object
 @category Array
--- a/source/partial-deep.d.ts
+++ b/source/partial-deep.d.ts
@@ -12,11 +12,11 @@ export type PartialDeepOptions = {
 	readonly recurseIntoArrays?: boolean;
 
 	/**
-	Whether to authorize undefined values in arrays.
+    Allows undefined values in non-tuple arrays. When set to `true`, elements of non-tuple arrays can be `undefined`. When set to `false`, only explicitly defined elements are allowed in non-tuple arrays, ensuring stricter type checking. Defaults to `true` if not specified.
 
-	@default true
+    @default true
 	*/
-	readonly allowUndefinedInArrays?: boolean;
+	readonly allowUndefinedInNonTupleArrays?: boolean;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


som-sm:
Will this create confusion with tuples, because `allowUndefinedInArrays` wouldn't do anything with tuples. 

For eg,

```ts

type Test = PartialDeep<
  // ^? { a?: [string?] }
  {a: [string]},
  {recurseIntoArrays: true; allowUndefinedInArrays: false}
>;

const test: Test = {a: [undefined]};
```

In the example above, `undefined` is allowed even though `allowUndefinedInArrays` is set to `false`. And this is because `allowUndefinedInArrays` doesn't seem to do anything for tuples. For tuples, allowance of `undefined` is decided by `exactOptionalPropertyTypes` setting.

@PicoI2 WDYT?

som-sm:
IMO, there's no concept of optionality with non-tuple arrays. A `string[]` is also in a way optional, because it either holds strings or nothing, so `(string | undefined)[]` isn't IMO the optional version of `string[]`.

Although, adding the optional modifier (`?`) in a mapped type and instantiating it with a non-tuple array like `T[]`, does produce `(T | undefined)[]`, like `Partial<string[]>` gives `(string | undefined)[]`. But this feels incorrect to me unless there's a way to prevent array items from being `undefined` in `(string | undefined)[]`.

So, IMO, `allowUndefinedInArrays` in arrays should be the default behaviour.

PicoI2:
> ```ts
> type Test = PartialDeep<
>   // ^? { a?: [string?] }
>   {a: [string]},
>   {recurseIntoArrays: true; allowUndefinedInArrays: false}
> >;
> 
> const test: Test = {a: [undefined]};
> ```

You are right, I did not saw this case. But as the option is `allowUndefinedInArrays`and not `allowUndefinedInArraysAndTuples` it could give a hint. Or maybe we could found another name for the option to be more explicit.


> IMO, there's no concept of optionality with non-tuple arrays. A `string[]` is also in a way optional, because it either holds strings or nothing, so `(string | undefined)[]` isn't IMO the optional version of `string[]`.
> 
> Although, adding the optional modifier (`?`) in a mapped type and instantiating it with a non-tuple array like `T[]`, does produce `(T | undefined)[]`, like `Partial<string[]>` gives `(string | undefined)[]`, but this feels incorrect to me.
> 
> So, IMO, `allowUndefinedInArrays` in arrays should be the default behaviour.

I totally agree, but making `allowUndefinedInArrays` default value to false would be a breaking change, and nobody likes breaking changes. IMO, the default value should be true, until the release of the version 5 of type-fest, where it could be noticed in the migration guide.

som-sm:
> I totally agree, but making `allowUndefinedInArrays` default value to false would be a breaking change, and nobody likes breaking changes. IMO, the default value should be true, until the release of the version 5 of type-fest, where it could be noticed in the migration guide.

Yeah true, setting the default value to `false` would indeed be a breaking change. However, my rationale was that most users would likely be using the `recurseIntoArrays` option to make the items within the array optional, rather than adding `undefined` to the array itself, so it wouldn't be noticeable in majority of the cases.

But yeah, to be on the safer side we can keep it as an option, and in v5 we can either totally remove the option or flip the default value.

source/partial-deep.d.ts
```diff
@@ -10,6 +10,13 @@ export type PartialDeepOptions = {
 	@default false
 	*/
 	readonly recurseIntoArrays?: boolean;
+
+	/**
+	Whether to authorize undefined values in arrays and tuples.
```
som-sm, line 15:
```suggestion
	Whether to authorize undefined values in arrays.
```

som-sm (changes requested):

som-sm (approved):

source/partial-deep.d.ts
```diff
@@ -10,6 +10,13 @@ export type PartialDeepOptions = {
 	@default false
 	*/
 	readonly recurseIntoArrays?: boolean;
+
+	/**
+	Whether to authorize undefined values in arrays.
+
+	@default true
+	*/
+	readonly allowUndefinedInArrays?: boolean;
```
sindresorhus, line 19: Should be `allowUndefinedInNonTupleArrays` since it does not apply to tuples and the above option using `*array` in the name does.

source/partial-deep.d.ts
```diff
@@ -10,6 +10,13 @@ export type PartialDeepOptions = {
 	@default false
 	*/
 	readonly recurseIntoArrays?: boolean;
+
+	/**
+	Whether to authorize undefined values in arrays.
```
sindresorhus, line 15: "authorize" is not the correct word here. Also needs more extensive explanation of how it works.

source/partial-deep.d.ts
```diff
@@ -54,6 +61,21 @@ const partialSettings: PartialDeep<Settings, {recurseIntoArrays: true}> = {
...
+interface Settings {
+	languages: string[];
+}
+
+const partialSettings: PartialDeep<Settings, {recurseIntoArrays: true; allowUndefinedInArrays: false}> = {
+	languages: [undefined] 	// Error
+	languages: []			// OK
+};
+```
+
```
sindresorhus, line 78: Should should be moved to the docs for the option.

sindresorhus (changes requested):

source/partial-deep.d.ts
PicoI2: Yes you are right

source/partial-deep.d.ts
PicoI2: Done, I hope the new explanation will be OK, do not hesitate to modify if you think you could provide a better one.

source/partial-deep.d.ts
PicoI2: I move the example in the option documentation but I am not sure this was what you expected.
Moreover, in vs-code, when I put my cursor over the `PartialDeep` type, I see the documentation and example, but I do not see what is inside the `PartialDeepOptions` type, and so I do not see my example anymore.
I have done it in a separate commit to be able to rollback it if needed.

som-sm: @PicoI2 We link the option to the type, so the option is reachable from the type and the type is reachable from the option.

Refer [this commit](https://github.com/sindresorhus/type-fest/pull/1019/commits/bb180f4faba7a907c90cb47df360ddfde1acc56a).

source/partial-deep.d.ts
```diff
@@ -10,6 +10,26 @@ export type PartialDeepOptions = {
 	@default false
 	*/
 	readonly recurseIntoArrays?: boolean;
+
+	/**
+	Allows undefined values in non-tuple arrays. When set to `true`, elements of non-tuple arrays can be `undefined`. When set to `false`, only explicitly defined elements are allowed in non-tuple arrays, ensuring stricter type checking. Defaults to `true` if not specified.
```
som-sm, line 15: Would be more readable if you change the `true`/`false` statements into points. Also, the default statement could be removed, since we already have `@default`.

```suggestion
	Allows `undefined` values in non-tuple arrays.

	- When set to `true`, elements of non-tuple arrays can be `undefined`.
	- When set to `false`, only explicitly defined elements are allowed in non-tuple arrays, ensuring stricter type checking.

	@default true
	@example
	You can prevent `undefined` values in non tuple arrays by passing `{recurseIntoArrays: true; allowUndefinedInNonTupleArrays: false}` as the second type argument:
```

som-sm (changes requested):

source/partial-deep.d.ts
```diff
@@ -10,6 +10,26 @@ export type PartialDeepOptions = {
...
+
+	@default true
+	@example
+	You can prevent undefined values in non tuple arrays by passing `{recurseIntoArrays: true; allowUndefinedInNonTupleArrays: false}` as the second type argument:
```
som-sm, line 19: Would be more readable if you change the `true`/`false` statements into points. Also, the default statement could be removed, since we already have `@default`.

```suggestion
	Allows `undefined` values in non-tuple arrays.

	- When set to `true`, elements of non-tuple arrays can be `undefined`.
	- When set to `false`, only explicitly defined elements are allowed in non-tuple arrays, ensuring stricter type checking.

	@default true
	@example
	You can prevent `undefined` values in non tuple arrays by passing `{recurseIntoArrays: true; allowUndefinedInNonTupleArrays: false}` as the second type argument:
```

source/partial-deep.d.ts
```diff
@@ -10,6 +10,26 @@ export type PartialDeepOptions = {
...
+
+	```
+	interface Settings {
+		languages: string[];
+	}
+
+	const partialSettings: PartialDeep<Settings, {recurseIntoArrays: true; allowUndefinedInNonTupleArrays: false}> = {
+		languages: [undefined]	// Error
+		languages: []		// OK
+	};
```
som-sm, line 29: The example is syntactically incorrect, you can't set `languages` twice. 

Better to update to this:

```suggestion
	type Settings = {
		languages: string[];
	};

	declare const partialSettings: PartialDeep<Settings, {recurseIntoArrays: true; allowUndefinedInNonTupleArrays: false}>;

	partialSettings.languages = [undefined]; // Error
	partialSettings.languages = []; // Ok
```

PR #1024, opened 2024-12-30
Add `IsTuple` type

### LINKED ISSUE #1022: Proposal: `IsTuple<Type>` & `IsFixedLengthTuple<Type>`

I’m surprised that such a type hasn’t been added to type-fest yet. Handling tuple types is quite common in advanced type definitions, and patterns like `number extends T['length']` have already appeared multiple times in this project.

I propose adding two types for checking tuple characteristics: `IsTuple<Type>` and `IsFixedLengthTuple<Type>`. The former determines whether a type is a tuple, including those with rest elements, while the latter is potentially more useful. It checks whether a tuple is fixed-length by verifying if its `length` property is a numeric literal, excluding tuples with rest elements.

### Type source

I’ve attempted to implement these two types, but their type constraints and naming might still need further discussion. Perhaps these types would be better suited to take `ArrayType extends readonly unknown[]` as input. For `IsFixedLengthTuple`, there might also be other names that are more concise yet precise.

~~~typescript
/**
Returns a boolean for whether the given type is a tuple.

Note that a tuple can include rest elements. If you need to check whether a type is a fixed-length tuple (i.e., a tuple without rest elements but possibly containing optional elements), use {@link IsFixedLengthTuple} instead.

@example
```
import type {IsTuple} from 'type-fest';

type Simple = IsTuple<[string, number, boolean]>;
// => true

type Readonly = IsTuple<readonly [number, boolean]>;
// => true

type WithRest = IsTuple<[string, number, ...boolean[]]>;
// => true

type WithOptional1 = IsTuple<[string, number?]>;
// => true

type WithOptional2 = IsTuple<[foo: string, bar?: number]>;
// => true

type WithBoth1 = IsTuple<[string, number?, ...boolean[]]>;
// => true

type WithBoth2 = IsTuple<[string?, ...string[]]>;
// => true

type NonTuple = IsTuple<number[]>;
// => false
```

@category Type Guard
@category Utilities
*/
type IsTuple<Type> = Type extends unknown[]
	? IsEqual<Type, Type[number][]> extends true
		? false
		: true
	: Type extends readonly unknown[]
		? IsEqual<Type, readonly Type[number][]> extends true
			? false
			: true
		: false;

/**
Returns a boolean for whether the given type is a tuple with a fixed number of elements. 

A fixed-length tuple may include optional elements but does not include rest elements.

@example
```
import type {IsFixedLengthTuple} from 'type-fest';

type Simple = IsFixedLengthTuple<[string, number, boolean]>;
// => true

type Readonly = IsFixedLengthTuple<readonly [number, boolean]>;
// => true

type WithOptional1 = IsFixedLengthTuple<[string, number?]>;
// => true

type WithOptional2 = IsFixedLengthTuple<[foo: string, bar?: number]>;
// => true

type NonTuple = IsFixedLengthTuple<number[]>;
// => false

type NonFixedLength = IsFixedLengthTuple<[string, ...boolean[]]>;
// => false
```

@category Type Guard
@category Utilities
*/
type IsFixedLengthTuple<Type> = Type extends readonly unknown[]
	? number extends Type['length']
		? Type['length'] extends number
			? false // `any`
			: true
		: true
	: false;
~~~

- [X] I tried my best to look for it

som-sm:
Fixes #1022

### DIFF
--- a/source/is-tuple.d.ts
+++ b/source/is-tuple.d.ts
@@ -9,20 +9,20 @@ export type IsTupleOptions = {
 	/**
 	Consider only fixed length arrays as tuples.
 
-	- When set to `true`, arrays with rest elements (e.g., `[1, ...number[]]`) are _not_ considered as tuples.
-	- When set to `false` (default), arrays with at least one non-rest element (e.g., `[1, ...number[]]`) are considered as tuples.
+	- When set to `true` (default), arrays with rest elements (e.g., `[1, ...number[]]`) are _not_ considered as tuples.
+	- When set to `false`, arrays with at least one non-rest element (e.g., `[1, ...number[]]`) are considered as tuples.
 
-	@default false
+	@default true
 
 	@example
 	```ts
 	import type {IsTuple} from 'type-fest';
 
-	type Example1 = IsTuple<[number, ...number[]], {fixedLengthOnly: false}>;
-	//=> true
-
-	type Example2 = IsTuple<[number, ...number[]], {fixedLengthOnly: true}>;
+	type Example1 = IsTuple<[number, ...number[]], {fixedLengthOnly: true}>;
 	//=> false
+
+	type Example2 = IsTuple<[number, ...number[]], {fixedLengthOnly: false}>;
+	//=> true
 	```
 	*/
 	fixedLengthOnly?: boolean;
@@ -49,11 +49,11 @@ type NotTuple = IsTuple<number[]>;
 type TupleWithOptionalItems = IsTuple<[1?, 2?]>;
 //=> true
 
-type RestItemsAllowed = IsTuple<[1, 2, ...number[]]>;
-//=> true
-
-type RestItemsNotAllowed = IsTuple<[1, 2, ...number[]], {fixedLengthOnly: true}>;
+type RestItemsNotAllowed = IsTuple<[1, 2, ...number[]]>;
 //=> false
+
+type RestItemsAllowed = IsTuple<[1, 2, ...number[]], {fixedLengthOnly: false}>;
+//=> true
 ```
 
 @see {@link IsTupleOptions}
@@ -63,16 +63,16 @@ type RestItemsNotAllowed = IsTuple<[1, 2, ...number[]], {fixedLengthOnly: true}>
 */
 export type IsTuple<
 	TArray extends UnknownArray,
-	Options extends IsTupleOptions = {fixedLengthOnly: false},
+	Options extends IsTupleOptions = {fixedLengthOnly: true},
 > =
 	IfAny<TArray, boolean, IfNever<TArray, false,
 	TArray extends unknown // For distributing `TArray`
 		? number extends TArray['length']
-			? Options['fixedLengthOnly'] extends true
-				? false
-				: IfNever<keyof TArray & `${number}`,
+			? Options['fixedLengthOnly'] extends false
+				? IfNever<keyof TArray & `${number}`,
 				TArray extends readonly [...any, any] ? true : false, // To handle cases where a non-rest element follows a rest element, e.g., `[...number[], number]`
 				true>
+				: false
 			: true
 		: false
 	>>;
--- a/source/is-tuple.d.ts
+++ b/source/is-tuple.d.ts
@@ -1,0 +1,65 @@
+import type {IfAny} from './if-any';
+import type {IfNever} from './if-never';
+import type {UnknownArray} from './unknown-array';
+
+type IsTupleOptions = {
+	/**
+	Consider only fixed length arrays as tuples.
+
+	- When set to `true`, arrays with rest elements (e.g., `[1, ...number[]]`) are _not_ considered as tuples.
+	- When set to `false` (default), arrays with at least one non-rest element (e.g., `[1, ...number[]]`) are considered as tuples.
+
+	@example
+	```ts
+	type Example1 = IsTuple<[number, ...number[]], {includeFixedLengthTuplesOnly: false}>;
+	//=> true
+
+	type Example2 = IsTuple<[number, ...number[]], {includeFixedLengthTuplesOnly: true}>;
+	//=> false
+	```
+
+	@default false
+	*/
+	includeFixedLengthTuplesOnly?: boolean;
+};
+
+/**
+Returns a boolean for whether the given array is a tuple.
+
+Use-case:
+- If you want to make a conditional branch based on the result of whether an array is a tuple or not.
+
+@example
+```ts
+type Tuple = IsTuple<[1, 2, 3]>;
+// => true
+
+type NotTuple = IsTuple<number[]>;
+// => false
+
+type TupleWithOptionalItems = IsTuple<[1?, 2?]>;
+// => true
+
+type RestItemsAllowed = IsTuple<[1, 2, ...number[]]>;
+// => true
+
+type RestItemsNotAllowed = IsTuple<[1, 2, ...number[]], {includeFixedLengthTuplesOnly: true}>;
+// => false
+```
+
+@category Type Guard
+@category Utilities
+*/
+export type IsTuple<
+	TArray extends UnknownArray,
+	Options extends IsTupleOptions = {includeFixedLengthTuplesOnly: false},
+> =
+	IfAny<TArray, boolean, IfNever<TArray, false,
+	TArray extends unknown // For distributing `TArray`
+		? number extends TArray['length']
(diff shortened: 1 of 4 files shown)


test-d/is-tuple.ts
```diff
@@ -0,0 +1,47 @@
...
+
+// Tuples
+expectType<IsTuple<[]>>(true);
+expectType<IsTuple<[1]>>(true);
+expectType<IsTuple<[1, 2]>>(true);
+expectType<IsTuple<[1, 2, ...number[]]>>(true);
+expectType<IsTuple<[1?]>>(true);
+expectType<IsTuple<[1?, 2?]>>(true);
+expectType<IsTuple<[1?, 2?, ...number[]]>>(true);
+expectType<IsTuple<[never]>>(true);
```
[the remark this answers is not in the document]
sindresorhus: Actually some more:

```typescript
// Labeled tuples
expectType<IsTuple<[x: string, y: number]>>(true);
expectType<IsTuple<[first: string, ...rest: number[]]>>(true);
expectType<IsTuple<readonly [name: string, age?: number]>>(true);

// Mixed optional/required elements
expectType<IsTuple<[string, number?]>>(true);
expectType<IsTuple<[string?, number]>>(true);
expectType<IsTuple<readonly [string, number?, ...boolean[]]>>(true);

// Complex unions
expectType<IsTuple<[string] | [number] | number[]>>({} as boolean);
expectType<IsTuple<[string, ...number[]] | [number?, string?]>>(true);
expectType<IsTuple<[] | [number] | readonly [string]>>(true);
```

source/is-tuple.d.ts
```diff
@@ -0,0 +1,71 @@
+import type {IfAny} from './if-any';
+import type {IfNever} from './if-never';
+import type {UnknownArray} from './unknown-array';
+
+type IsTupleOptions = {
```
sindresorhus, line 8: Should be exported.

sindresorhus (changes requested):

source/is-tuple.d.ts
```diff
@@ -0,0 +1,67 @@
...
+	type Example1 = IsTuple<[number, ...number[]], {includeFixedLengthTuplesOnly: false}>;
+	//=> true
+
+	type Example2 = IsTuple<[number, ...number[]], {includeFixedLengthTuplesOnly: true}>;
+	//=> false
+	```
+
+	@default false
+	*/
+	includeFixedLengthTuplesOnly?: boolean;
```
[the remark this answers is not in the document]
som-sm: Yeah makes sense, the option name can be simplified since the documentation is quite clear.

source/is-tuple.d.ts
som-sm: Addressed in [da2888](https://github.com/sindresorhus/type-fest/pull/1024/commits/da28883e7d74799b7d4caabc37f08c36ffe72c05).

test-d/is-tuple.ts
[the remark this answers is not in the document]
som-sm: Added in [797116f](https://github.com/sindresorhus/type-fest/pull/1024/commits/797116ffabe546fca28c099bb4252796887d956b). Had to remove the second test case in `Mixed optional/required elements`, because the tuple `[string?, number]` isn't valid (`A required element cannot follow an optional element`).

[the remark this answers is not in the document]
som-sm: Also, replaced all `1`s and `2`s with `number`s and `string`s, just to make the tests a bit more diverse.

Refer [fc451bf](https://github.com/sindresorhus/type-fest/pull/1024/commits/fc451bfe0dae2e3e131a695c9e469b2641626ad8).

source/is-tuple.d.ts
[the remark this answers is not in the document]
som-sm: Done in [3a8ea04](https://github.com/sindresorhus/type-fest/pull/1024/commits/3a8ea0477bdb415602931c2ae41dabee34c48f68).

test-d/is-tuple.ts
```diff
@@ -0,0 +1,59 @@
...
+expectType<IsTuple<readonly [number]>>(true);
+expectType<IsTuple<readonly [number, string, ...number[]]>>(true);
+expectType<IsTuple<readonly [number?, string?, ...number[]]>>(true);
+expectType<IsTuple<readonly [number?]>>(true);
+expectType<IsTuple<readonly [never]>>(true);
+
+// Non-tuples
+expectType<IsTuple<number[]>>(false);
+expectType<IsTuple<readonly number[]>>(false);
+expectType<IsTuple<[...number[]]>>(false);
```
Emiyaaaaa, line 25: What's these expected results?
```ts
IsTuple<[...number[], string]
IsTuple<[...number[], number]
IsTuple<[number, string, ...number[]]
IsTuple<[number, ...number[]]
```

som-sm: @Emiyaaaaa Interesting!

The bottom two cases are tuples when `fixedLengthOnly` is `false`, and they are not otherwise. And there's already test cases for these.

The top two cases seem like they should be tuples, but the current implementation doesn't handle these. Let me find a fix!

Emiyaaaaa: Wait, what's these expected results?
```
IsTuple<number[]>
IsTuple<[...number[]]
IsTuple<[...number[], number]
```

Emiyaaaaa: > Wait, what's these expected results?
> 
> ```ts
> IsTuple<number[]>
> IsTuple<[...number[]]
> IsTuple<[...number[], number]
> ```


Maybe all non-fixed-length array sould not be tuple?

som-sm: The first two should always be `false`. 
The third one should be a tuple when `fixedLengthOnly` is `false`.

Emiyaaaaa: Meybe all non-fixed-length array should not be tuple? The definition of 'tuple' is an immutable array.

som-sm: There's an option to control this behaviour. In TS, there might be cases where distinguishing between `number[]` and `[number, ...number[]]` makes sense.

Maybe we can the default behaviour though?

Emiyaaaaa: > there might be cases where distinguishing between `number[]` and `[number, ...number[]]` would make sense.

But why use `IsTuple` to distinguishing between `number[]` and `[number, ...number[]]` and not something else? look like confusion.

Some related type: [StaticPartOfArray](https://github.com/sindresorhus/type-fest/blob/3d546271332116970e5facde69b4447d2e64b45f/source/internal/array.d.ts#L42), [VariablePartOfArray](https://github.com/sindresorhus/type-fest/blob/3d546271332116970e5facde69b4447d2e64b45f/source/internal/array.d.ts#L61C13-L61C32)

> Maybe we can update the default behaviour though?
I agree

som-sm: > But why use IsTuple to distinguishing between number[] and [number, ...number[]] and not something else? look like confusion.

Umm...I meant just for distinguishing, like either or cases. True that [StaticPartOfArray](https://github.com/sindresorhus/type-fest/blob/3d546271332116970e5facde69b4447d2e64b45f/source/internal/array.d.ts#L42), [VariablePartOfArray](https://github.com/sindresorhus/type-fest/blob/3d546271332116970e5facde69b4447d2e64b45f/source/internal/array.d.ts#L61C13-L61C32) can also be used, but IMO it doesn't hurt to give that option with `IsTuple` as well.

WDTY? @sindresorhus @Snowflyt

som-sm: @Emiyaaaaa 
> The top two cases seem like they should be tuples, but the current implementation doesn't handle these. Let me find a fix!

Pushed a fix for this in [985ee21](https://github.com/sindresorhus/type-fest/pull/1024/commits/985ee21e35d47ec2850d11c2723b319376a3ef04), not the most elegant fix, but it works. WDYT?

Emiyaaaaa: great!

Emiyaaaaa: > but IMO it doesn't hurt to give that option with `IsTuple` as well.

we can set default options to true first.

som-sm: Updated the default in [dd630e6](https://github.com/sindresorhus/type-fest/pull/1024/commits/dd630e60fa112389a9fd0a25e3af48af56898d49).

PR #1031, opened 2025-01-03
`Get`: Optimize performance

Emiyaaaaa:
Limit suggestions to 'a.b.c' to optimize performance. That is enough for most cases.

### DIFF
--- a/source/get.d.ts
+++ b/source/get.d.ts
@@ -205,6 +205,6 @@ export type Get<
 	BaseType,
 	Path extends
 	| readonly string[]
-	| LiteralStringUnion<ToString<Paths<BaseType, {bracketNotation: false}> | Paths<BaseType, {bracketNotation: true}>>>,
+	| LiteralStringUnion<ToString<Paths<BaseType, {bracketNotation: false; maxRecursionDepth: 2}> | Paths<BaseType, {bracketNotation: true; maxRecursionDepth: 2}>>>,
 	Options extends GetOptions = {}> =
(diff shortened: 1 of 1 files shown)


marcomuser:
Even with this change it will remain configurable by passing in Paths generic with a higher maxRecursionDepth, right? I need a higher maxRecursionDepth than 2. :)

sindresorhus:
This also should be done in v5. And if we decide to do it, it must be possible to increase it. I'm just wondering whether the performance increase is worth the inconvenience of potentially not having all things work directly with `Get` without having to set an option.

som-sm:
> I'm just wondering whether the performance increase is worth the inconvenience of potentially not having all things work directly with Get without having to set an option.

@sindresorhus I think there's some confusion, everything will still continue to work, right? The second type argument still effectively accepts any `string`.

<br>

> And if we decide to do it, it must be possible to increase it.
Not sure, if we really need an option to control the autocomplete behaviour. The way it's used in the JSDoc example (which I feel is how this would mostly be used), the autocomplete is anyways not dependent on our code but on user's code.
```ts
// `Path` here decides the autocomplete behaviour
const get = <BaseType, Path extends string | readonly string[]>(object: BaseType, path: Path): Get<BaseType, Path> =>
	lodash.get(object, path);
```

sindresorhus:
Ah, I didn't realize this is only about auto-complete. Then it's totally fine.

Emiyaaaaa:
> Even with this change it will remain configurable by passing in Paths generic with a higher maxRecursionDepth, right? I need a higher maxRecursionDepth than 2. :)

sure

PR #1036, opened 2025-01-13
`StringSlice`: Fix return type when passing in non-literal

### LINKED ISSUE #1035: `StringSlice<S, Start, End>` returns incorrect type when S is non-literal string
### Bug description

```ts
import type {StringSlice} from 'type-fest';

type Slice = StringSlice<string, 0, 1>; // string[]
```

https://github.com/sindresorhus/type-fest/blob/cfb594777f5ede903c151a45c2db250b057902d7/source/string-slice.d.ts#L29-L37

Line 34 should be modified to replace `string[]` with `string`.

### Repro

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBDAnmApnA3gZRlYA7Ac0wBtgBjFAXzgDMoIQ4ByJVAWhpQGcYmBuALAAoYazQlyaALxxsuQhIoAeHvIIAaOAAZNARgB8goQEgA9KbgA9APzCzF1fgJxgXOCgAeqMjBQATTTwIeEdCAG0AXWEgA

ghaaj:
Closes #1035
This PR modifies the behavior of `StringSlice` so that it returns `string` instead of `string[]` when S is non-literal string.

### DIFF
--- a/source/string-slice.d.ts
+++ b/source/string-slice.d.ts
@@ -31,7 +31,7 @@ export type StringSlice<
 	Start extends number = 0,
 	End extends number = StringToArray<S>['length'],
 > = string extends S
-	? string[]
+	? string
 	: ArraySlice<StringToArray<S>, Start, End> extends infer R extends readonly string[]
 		? Join<R, ''>
 		: never;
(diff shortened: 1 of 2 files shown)


som-sm (approved):
Makes sense!

PR #1037, opened 2025-01-13
`SetRequiredDeep`: Fix handling of unions in nested keys

### LINKED ISSUE #1026: `SetRequiredDeep` doesn't remove optional modifiers from tuples
### Bug description

Refer to the note in [this comment](https://github.com/sindresorhus/type-fest/pull/939#issuecomment-2565090067).

```ts
import {SetRequiredDeep} from "type-fest";

type Test = SetRequiredDeep<{a: [string?]}, "a.0">; // doesn't remove the `?` modifier
//  actual: {a: [string?]} // returns the same thing
//  expected: {a: [string]}
```

### Repro

https://tsplay.dev/NajlBm

som-sm:
`SetRequiredDeep` currently fails to preserve unions in nested keys:
```ts
type Test = SetRequiredDeep<{ size?: { w: number; h: number } | { width: number; height: number } }, "size">;
//   ^? type Test = {
//        size:
//          | { w: number; h: number }
//          | { width: number; height: number }
//          | { w: number; h: number; width: number; height: number } // Extraneous entry
//          | { width: number; height: number; w: number; h: number } // // Extraneous entry
//      }
```
It should have simply removed the optional modifier from `size` and kept the union as is.

<br>

And this happens because `SimplifyDeep<{s?: {x: 1} | {w: 1}} & {s: {x: 1} | {w: 1}}>` doesn't simply remove the optional modifier from `s`, instead, it resolves to the following:
```ts
type Test = SimplifyDeep<{s?: {x: 1} | {w: 1}} & {s: {x: 1} | {w: 1}}>
//   ^? type Test = {
//        s: { x: 1 } | { x: 1; w: 1 } | { w: 1; x: 1 } | { w: 1 }
//      }
```
It seems to ignore the fact that the value of `s` is same in both the objects. Instead, it computes the intersection of `{x: 1} | {w: 1}` with itself and then expands it to `{x: 1} | {x: 1; w: 1} | {w: 1; x: 1} | {w: 1}`.

<hr>
<br>

I tried a couple of solutions but nothing worked, so I had to refactor the implementation of `SetRequiredDeep`. 

This implementation first converts the input keys into an array, then iterates over them and transforms the input object one path at a time.

IMO, this implementation is much easier to understand and manage. Plus, it automatically fixes #1026 since it uses `SetRequired` under the hood.

### DIFF
--- a/source/set-required-deep.d.ts
+++ b/source/set-required-deep.d.ts
@@ -1,4 +1,4 @@
-import type {IsAny} from 'expect-type';
+import type {IsAny} from './is-any';
 import type {NonRecursiveType, StringToNumber} from './internal';
 import type {Paths} from './paths';
 import type {SetRequired} from './set-required';
--- a/source/set-required-deep.d.ts
+++ b/source/set-required-deep.d.ts
@@ -32,6 +32,10 @@ type SomeRequiredDeep = SetRequiredDeep<Foo, 'a' | `c.${number}.d`>;
 // 		d: number // Is now required
 // 	}[]
 // }
+
+// Set specific indices in an array to be required.
+type ArrayExample = SetRequiredDeep<{a: [number?, number?, number?]}, 'a.0' | 'a.1'>;
+//=> {a: [number, number, number?]}
 ```
 
 @category Object
--- a/source/set-required-deep.d.ts
+++ b/source/set-required-deep.d.ts
@@ -1,6 +1,10 @@
+import type {IsAny} from 'expect-type';
 import type {NonRecursiveType, StringToNumber} from './internal';
 import type {Paths} from './paths';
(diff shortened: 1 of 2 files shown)


source/set-required-deep.d.ts
```diff
@@ -1,6 +1,10 @@
+import type {IsAny} from 'expect-type';
```
sindresorhus, line 1:
```suggestion
import type {IsAny} from './is-any';
```

sindresorhus (changes requested):

source/set-required-deep.d.ts
som-sm: Done!

Issue #959, opened 2024-09-26
`OmitDeep` for array item


Lonli-Lokli:
Assuming 
```ts
type PanelGroup = {
id: string;
panels: Panel[]
}

type Panel = {
panelId: string;
titles: string[]
}
```

There is not way to create a Panel Group type, where panels do not have titles

## Upvote & Fund

- We're using [Polar.sh](https://polar.sh/sindresorhus) so you can upvote and help fund this issue.
- The funding will be given to active contributors.
- Thank you in advance for helping prioritize & fund our backlog.

<a href="https://polar.sh/sindresorhus/type-fest/issues/959">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/959/pledge.svg?darkmode=1">
  <img alt="Fund with Polar" src="https://polar.sh/api/github/sindresorhus/type-fest/issues/959/pledge.svg">
</picture>
</a>

suchak-dwivedi:
@som-sm How does one omits multiple fields from nested array elements?

Tried below which is not working
```typescript
type Info = {
      a: string
      arr: {i: string, useless: string, useless2: string}[]
}
type Derived = OmitDeep<Info, `arr.${number}.useless` | `arr.${number}.useless2`>
const value: Derived = {a: 'a', arr: [{i: 'i'}]}
```

Only way I could make it work is like below, which is not readable in my opinion

```typescript
type Derived = OmitDeep<OmitDeep<Info, `arr.${number}.useless`>, `arr.${number}.useless2`>
const value: Derived = {a: 'a', arr: [{i: 'i'}]}
```

Is there something I am missing?

som-sm:
> How does one omits multiple fields from nested array elements?

@suchak-dwivedi This was not working as expected, I've opened up a PR that would fix it.

PR #1049, opened 2025-02-07
`OmitDeep`: Fix removal of multiple paths within arrays

som-sm:
Fixes https://github.com/sindresorhus/type-fest/issues/959#issuecomment-2643290328

Just enabling `recurseIntoArrays` for `SharedUnionDeep` within `OmitDeep` fixes the above issue, however, it causes some existing test cases to fail, as `SharedUnionDeep` currently fails when instantiated with arrays where the non-fixed part precedes the fixed part (will open an issue for this).
```ts
import { SharedUnionFieldsDeep } from "type-fest";

type Test = SharedUnionFieldsDeep<{a: [...{a: 1}[], {b: 1}]}, {recurseIntoArrays: true}>
//   ^? type Test = { a: {}[]; }
```

So, this PR removes usage of `SharedUnionDeep` from `OmitDeep` and replaces it with logic similar to what is used in `SetRequiredDeep`.
https://github.com/sindresorhus/type-fest/blob/81a05404c6c60583ff3dfcc0e4b992c62e052626/source/set-required-deep.d.ts#L47-L55

### DIFF
--- a/source/omit-deep.d.ts
+++ b/source/omit-deep.d.ts
@@ -1,10 +1,10 @@
+import type {UnionToTuple} from 'expect-type';
 import type {ArraySplice} from './array-splice';
 import type {ExactKey, IsArrayReadonly, NonRecursiveType, SetArrayAccess, ToString} from './internal';
 import type {IsEqual} from './is-equal';
 import type {IsNever} from './is-never';
 import type {LiteralUnion} from './literal-union';
 import type {Paths} from './paths';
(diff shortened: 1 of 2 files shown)


sindresorhus (approved):

viceice:
this is a breaking change. it depends on a dev dependency `expect-type` which isn't installed by default

viceice:
I see it's already fixed, sorry for the noise

Issue #1051, opened 2025-02-09
A test dependency is used in an exported code


borisvu:
### Bug description

In `omit-deep.d.ts` there is a new line that uses `expect-type` which is a devDependency.
```
import type {UnionToTuple} from 'expect-type';
```
This causes this TS validation error:
```
> tsc -p tsconfig.build.json

node_modules/type-fest/source/omit-deep.d.ts:1:33 - error TS2307: Cannot find module 'expect-type' or its corresponding type declarations.

1 import type {UnionToTuple} from 'expect-type';
                                  ~~~~~~~~~~~~~


Found 1 error in node_modules/type-fest/source/omit-deep.d.ts:1
```

### Repro

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBDAnmApnA3geRMGARFFMAXzgDMoIQ4ByJVAWjJQGcYaBuAWAChf60ASQB2ZCAEY4AXgy8AkAEMAJkqisWALjgBteXPR65bNShhbjwYQHM9xADR6DPOS+OEYAJnMwolqw+cXMQhvX2tbALkAXW4eYn5kNABBFTUWFhExaThsXAIiAB5MiTtaZVV1ADpxSuCaAD5YoA

PR #1053, opened 2025-02-09
Add lint rule to prevent unintended imports in declaration files

som-sm:
VSCode autocomplete suggests imports from `expect-type` before local modules, leading to unintended imports from `expect-type` in declaration files, like:
- https://github.com/sindresorhus/type-fest/pull/1037#discussion_r1914295743
- https://github.com/sindresorhus/type-fest/issues/1051

image<br><br>

This PR adds a lint rule that will catch such issues, causing [CI to fail](https://github.com/sindresorhus/type-fest/actions/runs/13226754137/job/36918575333?pr=1053).

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,11 @@
 			{
 				"files": "source/*.d.ts",
 				"rules": {
-					"no-restricted-imports": ["error", "tsd", "expect-type"]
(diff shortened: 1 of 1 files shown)


sindresorhus:
Good idea 👍

PR #1055, opened 2025-02-11
Add `TupleToObject` type

### LINKED ISSUE #956: Proposal: `TupleToObject<TUPLE>`
## Summary
Given
TUPLE: Any tuple

Provide
A type that defines an object type whose keys and the corresponding value types would be the tuple indices and the corresponding tuple element types.

## Potential definition
```typescript
type TupleToObject<TUPLE> = { [I in keyof TUPLE & `${number}`]: TUPLE[I] }
```
Reference
https://stackoverflow.com/a/75746375/22340824
https://stackoverflow.com/a/70398429/22340824

## Use case
Get union of bool elements indices.
```typescript
const arr = [true, false, 'test'] as const;
type BoolIndices = StringToNumber<ConditionalKeys<TupleToObject<typeof arr>, boolean>>;
```

## Upvote & Fund

<a href="https://polar.sh/sindresorhus/type-fest/issues/956">
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://polar.sh/api/github/sindresorhus/type-fest/issues/956/pledge.svg?darkmode=1">
  Fund with Polar
</picture>
</a>

som-sm:
Closes #956

### DIFF
--- a/source/tuple-to-object.d.ts
+++ b/source/tuple-to-object.d.ts
@@ -1,0 +1,36 @@
+import type {IsTuple} from './is-tuple';
+import type {UnknownArray} from './unknown-array';
+import type {IfAny} from './if-any';
+
+/**
+Transforms a tuple into an object, mapping each tuple index to its corresponding type as a key-value pair.
+
+@example
+```
+type Example1 = TupleToObject<[number, string, boolean]>;
+//=> { 0: number; 1: string; 2: boolean }
+
+// Tuples with optional indices
+type Example2 = TupleToObject<[number, string?, boolean?]>;
+//=> { 0: number; 1?: string; 2?: boolean }
+
+// Readonly tuples
+type Example3 = TupleToObject<readonly [number, string?]>;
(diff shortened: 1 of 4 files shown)


source/tuple-to-object.d.ts
```diff
@@ -0,0 +1,36 @@
+import type {IsTuple} from './is-tuple';
+import type {UnknownArray} from './unknown-array';
+import type {IfAny} from './if-any';
+
+/**
+Transforms a tuple into an object, mapping each tuple index to its corresponding type as a key-value pair.
+
```
sindresorhus, line 7: It should point out that it cannot preserve the label of named tuples.

som-sm: Yeah, added in [6c4104b](https://github.com/sindresorhus/type-fest/pull/1055/commits/6c4104bea41675b1e0fae37555b3eb622a42eb55).

sindresorhus:
Looks good

PR #1073, opened 2025-03-03
`DelimiterCase` / `SnakeCase` / `ScreamingSnakeCase` / `KebabCase`: Fix default value for `splitOnNumbers` option

som-sm:
Fixes https://github.com/sindresorhus/type-fest/pull/930#discussion_r1977964498.

The `Words` type was already splitting on numbers, so for it `splitOnNumbers` correctly defaults to `true`.

However, `DelimiterCase` and its derived types didn't previously split on numbers, so their default value for `splitOnNumbers` should be `false`.

### DIFF
--- a/source/delimiter-case.d.ts
+++ b/source/delimiter-case.d.ts
@@ -61,7 +61,7 @@ const rawCliOptions: OddlyCasedProperties<SomeOptions> = {
 export type DelimiterCase<
 	Value,
 	Delimiter extends string,
-	Options extends WordsOptions = {},
+	Options extends WordsOptions = {splitOnNumbers: false},
 > = Value extends string
 	? IsStringLiteral<Value> extends false
 		? Value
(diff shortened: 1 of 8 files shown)


source/delimiter-case.d.ts
```diff
@@ -61,7 +61,7 @@ const rawCliOptions: OddlyCasedProperties<SomeOptions> = {
 export type DelimiterCase<
 	Value,
 	Delimiter extends string,
-	Options extends WordsOptions = {},
+	Options extends WordsOptions = {splitOnNumbers: false},
```
som-sm, line 64: I'm working on a PR that'll improve how we handle options in our codebase, so I've kept it simple here.

som-sm:
@sindresorhus This should have been a patch release for `v4.36`, right?

sindresorhus:
No, there were non-patch changes, so it ended up being 4.37.

PR #1079, opened 2025-03-04
`Paths`: Add `maxCircularDepth` option

LaurynasGr:
Adds `maxCircularDepth` option to `Paths`


som-sm:
@LaurynasGr Types like the following shouldn't be affected by `maxCircularDepth`, but they currently do.

```ts
type Test = {
	foo: string;
	bar: {
		foo: string;
		bar: {};
	};
};

type P = Paths<Test, {maxCircularDepth: 0}>;
//   ^? type P = 'foo' | 'bar'
```

som-sm:
On second thought, I think this option is just going to be more confusing than helpful.

- I doubt if there's a proper way to solve https://github.com/sindresorhus/type-fest/pull/1079#issuecomment-2699848872, because how would you differentiate b/w structurally similar and truly recursive cases?

- The fact that `maxRecursionDepth` and `maxCircularDepth` don't target the same depth could lead to confusions. For e.g.,
  ```ts
  type Test = {
	  a: string;
	  b: {
		  c: Test;
	  };
  };
  
  type P1 = Paths<Test, {maxRecursionDepth: 2}>;
  //   ^? type P1 = "a" | "b" | "b.c" | "b.c.a" | "b.c.b"
  
  type P2 = Paths<Test, {maxCircularDepth: 1}>; // This needs to be `1` to get the same results
  //   ^? type P2 = "a" | "b" | "b.c" | "b.c.a" | "b.c.b" | "b.c.b.c"
  ```
  
  Ignore the last `"b.c.b.c"` in `P2`—that's probably a bug.

LaurynasGr:
> @LaurynasGr Types like the following shouldn't be affected by `maxCircularDepth`, but they currently do.
> 
> ```ts
> type Test = {
> 	foo: string;
> 	bar: {
> 		foo: string;
> 		bar: {};
> 	};
> };
> 
> type P = Paths<Test, {maxCircularDepth: 0}>;
> //   ^? type P = 'foo' | 'bar'
> ```

@som-sm sorted this one  - was a simple `a extends b` not the same as `b extends a` issue (see https://github.com/sindresorhus/type-fest/pull/1079/commits/54691f132533b4632bcb50a5a81e284f0ea289b4). Added tests to cover this as well.

Will respond to the other comment tomorrow with a real-world example where I need this specific option to stop VS Code from crashing without limiting myself to a very shallow depth with the general recursion depth option.

som-sm:
@LaurynasGr Great, I'll give this a read sometime tomorrow then!

som-sm:
@LaurynasGr Also, since `0` depth currently means root level keys, like:
```ts
type Test = {a: Test};
type P = Paths<Test, {maxRecursionDepth: 0}>;
//   ^? type P = "a"
```

Should depth `0` for `maxCircularDepth` mean recursing into the circular reference once?
```ts
// Current behaviour
type Test = {a: Test};
type P = Paths<Test, {maxCircularDepth: 0}>;
//   ^? type P = "a"
```

```ts
// Suggested behaviour
type Test = {a: Test};
type P = Paths<Test, {maxCircularDepth: 0}>;
//   ^? type P = "a" | "a.a"
```

LaurynasGr:
> @LaurynasGr Also, since `0` depth currently means root level keys, like:

@som-sm I think that having `maxRecursionDepth: 0` return keys at level `0` makes sense because there's nothing below it you might want (i.e. the step below it is no keys at all)

Whereas `maxCircularDepth: 0` essentially turns off recursion into circular references. If we were to change it so that `maxCircularDepth: 0` would mean recursing into the circular reference once, if you wanted to fully turn off circular recursion you would have to provide `maxCircularDepth: -1` which in my eyes is not a logical thing to provide for it

LaurynasGr:
> On second thought, I think this option is just going to be more confusing than helpful.
> <.....>
> * The fact that `maxRecursionDepth` and `maxCircularDepth` don't target the same depth could lead to confusions. For e.g.,
>   ```ts
>   type Test = {
>       a: string;
>       b: {
>     	  c: Test;
>       };
>   };
>   
>   type P1 = Paths<Test, {maxRecursionDepth: 1}>;
>   //   ^? type P1 = "a" | "b" | "b.c"
>   
>   type P2 = Paths<Test, {maxCircularDepth: 1}>;
>   //   ^? type P2 = "a" | "b" | "b.c" | "b.c.a" | "b.c.b" | "b.c.b.c"
>   ```

If this is related to
> @LaurynasGr Also, since 0 depth currently means root level keys, like:
> <...>
> Should depth 0 for maxCircularDepth mean recursing into the circular reference once?

I answered it here https://github.com/sindresorhus/type-fest/pull/1079#issuecomment-2703519336

In other words the whole point of this option is to detach it from `maxRecursionDepth` - they do different things. Below is a real world example, why I taken on to add this option:

I have this state type
```ts
type SomeState = {
  a: {
    b: {
      c: {
        d: boolean;
        e: string;
        f: {
          g: number;
          h: boolean;
        };
      };
    };
    e: number;
  };
  connection: Connection;
  something: string;
  oneMoreThing: {
    a: string;
    b: {
      c: string[];
      d: {
        e: string;
        f: {
          g: number;
          h: boolean;
        };
      };
    };
  };
  getSomething: (param?: string) => Promise<Something>;
  // <....>
};
```
Note that the `connection: Connection` is referring to this type https://github.com/microsoft/vscode-languageserver-node/blob/main/server/src/common/server.ts#L1359 which has a bunch of properties and many of them recursively refers back to `Connection` by having it's own `connection: Connection` reference (e.g. `console`, `tracer`, `telemetry` etc.).

Now if I were to try to get `Paths` for `SomeState`
```ts
type SomeStateKeys = Paths<SomeState>;
```
my VS Code (running on Intel i9-13900k with 128GB RAM) will immediately choke
image

and then after thinking for a good while, it will just give up and say "No, sorry mate, this is too much for me"
image

The only solution I have right now is to limit the whole recursion to some arbitrary small number
image
image

But the issue here is that I loose all of the state keys that I myself defined and still get large numbers of keys I do not care about rom the `Connection` recursion.

Now if I just use `maxCircularDepth: 0` I get all of my own keys, all of the non-recursive `Connection` keys and none of the circular ones I don't care about
image

som-sm:
@LaurynasGr Thanks for the detailed explanation!

<br>

Haven't had a chance to properly review this PR yet, but it looks like `maxCircularDepth` doesn't do anything in this case:
```ts
type Test = [...Test[], string];

type P = Paths<Test, {maxCircularDepth: 0}>;
```
image

LaurynasGr:
> But can you please add a note in the description, similar to `depth`, clarifying what `depth: 0` means.

@som-sm Sure - added the note here https://github.com/sindresorhus/type-fest/pull/1079/commits/b0252206447f35445ba61e8e982d264446fb1302. The first example also clearly shows this, so we should be OK.
Screenshot 2025-03-06 at 16 35 29

sindresorhus:
Add this test:

```ts
type OptionalCircular = {value: string; next?: OptionalCircular};

// Should only expose the first hop when circular recursion is disabled
expectType<'value' | 'next'>({} as Paths<OptionalCircular, {maxCircularDepth: 0}>);

// The moment we allow one circular hop, the deeper path must appear
expectAssignable<Paths<OptionalCircular, {maxCircularDepth: 1}>>({} as 'next.value');
```

LaurynasGr:
@sindresorhus apologies for leaving this unfinished. Ran into another issue, where https://github.com/sindresorhus/type-fest/issues/1304#issuecomment-3592904443 would be super helpful, so will finish up this PR first before looking into the `exclude` option for `Paths`.

@som-sm you haven't started working on `exclude` have you?

som-sm:
> @som-sm you haven't started working on `exclude` have you?

No, feel free to open a PR for the same!

LaurynasGr:
@som-sm @sindresorhus once again apologies for the delay.

Brought this up to date with the new major release and resolved all issues & comments. Checked the related changes since v4.37.0 and made sure it's following the new conventions. Reviewed manually and did a few independent CR passes with Claude Fable and GPT Sol.

Let me know if you spot any issues, will sort it out right away.

PR #1080, opened 2025-03-05
`DelimiterCase` / `SnakeCase` / `ScreamingSnakeCase` / `KebabCase`: Fix instantiations containing punctuations

### LINKED ISSUE #1074: KebabCasedProperties transforming `:` to `-:`
### Bug description

The `KebabCasedProperties` is hyphening before colons. I would expect the property name `onDialog:close` to become `on-dialog:close`, not `on-dialog-:close`.

### Repro

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBDAnmApnA3nA0igRgQ1wGF8BnFAEwAUoJVZgVS4BfOAM1pDgHIlUAtOyYweAbgCwAKGnAAdjBRR2+AMZoAYhAgZpcfbwhyAIsHwAbCAHMAXKsvkeNuAAoAlHAC8APjgA3CGAKSSkWEOl+NAAhfCgvbDxCEnJqWnoYRlIAHi0IbxCDOAB6IrgAPQB+aSA

som-sm:
Fixes #1074.

<br>

Previously, `DelimiterCase` was not built using `Words`, so their behaviours with punctuations were different.
```ts
// v4.5
import type {Words, DelimiterCase} from "type-fest";

type T1 = Words<"onDialog::close">; // Punctuations cause split (kinda)
//   ^? type T1 = ["on", "Dialog", ":", ":close"]

type T2 = DelimiterCase<"onDialog::close", "-">; // No split on punctuations
//   ^? type T2 = "on-dialog::close"
```

<br>

But now, since `DelimiterCase` is built using `Words` and we want to avoid breaking changes, this PR includes a workaround to fix only the behaviour of `DelimiterCase`.

<br>

As mentioned in https://github.com/sindresorhus/type-fest/issues/1074#issuecomment-2698729943, in V5, IMO we should align this behaviour with Lodash. So, `Words<"onDialog::close">` should return  `["on", "Dialog", "close"]` and `KebabCase<"onDialog::close">` should return `on-dialog-close`.


som-sm:
@filipw01 Would appreciate if you could give this a read!

source/internal/characters.d.ts
```diff
@@ -31,3 +31,8 @@ export type Whitespace =
 	| '\u{FEFF}';
 
 export type WordSeparators = '-' | '_' | Whitespace;
+
+export type Punctuations =
```
filipw01, line 35: Where does this list come from? Are we sure it's complete?

filipw01 (approved):

source/internal/characters.d.ts
som-sm: > Are we sure it's complete?

@filipw01 Not really, it's GPT generated.

sindresorhus, line 35:
```suggestion
export type AsciiPunctuation =
```

source/internal/characters.d.ts
```diff
@@ -31,3 +31,8 @@ export type Whitespace =
...
+  | '!' | '"' | '#' | '$' | '%' | '&' | '\'' | '(' | ')' | '*' | '+' | ',' | '-' | '.' | '/'
+  | ':' | ';' | '<' | '=' | '>' | '?' | '@' | '[' | '\\' | ']' | '^' | '_' | '`'
+  | '{' | '|' | '}' | '~';
```
sindresorhus, line 38: One item per line and use tab-indentation.

sindresorhus (changes requested):

source/internal/characters.d.ts
som-sm: Addressed in [399cf03](https://github.com/sindresorhus/type-fest/pull/1080/commits/399cf038f462866faf9bf6a52ba73fecdcb12c09).

source/internal/characters.d.ts
som-sm: Addressed in [399cf03](https://github.com/sindresorhus/type-fest/pull/1080/commits/399cf038f462866faf9bf6a52ba73fecdcb12c09).

PR #1081, opened 2025-03-05
Refactor how options are handled for types

som-sm:
This PR streamlines how type options are handled. Currently, there’s no consistent approach—some types define default options inline as the default type argument, which only works well when there’s a single option. Others don’t specify any defaults at all, which isn’t ideal.

To address this, this PR introduces an internal type called `ApplyDefaultOptions`. It takes three parameters: the options schema, the default options, and the user-specified options and returns an object ensuring that **all options have defined values**, using defaults where necessary.

An example usage of this would be as follows:
https://github.com/sindresorhus/type-fest/blob/382eb878280afd4c4363a719e0f401ddc11e83ad/source/paths.d.ts#L179-L181
github com_sindresorhus_type-fest_pull_1081_files (2)

And the type also enforces the following checks:
https://github.com/sindresorhus/type-fest/blob/382eb878280afd4c4363a719e0f401ddc11e83ad/test-d/internal/apply-default-options.ts#L48-L65

<br>

**NOTE:**
1. I haven't updated `Get` to use `ApplyDefaultValue` because its option handling is a bit complex. It's better to address that in a separate PR.
2. `Schema` also couldn't be updated to use `ApplyDefaultValue` because `SchemaOptions` contains an optional property that explicitly allows `undefined` (`recurseIntoArrays?: boolean | undefined`). This becomes a problem because `ApplyDefaultValue` is designed to overwrite an _optional_ option if its user-specified type is `undefined`—which I believe is the ideal behaviour.
https://github.com/sindresorhus/type-fest/blob/f26ff0b766af8a7cd8bf55edb3ccf514fb38f97a/test-d/internal/apply-default-options.ts#L35-L37
However, this also means that explicitly defined `undefined` values also get overwritten, because I don't think there's a way to differentiate between something like `{ option?: string }` and `{ option?: string | undefined }`. Let me know your thoughts on this.
https://github.com/sindresorhus/type-fest/blob/f26ff0b766af8a7cd8bf55edb3ccf514fb38f97a/test-d/internal/apply-default-options.ts#L39-L41
IMO, we could just remove `undefined` from `recurseIntoArrays` option in `v5`.

### DIFF
--- a/source/camel-case.d.ts
+++ b/source/camel-case.d.ts
@@ -1,3 +1,4 @@
+import type {ApplyDefaultOptions} from './internal';
 import type {Words} from './words';
 
 /**
@@ -14,12 +15,16 @@ export type CamelCaseOptions = {
 	preserveConsecutiveUppercase?: boolean;
 };
 
+export type DefaultCamelCaseOptions = {
+	preserveConsecutiveUppercase: true;
+};
+
 /**
 Convert an array of words to camel-case.
 */
 type CamelCaseFromArray<
 	Words extends string[],
-	Options extends CamelCaseOptions,
+	Options extends Required<CamelCaseOptions>,
 	OutputString extends string = '',
 > = Words extends [
 	infer FirstWord extends string,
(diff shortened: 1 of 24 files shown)


sindresorhus:
This looks really great! I'm so happy to have consistent and correct options handling.

---

*Sorry about the delay in reviewing. I have been traveling.*

sindresorhus:
> IMO, we could just remove undefined from recurseIntoArrays option in v5.

Agreed. I added it to the v5 list.

PR #1085, opened 2025-03-25
Add `ArrayFlat` type

### LINKED ISSUE #1075: Proposal: `Flat` (equivalent to `Array.prototype.flat()`)

I'd like to propose something like `Array.prototype.flat()` but for types.

### Type source

```ts
type Flat<T extends any[][]> = T extends (infer U)[][] ? U : never;

type Input = [
  ['A', 'B'],
  ['C'],
]

type Output = Flat<Input>
// => ['A', 'B', 'C']
```

Emiyaaaaa:
- closed #1075


test-d/array-flat.ts
```diff
@@ -0,0 +1,60 @@
...
+expectType<ArrayFlat<readonly [readonly [1, 2], readonly [3, 4]]>>([1, 2, 3, 4]);
+
+// Edge cases
+expectType<ArrayFlat<Array<[]>>>([]);
+expectType<ArrayFlat<[[]]>>([]);
+expectType<ArrayFlat<[[], []]>>([]);
+expectType<ArrayFlat<undefined[]>>(null! as undefined[]);
+expectType<ArrayFlat<any[]>>(null! as any[]);
+expectType<ArrayFlat<unknown[]>>(null! as unknown[]);
+expectType<ArrayFlat<never[]>>(null! as never[]);
```
sindresorhus, line 60: Some more potential tests (AI generated, but I looked over them):

```js
// Test specifically for non-tuple array handling (Array<T> vs [T, T])
type GenericNumberArrays = Array<Array<number>>;
const genericArrayTest: ArrayFlat<GenericNumberArrays> = [1, 2, 3, 4, 5];

// Test for deeply nested non-tuple arrays with specific depths
type DeepGenericArray = Array<Array<Array<Array<number>>>>;
const deepGenericTest1: ArrayFlat<DeepGenericArray, 2> = [[1, 2], [3, 4]];
const deepGenericTest2: ArrayFlat<DeepGenericArray, 3> = [1, 2, 3, 4];

// Test for mixed generic arrays and tuples
type MixedGenericAndTuple = Array<[number, string]>;
const mixedGenericTupleTest: ArrayFlat<MixedGenericAndTuple> = [1, 'a', 2, 'b'];

// Test for array with optional elements in nested structure
type NestedOptional = [number, [string?, number?][]];
const nestedOptionalTest: ArrayFlat<NestedOptional> = [1, undefined, 2, 'a', 3];

// Test for arrays with rest elements in nested structure
type NestedRest = [string, [...number[]][]];
const nestedRestTest: ArrayFlat<NestedRest> = ['a', 1, 2, 3, 4, 5];

// Test for flattening arrays with union types in nested structures
type NestedUnion = Array<Array<string | number>>;
const nestedUnionTest: ArrayFlat<NestedUnion> = ['a', 1, 'b', 2];

// Test for empty array in complex structure
type ComplexWithEmpty = [number, Array<[]>, string];
const complexEmptyTest: ArrayFlat<ComplexWithEmpty> = [1, 'string'];

// Test for array with undefined/null elements
type ArrayWithNullish = [number, [undefined, null]];
const nullishTest: ArrayFlat<ArrayWithNullish> = [1, undefined, null];

// Test for array with mixed depth elements
type MixedDepthArray = [number, Array<string>, [[boolean]]];
const mixedDepthTest: ArrayFlat<MixedDepthArray, 1> = [1, 'a', 'b', [true]];

// Test for readonly nested arrays with different depths
type ReadonlyNestedComplex = readonly [number, readonly (readonly string[])[]];
const readonlyComplexTest: ArrayFlat<ReadonlyNestedComplex, 2> = [1, 'a', 'b', 'c'];

// Test for recursive flattening with non-array elements
type RecursiveWithNonArray = [number, [string, {a: number}]];
const recursiveNonArrayTest: ArrayFlat<RecursiveWithNonArray> = [1, 'string', {a: 42}];
```

test-d/array-flat.ts
```diff
@@ -0,0 +1,101 @@
...
+
+// Test with different element types
+expectType<ArrayFlat<[string, [number, boolean]]>>([null! as string, null! as number, null! as boolean]);
+expectType<ArrayFlat<[{a: number}, [{b: string}]]>>([null! as {a: number}, null! as {b: string}]);
+
+// Test with union types
+expectType<ArrayFlat<[1, 2] | [[3, 4]]>>([1, 2] as [1, 2] | [3, 4]);
+expectType<ArrayFlat<[1, [2]] | [[3], 4]>>([1, 2] as [1, 2] | [3, 4]);
+
+expectType<ArrayFlat<[...Array<[number, string]>]>>(null! as [number?, string?, number?, string?, number?, string?, number?, string?, number?, string?, number?, string?, number?, string?, number?, string?, number?, string?, number?, string?]);
```
som-sm, line 37: But the output of `[number?, string?, number?, ...]` wouldn't work if the input were `[...Array<[number, string?]`.

Consider this example:
```ts
const input: [...Array<[number, string?]>] = [[1], [3], [5]];
const output: ArrayFlat<[...Array<[number, string?]>]> = [1, 3, 5]; // Errors, but it shouldn't
```

test-d/array-flat.ts
```diff
@@ -0,0 +1,101 @@
...
+expectType<ArrayFlat<[1, [2, ...Array<3>], 4]>>([1, 2, ...(null! as Array<3>), 4]);
+
+// Test with mixed arrays and tuples
+expectType<ArrayFlat<[1, number[], 3]>>([1, ...(null! as number[]), 3]);
```
som-sm, line 41: Add these cases:

```ts
ArrayFlat<[1, number[], string[], true]>
ArrayFlat<[1, number[], true, string[]]>
ArrayFlat<[string[], [...boolean[], string]]>
```

som-sm, line 41: Fails in this case: `ArrayFlat<[...number[], string[]]>`

Returns `[...number[], string[]]` which isn't correct, should probably return `(number | string)[]` instead.

```ts
const input: [...number[], string[]] = [1, 2, 3, ['a', 'b', 'c']];
const output: ArrayFlat<[...number[], string[]]> = [1, 2, 3, 'a', 'b', 'c']; // Errors, but it shouldn't
```

test-d/array-flat.ts
```diff
@@ -0,0 +1,101 @@
...
+// Test for deeply nested non-tuple arrays with specific depths
+type DeepGenericArray = number[][][][];
+expectAssignable<ArrayFlat<DeepGenericArray, 2>>([[1, 2], [3, 4]]);
+expectAssignable<ArrayFlat<DeepGenericArray, 3>>([1, 2, 3, 4]);
+
+// Test for mixed generic arrays and tuples
+type MixedGenericAndTuple = Array<[number, string]>;
+expectAssignable<ArrayFlat<MixedGenericAndTuple>>([1, 'a', 2, 'b']);
+
+// Test for array with optional elements in nested structure
```
som-sm, line 71: Fails in this case:

```ts
const input: [number, [string, boolean?], number] = [1, ['a'], 2];
const output: ArrayFlat<[number, [string, boolean?], number]> = [1, 'a', 2]; // Errors, but it shouldn't
```

type-fest@4.37.0
som-sm, line 1: What's this file?

som-sm (changes requested):
@Emiyaaaaa Added comments on the tests — haven’t reviewed the implementation yet.

test-d/array-flat.ts
Emiyaaaaa: `ArrayFlat<[number, [string, boolean?], number]>` should be `[number, string, boolean?, number?]` or `[number, string, number] | [number, string, boolean, number]`

som-sm: @Emiyaaaaa 
> `[number, string, boolean?, number?]`

This isn't correct, consider this example:
```ts
const input: [number, [string, boolean?], number] = [1, ['a'], 2];
const output: [number, string, boolean?, number?] = [1, 'a', 2]; // Errors, but it shouldn't
```

<br>

> `[number, string, number] | [number, string, boolean | undefined, number]` 

This feels correct to me. Would be nice if the `| undefined` is present only when `exactOptionalPropertyTypes` is disabled.

Emiyaaaaa: > This feels correct to me
I agree

> Would be nice if the | undefined is present only when exactOptionalPropertyTypes is disabled.

I think `| undefined` is necessary.
```ts
const input: [number, [string, boolean?], number] = [1, ['a', undefined], 2];
const output: [number, string, boolean, number] = [1, 'a', undefined, 2]; // Errors, but it shouldn't

```

som-sm: > I think | undefined is necessary.

But if `exactOptionalPropertyTypes` is enabled, the `input` itself would error. `undefined` for optional indices will _only_ be allowed if `exactOptionalPropertyTypes` is disabled.

```ts

const input: [number, [string, boolean?], number] = [1, ['a', undefined], 2]; // Errors
```

Emiyaaaaa: Okay I got, but there seems to be no way to know `exactOptionalPropertyTypes` value

som-sm: Yeah, there isn't a way to know the value of `exactOptionalPropertyTypes`.

But in some cases the compiler automatically decides whether to add `| undefined` or not depending on the value of `exactOptionalPropertyTypes`, like:

```ts

type Input = [number, boolean?]

type Nullify<T> = {
    [P in keyof T]: T[P] | null
}

type T = Nullify<Input>
//   ^? type T = [number | null, (boolean | null | undefined)?]
```

```ts

type Input = [number, boolean?]

type Nullify<T> = {
    [P in keyof T]: T[P] | null
}

type T = Nullify<Input>
//   ^? type T = [number | null, (boolean | null)?]
```

But don't worry too much about this—just see once if the compiler is automatically doing the trick, if not, it's okay to explicitly add the `| undefined`.

Emiyaaaaa: Thanks you for your inspiration, I find a way to get `exactOptionalPropertyTypes` option value:
```ts
type ExactOptionalPropertyTypesEnable = [(string | undefined)?] extends [string?] ? false : true;
```

som-sm: @Emiyaaaaa That's pretty cool ✨

sindresorhus:
@Emiyaaaaa Can you fix the conflicts?

Emiyaaaaa:
> @Emiyaaaaa Can you fix the conflicts?

done

source/array-flat.d.ts
```diff
@@ -0,0 +1,256 @@
...
+Return true if the number is 0
+*/
+type IsZero<T extends number> = [T] extends [0] ? true : false;
+
+/**
+Options for the `ArrayFlat` type.
+*/
+type ArrayFlatOptions = {
+	/**
+	 * The number of times to repeat the array items when flattening an non-fixed length array.
```
sindresorhus, line 20: Don't prefix the doc comments with `*`

source/array-flat.d.ts
```diff
@@ -0,0 +1,256 @@
...
+	 * ```
+	 * type FlatArr0 = ArrayFlat<Array<number, string>, 1, { maxRepeat: 3 }>;
+	 * //=> type FlatArr0 =
+	 * []
+	 * | [number, string]
+	 * | [number, string, number, string]
+	 * | [number, string, number, string, number, string];
+	 * ```
+	 *
+	 * @default 5
```
sindresorhus, line 32: Default should be before example

source/array-flat.d.ts
```diff
@@ -0,0 +1,256 @@
...
+};
+
+type DefaultArrayFlatOptions = {
+	maxRepeat: 5;
+};
+
+/**
+Creates a new array type by flattening an array to a specified depth.
+
+Use-case: Flatten an array type to a specified depth.
```
sindresorhus, line 44: This just repeats the description. Provide a real-world use-case.

sindresorhus:
I recommend copy-pasting the type into ChatGPT and ask it to "review, find bugs, and suggest improvements".

source/internal/array.d.ts
som-sm, line 1: Also add tests for these helpers. 

And, if these helpers are very specific to `ArrayFlat` then it doesn't make sense to put them here. Because, for example, if I see something like `IsLeadingSpreadArray` as a helper, I'd expect it to work with all types of arrays, not just a specific subset.

som-sm (changes requested):

sindresorhus:
@Emiyaaaaa Can you fix the merge conflict? (Don't overwrite my commits)

test-d/array-flat.ts
```diff
@@ -0,0 +1,146 @@
...
+// Test with readonly arrays
+expectType<ArrayFlat<readonly [1, readonly [2, 3]]>>([1, 2, 3]);
+expectType<ArrayFlat<readonly [readonly [1, 2], readonly [3, 4]]>>([1, 2, 3, 4]);
+
+// Edge cases
+expectType<ArrayFlat<Array<[]>>>([]);
+expectType<ArrayFlat<[[]]>>([]);
+expectType<ArrayFlat<[[], []]>>([]);
+expectType<ArrayFlat<undefined[]>>(null! as undefined[]);
+expectType<ArrayFlat<unknown[]>>(null! as unknown[]);
```
sindresorhus, line 101: some more:

```ts
expectType<ArrayFlat<any>>(null! as any);
expectType<ArrayFlat<never>>(null! as never);
expectType<ArrayFlat<any[]>>(null! as any[]);
expectType<ArrayFlat<never[]>>(null! as never[]);
expectType<ArrayFlat<[any, [never]]>>(null! as [any, never]);
```

sindresorhus:
Some things:

- Depth handling does not currently match Array#flat for negative and non-integer depths. In JS, depth is normalized, so values like -1 and 0.5 should effectively behave like 0 (no flattening). Right now those cases can still flatten, and fractional depth can also trigger excessive type instantiation.

- A few tuple type tests have extra internal whitespace before ].

test-d/array-flat.ts
```diff
@@ -0,0 +1,154 @@
...
+type MixedDepthArray = [number, string[], [[boolean]]];
+expectAssignable<ArrayFlat<MixedDepthArray>>([1, 'a', 'b', [true]]);
+
+// Test for readonly nested arrays with different depths
+type ReadonlyNestedComplex = readonly [number, ReadonlyArray<readonly string[]>];
+expectAssignable<ArrayFlat<ReadonlyNestedComplex, 2>>([1, 'a', 'b', 'c']);
+
+// Test for recursive flattening with non-array elements
+type RecursiveWithNonArray = [number, [string, {a: number}]];
+expectAssignable<ArrayFlat<RecursiveWithNonArray>>([1, 'string', {a: 42}]);
```
sindresorhus, line 157:
```suggestion
expectAssignable<ArrayFlat<RecursiveWithNonArray>>([1, 'string', {a: 42}]);

// Test options.maxRepeat for non-fixed length arrays
type RepeatedPair = ArrayFlat<Array<[number, string]>, 1, {maxRepeat: 2}>;
expectAssignable<RepeatedPair>([]);
expectAssignable<RepeatedPair>([1, 'a']);
expectAssignable<RepeatedPair>([1, 'a', 2, 'b']);
expectNotAssignable<RepeatedPair>([1]);
expectNotAssignable<RepeatedPair>([1, 'a', 2]);
expectNotAssignable<RepeatedPair>([1, 'a', 2, 'b', 3, 'c']);

type SingleRepeatedPair = ArrayFlat<Array<[number, string]>, 1, {maxRepeat: 1}>;
expectAssignable<SingleRepeatedPair>([]);
expectAssignable<SingleRepeatedPair>([1, 'a']);
expectNotAssignable<SingleRepeatedPair>([1, 'a', 2, 'b']);

type NoRepeatedPair = ArrayFlat<Array<[number, string]>, 1, {maxRepeat: 0}>;
expectType<NoRepeatedPair>([]);
expectNotAssignable<NoRepeatedPair>([1, 'a']);
```

sindresorhus (changes requested):

sindresorhus:
Doesn't seem to handle this: https://github.com/sindresorhus/type-fest/pull/1085/commits/302efd8962e0756c079599ed62156ed5d53ce66f Which the runtime version handles.

Emiyaaaaa:
> Doesn't seem to handle this: [302efd8](https://github.com/sindresorhus/type-fest/commit/302efd8962e0756c079599ed62156ed5d53ce66f) Which the runtime version handles.

I could implement like `math.floor(depth)` and `math.max(0, depth)` on depth, but I feel it's over design, almost no one would attempt to use decimals or negative values in the depth parameter.

sindresorhus:
You are probably right. I think the tests are still valuable to show behavior though, just fix them up to pass.

Emiyaaaaa:
> You are probably right. I think the tests are still valuable to show behavior though, just fix them up to pass.

and another reason is this would add extra performance overhead. I don’t think it’s worth introducing guaranteed performance issues just for something that almost never happens.

PR #1088, opened 2025-03-27
`WritableKeysOf` / `ReadonlyKeysOf`: Fix behaviour with unions and optional properties

### LINKED ISSUE #1087: WritableKeysOf return type has undefined
### Bug description

```tsx
interface User {
    name: string;
    surname?: string;
    readonly id: number;
}
type _User = WritableKeysOf<User>; // return  "name" | "surname" | undefined
const user: _User = undefined; // not error
```

### Repro

https://www.typescriptlang.org/play/?#code/PTDECcFMHMEsHsB2ACALgTwA6QLQDNIBnVAWAChyNtkBJQgUQEcBXAQwBsAeAQQBpkAQgD5kAXnIBIABScA4kKkBKMSNnJIAD1SREAE0LJuyAGTI1AHzPIA-MgCMyAFzIATMs3a9hSTPlKVVh46+oImVpZqtg7ObpK2qODMkJLOeByEkADc5JRYkMgA6uCwqKwARuyQANKQ6IQA8nicACoiosgAckgdzOzs5ZWcAN6SANoACsiwKADWtfB4yM0Aus50TGxcQ8ijAIpTKOOrSxPLyAC+-NtQrLpI7Og7+9PIR87NpxciQV7IaewZGyvJzIRCQABukHA5HOozm6AWS2WQmyFDIIFAwSgcCQOTI0204DSAGN8gBVDLgZAjMgSRCsAC2kGcxGKiGgqIkhGY4HpTOsLIS0w5khud0QDymumciGYDLKUNR51y1AA+hSoWJCsVShVqrUGk0NeAUcgQMgoKgeShkAAiPmQW3ISy27m8xmO53IZh6SB4aaQXTkYlIYjeynOdWUrU+3R+gO6TJm4Cg+CodTgcDwcBAA

som-sm:
Fixes #1087

Looks like there's some issue in TS `v5.4` and later, because it works fine in `v5.3` and below ([playground link](https://tsplay.dev/w1vnKN)). Regardless, the updated implementation should work fine across all versions.

Also, fixes behaviour with unions.

<br />

The updated implementation also improves behaviour with arrays:
```ts
type CurrentBehaviour = WritableKeysOf<[string, number]>;
//   ^? type CurrentBehaviour = 2 | "0" | "1" | (() => ArrayIterator<"0" | "1">) | {
//          [x: number]: boolean | undefined;
//          length?: boolean | undefined;
//          toString?: boolean | undefined;
//          toLocaleString?: boolean | undefined;
//          ... 28 more ...;
//          readonly [Symbol.unscopables]?: boolean | undefined;
//      } | ... 28 more ... | ((searchElement: "0" | "1", fromIndex?: number) => boolean)

type UpdatedBehaviour = WritableKeysOf<[string, number]>;
//   ^? type T = number | typeof Symbol.iterator | "0" | "1" | "length" | "toString" | "toLocaleString" | "pop" | "push" | "concat" | "join" | "reverse" | "shift" | "slice" | "sort" | "splice" | ... 26 more ... | "with"
```

However, I had to comment out the array cases because looks like there's some issue with `readonly` arrays in `v5.1` and `v5.2`.

In `v5.1` and `v5.2`, the `number` key loses it's `readonly` modifier while mapping, causing `WritableKeys<readonly [string]>` to incorrectly include `number` in its result.

```ts
type Test<T> = {
	[P in keyof T as number extends P ? P : never]: {[Q in P]: T[P]}
};

type T1 = Test<readonly [string]>;
//   ^? type T1 = {
//          [x: number]: {
//              [x: number]: string;
//          };
//      }
```

In `v5.3` and higher, the `number` key correctly preserves it's `readonly` modifier while mapping, so `WritableKeys<readonly [string]>` no longer has `number` in it's result.
```ts
type Test<T> = {
	[P in keyof T as number extends P ? P : never]: {[Q in P]: T[P]}
};

type T1 = Test<readonly [string]>;
//   ^? type T1 = {
//          readonly [x: number]: {
//              readonly [x: number]: string;
//          };
//      }
```

We'll probably have to figure out a way in which we can run certain test cases only for certain specific versions.

### DIFF
--- a/source/readonly-keys-of.d.ts
+++ b/source/readonly-keys-of.d.ts
@@ -1,4 +1,4 @@
-import type {IsEqual} from './is-equal';
+import type {WritableKeysOf} from './writable-keys-of';
 
 /**
 Extract all readonly keys from the given type.
@@ -24,6 +24,4 @@ const update1: UpdateResponse<User> = {
 
 @category Utilities
 */
-export type ReadonlyKeysOf<T> = NonNullable<{
-	[P in keyof T]: IsEqual<{[Q in P]: T[P]}, {readonly [Q in P]: T[P]}> extends true ? P : never
-}[keyof T]>;
+export type ReadonlyKeysOf<T> = T extends unknown ? Exclude<keyof T, WritableKeysOf<T>> : never;
--- a/source/readonly-keys-of.d.ts
+++ b/source/readonly-keys-of.d.ts
@@ -1,3 +1,4 @@
+import type {KeysOfUnion} from './keys-of-union';
 import type {WritableKeysOf} from './writable-keys-of';
 
 /**
@@ -24,4 +25,4 @@ const update1: UpdateResponse<User> = {
 
 @category Utilities
 */
-export type ReadonlyKeysOf<T> = T extends unknown ? Exclude<keyof T, WritableKeysOf<T>> : never;
+export type ReadonlyKeysOf<T> = Exclude<KeysOfUnion<T>, WritableKeysOf<T>>;
--- a/source/writable-keys-of.d.ts
+++ b/source/writable-keys-of.d.ts
@@ -25,6 +25,6 @@ const update1: UpdateRequest<User> = {
 
 @category Utilities
 */
-export type WritableKeysOf<T> = NonNullable<{
-	[P in keyof T]: IsEqual<{[Q in P]: T[P]}, {readonly [Q in P]: T[P]}> extends false ? P : never
-}[keyof T]>;
+export type WritableKeysOf<T> = keyof {
+	[P in keyof T as IsEqual<{[Q in P]: T[P]}, {readonly [Q in P]: T[P]}> extends false ? P : never]: never
+};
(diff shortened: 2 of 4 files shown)


som-sm:
> We'll probably have to figure out a way in which we can run certain test cases only for certain specific versions.

I had to [do a similar thing](https://github.com/ts-essentials/ts-essentials/pull/416) in `ts-essentials` as well.

som-sm:
Also, @sindresorhus in `v5`, I feel we should add a constraint of `object` to `T` for `WritableKeysOf` and `ReadonlyKeysOf`, just like we do in `OptionalKeysOf` and `RequiredKeysOf`.

https://github.com/sindresorhus/type-fest/blob/a6590b961a2485a5f6ecf7acb1c6bdfd1075ab3e/source/optional-keys-of.d.ts#L34

sindresorhus:
> I feel we should add a constraint of object to T for WritableKeysOf and ReadonlyKeysOf

Agreed. Added to the list.

PR #1089, opened 2025-03-27
`OptionalKeysOf` / `RequiredKeysOf`: Fix instantiations with unions and arrays

som-sm:
Fixes behaviour `OptionalKeysOf` / `RequiredKeysOf` with unions and also improves the behaviour with arrays.

### DIFF
--- a/source/readonly-keys-of.d.ts
+++ b/source/readonly-keys-of.d.ts
@@ -1,4 +1,3 @@
-import type {KeysOfUnion} from './keys-of-union';
 import type {WritableKeysOf} from './writable-keys-of';
 
 /**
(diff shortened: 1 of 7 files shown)


source/readonly-keys-of.d.ts
```diff
@@ -25,4 +24,7 @@ const update1: UpdateResponse<User> = {
 
 @category Utilities
 */
-export type ReadonlyKeysOf<T> = Exclude<KeysOfUnion<T>, WritableKeysOf<T>>;
+export type ReadonlyKeysOf<T> =
+	T extends unknown // For distributing `T`
+		? Exclude<keyof T, WritableKeysOf<T>>
+		: never; // Should never happen
```
som-sm, line 30: Had to revert back to the conditional approach because otherwise union cases like the following would fail:

```ts
expectType<'a' | 'b'>({} as ReadonlyKeysOf<{a: string; b: number} | {readonly a: string; readonly b: number}>);
```

With the `KeysOfUnion` approach the above case would incorrectly return `never`.

PR #1091, opened 2025-03-29
`ArrayTail`: Add `preserveReadonly` option

som-sm:
Adds `preserveReadonly` option to `ArrayTail` type.


sindresorhus:
Do you think it would make sense to make it `true` by default in the v5?

som-sm:
> Do you think it would make sense to make it true by default in the v5?

@sindresorhus Yeah, I think we should!

PR #1094, opened 2025-03-31
export `ConditionalSimplify` and `ConditionalSimplifyDeep`

### LINKED ISSUE #1005: Export `ConditionalSimplify` and `ConditionalSimplifyDeep`
Although these two types were added as experimental types, but more and more types in type-fest now using these two internal types and are also useful for external users.

related https://github.com/sindresorhus/type-fest/issues/436

romanywu:
closed #1005


sindresorhus:
There is never any need to open multiple pull requests. If you made a mistake, you can simply force push to the same branch. Multiple closed pull requests adds needless noise to the our notifications.

source/conditional-simplify-deep.d.ts
```diff
@@ -0,0 +1,62 @@
+/**
+Recursively simplifies a type while including and/or excluding certain types from being simplified.
```
sindresorhus, line 2: Please preserve the code example like it was originally, this includes empty lines.

Emiyaaaaa (approved):

source/conditional-simplify-deep.d.ts
sindresorhus: @romanywu I was talking in general, not just this example. All the code.

source/conditional-simplify-deep.d.ts
```diff
@@ -0,0 +1,63 @@
...
+
+@example
+```ts
+import type {ConditionalSimplifyDeep} from 'type-fest';
```
sindresorhus, line 6:
```suggestion
import type {ConditionalSimplifyDeep} from 'type-fest';

```

sindresorhus: Etc etc.

PR #1101, opened 2025-04-05
Make `IsFloat<T>` handle numbers represented using exponential notation

lvlte:
NB. This also fixes `IsInteger<T>`, `Float<T>` and `Integer<T>` as well since they rely on `isFloat<T>`.

### DIFF
--- a/source/is-float.d.ts
+++ b/source/is-float.d.ts
@@ -3,31 +3,41 @@ import type {Zero} from './numeric';
 /**
 Returns a boolean for whether the given number is a float, like `1.5` or `-1.5`.
 
-It returns `false` for `Infinity`.
-
 Use-case:
 - If you want to make a conditional branch based on the result of whether a number is a float or not.
 
 @example
 ```
-type Float = IsFloat<1.5>;
+type A = IsFloat<1.5>;
 //=> true
 
-type IntegerWithDecimal = IsInteger<1.0>;
+type B = IsFloat<-1.5>;
+//=> true
+
+type C = IsFloat<1e-7;
+//=> true
+
+type D = IsFloat<1.0>;
 //=> false
 
-type NegativeFloat = IsInteger<-1.5>;
-//=> true
+type E = IsFloat<PositiveInfinity>;
+//=> false
 
-type Infinity_ = IsInteger<Infinity>;
+type F = IsFloat<1.23+21>;
 //=> false
 ```
+
+@category Type Guard
+@category Numeric
 */
-export type IsFloat<T> =
-T extends number
-	? `${T}` extends `${infer _Sign extends '' | '-'}${number}.${infer Decimal extends number}`
-		? Decimal extends Zero
-			? false
-			: true
-		: false
+export type IsFloat<T> = T extends number
+	? `${T}` extends `${'' | '-'}${number}e${infer E extends '-' | '+'}${number}`
+		? E extends '-'
+			? true
+			: false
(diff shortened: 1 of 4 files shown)


source/is-integer.d.ts
```diff
@@ -3,38 +3,37 @@ import type {IsFloat} from './is-float';
...
-type Float = IsInteger<1.5>;
-//=> false
-
-// Supports non-decimal numbers
-
-type OctalInteger: IsInteger<0o10>;
+type D = IsInteger<1.23+21>;
 //=> true
 
-type BinaryInteger: IsInteger<0b10>;
```
sindresorhus, line 32: Why remove this?

lvlte: Because `0b10` is just another way for _writing_ the number `2`, the value is _interpreted_ as the decimal number `2`, not `0b10`. 

Binary, octal, and hexadecimal numbers are always read as decimal numbers, and they can't have a fractional part, so they are always integers.

[Playground](https://www.typescriptlang.org/play/?#code/MYewdgzgLgBAhjAvDADAIwIwoNwCgD0+MxMAegPy66iSxpKohZ6ElmXXjQzAMoAezAkRIUgA)

source/is-float.d.ts
```diff
@@ -3,31 +3,41 @@ import type {Zero} from './numeric';
...
 */
-export type IsFloat<T> =
-T extends number
-	? `${T}` extends `${infer _Sign extends '' | '-'}${number}.${infer Decimal extends number}`
-		? Decimal extends Zero
-			? false
-			: true
-		: false
+export type IsFloat<T> = T extends number
+	? `${T}` extends `${'' | '-'}${number}e${infer E extends '-' | '+'}${number}`
```
som-sm, line 34:
```suggestion
	? `${T}` extends `${number}e${infer E extends '-' | '+'}${number}`
```

You don't need to explicitly match the negative sign, `${number}` will take care of it.

source/is-float.d.ts
```diff
@@ -3,31 +3,41 @@ import type {Zero} from './numeric';
...
+		? E extends '-'
+			? true
+			: false
+		: `${T}` extends `${'' | '-'}${number}.${infer Decimal extends number}`
```
som-sm, line 38:
```suggestion
		: `${T}` extends `${number}.${infer Decimal extends number}`
```

source/is-float.d.ts
```diff
@@ -3,31 +3,41 @@ import type {Zero} from './numeric';
...
+			? Decimal extends Zero
+				? false
+				: true
```
som-sm, line 41:
```suggestion
		: `${T}` extends `${'' | '-'}${number}.${number}`
			? true
```
<br>

`` `${T}` `` automatically removes unnecessary zeroes, for ex:
```ts
type T = `${5.00}`;
//   ^? type T = "5"
```

som-sm (changes requested):

lvlte:
Ok, is there anything to do on this one before merging ?

source/is-integer.d.ts
lvlte: But yeah this may might not be obvious for everyone, do you think we should keep them ?

som-sm: @lvlte Yeah, I think we should keep it!

som-sm (approved):

sindresorhus:
It's good. Thanks for contributing 🙏

PR #1103, opened 2025-04-08
Add `NonEmptyString` type

### LINKED ISSUE #946: Non-Empty String Type
It seems pretty common that we run into issues with a string being empty when it's not supposed to. These are some common issues I see:

1. An API parameter is meant to be a string, but an empty string isn't valid.
1. We mean to default an empty string to something else, such as `'-1'`, but our ESLint config encourages us to use nullish coalescing `??` instead of `||`. Obviously these don't work the same in cases where you have an empty string.

anyways, my suggestion is to create a type called `NonEmptyString`. According to [this Stackoverflow](https://stackoverflow.com/questions/46253340/how-to-write-a-string-type-that-does-not-contain-an-empty-string-in-typescript), the best way to accomplish this is with a very verbose template literal with any non-empty character followed by a string

jor1ken:
closed #946


test-d/non-empty-string.ts
```diff
@@ -0,0 +1,11 @@
...
+import type {NonEmptyString} from '../index';
+
+declare const empty: NonEmptyString<''>;
+expectNever(empty);
+
+declare const a: NonEmptyString<'a'>;
+expectType<NonEmptyString<'a'>>(a);
+
+declare const b: NonEmptyString<'b' | 'c'>;
+expectType<NonEmptyString<'b' | 'c'>>(b);
```
som-sm, line 11: The expected type and the actual type are identical, making these tests useless.

```suggestion
declare const a: NonEmptyString<'a'>;
expectType<'a'>(a);

declare const b: NonEmptyString<'b' | 'c'>;
expectType<'b' | 'c'>(b);
```

som-sm, line 11: Also, add these test cases:

```ts
expectType<'a'>({} as NonEmptyString<'' | 'a'>);
expectType<string>({} as NonEmptyString<string>);
expectType<Uppercase<string>>({} as NonEmptyString<Uppercase<string>>);
expectType<`on${string}`>({} as NonEmptyString<`on${string}`>);
```

test-d/non-empty-string.ts
```diff
@@ -0,0 +1,11 @@
+import {expectNever, expectType} from 'tsd';
+import type {NonEmptyString} from '../index';
+
+declare const empty: NonEmptyString<''>;
+expectNever(empty);
```
som-sm, line 5: Better to use `expectType` instead of `expectNever`.

```suggestion
declare const empty: NonEmptyString<''>;
expectType<never>(empty);
```

som-sm (changes requested):

sindresorhus:
I would like to see some more tests. Think of more edge-cases to test.

---

Here are some AI-generated suggestions:

```ts
// Tests that string types from mapped types work correctly
type StringMap = { [K in 'a' | 'b']: string };
type StringMapValues = StringMap[keyof StringMap];
expectType<string>({} as NonEmptyString<StringMapValues>);

// Tests handling of conditional types that resolve to empty string or string
type MaybeEmpty<T> = T extends number ? '' : string;
expectType<never>({} as NonEmptyString<MaybeEmpty<number>>);
expectType<string>({} as NonEmptyString<MaybeEmpty<boolean>>);

// Tests distribution over unions in conditional types
type UnionResult<T> = T extends string ? T | '' : never;
expectType<string>({} as NonEmptyString<UnionResult<string>>);

// Tests with generic constraints - this shows behavior with type parameters
declare function genericTest<T extends string, U extends ''>(
    a: NonEmptyString<T>,
    b: NonEmptyString<U>
): void;
```

som-sm:
> Here are some AI-generated suggestions:

@sindresorhus These AI generated tests aren't actually doing anything different — they just look complicated.

Let's break them down:
> ```ts 
> // Tests that string types from mapped types work correctly
> type StringMap = { [K in 'a' | 'b']: string };
> type StringMapValues = StringMap[keyof StringMap];
> expectType<string>({} as NonEmptyString<StringMapValues>);
> ```

This is effectively just `NonEmptyString<string>`, since `StringMapValues` is simply `string`. Probably the actual intention here was to test `NonEmptyString<"a" | "b">`, but even that wouldn't be something different.

<br>

> ```ts
> // Tests handling of conditional types that resolve to empty string or string
> type MaybeEmpty<T> = T extends number ? '' : string;
> expectType<never>({} as NonEmptyString<MaybeEmpty<number>>);
> expectType<string>({} as NonEmptyString<MaybeEmpty<boolean>>);
> ```

This just tests `NonEmptyString<''>` and `NonEmptyString<string>`. Nothing new here either!

<br>

> ```ts
> // Tests distribution over unions in conditional types
> type UnionResult<T> = T extends string ? T | '' : never;
> expectType<string>({} as NonEmptyString<UnionResult<string>>);
> ```

Again, this simplifies down to `NonEmptyString<string>`. Most likely, the intention here was to test something like `NonEmptyString<'' | 'a'>`, but again that wouldn't be anything new.

<br>

> ```ts
> // Tests with generic constraints - this shows behavior with type parameters
> declare function genericTest<T extends string, U extends ''>(
> 	a: NonEmptyString<T>,
> 	b: NonEmptyString<U>
> ): void;
> ```

And this one doesn't seem to test anything at all.

<br>

The only meaningful edge case I can think of is testing for generic assignability.
```ts
// `NonEmptyString<S>` should be assignable to `string`
type Assignability1<_S extends string> = unknown;
type Test1<S extends string> = Assignability1<NonEmptyString<S>>;

// `string` should NOT be assignable to `NonEmptyString<S>`
type Assignability2<_S extends string, _SS extends NonEmptyString<_S>> = unknown;
// @ts-expect-error
type Test2<S extends string> = Assignability2<S, S>;
```

sindresorhus:
> @sindresorhus @jor1ken WDYT?

👍

som-sm (approved):
@sindresorhus I've updated the implementation to handle [this](https://github.com/sindresorhus/type-fest/pull/1103#issuecomment-2788744549) and also updated the tests, please have a look!

PR #1117, opened 2025-04-30
Add `SetNonNullableDeep` type

### LINKED ISSUE #795: NonNullableDeep
I'd like `NonNullable` but for deep structures. This is mentioned in #117 but it isn't clear that's the solution the issue creator is looking for so I'm creating a separate issue for this one as I'd like to see it added and used.

som-sm:
Closes #795 

Notes:
- Currently, there’s no way to remove nullables _just_ from the rest element in an array type. For instance, you can’t get `[string | null, …number[]]` from `[string | null, …(number | null)[]]`.

  Furthermore, it’s not very obvious what `KeyPaths` should be used to target the rest element. Using `number` would target all elements, not just the rest element (refer to the next point).

- Rest elements in array types behave slightly differently across versions when `KeyPaths` includes `number`. Refer the example below:
    ```ts
    // In v5.3 and below
    // `number` in `KeyPaths` DOESN'T affect the rest element
    type Test = SetNonNullableDeep<{a: [string | null, number | null, ...Array<boolean | null>]}, `a.${number}`>;
    //=> { a: [string, number, ...(boolean | null)[]]; }
    ```

    ```ts
    // In v5.4 and above
    // `number` in `KeyPaths` DOES affect the rest element
    type Test = SetNonNullableDeep<{a: [string | null, number | null, ...Array<boolean | null>]}, `a.${number}`>;
    //=> { a: [string, number, ...NonNullable<boolean | null>[]]; }
    ```

- `SetNonNullableDeep` currently doesn't deeply remove all nullables if `KeyPaths` is `any`. This will require a separate `NonNullableDeep` type, which I'll create it a separate PR.

### DIFF
--- a/test-d/set-non-nullable-deep.ts
+++ b/test-d/set-non-nullable-deep.ts
@@ -1,0 +1,158 @@
+import {expectType} from 'tsd';
+import type {SetNonNullableDeep} from '../source/set-non-nullable-deep';
+
+expectType<{a: number}>({} as SetNonNullableDeep<{a: number | null}, 'a'>);
+expectType<{a: number}>({} as SetNonNullableDeep<{a: number | undefined}, 'a'>);
+expectType<{a: number}>({} as SetNonNullableDeep<{a: number | null | undefined}, 'a'>);
+expectType<{a?: number}>({} as SetNonNullableDeep<{a?: number | null}, 'a'>);
+expectType<{a?: number}>({} as SetNonNullableDeep<{a?: number | undefined}, 'a'>);
+expectType<{a?: number}>({} as SetNonNullableDeep<{a?: number | null | undefined}, 'a'>);
+
+expectType<{a: number; b: {c: string}}>({} as SetNonNullableDeep<{a: number; b: {c: string | null}}, 'b.c'>);
+expectType<{a: number; b: {c: string}}>({} as SetNonNullableDeep<{a: number; b: {c: string | undefined}}, 'b.c'>);
+expectType<{a: number; b: {c: string}}>({} as SetNonNullableDeep<{a: number; b: {c: string | null | undefined}}, 'b.c'>);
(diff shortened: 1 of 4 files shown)


ogroppo:
Thanks for looking into this!

test-d/set-non-nullable-deep.ts
```diff
@@ -0,0 +1,158 @@
...
+{a: number | undefined; b: null | {readonly c: {1: number[] | undefined} | undefined} | {d: {1: number[] | undefined} | null}}, 'a' | `b.${'c' | 'd'}.1`
+>);
+
+// Non recursive types
+expectType<{a: {b: never} | Set<number | null>}>({} as SetNonNullableDeep<{a: {b: null} | Set<number | null>}, 'a.b'>);
+expectType<{a: {b: {c: string} | Map<string, string>}}>({} as SetNonNullableDeep<{a: {b: {c: string} | Map<string, string> | null}}, 'a.b'>);
+
+// === Arrays ===
+expectType<[string, number | null, boolean]>({} as SetNonNullableDeep<[string | null, number | null, boolean | undefined], '0' | '2'>);
+expectType<{a?: [string, number, (boolean | null)?] | null}>({} as SetNonNullableDeep<{a?: [string | undefined, number | undefined, (boolean | null)?] | null}, 'a.0' | 'a.1'>);
```
som-sm, line 106: Something really strange is happening here. If I remove the test on line 106, some tests in `set-required-deep.ts` start failing—tests that are completely unrelated to this. And there's no changes in `SetRequiredDeep` related files.

And this doesn't just fail locally, it also fails in CI.

@sindresorhus @Emiyaaaaa Do you have any idea why this could be happening?

https://github.com/user-attachments/assets/6f6de238-7f6f-4721-9e3c-3afd64a85471

sindresorhus: No idea. Probably some compiler quirk.

The minimum. Any more simplifications of this and it triggers the issue.

```ts
expectType<{a?: [string]}>({} as SetNonNullableDeep<{a?: [string]}, 'a.0'>);
```

PR #1118, opened 2025-05-01
docs: add `-D` flag to the install command

ShlomoCode:
`type-fest` is a typescript-only library, so [like typescript itself](https://www.npmjs.com/package/typescript#:~:text=latest%20stable%20version%3A-,npm%20install%20%2DD%20typescript,-For%20our%20nightly), it should be added to ``devDependencies`` and not to ``dependencies``.


sindresorhus:
Duplicate of #851

PR #1119, opened 2025-05-03
Add `ExtractStrict` type

nigeltroy:
Adds `ExtractStrict` type as per issue https://github.com/sindresorhus/type-fest/issues/222.

How it differs from PR https://github.com/sindresorhus/type-fest/pull/291:
1. Originally, the type condition for `Union` was `Union extends Type`. This meant that you always had to supply the entire type when extracting, which seemed cumbersome.

```ts
type A = { a: string; b: number };
type B = { c: boolean };
type C = A | B;

type AByA = ExtractStrict<C, { a: string }>; // FAIL when `Union extends Type`, should PASS because `a` is a valid field
```

I have fixed this by slightly modifying the type condition and have added a bunch of tests.


index.d.ts
```diff
@@ -175,3 +175,6 @@ export type {LastArrayElement} from './source/last-array-element';
 export type {GlobalThis} from './source/global-this';
 export type {PackageJson} from './source/package-json';
 export type {TsConfigJson} from './source/tsconfig-json';
+
+// Improved Built-in
```
nigeltroy, line 179: As per this comment: https://github.com/sindresorhus/type-fest/pull/291#discussion_r780633068, I have added this new section/category.

source/extract-strict.d.ts
```diff
@@ -0,0 +1,56 @@
+import type {Exact} from './exact';
+
+/**
+Extract members of a union type `Type` based on the
+fields in the given union type `Union`, where each
+union member of `Union` is only allowed to be a subset
+of some union member of `Type`.
+
+Constraint: ∀ U ∈ Union, U ⊆ T, where T ∈ Type
```
nigeltroy, line 9: I can remove if this overcomplicates things, but I found this mathematical notation pretty useful

source/extract-strict.d.ts
```diff
@@ -0,0 +1,56 @@
+import type {Exact} from './exact';
+
+/**
+Extract members of a union type `Type` based on the
+fields in the given union type `Union`, where each
+union member of `Union` is only allowed to be a subset
+of some union member of `Type`.
```
som-sm, line 7: > where each
union member of `Union` is only allowed to be a subset
of some union member of `Type`.

This is _not_ true, consider the following example:
```ts
type T = Extract<{foo: 1; bar: 1} | {baz: 1}, {foo: 1}>;
```
The above instantiation of `Extract` is valid but `{foo: 1}` is neither a subset of `{foo: 1; bar: 1}` nor a subset of `{baz: 1}`. It is infact a superset of `{foo: 1; bar: 1}`.

source/extract-strict.d.ts
```diff
@@ -0,0 +1,56 @@
...
+*/
+export type ExtractStrict<
+	Type,
+	/**
+	 * Only allow keys that are in some union member
+	 * of `Type`. Thus, each union member of `Union`
+	 * is only allowed to be a subset of some union
+	 * member of `Type`.
+	 */
+	Union extends Partial<Exact<Type, Union>>,
```
som-sm, line 55: Also, is `ExtractStrict` not meant to be used in cases like these:

```ts
type Line =
	| {x1: number; y1: number; x2: number; y2: number}
	| [[x1: number, y1: number], [x2: number, y2: number]]
	| [x1: number, y1: number, x2: number, y2: number];
	
type LineArrayNotations = ExtractStrict<Line, unknown[]>;
//   ^? type LineArrayNotations = [[x1: number, y1: number], [x2: number, y2: number]] | [x1: number, y1: number, x2: number, y2: number];
```

The above instantiation of `ExtractStrict` errors even though `LineArrayNotations` is not `never`.

som-sm: Can the strictness be simplified to just ensure if `Union` is able to successfully extract something from `Type`, like:

```ts
type ExtractStrict<
  Type,
  Union extends Extract<Type, Union> extends never ? never : unknown,
> = Extract<Type, Union>;

type T1 = ExtractStrict<1 | 2 | 3 | 4, 1 | 2 | 3>;
//   ^? type T1 = 1 | 2 | 3

// @ts-expect-error
type T2 = ExtractStrict<1 | 2 | 3 | 4, "one" | "two">;

// But something like this will also pass
type T3 = ExtractStrict<1 | 2 | 3 | 4, 1 | 2 | "three">;
//   ^? type T3 = 1 | 2
```
Playground: https://tsplay.dev/wQDG1W

som-sm: Or maybe we can ensure that every member of `Union` is able to do some extraction from `Type`, like:

```ts
type ExtractStrict<
  Type,
  Union extends [Union] extends [
    Union extends unknown ? (Extract<Type, Union> extends never ? never : Union) : never,
  ]
    ? unknown
    : never,
> = Extract<Type, Union>;

type T1 = ExtractStrict<1 | 2 | 3 | 4, 1 | 2 | 3>;
//   ^? type T1 = 1 | 2 | 3

// @ts-expect-error
type T2 = ExtractStrict<1 | 2 | 3 | 4, 1 | 2 | "three">;
```
Playground: https://tsplay.dev/Wz0MRN

source/extract-strict.d.ts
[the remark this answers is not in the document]
nigeltroy: Thank you for the detailed review!

> The above instantiation of Extract is valid, but {foo: 1} is neither a subset of {foo: 1; bar: 1} nor a subset of {baz: 1}. It is infact a superset of {foo: 1; bar: 1}.

You're right here in terms of the type, but I'm open to amending the language to whatever you think makes sense to most users. I was thinking a subset of fields, not of the type. Let `U` be a union member of `Union`, `T` be a union member of `Type`. For a type, `U` should be a superset of `T` (because `U` is more loose than `T`), but in terms of fields of `U`, the fields are a subset of the fields of `T` (basically, a `Partial`). I can try to make either approach more clear, or am open to whatever language you think makes sense.

> If I understand the ask correctly, 'foo' should ideally not be allowed here, because it doesn't do any extraction.

100%. This is a great test case for me to add and calls for a better `extends` condition for `Union`

source/extract-strict.d.ts
nigeltroy: I like your idea of additionally checking `Extract<Type, Union> extends never`. Let me add that in and see what works best. Thanks!

som-sm: @nigeltroy I'd rather prefer the [second approach](https://github.com/sindresorhus/type-fest/pull/1119#discussion_r2072332942) that I suggested.

source/extract-strict.d.ts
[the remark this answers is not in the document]
som-sm: IMO, it'll be difficult to clearly explain this using technical terms, so it'd be better if we keep the text simple and explain it with examples. Like:

````ts
/**
A stricter version of {@link Extract} that ensures every member of `U` can successfully extract something from `T`.

For e.g., `StrictExtract<string | number | boolean, number | bigint>` will error because `bigint` cannot extract anything from `string | number | boolean`.

@example
```
// Valid Examples

type Example1 = ExtractStrict<{status: 'success'; data: string[]} | {status: 'error'; error: string}, {status: 'success'}>;
//=> {status: 'success'; data: string[]}

type Example2 = ExtractStrict<'xs' | 's' | 'm' | 'l' | 'xl', 'xs' | 's'>;
//=> 'xs' | 's'

type Example3 = ExtractStrict<{x: number; y: number} | [number, number], unknown[]>;
//=> [number, number]
```

@example
```
// Invalid Examples

type Example1 = ExtractStrict<'xs' | 's' | 'm' | 'l' | 'xl', 'xl' | 'xxl'>;
//                                                           ~~~~~~~~~~~~
//=> Error: Type "'xl' | 'xxl'" does not satisfy the constraint 'never'.

type Example2 = ExtractStrict<{x: number; y: number} | {x: string; y: string}, unknown[]>;
//                                                                             ~~~~~~~~~
//=> Error: Type 'unknown[]' does not satisfy the constraint 'never'.
```

@category Improved Builtin
*/
````

som-sm:
@nigeltroy I've updated the PR with the suggested changes, please have a look!

index.d.ts
som-sm: @sindresorhus Do you still want to introduce a new category? If so, maybe we should consider moving some other types to this category as well.

nigeltroy:
hey, thanks! I've been busy with work so didn't get the chance to fully implement all the changes, so thank you for going ahead and doing it! I will take a look later today

index.d.ts
sindresorhus: Yeah, makes sense.

PR #1121, opened 2025-05-06
SetRequiredDeep: fix incorrect docs

larrifax:
The docblock for `SetRequiredDeep` insinuates that it changes the requiredness of the entire path given to it, while in reality it just changes the leaf.


som-sm (approved):
Nice catch!

PR #1124, opened 2025-05-08
`Split`: Enable `StrictLiteralChecks` option by default

som-sm:
Completes point no. 7 in #450.


som-sm:
@sindresorhus Would it be better to create separate PRs for each task, or shall I group certain tasks together, like updating defaults?

sindresorhus:
One PR per thing would be preferable to make it easy to generate the release notes later on.

PR #1129, opened 2025-05-08
export jsonify local types

### LINKED ISSUE #1128: Infer Jsonify get never for nested object
TS verision: 5.8.3

```ts
type Tag = {
  name: string;
  create: Date;
};

type User = {
  id: string;
  date: Date;
  // if no nested object, infer `Jsonify` works
  tag: Tag[];
};

// if change `Jsonify` to `JsonifyObject`, later infer works
type Format<T extends object> = RequiredDeep<Jsonify<T>>;

type FormatUser = Format<User>;

type InferFormat<T> = T extends Format<infer U> ? U : never;

type InferUser = InferFormat<FormatUser>;
```

In above example the final type of `InferUser` is `never`. The output of `FormatUser` is direct referring `JsonifyObject` which is local type in `jsonify.ts`.

```ts
type FormatUser = {
  id: string;
  date: string;
  tag: RequiredObjectDeep<
    JsonifyObject<{
      name?: string;
      create?: Date;
    }>
  >[];
};
```

After export `JsonifyObject` and replace to use it, the final infer works as expected.  

Didn't dig into detail how to make `Jsonify` work with nested object, but make those local types be exported is somehow good enough.

NeoJRotary:
closes #1128

### DIFF
--- a/source/jsonify.d.ts
+++ b/source/jsonify.d.ts
@@ -9,13 +9,13 @@ import type {TypedArray} from './typed-array.d.ts';
 import type {UnknownArray} from './unknown-array.d.ts';
 
 // Note: The return value has to be `any` and not `unknown` so it can match `void`.
-type NotJsonable = ((...arguments_: any[]) => any) | undefined | symbol;
+export type NotJsonable = ((...arguments_: any[]) => any) | undefined | symbol;
 
-type NeverToNull<T> = IsNever<T> extends true ? null : T;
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks for the suggestion, but I'm going to close as this as it is not the correct solution. Exporting from the file does nothing unless they are also re-exporting in the index. But still not the correct solution. It's also missing a test to prove it fixes the problem. Lets continue in the issue.

PR #1132, opened 2025-05-09
`Schema`: Remove `undefined` as a valid value for `recurseIntoArrays` option

som-sm:
Completes point no. 11 of #450.

<br>

Also, discovered the following issues with `Schema`:
- Adds extraneous type for nested structures

  ```ts
  type Current = SimplifyDeep<Schema<{foo: {bar: number; baz: number}}, boolean>>;
  //   ^? type Current = {
  //          foo: boolean | {
  //              bar: boolean;
  //              baz: boolean;
  //          };
  //      }
  
  type Ideal = SimplifyDeep<Schema<{foo: {bar: number; baz: number}}, boolean>>;
  //   ^? type Ideal = {
  //          foo: {
  //              bar: boolean;
  //              baz: boolean;
  //          };
  //      }
  ```
  
  Not sure if this is intended, probably not, because I don't see any tests verifying this behaviour.
  
- Doesn't preserve array structures
  ```ts
  type Current = SimplifyDeep<Schema<{foo: [string, string]}, boolean>>;
  //   ^? type Current = {
  //          foo: boolean[];
  //      }
  
  type Ideal = SimplifyDeep<Schema<{foo: [string, string]}, boolean>>;
  //   ^? type Ideal = {
  //          foo: [boolean, boolean];
  //      }
  ```
  
Will create separate PRs for both!

### DIFF
--- a/source/schema.d.ts
+++ b/source/schema.d.ts
@@ -11,25 +11,22 @@ export type SchemaOptions = {
 	```
 	import type {Schema} from 'type-fest';
 
-	type Form = {
-		id: number;
-		personal: {
-			name: string;
-			email: string;
-		};
-		skills: string[];
+	type Participants = {
+		attendees: string[];
+		speakers: string[];
 	};
 
-	type FormFieldVisibility = Schema<Form, boolean, {recurseIntoArrays: false}>;
+	type ParticipantsWithMetadata = Schema<Participants, {id: number; name: string}, {recurseIntoArrays: true}>;
+	//=> {
+	// 	attendees: Array<{id: number; name: string}>;
+	// 	speakers: Array<{id: number; name: string}>;
+	// };
 
-	const formFieldVisibility: FormFieldVisibility = {
-		id: true,
-		personal: {
-			name: true,
-			email: false,
-		},
-		skills: false,
-	};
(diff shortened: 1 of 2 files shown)


som-sm:
> Not sure if this is intended, probably not, because I don't see any tests verifying this behaviour.

  Looks like this is intentional, but it still feels incorrect to me!
  https://github.com/sindresorhus/type-fest/blob/a6efbe0a438b75d4dcef60f15ebea452f0b873c4/source/schema.d.ts#L82
  
  <br>
  
  And if the current behaviour is ideal, then it should be consistent with arrays as well.
  ```ts
  type Current = SimplifyDeep<Schema<{foo: string[]}, boolean>>;
  //   ^? type Current = {
  //          foo: boolean[];
  //      }
  
  type Consistent = SimplifyDeep<Schema<{foo: string[]}, boolean>>;
  //   ^? type Consistent = {
  //          foo: boolean | boolean[];
  //      }  
  ```
  @sindresorhus WDYT?

sindresorhus:
> Looks like this is intentional, but it still feels incorrect to me!

Yeah, seems incorrect to me too.

> Doesn't preserve array structures

👍

PR #1134, opened 2025-05-11
`ApplyDefaultOptions`: Remove `strictNullChecks: false` workaround

som-sm:
Completes point no. 14 of #450.

> [!NOTE]
> This means that none of the types that have options would work properly if `strictNullChecks` is disabled.

@sindresorhus If you feel this is too much, we can skip this change—the workaround isn't that bad.


sindresorhus:
I think it's fine. We don't test on `strictNullChecks: false`, so it's likely a lot other types are not working either.

Issue #788, opened 2024-01-02
IfType type guards


ZachHaber:
The readme mentions that there should be a corresponding `IfType` guard for every `IsType` check.

> For every `IsT` type (e.g. `IsAny`), there is an associated `IfT` type that can help simplify conditional types.

I'm only seeing a few `IfType` guards in the package (`IfNever`, `IfAny`, and `IfUnknown`).

Potential fixes (that I can think of at the moment):
1. Add the missing `IfType` guards
2. Update the readme to specifically call out those three guards

sindresorhus:
PR welcome for types starting with `Is` that are missing an `If` type.

sindresorhus:
I'm starting to question whether it was the correct choice to add `If` equivalents. It seemed nice at first, but there is now so many extra types and all of them require extra docs and maintenance. Maybe we should instead have supported the extra parameters on `Is*` or just referred people to a `If` type. Thoughts?

// @ikelax

sindresorhus:
// @Emiyaaaaa @voxpelli

ikelax:
Sorry for the late response.

I also think that the `If` types create a lot of extra work. For me, it would be fine if we support the extra parameters on `Is*` or just refer people to a `If` type. However, I would prefer later because the naming helps a lot.

ikelax:
// @sindresorhus

voxpelli:
> I'm starting to question whether it was the correct choice to add `If` equivalents. It seemed nice at first, but there is now so many extra types and all of them require extra docs and maintenance. Maybe we should instead have supported the extra parameters on `Is*` or just referred people to a `If` type. Thoughts?

I'm 👍 on this, and eg. the `IfAny` already works as `IsAny` if no values are given:

```ts
type IfAny<T, TypeIfAny = true, TypeIfNotAny = false>
```

Though internally it uses `IsAny`:

https://github.com/sindresorhus/type-fest/blob/3d473d1f8b0ec5820e1f3aaf7965b68784423778/source/if-any.d.ts#L22-L24

I would deprecate the `Is*` and refer people to the `If*`

sindresorhus:
> I would deprecate the Is* and refer people to the If*

Why not the other way around? `Is` instead of `If`?

voxpelli:
[...]

PR #1135, opened 2025-05-12
Remove `If*` types in favour of a single `If` type

som-sm:
Completes point no. 16 of #450.

Removes all `If*` types (`IfAny`, `IfNever`, `IfUnknown`, `IfNull` & `IfEmptyObject`) in favour of a single `If` type as suggested [here](https://github.com/sindresorhus/type-fest/issues/788#issuecomment-2871115920).

NOTE:
- I have kept the `IfNotAnyOrNever` type because I feel it's still a nice abstraction as it handles two checks, and the defaults of `any` and `never` are mostly useful.


readme.md
```diff
@@ -375,6 +351,7 @@ type ShouldBeNever = IfAny<'not any', 'not never', 'never'>;
 - `SetValues` - See [`IterableElement`](source/iterable-element.d.ts)
 - `PickByTypes` - See [`ConditionalPick`](source/conditional-pick.d.ts)
 - `HomomorphicOmit` - See [`Except`](source/except.d.ts)
+- `IfAny` / `IfNever` / `If*` - See [`If`](source/if.d.ts)
```
som-sm, line 354: Not sure if this the perfect place for mentioning this. `If` can be used as an alternative for `IfAny`, but it's not a one-to-one replacement.

readme.md
```diff
@@ -375,6 +351,7 @@ type ShouldBeNever = IfAny<'not any', 'not never', 'never'>;
 - `SetValues` - See [`IterableElement`](source/iterable-element.d.ts)
 - `PickByTypes` - See [`ConditionalPick`](source/conditional-pick.d.ts)
 - `HomomorphicOmit` - See [`Except`](source/except.d.ts)
+- `IfAny`, `IfNever`, `If*` - See [`If`](source/if.d.ts)
```
som-sm, line 354: Not sure if this the perfect place for mentioning this. `If` can be used as an alternative for `IfAny`, but it's not a one-to-one replacement.

sindresorhus:
> I have kept the IfNotAnyOrNever type because I feel it's still a nice abstraction as it handles two checks, and the defaults of any and never are mostly useful.

👍

readme.md
sindresorhus: It's fine 👍

sindresorhus:
The deprecated comment could say it will be removed in the next major version.

som-sm:
> I think it would be nice to just keep alias to the existing `If*` types using the `If` type with a deprecated doc comment, in index.d.ts. That would make it easier for people to migrate.

Yeah, I too thought so, will update the PR.

source/if-any.d.ts
```diff
@@ -3,6 +3,8 @@ import type {IsAny} from './is-any.d.ts';
 /**
 An if-else-like type that resolves depending on whether the given type is `any`.
 
+@deprecated This type will be removed in the next major version. Use the {@link If} type instead.
+
 @see {@link IsAny}
```
som-sm, line 8: Note: The `If` link is still not clickable, because `If` is not imported in this file.

And this happens at many places (like the `IsAny` link), I'll look into this separately.

som-sm:
Updated the PR to just deprecate the `If*` types.

Properly gets flagged while importing these types:
image

PR #1144, opened 2025-05-20
`SetRequired` / `SetOptional` / `SetReadonly`: Fix behaviour with functions containing properties

som-sm:
Similar to #1108, this PR fixes behaviour of `SetRequired`, `SetOptional` and `SetReadonly` types when instantiated with functions containing properties.

```ts
type Current = SetRequired<{ (x: string): number; p1?: number; p2: number }, 'p1'>; // The call signature gets lost
//   ^? type Current = { p2: number; p1: number; }

type Updated = SetRequired<{ (x: string): number; p1?: number; p2: number }, 'p1'>; // The call signature is preserved
//   ^? type Updated = { (x: string): number; p2: number; p1: number; }  
```

Functions w/o properties are now returned as is, instead of being returned as an empty object.
```ts
type Current = SetRequired<(x: string): number, never>;
//   ^? type Current = {}

type Updated = SetRequired<(x: string): number, never>;
//   ^? type Updated = (x: string): number
```

<br>

Points to discuss:
1. Should these types have a constraint of `object` on their first type argument, like:
    ```diff
    - type SetReadonly<BaseType, Keys extends keyof BaseType>
    + type SetReadonly<BaseType extends object, Keys extends keyof BaseType>
    ```

2. Also, there's no `SetWritable` type, it's actually called just `Writable`. IMO, we should rename it to `SetWritable`, and drop support for second type argument from `Writable`. This would make it more consistent with built-in utility types like `Partial`, `Readonly`, and `Required`, which also don’t take a second type argument.

    ```ts
    type Writable<BaseType> = SetWritable<BaseType, any>
    ```

### DIFF
--- a/source/set-optional.d.ts
+++ b/source/set-optional.d.ts
@@ -28,6 +28,12 @@ type SomeOptional = SetOptional<Foo, 'b' | 'c'>;
 @category Object
 */
 export type SetOptional<BaseType, Keys extends keyof BaseType> =
+	(BaseType extends (...arguments_: never) => any
+		? (...arguments_: Parameters<BaseType>) => ReturnType<BaseType>
+		: unknown)
+	& _SetOptional<BaseType, Keys>;
+
+type _SetOptional<BaseType, Keys extends keyof BaseType> =
 	BaseType extends unknown // To distribute `BaseType` when it's a union type.
 		? Simplify<
 		// Pick just the keys that are readonly from the base type.
--- a/source/set-readonly.d.ts
+++ b/source/set-readonly.d.ts
@@ -28,10 +28,13 @@ type SomeReadonly = SetReadonly<Foo, 'b' | 'c'>;
 @category Object
 */
 export type SetReadonly<BaseType, Keys extends keyof BaseType> =
-	// `extends unknown` is always going to be the case and is used to convert any
-	// union into a [distributive conditional
-	// type](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#distributive-conditional-types).
-	BaseType extends unknown
+	(BaseType extends (...arguments_: never) => any
+		? (...arguments_: Parameters<BaseType>) => ReturnType<BaseType>
+		: unknown)
+	& _SetReadonly<BaseType, Keys>;
+
+export type _SetReadonly<BaseType, Keys extends keyof BaseType> =
+	BaseType extends unknown // To distribute `BaseType` when it's a union type.
 		? Simplify<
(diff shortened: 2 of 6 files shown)


sindresorhus:
1. 👍
2. I agree. We should rename and remove parameter from `Writable`.

som-sm:
> 1. 👍
> 2. I agree. We should rename and remove parameter from `Writable`.

Cool, I'll open separate PRs for those.

som-sm:
> Should these types have a constraint of object on their first type argument, like:

Looks like doing this might introduce subtle gotchas. So, once the constraint is added, at certain places we'll have to intersect values with `object` to satisfy it, something like `T & object`. While this might look harmless, it can unexpectedly change the resulting type.

For ex:
```ts
type NewSetRequired<T extends object, K extends keyof T> = SetRequired<T, K>;

type Test<T, K extends keyof T> = NewSetRequired<T & object, K>;

type T1 = Test<[number?], '0'>;
//   ^? type T1 = [number, ...unknown[]]
```

Here's a simpler example:
```ts
type Test1 = SetRequired<[number?], '0'>;
//   ^? type Test1 = [number]

type Test2 = SetRequired<[number?] & object, '0'>;
//   ^? type Test2 = [number, ...unknown[]]
```

While this issue can be avoided by using a conditional instead of intersecting with `object`, it's an easy detail to miss and can lead to confusion.
```ts
type NewSetRequired<T extends object, K extends keyof T> = SetRequired<T, K>;

type Test<T, K extends keyof T> = T extends object
  ? NewSetRequired<T, K>
  : never;

type T1 = Test<[number?], '0'>;
//   ^? type T1 = [number]
```

I’ll explore this more, but at this point, I think a better approach would be to leave out the `object` constraint and instead ensure that primitives are returned as-is from these types.

som-sm:
> it can unexpectedly change the resulting type.

This likely happens because when an array is intersected with `object` and passed to a mapped type, the mapped type—although homomorphic—does not preserve the array structure.

```ts
type ValuesToString<T> = {
    [P in keyof T]: string;
}

type T1 = ValuesToString<[number, number]>;
//   ^? type T1 = [string, string]

type T2 = ValuesToString<[number, number] & object>;
//   ^? type T2 = { [x: number]: string; 0: string; 1: string; length: string; toString: string; toLocaleStr…
```

sindresorhus:
I think a summary of what you wrote above would be useful as a code comment on these types on why they don't constrain `object`, so we don't try to do it again in the future.

PR #1145, opened 2025-05-20
Add `LoosenUnion` type

benzaria:
Create a union by removing all Non-Literal primitive types from a union, while retaining literal types.

This utility helps you extract only the literal members from a "literal union" type
(e.g., `'foo' | 'bar' | string` becomes `'foo' | 'bar'`), saving you from defining separate types for literals and unions.

It works with all primitive and tagged types, and supports two strictness modes:

- **Strict mode (`Strict = true`)**: Removes any infinite signature type (e.g., `abc${string}`, `123${number}`), keeping only genuine literals.
- **Non-strict mode (`Strict = false`)**: Removes only wide primitive types (e.g., `string`, `number`).

default: false

```ts
import type { LoosenUnion } from 'type-fest';

// String example:
type Pet = LiteralUnion<'dog' | 'cat' | `${string}Dog`, string>;
//    ^? type Pet = 'dog' | 'cat' | `${string}Dog` | (string & {})
type PetLiteralNonStrict = LoosenUnion<Pet>;
//    ^? type PetLiteralNonStrict = 'dog' | 'cat' | `${string}Dog`
type PetLiteralStrict = LoosenUnion<Pet, true>;
//    ^? type PetLiteralStrict = 'dog' | 'cat'

// Number example:
type Nums = LiteralUnion<0 | 1 | 2, number>;
//    ^? type Nums = 0 | 1 | 2 | (number & {})
type NumsLiteral = LoosenUnion<Nums>;
//    ^? type NumsLiteral = 0 | 1 | 2

// Symbol example:
declare const sym1: unique symbol;
declare const sym2: unique symbol;
type Symbols = LiteralUnion<typeof sym1 | typeof sym2, symbol>;
//    ^? type Symbols = typeof sym1 | typeof sym2 | (symbol & {})
type SymbolsLiteral = LoosenUnion<Symbols>;
//    ^? type SymbolsLiteral = typeof sym1 | typeof sym2

// BigInt example:
type Big = LiteralUnion<1n | 2n, bigint>;
//    ^? type Big = 1n | 2n | (bigint & {})
type BigLiteral = LoosenUnion<Big>;
//    ^? type BigLiteral = 1n | 2n

```


sindresorhus:
The name is incorrect. Should be something like `ExtractLiterals`.

readme.md
```diff
@@ -122,6 +122,7 @@ Click the type names for complete docs.
 - [`UndefinedOnPartialDeep`](source/undefined-on-partial-deep.d.ts) - Create a deep version of another type where all optional keys are set to also accept `undefined`.
 - [`ReadonlyDeep`](source/readonly-deep.d.ts) - Create a deeply immutable version of an `object`/`Map`/`Set`/`Array` type. Use [`Readonly<T>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#readonlytype) if you only need one level deep.
 - [`LiteralUnion`](source/literal-union.d.ts) - Create a union type by combining primitive types and literal types without sacrificing auto-completion in IDEs for the literal type part of the union. Workaround for [Microsoft/TypeScript#29729](https://github.com/Microsoft/TypeScript/issues/29729).
+- [`LoosenUnion`](source/loosen-union.d.ts) - Create a union by removing all Non-Literal primitive types from a union, while retaining literal types.
```
sindresorhus, line 125: Needs a better description.

benzaria: i will work on that 👍

benzaria:
> ExtractLiterals

What about `literalOf` ? that seems good enough

source/literal-of.d.ts
som-sm, line 1: @benzaria Fails if the union is created using `{}` instead of `Record<never, never>`.

```ts
type Test = LoosenUnion<'foo' | 'bar' | (string & {})>;
//   ^? type Test = 'foo' | 'bar' | (string & {})
```

The implementation shouldn't be tied to `Record<never, never>`.

source/literal-of.d.ts
```diff
@@ -0,0 +1,93 @@
...
+			? Not<IsLiteral<
+			K extends TagContainer<{[J in PropertyKey]: any}>
+				? UnwrapTagged<K>
+				: K
+			>>
+			: IsEqual<K, (P & Record<never, never>)>
+	) extends true
+		? never
+		: K
+}>;
```
som-sm, line 80: The implementation could be simplified to this:

```ts
export type LoosenUnion<LiteralUnion extends Primitive> =
LiteralUnion extends unknown
	? (
		IsLiteral<
		LiteralUnion extends TagContainer<{[J in PropertyKey]: any}>
			? UnwrapTagged<LiteralUnion>
			: LiteralUnion
		>
	) extends true
		? LiteralUnion
		: never
	: never;
```

source/literal-of.d.ts
som-sm: Also doesn't work properly with tagged types when `strict` is `false`.

```ts
type T = LoosenUnion<Tagged<LiteralUnion<'hover' | 'click', string>, 'events'>>;
//   ^? type T = Tagged<LiteralUnion<'hover' | 'click', string>, 'events'>
```
`T` should have instead been `Tagged<'hover' | 'click', 'events'>`.

source/literal-of.d.ts
```diff
@@ -0,0 +1,93 @@
...
+/**
+Creates a union of the literal members from a given union type, removing wide primitive or infinite types.
+
+This utility helps you extract only the literal members from a "literal union" type
+(e.g., `'foo' | 'bar' | string` becomes `'foo' | 'bar'`), saving you from defining separate types for literals and unions.
+
+It works with all primitive and tagged types, and supports two strictness modes:
+
+- **Strict mode (`Strict = true`)**: Removes any infinite signature type (e.g., `abc${string}`, `123${number}`).
+- **Non-strict mode (`Strict = false`)**: Removes only wide primitive types (e.g., `string`, `number`).
```
som-sm, line 17: This is _not_ how we handle options. Please refer to other types in the repo to see how we do it.

Search for `ApplyDefaultOptions`.

som-sm:
> The name is incorrect. Should be something like `ExtractLiterals`. Same with the docs and descriptions.

I agree that `ExtractLiterals` makes more sense.

> And I think it should be strict by default.

Also, agree on this.

source/literal-of.d.ts
benzaria: although the `ApplyDefaultOptions` type is useful in multiple options in this case there only one which is Strict and a syntax like `ExtractLiterals<Union, false>` in cleaner that `ExtractLiterals<Union, {strict: false>>`

IDK u tell me

source/literal-of.d.ts
benzaria: > Maybe `IsLiteral` itself should accept a `strict` option, but I guess it's fine for now.

U know what that actually what i was sort of thinking, i was hopping that `IsPrimitive` type can detect only pure Primitive types like `string`, But what i found is that it match any literal too which made no sense it's just an `extends` in disguise. So i just fixed That.

benzaria:
Hey guys👋, any updates or reviews ?

som-sm:
@benzaria Looks like the PR does so many things now. Anyways, here are my thoughts:

1. I don't think `IsPrimitive` should return `false` for primitive literals. After all, primitive literals are still primitives. At the very least, that shouldn’t be the default behaviour.
    ```ts
    type T = IsPrimitive<"foo">;
    //   ^? type T = false
    ```

2. > Maybe `IsLiteral` itself should accept a `strict` option, but I guess it's fine for now. 
    
    To clarify, what I meant is that we should consider having a `strict` option not just in `ExtractLiterals`, but also in `IsLiteral` and `IsStringLiteral`. Ideally, the `strict` flag from `ExtractLiterals` should flow through to `IsLiteral`, and from there to `IsStringLiteral`. That way, the behaviour would remain consistent across all related types.

For now, maybe we can keep it simple and just add `ExtractLiterals` without any options, like [this](https://github.com/sindresorhus/type-fest/pull/1145#discussion_r2100132492), and introduce the `strict` option in a follow-up PR.

<hr>

Also, just curious, do you have any use cases where treating infinite string types like ` ``on${string}`` `, `` `${number}` ``, or `Uppercase<string>` as literals (i.e. `strict: false`) would actually be preferred?

som-sm (commented):
Please try to keep the PR focused on a single thing, bundling too many changes together makes it harder to review and understand the overall impact.

benzaria:
@som-sm 

So After thinking `IsLiteral` should have a strict options But also for `IsPrimitive` and they both will be strict = true

PR #1151, opened 2025-05-24
Enforce `.d.ts` extension for imports and source files

benzaria:
# Commit

 - This commit migrates test scripts to TypeScript.
 - Adds argument parsing and utilities.
 - Enforces `.d.ts` extensions for source files.
 - Ensures import statements reference `.d.ts` files.
 - Updates dependencies and scripts to support the new workflow.

<details>
  <summary><h2>See: Scripts In Action ?<h2></summary>

### Import-files-extension
Screenshot 2025-05-24 145701
Screenshot 2025-05-24 152841

### Source-file-extension
Screenshot 2025-05-24 161151
Screenshot 2025-05-24 161402

</details>

# Current type-fest stat:

Screenshot 2025-05-24 162052

PR #1152, opened 2025-05-25
Add lint rule to prevent `.js` / `.ts` extensions in import statements

som-sm:
Noticed in #1151 that a lot of our recent changes had imports using the `.js` extension, like:
- https://github.com/sindresorhus/type-fest/blob/eb377995637c1c63b7d18194f4e4636a2909c4eb/source/array-tail.d.ts#L1
- https://github.com/sindresorhus/type-fest/blob/eb377995637c1c63b7d18194f4e4636a2909c4eb/source/conditional-keys.d.ts#L1-L2

<br>

This is because, after the `V5` changes, VSCode auto-imports now add a `.js` extension:
Screenshot 2025-05-25 at 4 26 05 PM

<br>

This PR adds lint rule to prevent `.js` and `.ts` extensions in import statements. And, if there's no extension, TS already complains, so that's already covered.

Note: Every time VSCode auto-imports, we'll have to manually correct it, that could be painful!


benzaria:
@som-sm 
> Note: Every time VSCode auto-imports, we'll have to manually correct it, that could be painful!


Love what u added, and it will be nice not having to change all extension manually, a script that checks for it and fix can be helpful. #1151

sindresorhus:
> Note: Every time VSCode auto-imports, we'll have to manually correct it, that could be painful!

This sounds like a VSCode bug and should be reported there. It should not add imports for files that don't exist.

som-sm:
@benzaria Even with a custom script, the DX wouldn't improve much. Ideally, this should have been autofixable, so we could auto-import, hit save, and have everything fixed automatically.

som-sm:
> This sounds like a VSCode bug and should be reported there. It should not add imports for files that don't exist.

VSCode probably suggests a `.js` extension because when `moduleResolution` is set to `node16` or `nodeNext`, TS wants relative import paths to have an extension, and having a `.js` import (although non-existent) makes sense because in the compiled output there would be a corresponding `.js` file. And this is also something that TS itself suggests.

Screenshot 2025-05-26 at 11 38 51 PM<br>



However, in case of a `.d.ts` file, there wouldn't be a corresponding `.js` file, so suggesting a `.js` import wouldn't make sense anymore. And, TS probably senses this, because if you look at the screenshot below, you'll notice that when importing from a `.d.ts` file, it doesn't have the `Did you mean './foo.js'?` remark.

Screenshot 2025-05-26 at 11 36 27 PM

sindresorhus:
Wonder if it would help with https://www.typescriptlang.org/tsconfig/#allowImportingTsExtensions

PR #1153, opened 2025-05-25
`IsLiteral`: Adding `strictStringCheck` option

benzaria:
Adding `strictStringCheck` option to control the behavior of `IsStringLiteral` against wide patterns such as template literals (`on${string}`), or utility type results like `Uppercase<string>`.

 - When `strictStringCheck` is `true`, types like `on${string}` or `Uppercase<string>` are considered **not** string literals.
 - When set to `false`, those cases are considered **valid** literals.

Fixes #1145

### DIFF
--- a/source/is-literal.d.ts
+++ b/source/is-literal.d.ts
@@ -12,14 +12,14 @@ Supports all {@link IsStringLiteralOptions} options.
 
 @see {@link IsLiteral}
 */
-type IsLiteralOptions =
+export type IsLiteralOptions =
 	& IsStringLiteralOptions;
 //  & Is*LiteralOptions...
 
 /**
 @see {@link IsStringLiteral}
 */
-type IsStringLiteralOptions = {
+export type IsStringLiteralOptions = {
 	/**
 	Controls whether wide string patterns such as template literals (`on${string}`),
 	or utility type results like `Uppercase<string>`, should be treated as **non-literals**.
@@ -164,7 +164,6 @@ type L2 = Length<`${number}`>;
 //=> number
 ```
 
-@see IsStringPrimitive
 @category Type Guard
 @category Utilities
 */
@@ -230,7 +229,6 @@ endsWith('abc123', end);
 //=> boolean
 ```
 
-@see IsNumericPrimitive
 @category Type Guard
 @category Utilities
 */
@@ -271,7 +269,6 @@ const eitherId = getId({asString: runtimeBoolean});
 //=> number | string
 ```
 
-@see IsBooleanPrimitive
 @category Type Guard
 @category Utilities
 */
@@ -307,7 +304,6 @@ get({[symbolValue]: 1} as const, symbolValue);
 //=> number
 ```
 
-@see IsSymbolPrimitive
 @category Type Guard
 @category Utilities
 */


som-sm:
@benzaria Please keep the PRs focused on a single change. In this case, just add the `strict` option to `IsLiteral` and `IsStringLiteral`, and avoid fixing any unrelated types in the same PR.

Keeping changes atomic makes it much easier to review and provide feedback. Hope that makes sense!

test-d/is-literal.ts
```diff
@@ -75,12 +91,32 @@ expectType<IsStringLiteral<`ab${'c' | 'd' | 'e'}`>>(true);
 expectType<IsStringLiteral<Uppercase<'a' | 'b'> | 'C' | 'D'>>(true);
 expectType<IsStringLiteral<Lowercase<'xyz'> | Capitalize<'abc'>>>(true);
 
-// Strings with union of literals and non-literals return `boolean`
-expectType<IsStringLiteral<Uppercase<string> | 'abc'>>({} as boolean);
-expectType<IsStringLiteral<Lowercase<string> | 'Abc'>>({} as boolean);
-expectType<IsStringLiteral<null | '1' | '2' | '3'>>({} as boolean);
-expectType<IsStringLiteral<1 | 2 | '3'>>({} as boolean);
-expectType<IsStringLiteral<'foo' | 'bar' | number>>({} as boolean);
```
som-sm, line 83: Why are these tests removed? And why do these cases now return `false`?

Please review the PR once to ensure that the changes don't cause any unintended side effects.

benzaria: @som-sm I Unified the behavior of all `Is*Literal`s all those types return false on diff union types 
like `IsBooleanLiteral<'abc' | true> //=> false` that was the default. Except for `IsStringLiteral` return's boolean.
So made it return false as the others.

som-sm: I think returning `boolean` in these cases makes more sense. So, types that don't currently return `boolean` need to be fixed.

However, in any case, this should be done in a separate dedicated PR.

benzaria: Okay i will remove the changes on that issues and re-push !

som-sm:
> So I can remove IsTrue from this [line](https://github.com/sindresorhus/type-fest/blob/55557e14760e43ef5da73ad343412513af3922e5/source/is-literal.d.ts#L140) to bring the boolean back but that will only be in Strict true (uses _IsStringLiteral), if it's not the LiteralCheck type will return false anyway.

Ideally, all the logic should remain in `_IsStringLiteral`, and it should receive the `ResolvedOptions` from `IsStringLiteral`.

```ts
export type IsStringLiteral<S, Options extends IsStringLiteralOptions = {}> =
	ApplyDefaultOptions<IsStringLiteralOptions, DefaultIsStringLiteralOptions, Options> extends infer ResolvedOptions extends Required<IsStringLiteralOptions>
		? IfNotAnyOrNever<S,
			_IsStringLiteral<CollapseLiterals<S extends TagContainer<any> ? UnwrapTagged<S> : S>, ResolvedOptions>,
			false, false>
		: never;

type _IsStringLiteral<S, Options extends Required<IsStringLiteralOptions>> =
	// If `T` is an infinite string type (e.g., `on${string}`), `Record<T, never>` produces an index signature,
	// and since `{}` extends index signatures, the result becomes `false`.
	S extends string
		? Options['strict'] extends false
			? LiteralChecks<S, string>
			: {} extends Record<S, never>
				? false
				: true
		: false;
```

And, this would ensure you get `boolean` even after using `LiteralChecks`, because the type would’ve been distributed before reaching `LiteralChecks`. So, the only change required here was to pass `ResolvedOptions` from `IsStringLiteral` to `_IsStringLiteral`, and then call `LiteralChecks` when `strict` is `false`.

benzaria:
@som-sm i did ur suggestion. and also later on we can modify `LiteralChecks` to do the distribution for other types.

benzaria:
@som-sm Hi, I hope u review the latest changes, and maybe take a look at #1158 as well.

benzaria:
This PRs will be ready to review after Merge of `ExtendsStrict` branch

benzaria:
> @benzaria Please address this [#1153 (comment)](https://github.com/sindresorhus/type-fest/pull/1153#issuecomment-2909025756), this PR still seems to have changes that are not related to adding strict option to `IsLiteral` and `IsStringLiteral`.

@som-sm Are u referring to the changes on `IsNumericLiteral` ?

benzaria:
@som-sm ??

source/is-literal.d.ts
```diff
@@ -1,8 +1,28 @@
...
+/**
+@see {@link IsLiteral}
+*/
+type IsLiteralOptions = {
+	/**
+	Whether to match only finite literal types.
+
+	@default true
+	*/
+	strict?: boolean;
```
som-sm: All options need examples.

source/is-literal.d.ts
```diff
@@ -47,20 +65,26 @@ LiteralChecks<1n, Numeric>
...
+			? LiteralCheck<T, LiteralUnionType>
+			: never
 	>
 );
 
 /**
 Returns a boolean for whether the given type is a `string` [literal type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types).
 
+By default, it's Strict only matching finite literals, See {@link IsLiteralOptions.strict strict} option to change this behaviour.
+
```
som-sm: Run it through AI please.

source/is-literal.d.ts
```diff
@@ -170,10 +203,16 @@ endsWith('abc123', end);
...
 @category Type Guard
 @category Utilities
 */
-export type IsNumericLiteral<T> = LiteralChecks<T, Numeric>;
+export type IsNumericLiteral<T> =
+	T extends number
+		? T extends bigint
+			? LiteralCheck<T, Numeric>
+			: LiteralCheck<T, number>
+		: LiteralCheck<T, bigint>;
```
som-sm: This looks unrelated to adding the `strict` option, move it to a separate PR.

source/is-literal.d.ts
```diff
@@ -291,10 +334,12 @@ stripLeading(str, 'abc');
...
 
+@see IsPrimitive
 @category Type Guard
 @category Utilities
 */
-export type IsLiteral<T> =
-	IsPrimitive<T> extends true
-		? IsNotFalse<IsLiteralUnion<T>>
-		: false;
+export type IsLiteral<T, Options extends IsLiteralOptions = {}> = (
```
som-sm, line 371: Use `ApplyDefaultOptions`.

source/is-literal.d.ts
```diff
@@ -111,21 +135,30 @@ type L2 = Length<`${number}`>;
...
-
-export type _IsStringLiteral<S> =
-// If `T` is an infinite string type (e.g., `on${string}`), `Record<T, never>` produces an index signature,
-// and since `{}` extends index signatures, the result becomes `false`.
-S extends string
-	? {} extends Record<S, never>
-		? false
-		: true
-	: false;
+export type IsStringLiteral<T, Options extends IsLiteralOptions = {}> = (
```
som-sm: This shouldn't accept all `IsLiteralOptions`, there should be a separate `IsStringLiteralOptions`.

source/is-literal.d.ts
som-sm: `strict` name is ambiguous here, since `IsLiteral` doesn't involve _just_ strings.

<br>

I was thinking of referencing the `IsStringLiteralOptions` in `IsLiteralOptions`, so we'd _not_ have to repeat ourself.

```ts
type IsStringLiteralOptions = {
	strict?: boolean;
};

/**
Supports all {@link IsStringLiteralOptions} options.
*/
type IsLiteralOptions = IsStringLiteralOptions;
```

In future if we add options to any other `Is*Literal` type, we can simply `&` it here and add a link, like:
```ts
/**
Supports all {@link IsStringLiteralOptions} and {@link IsNumberLiteralOptions} options.
*/
type IsLiteralOptions = IsStringLiteralOptions & IsNumberLiteralOptions;
```
Only downside is that we would have ensure there's no conflicting names. @sindresorhus WDYT?

som-sm (commented):
Please move `Fixing unpredictable behaviors` part in a separate PR.

source/is-literal.d.ts
benzaria: Sure ! after this one

source/is-literal.d.ts
benzaria: That sound amazing actualy. great thinking !

sindresorhus: 👍

benzaria: I Did it, and test it with and other options and worked flawlessly

benzaria: Done!

source/is-literal.d.ts
benzaria: Check the new doc ! and tell me if it any good

source/is-literal.d.ts
benzaria: @sindresorhus @som-sm Hey guys what should we keep the current `strictStringCheck` options name as is or change it?
here are some suggestion I'm thinking of:
 - [x] `strictStringLiteral`
 - [ ] `strictLiteralCheck`

source/is-literal.d.ts
```diff
@@ -170,6 +230,7 @@ endsWith('abc123', end);
 //=> boolean
 ```
 
+@see IsNumericPrimitive
```
sindresorhus: Does not exist and wrong syntax.

source/is-literal.d.ts
sindresorhus: > strictStringLiteral

👍

PR #1154, opened 2025-05-26
Add `AndAll` and `OrAll` types

benzaria:
**External**:
 - Adding `AndAll` Returns a boolean for whether All given types are true.
 - Adding `OrAll` Returns a boolean for whether either of All given types are true.
 - Changing: `And<A, B>`, `Or<A, B>` to use `AndAll<[A, B]>`, `OrAll<[A, B]>`.

 **Internal**:
 - Adding `IsTrue` Check if all value in giving union are true,
 - Adding `IsFalse` Check if all value in giving union are false,
 - Adding `IsAnyOrNever` Check if a value Any or Never,
 - Adding `Extends` Check if value A extends value B,

> [!NOTE]
> `Xor`, `XorAll`: depends on #1155

### DIFF
--- a/source/and.d.ts
+++ b/source/and.d.ts
@@ -1,5 +1,5 @@
 import type {IsTrue} from './internal/type.d.ts';
-import type {Includes} from'./includes.d.ts';
+import type {Includes} from './includes.d.ts';
 import type {IsAny} from './is-any.d.ts';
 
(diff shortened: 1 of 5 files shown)


benzaria:
@som-sm can u review Please. Thanks a lot!

PR #1155, opened 2025-05-26
Add `IndexOf`, `LastIndexOf`, `AllIndexOf`, `CountOf` and `ArrayReverse` types

benzaria:
**External**
 - `IndexOf` Returns the index of the first occurrence of a value in an array, or -1 if it is not present.
 - `LastIndexOf` Returns the index of the last occurrence of a value in an array, or -1 if it is not present.
 - `AllIndexOf` Return an array of indexes of all the occurrences of a value in an array, or [] if it is not present.
 - `CountOf` Return the count of all the occurrences of a value in an array, or 0 if it is not present.
 - `ArrayReverse` Reverse an array.

> [!NOTE]
> For now there is no tests or docs until approval!

### DIFF
--- a/source/array-reverse.d.ts
+++ b/source/array-reverse.d.ts
@@ -1,0 +1,7 @@
+import type { UnknownArray } from './unknown-array.d.ts'
+
+export type ArrayReverse<Array_ extends UnknownArray> = (
+	Array_ extends [...infer Head, infer Tail]
+		? [Tail, ...ArrayReverse<Head>]
+		: []
+)
--- a/source/index-of.d.ts
+++ b/source/index-of.d.ts
@@ -1,0 +1,57 @@
+import type { SplitArrayByIndex } from './array-splice.d.ts'
+import type { SubtractPositives } from './subtract.d.ts'
+import type { ArrayReverse } from './array-reverse.d.ts'
+import type { UnknownArray } from './unknown-array.d.ts'
+import type { SumPositives } from './sum.d.ts'
+import type { IsEqual } from './is-equal.d.ts'
+
+/**
+Simpler version of Sum<T, 1>, without the extra logic.
+*/
+export type Increment<T extends number> = SumPositives<T, 1>
+
+type _IndexOf<
+	Array_ extends UnknownArray, Item, 
+	FromIndex extends number = 0, 
+	Index extends number = 0
+> = (
+	Array_ extends readonly [infer Head, ...infer Tail]
+		? IsEqual<Head, Item> extends true
+			? SumPositives<Index, FromIndex>
+			: _IndexOf<Tail, Item, FromIndex, Increment<Index>>
+		: -1 // Same as `indexOf`
+)
+
+type _AllIndexOf<
+	Array_ extends UnknownArray, Item,
+	FromIndex extends number = 0,
+	Indexs extends number[] = []
+> = (
+	IndexOf<Array_, Item, FromIndex> extends infer Index extends number
+		? Index extends -1
+			? Indexs
+			: _AllIndexOf<Array_, Item, Increment<Index>, [...Indexs, Index]>
+		: never
+)
+
+export type IndexOf<
+	Array_ extends UnknownArray, Item, 
+	FromIndex extends number = 0
+> = _IndexOf<SplitArrayByIndex<Array_, FromIndex>[1], Item, FromIndex>
+	// return's never If `FromIndex > ArrayLength`
+
+export type AllIndexOf<
+	Array_ extends UnknownArray, Item, 
+	FromIndex extends number = 0
+> = _AllIndexOf<Array_, Item,FromIndex>
+
+export type LastIndexOf<
+	Array_ extends UnknownArray, Item,
+	FromIndex extends number = 0
+> = (
+	IndexOf<ArrayReverse<Array_>, Item, FromIndex> extends infer Index extends number
+		? Index extends -1
+			? -1
+			: SubtractPositives<Array_['length'], Increment<Index>>
+		: never
+)
--- a/source/internal/type.d.ts
+++ b/source/internal/type.d.ts
@@ -1,7 +1,8 @@
-import type {If} from '../if.d.ts';
-import type {IsAny} from '../is-any.d.ts';
-import type {IsNever} from '../is-never.d.ts';
 import type {Primitive} from '../primitive.d.ts';
+import type {IsNever} from '../is-never.d.ts';
+import type {IsAny} from '../is-any.d.ts';
+import type {Or} from '../or.d.ts';
+import type {If} from '../if.d.ts';
 
 /**
 Matches any primitive, `void`, `Date`, or `RegExp` value.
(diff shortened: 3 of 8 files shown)


benzaria:
@som-sm

benzaria:
@som-sm Hi, can u review this Please. Thanks !

source/array-reverse.d.ts
```diff
@@ -0,0 +1,6 @@
+import type {UnknownArray} from './unknown-array.d.ts';
+
+export type ArrayReverse<Array_ extends UnknownArray> =
+	Array_ extends [...infer Head, infer Tail]
+		? [Tail, ...ArrayReverse<Head>]
+		: [];
```
sindresorhus: Should this be:

```ts
export type ArrayReverse<Array_ extends UnknownArray> =
	Array_ extends readonly [...infer Head, infer Tail]
		? readonly [...ArrayReverse<Head>, Tail]
		: readonly [];
```

?

sindresorhus:
Looks ok to me 👍

We cannot really review in detail without tests.

---

`AllIndexOf` => `IndicesOf`
`ArrayReverse` => `Reverse`

source/index-of.d.ts
```diff
@@ -0,0 +1,60 @@
...
+import type {SubtractPositives} from './subtract.d.ts';
+import type {ArrayReverse} from './array-reverse.d.ts';
+import type {UnknownArray} from './unknown-array.d.ts';
+import type {SumPositives} from './sum.d.ts';
+import type {IsEqual} from './is-equal.d.ts';
+
+/**
+Simpler version of Sum<T, 1>, without the extra logic.
+*/
+export type Increment<T extends number> = SumPositives<T, 1>;
```
sindresorhus, line 11: Why are you exporting it? It's not used elsewhere.

source/internal/type.d.ts
```diff
@@ -13,6 +14,13 @@ Matches non-recursive types.
 */
 export type NonRecursiveType = BuiltIns | Function | (new (...arguments_: any[]) => unknown);
 
+/**
+ * Checks if one type extends another. Note: this is not quite the same as `Left extends Right` because:
+ * 1. If either type is `never`, the result is `true` iff the other type is also `never`.
+ * 2. Types are wrapped in a 1-tuple so that union types are not distributed - instead we consider `string | number` to _not_ extend `number`. If we used `Left extends Right` directly you would get `Extends<string | number, number>` => `false | true` => `boolean`.
+ */
+export type Extends<Left, Right> = IsNever<Left> extends true ? IsNever<Right> : [Left] extends [Right] ? true : false;
```
sindresorhus: I think this should be named `ExtendsStrict` or something, because it doesn't follow the typical TS "extends" behavior.

source/internal/type.d.ts
```diff
@@ -13,6 +14,13 @@ Matches non-recursive types.
 */
 export type NonRecursiveType = BuiltIns | Function | (new (...arguments_: any[]) => unknown);
 
+/**
+ * Checks if one type extends another. Note: this is not quite the same as `Left extends Right` because:
```
sindresorhus: Don't prefix with `*`

source/internal/type.d.ts
```diff
@@ -40,9 +48,19 @@ export type HasMultipleCallSignatures<T extends (...arguments_: any[]) => unknow
 		: false;
 
 /**
-Returns a boolean for whether the given `boolean` is not `false`.
+Returns a boolean for whether the given `boolean` Union containe's `false`.
```
sindresorhus: typo

sindresorhus (changes requested):

source/index-of.d.ts
benzaria: Ohh i forgot about it i was using it in a arithmetic types I'm working on to Improve `Sum`, `Substract` with a range of 10<sup>999</sup>  number, with also Multiplication, Division. do u consider adding does types or it just to much ?

source/internal/type.d.ts
benzaria: i can do that I just copy past it from exept-type

Edit: is there a chance to keep the name as is it really just fits. and considering it's a dedicated types named `Extends` contributors will not assume it behave as a normal `extends` and the Doc its just super verbose about what the type does.

source/array-reverse.d.ts
benzaria: Is the readonly in output nessecary? i think a type like :
```ts
export type Reverse<Array_ extends UnknownArray> =
	Array_ extends readonly [...infer Head, infer Tail]
		? [Tail, ...Reverse<Head>]
		: [];
```

sindresorhus: A test can answer that.

som-sm:
> For now there is no tests or docs until approval!

@benzaria If you want to understand the feasibility of a type, please consider opening an [issue first](https://github.com/sindresorhus/type-fest/blob/main/.github/ISSUE_TEMPLATE/1%20new%20type.yml). Because without proper tests it doesn't really make sense to review the PR.

source/array-reverse.d.ts
som-sm: Array types need to be tested against various different cases. Here's two cases where the current implementation fails:
```ts
type T1 = ArrayReverse<[1?, 2?]>
//   ^? type T1 = []

type T2 = ArrayReverse<[1, 2, ...number[]]>
//   ^? type T2 = []
```
Please refer some existing tests in the repo for reference.

source/internal/type.d.ts
som-sm: > 1. If either type is `never`, the result is `true` iff the other type is also `never`.

This sentence isn't very clear.

som-sm: > I think this should be named ExtendsStrict or something, because it doesn't follow the typical TS "extends" behavior.

I agree!

som-sm (changes requested):
@benzaria A proper review would only be possible when tests are added. Also, every type needs proper documentation with examples.

som-sm:
Also, please submit one type per PR, unless the types are closely related. Once you start handling various edge cases, the implementations can become quite verbose and complex, so separating them into individual PRs makes a lot of sense.

And even for simpler types, it's best to keep them in separate PRs whenever possible.

source/array-reverse.d.ts
benzaria: Ahh yes i noticed that yesterday I'm working on it

som-sm (changes requested):
> Also, please submit one type per PR, unless the types are closely related. Once you start handling various edge cases, the implementations can become quite verbose and complex, so separating them into individual PRs makes a lot of sense.
> 
> And even for simpler types, it's best to keep them in separate PRs whenever possible.

Doesn't make sense to add seven different types in one single PR. I'd rather prefer seven independent PRs.

Here’s what I’d suggest:
Pick one of the seven types, open a PR just for that, and get it reviewed and merged. Once you're familiar with the process and expectations, you can move on to opening PRs for the remaining types.

CC: @sindresorhus

benzaria:
@som-sm @sindresorhus Hey guys, I'm so sorry for the many types I made in this PRs. I know it's better to separate them in diff PRs. but when a type relies on the other it's not ideal take for example the [`Xor-XorAll`](https://github.com/sindresorhus/type-fest/pull/1156) which relay on this PRs to work and now both of them is pending approval. and `Xor-XorAll` can't pass tests unless the `CountOf` type is present.

Making a PRs for each individual type and wait to merge will take a lot of time, and will delay the contribution.

the `SplitOnSpread` was intentionally to be an internal type for making `Reverse` work, but I thought it can be useful for others. What u think guys shood we expose it ?

PR #1157, opened 2025-05-27
Migrate to XO v1.0.0

som-sm:
Initially, I assumed the migration would simply involve upgrading to flat config and checking whether any of the [newly added rules](https://github.com/xojs/xo/releases/tag/v1.0.0) conflict with our workflow.

However, it turns out there’s a bit more to it. After migrating to `v1`, I saw nearly ~400 errors—most of which came from rules not mentioned in the [release notes](https://github.com/xojs/xo/releases/tag/v1.0.0).

Here are a few quick observations:
- Post-migration, I started seeing errors from the `@typescript-eslint/no-empty-object-type` rule. This appears to be newly enabled, since the current version of `xo` doesn’t report these errors unless the rule is explicitly turned on.
  ```json
  "@typescript-eslint/no-empty-object-type": "error"
  ```
- I’m also seeing a number of errors from `@stylistic` rules. When I try to add these same rules in our current version of `xo`, I get complains that they don’t exist—suggesting that these are new additions as well.

Related: https://github.com/xojs/xo/issues/800

<br>

I’m not sure what I’m missing, but it seems there are more changes than the ones listed in the release notes. Although, I might be completely wrong.


som-sm:
Looks like there's some conflicting behaviour b/w `@stylistic/indent` and `@stylistic/indent-binary-ops`.

image
image<br><br>

But moving union in a single line fixed this. Refer [this commit](https://github.com/sindresorhus/type-fest/pull/1157/commits/0a58aa8482f8f1870ab4af649f71e64078b832b7).

image

source/paths.d.ts
```diff
@@ -205,10 +205,8 @@ type InternalPaths<T, Options extends Required<PathsOptions>> =
 								? IsNumberLike<Key> extends true
 									? `[${Key}]`
 									: (Key | ToString<Key>)
-									: never
-								|
-										Options['bracketNotation'] extends false
-							// If `Key` is a number, return `Key | `${Key}``, because both `array[0]` and `array['0']` work.
+								: Options['bracketNotation'] extends false
+								// If `Key` is a number, return `Key | `${Key}``, because both `array[0]` and `array['0']` work.
```
som-sm, line 209: This is the _only_ implementation change in this entire PR, had to do this because otherwise there were conflicting lint errors similar to the one mentioned [here](https://github.com/sindresorhus/type-fest/pull/1157#issuecomment-2912459650).

And, the `never |` is unnecessary over here.

som-sm:
[#2fd243d](https://github.com/sindresorhus/type-fest/pull/1157/commits/2fd243d1747320eb667194bf2199e791ae9b12a5) has changes as per the `@stylistic/type-generic-spacing` rule.

This is a nice rule, although it seems to not like comments after type parameters, which is weird.
image

Also, the autofix behaviour with this rule tends to remove comments, which is most likely a bug.

https://github.com/user-attachments/assets/024645f8-660a-4142-9774-807f590c9276

test-d/set-non-nullable-deep.ts
som-sm, line 1: These changes are because of the `@stylistic/max-len` rule.

sindresorhus:
Thanks for doing this 👍

PR #1161, opened 2025-06-03
`And`: Fix behaviour with `boolean`, `never` and `any`

som-sm:
This PR fixes the behavior of `And` in the following cases:

1. Currently, `And<boolean, false>` correctly returns `false`, but `And<boolean, true>` returns `never` instead of `boolean`.
```diff
  type T1 = And<true, boolean>;
- //   ^? type T1 = never
+ //   ^? type T1 = boolean

  type T2 = And<boolean, true>;
- //   ^? type T2 = never
+ //   ^? type T2 = boolean

  type T5 = And<boolean, boolean>;
- //   ^? type T5 = never
+ //   ^? type T5 = boolean
```

2. Currently, `And<true, never>` returns `never`, but I feel it should instead return `false`. In this case, it'd be better if we treat `never` as `false` rather than treating it as empty, consider this example:
```diff
  type AreBothIntegers<A, B> = And<IsInteger<A>, IsInteger<B>>;

  type T1 = AreBothIntegers<never, 1>;
- //   ^? type T1 = true
+ //   ^? type T1 = false
```

```diff
  type T1 = And<true, never>;
- //   ^? type T1 = true
+ //   ^? type T1 = false

  type T2 = And<never, true>;
- //   ^? type T2 = true
+ //   ^? type T2 = false

  type T3 = And<never, never>;
- //   ^? type T3 = true
+ //   ^? type T3 = false
```

3. Behavior with `any` is streamlined as well.
```diff
type T1 = And<false, any>;
- //   ^? type T1 = boolean
+ //   ^? type T1 = false

type T2 = And<any, false>;
- //   ^? type T2 = boolean
+ //   ^? type T2 = false

type T3 = And<true, any>;
- //   ^? type T3 = true
+ //   ^? type T3 = boolean

  type T4 = And<any, true>;
- //   ^? type T4 = true
+ //   ^? type T4 = boolean

  type T5 = And<any, any>;
- //   ^? type T5 = true
+ //   ^? type T5 = boolean
```

I'll put up a similar PR fixing the `Or` type.


sindresorhus:
Good improvements. I agree with everything 👍

PR #1164, opened 2025-06-04
Add `AllExtend` type

som-sm:
This PR fixes all the known limitations of the internal `Every` type and makes it publicly available.
https://github.com/sindresorhus/type-fest/blob/a891143da61fb71df854936bb9697b11451ee79e/source/internal/array.d.ts#L103

<br>

Introduced a new `CollapseRestElement` type to handle arrays with rest and optional elements. This utility replaces the rest element with a single element that has the same type as the rest element. For example, `CollapseRestElement<[string, ...number[], boolean]>` returns `[string, number, boolean]`.

Before `Every` performs its computation, the input array is simplified using `CollapseRestElement`. This simplification shouldn't alter the result of `Every` because the expected output for `[number, …string[], boolean]` and `[number, string, boolean]` is the same.

### DIFF
--- a/test-d/internal/not.ts
+++ b/test-d/internal/not.ts
@@ -3,5 +3,4 @@ import type {Not} from '../../source/internal/index.d.ts';
 
 expectType<Not<true>>(false);
 expectType<Not<false>>(true);
-// FIXME
-expectType<Not<boolean>>(null! as boolean);
+expectType<Not<boolean>>({} as boolean);
--- a/source/all-extend.d.ts
+++ b/source/all-extend.d.ts
@@ -1,9 +1,10 @@
 import type {If} from './if.d.ts';
 import type {CollapseRestElement} from './internal/array.d.ts';
 import type {ApplyDefaultOptions} from './internal/object.d.ts';
-import type {IfNotAnyOrNever} from './internal/type.d.ts';
+import type {IfNotAnyOrNever, Not} from './internal/type.d.ts';
 import type {IsAny} from './is-any.d.ts';
 import type {IsNever} from './is-never.d.ts';
+import type {Or} from './or.d.ts';
 import type {UnknownArray} from './unknown-array.d.ts';
 
 /**
@@ -103,11 +104,10 @@ export type AllExtend<TArray extends UnknownArray, Type, Options extends AllExte
 type _AllExtend<TArray extends UnknownArray, Type, Options extends Required<AllExtendOptions>> = IfNotAnyOrNever<TArray, If<IsAny<Type>, true,
 	TArray extends readonly [infer First, ...infer Rest]
 		? IsNever<First> extends true
-			? IsNever<Type> extends true
+			? Or<IsNever<Type>, Not<Options['strictNever']>> extends true
+				// If target `Type` is also `never` OR `strictNever` is disabled, recurse further.
 				? _AllExtend<Rest, Type, Options>
-				: Options['strictNever'] extends true
-					? false
-					: _AllExtend<Rest, Type, Options>
+				: false
 			: First extends Type
 				? _AllExtend<Rest, Type, Options>
 				: false


sindresorhus:
Should it maybe be called `AllExtend`? I feel that is clearer than `Every` because it explicitly conveys the type-level intent: checking whether all elements extend a given type. `Every` is vague and sounds like an operation or transformation rather than a boolean predicate.

source/internal/type.d.ts
```diff
@@ -99,3 +99,10 @@ type C = IfNotAnyOrNever<never, 'VALID', 'IS_ANY', 'IS_NEVER'>;
 */
 export type IfNotAnyOrNever<T, IfNotAnyOrNever, IfAny = any, IfNever = never> =
 	If<IsAny<T>, IfAny, If<IsNever<T>, IfNever, IfNotAnyOrNever>>;
+
+/*
+Indicates the value of `exactOptionalPropertyTypes` compiler option.
+*/
+type ExactOptionalPropertyTypesEnabled = [(string | undefined)?] extends [string?]
```
sindresorhus:
```suggestion
export type ExactOptionalPropertyTypesEnabled = [(string | undefined)?] extends [string?]
```

I wonder why TS doesn't warn about this, even with ESM...

sindresorhus: And maybe prefix with `Is`

test-d/every.ts
```diff
@@ -0,0 +1,117 @@
...
+expectType<NonStrictNeverEvery<[1, 2, ...never[]], number>>(true);
+expectType<NonStrictNeverEvery<[...never[], 'a', 'b'], string>>(true);
+expectType<NonStrictNeverEvery<[never, never], never>>(true);
+expectType<NonStrictNeverEvery<[never, ...never[], never], any>>(true);
+expectType<NonStrictNeverEvery<[1, 2, never, ...never[]], any>>(true);
+expectType<NonStrictNeverEvery<[never, number], never>>(false);
+expectType<NonStrictNeverEvery<[never, any, never, any], never>>({} as boolean);
+
+expectType<Every<any, never>>(false);
+expectType<Every<never, any>>(false);
```
sindresorhus: Maybe also:

```ts
expectType<Every<unknown[], number>>({} as boolean);
expectType<Every<unknown[], any>>(true);

expectType<Every<[never, 1], any>>(true);
expectType<Every<[1, never], any>>(true);
expectType<Every<[never, ...never[], never], any>>(true);
```

source/internal/array.d.ts
benzaria, line 1: @som-sm I think u didn't take advantage of pre-existing types in the lib like `StaticPartOfArray` and `VariablePartOfArray`, I made a type that does quite the same:
https://github.com/sindresorhus/type-fest/blob/1f93d145e2472e25d3b0d9add1e43b335da046d0/source/split-on-spread.d.ts#L79-L90

```ts
type T = SplitOnSpread<[number, ...string[], boolean]>;
// => [[number], [string], [boolean]]

type U = SplitOnSpread<[...boolean[], string]>;
// => [[], [boolean], [string]]

type V = SplitOnSpread<[number, string]>;
// => [[number, string], [], []]
```

Which can be used as follows: (basic example)
```ts
export type CollapseRestElement<T extends UnknownArray> =
    SplitOnSpread<T> extends infer Result extends UnknownArray[]
      ? [
           ...Result[0],
           ...Result[1] extends readonly[] ? [] : [Result[1][number]],
           ...Result[2]
       ]
      : never
```

source/every.d.ts
```diff
@@ -0,0 +1,115 @@
...
+			? IsNever<Type> extends true
+				? _Every<Rest, Type, Options>
+				: Options['strictNever'] extends true
+					? false
+					: _Every<Rest, Type, Options>
+			: First extends Type
+				? _Every<Rest, Type, Options>
+				: false
+		: true
+>, false, false>;
```
benzaria: Don't repeat yourself.
```ts
type _Every<TArray extends UnknownArray, Type, Options extends Required<EveryOptions>> = 
    IfNotAnyOrNever<TArray, If<IsAny<Type>, true,
        TArray extends readonly [infer First, ...infer Rest]
            ? IsNever<First> extends true
                ? Or<
                    IsNever<Type>,
                    Not<Options['strictNever']>
                > extends true 
                    ? _Every<Rest, Type, Options>
                    : false
                : First extends Type
                    ? _Every<Rest, Type, Options>
                    : false
            : true
    >, false, false>;
```

source/internal/type.d.ts
som-sm: > I wonder why TS doesn't warn about this, even with ESM...

Umm... not sure why, maybe it's treating the module as `"ambient"`.

som-sm:
> Should it maybe be called `AllExtend`? I feel that is clearer than `Every` because it explicitly conveys the type-level intent: checking whether all elements extend a given type. `Every` is vague and sounds like an operation or transformation rather than a boolean predicate.

I chose the name `Every` to match the naming with `Array.prototype.every`, but then again, you can’t really use this utility to type the `every` method, so perhaps a different name would make more sense.

<br>

Updated it to `AllExtend`!

source/internal/type.d.ts
```diff
@@ -99,3 +99,10 @@ type C = IfNotAnyOrNever<never, 'VALID', 'IS_ANY', 'IS_NEVER'>;
 */
 export type IfNotAnyOrNever<T, IfNotAnyOrNever, IfAny = any, IfNever = never> =
 	If<IsAny<T>, IfAny, If<IsNever<T>, IfNever, IfNotAnyOrNever>>;
+
+/*
+Indicates the value of `exactOptionalPropertyTypes` compiler option.
+*/
+export type IsExactOptionalPropertyTypesEnabled = [(string | undefined)?] extends [string?]
+	? false
+	: true;
```
som-sm, line 108: Credits to @Emiyaaaaa for [suggesting this type](https://github.com/sindresorhus/type-fest/pull/1085#discussion_r2052018983) ✨

test-d/every.ts
som-sm: @sindresorhus Shouldn't `Every<unknown[], number>` return `false` and not `boolean`, because `unknown extends number` is not true.
```ts
type T = unknown extends number ? 'Y' : 'N';
//=> 'Y'
```

`Every<any[], number>` returns `boolean` because `any` distributes over conditionals.
```ts
type T = any extends number ? 'Y' : 'N';
//=> 'Y' | 'N'
```

source/every.d.ts
som-sm: Makes sense, updated in [f6a33a5](https://github.com/sindresorhus/type-fest/pull/1164/commits/f6a33a5f9c834cb1ad6a310111f6aac99f695f8b).

test-d/internal/not.ts
```diff
@@ -3,5 +3,4 @@ import type {Not} from '../../source/internal/index.d.ts';
 
 expectType<Not<true>>(false);
 expectType<Not<false>>(true);
-// FIXME
```
som-sm, line 6: There's nothing to fix here, `Not<boolean>` should return `boolean`.

source/internal/array.d.ts
som-sm: `StaticPartOfArray` and `VariablePartOfArray` don’t quite work the way I want for cases where there's a non-rest element after the rest element.

Examples:
```ts
type T1 = StaticPartOfArray<[...string[], number, boolean]>;
//=> []

type T2 = VariablePartOfArray<[...string[], number, boolean]>;
//=> [...string[], number, boolean]
```

Also, more importantly, I wanted special handling for optional elements, so if I used any of these utilities I'd then have to loop again and handle optional elements. So, I went for custom solution instead.

test-d/every.ts
sindresorhus: You're right 👍

PR #1165, opened 2025-06-05
Add `ExtendsStrict` type

benzaria:
A stricter version of `extends` that Checks Stricily if one type extends another without distribution.


benzaria:
@sindresorhus can u please review and merge this PRs is slowing down all the others i made. Thanks

source/extends-strict.d.ts
```diff
@@ -0,0 +1,34 @@
+import type {IsNever} from './is-never.d.ts';
+
+/**
+A stricter version of `extends` that Checks Stricily if one type extends another without distribution.
```
sindresorhus: Please run everything through ChatGPT to catch obvious typos and bugs. Also ask ChatGPT to improve descriptions. Applies to all your pull requests.

source/extends-strict.d.ts
```diff
@@ -0,0 +1,34 @@
...
+
+Note: this is not quite the same as `Left extends Right` because:
+
+1. Types are wrapped in a 1-tuple so that union types are not distributed - instead we consider `string | number` to _not_ extend `number`. If we used `Left extends Right` directly you would get `Extends<string | number, number>` => `false | true` => `boolean`. So it's return `true` if `[Left] extends [Right]`.
+
+2. Return's `true` if `Left` and `Right` are both `never`.
+
+@example
+```
```
sindresorhus, line 14: Should include import statement.

source/extends-strict.d.ts
```diff
@@ -0,0 +1,34 @@
...
+type T = ExtendsStrict<number | string, string>
```
sindresorhus:
```suggestion
type T = ExtendsStrict<number | string, string>;
```

source/extends-strict.d.ts
```diff
@@ -0,0 +1,34 @@
...
+//=> false
+
+type T = ExtendsStrict<string, number | string>
```
sindresorhus:
```suggestion
type T2 = ExtendsStrict<string, number | string>
```

sindresorhus: Code examples should be copy-pastable.

source/extends-strict.d.ts
sindresorhus: Also ask ChatGPT to:

> Add missing tests. Only things that isn't already tested.

sindresorhus: Basically, use AI as much as possible to catch mundane things, so humans can spend time on catching the hard stuff. Applies to all your pull requests.

sindresorhus (changes requested):

source/extends-strict.d.ts
benzaria: Sure !

source/extends-strict.d.ts
```diff
@@ -1,26 +1,28 @@
...
-1. Types are wrapped in a 1-tuple so that union types are not distributed - instead we consider `string | number` to _not_ extend `number`. If we used `Left extends Right` directly you would get `Extends<string | number, number>` => `false | true` => `boolean`. So it's return `true` if `[Left] extends [Right]`.
+1. Prevents distribution over union types by wrapping both types in tuples. For example, `ExtendsStrict<string | number, number>` returns `false`, whereas `string | number extends number` would result in `boolean`.
 
-2. Return's `true` if `Left` and `Right` are both `never`.
+2. Treats `never` as a special case: returns `true` if both `Left` and `Right` are `never`.
 
 @example
 ```
-type T = ExtendsStrict<number | string, string>
+import {ExtendsStrict} from 'type-fest'
```
sindresorhus:
```suggestion
import type {ExtendsStrict} from 'type-fest'
```

test-d/extends-strict.ts
```diff
@@ -1,21 +1,78 @@
...
+expectType<ExtendsStrict<string & {bar: string}, string>>(true);
+expectType<ExtendsStrict<string, string & {bar: string}>>(false);
+
+// Literal vs primitive
+expectType<ExtendsStrict<'foo', string>>(true);
+expectType<ExtendsStrict<string, 'foo'>>(false);
+
+// Arrays
+expectType<ExtendsStrict<string[], string[]>>(true);
+expectType<ExtendsStrict<string[], string[]>>(true);
```
sindresorhus: Duplicate

readme.md
```diff
@@ -296,6 +296,7 @@ Click the type names for complete docs.
 
 ### Improved built-in
 
+- [`ExtendsStrict`](source/extends-strict.d.ts) - A stricter version of `extends` that Checks Stricily if one type extends another without distribution.
```
sindresorhus: You have to update this too.

sindresorhus (changes requested):

readme.md
benzaria: Ohh yes may bad I was in a hurry !

test-d/extends-strict.ts
```diff
@@ -0,0 +1,78 @@
...
+
+// Never handling
+expectType<ExtendsStrict<never, never>>(true);
+expectType<ExtendsStrict<never, string>>(false);
+expectType<ExtendsStrict<string, never>>(false);
+
+// Any and unknown
+expectType<ExtendsStrict<any, any>>(true);
+expectType<ExtendsStrict<any, never>>(false);
+expectType<ExtendsStrict<never, any>>(false);
```
som-sm: I feel this should be `true`, everything should be assignable to `any`, including `never`.

source/extends-strict.d.ts
```diff
@@ -0,0 +1,36 @@
...
+//=> false
+
+type T2 = ExtendsStrict<string, number | string>;
+//=> true
+
+type T3 = ExtendsStrict<string, string>;
+//=> true
+
+type T4 = ExtendsStrict<never, never>;
+//=> true
```
som-sm: Add the `never` case where the behaviour differs from the `extends` keyword.

```ts
type T2 = ExtendsStrict<never, number>;
//=> false
```

som-sm (changes requested):

source/extends-strict.d.ts
benzaria: Sure !

source/extends-strict.d.ts
```diff
@@ -0,0 +1,36 @@
...
+
+Unlike the built-in `extends` keyword, `ExtendsStrict`:
+
+1. Prevents distribution over union types by wrapping both types in tuples. For example, `ExtendsStrict<string | number, number>` returns `false`, whereas `string | number extends number` would result in `boolean`.
+
+2. Treats `never` as a special case: returns `true` if both `Left` and `Right` are `never`.
+
+@example
+```
+import type {ExtendsStrict} from 'type-fest'
```
som-sm:
```suggestion
import type {ExtendsStrict} from 'type-fest';
```

test-d/extends-strict.ts
benzaria: Ohh yeah that valid

source/extends-strict.d.ts
```diff
@@ -0,0 +1,42 @@
...
+import type {IsAny} from './is-any.d.ts';
+
+/**
+A stricter, non-distributive version of `extends` for checking whether one type is assignable to another.
+
+Unlike the built-in `extends` keyword, `ExtendsStrict`:
+
+1. Prevents distribution over union types by wrapping both types in tuples. For example, `ExtendsStrict<string | number, number>` returns `false`, whereas `string | number extends number` would result in `boolean`.
+
+2. Treats `never` as a special case: returns `true` if both `Left` and `Right` are `never`.
```
som-sm: > if both `Left` and `Right` are `never`

Not completely true because `Right` could be `any` as well.

```suggestion
2. Treats `never` as a special case: `never` doesn't extend every other type, it only extends itself (or `any`). For example, `ExtendsStrict<never, number>` returns `false` whereas `never extends number` would result in `true`.
```

test-d/extends-strict.ts
```diff
@@ -0,0 +1,77 @@
...
+// Never handling
+expectType<ExtendsStrict<never, never>>(true);
+expectType<ExtendsStrict<never, string>>(false);
+expectType<ExtendsStrict<string, never>>(false);
+
+// Any and unknown
+expectType<ExtendsStrict<any, any>>(true);
+expectType<ExtendsStrict<any, never>>(true);
+expectType<ExtendsStrict<never, any>>(true);
+expectType<ExtendsStrict<any, unknown>>(true); // `any` is assignable to `unknown`
```
som-sm, line 29: Add case for `ExtendsStrict<any, number>`, that should also be `true`.

som-sm (changes requested):

source/extends-strict.d.ts
benzaria: True, the added `any` case change it.

PR #1166, opened 2025-06-05
Add `SplitOnRestElement`, `ExtractRestElement`, `ExcludeRestElement` types

benzaria:
- [`SplitOnRestElemnt`](source/split-on-rest-element.d.ts) - Splits an Array on its Rest element portion into three parts: `[[...Head], Rest[], [...Tail]]`.
- [`ExtractRestElemnt`](source/extract-rest-element.d.ts) - Returns the rest element type from and array.
- [`ExcludeRestElemnt`](source/exclude-rest-element.d.ts) - Create a array with the rest element removed.

### DIFF
--- a/source/split-on-rest-element.d.ts
+++ b/source/split-on-rest-element.d.ts
@@ -65,18 +65,11 @@ export type SplitOnRestElement<
 	Options extends SplitOnRestElementOptions = {},
 > =
 	Array_ extends unknown // For distributing `Array_`
-		? IfNotAnyOrNever<Array_,
-			_SplitOnRestElement<Array_,
-				ApplyDefaultOptions<
-					SplitOnRestElementOptions,
-					DefaultSplitOnRestElementOptions,
-					Options
-				>
-			>
-		> extends infer Result extends UnknownArray
-			? IsArrayReadonly<Array_> extends true
-				? Readonly<Result>
-				: Result
+		? IfNotAnyOrNever<Array_, _SplitOnRestElement<
+			Array_,
+			ApplyDefaultOptions<SplitOnRestElementOptions, DefaultSplitOnRestElementOptions, Options>
+		>> extends infer Result extends UnknownArray
+			? If<IsArrayReadonly<Array_>, Readonly<Result>, Result>
 			: never // Should never happen
 		: never; // Should never happen
 


test-d/extends-strict.ts
som-sm, line 1: Why is `ExtendsStrict` part of this PR? It has nothing to do with `SplitOnSpread`, `ExtractSpread` or `ExcludeSpread`.

source/split-on-spread.d.ts
som-sm, line 1: Have separate files for each publicly available type.

som-sm: Also separate test files.

test-d/extends-strict.ts
benzaria: @som-sm It's not needed here but rather in `Reverse` i just build after an other, because reverse need both ExtendsStrict and SplitOnSpread can u merge ExtendsStrict so i can rebase this branch

benzaria: Rebased!

source/split-on-spread.d.ts
benzaria: Done!

sindresorhus:
Please run everything through AI to find typos, bugs, mistakes, and if there are any missing tests.

source/extract-spread.d.ts
```diff
@@ -0,0 +1,33 @@
+import type {SplitOnSpread} from './split-on-spread.d.ts';
+import type {UnknownArray} from './unknown-array.d.ts';
+
+/**
+Return's the spread element type from and array.
```
sindresorhus: AI would have caught this:

```suggestion
Returns the spread element type from an array.
```

benzaria:
@som-sm can u review this !

benzaria:
@som-sm Ping... :)

sindresorhus:
Please run everything through AI. Same with your other pull requests.

source/extract-spread.d.ts
```diff
@@ -0,0 +1,33 @@
...
+// => boolean
+
+type T3 = ExtractSpread<[...'foo'[], true]>;
+// => 'foo'
+
+type T4 = ExtractSpread<[number, string]>;
+// => never
+```
+
+@author benzaria
```
sindresorhus: We don't use this tag.

source/extract-spread.d.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+Returns the spread element type from an array.
+
+@example
+```
+import type {ExtractSpread} from 'type-fest';
+
+type T1 = ExtractSpread<[number, ...string[], string, 'foo']>;
+// => string
```
sindresorhus:
```suggestion
//=> string
```

source/split-on-spread.d.ts
```diff
@@ -0,0 +1,88 @@
...
+export type SplitOnSpread<Array_ extends UnknownArray, Options extends SplitOnSpreadOptions = {}> = IfNotAnyOrNever<Array_,
+	DeconstructSpreadArray<Array_, ApplyDefaultOptions<SplitOnSpreadOptions, DefaultSplitOnSpreadOptions, Options>>
+>;
+
+/**
+Deconstruct an Array on its rest element and return the splited portions.
+
+See {@link SplitOnSpread SplitOnSpread} For info.
+
+@credit-to som-sm
```
sindresorhus: We don't use this tag.

benzaria: Noted ! and fixed

source/extract-spread.d.ts
benzaria: Noted ! and fixed

source/exclude-spread.d.ts
```diff
@@ -0,0 +1,36 @@
...
+// => [string]
+
+type T3 = ExcludeSpread<[...'foo'[], true]>;
+// => [true]
+
+type T4 = ExcludeSpread<[number, string]>;
+// => [number, string]
+```
+
+@author benzaria
```
som-sm: Remove this

source/extract-spread.d.ts
som-sm: It's _not_ yet fixed.

source/extract-spread.d.ts
som-sm: This is _not_ resolved yet, there shouldn't be a space b/w `//` and `=>`.

benzaria:
> @benzaria Please take the time to thoroughly check your PRs before requesting a review.

@som-sm So sorry for the mistake, we just had a **Code Blue** emergency in the block we needed to  move fast.

som-sm:
That’s fine, please take your time. Just make sure to double-check your work before requesting a review.

som-sm:
@benzaria Did you even go through anything? There are still obvious issues in the PR.

benzaria:
@som-sm I run it thru AI sanitizing any typos and bad grammar.

som-sm:
I'll try to give it a proper read tomorrow.

BTW, these types should probably be suffixed with `Rest` instead of `Spread`. The [TS handbook](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types) also refers to them as `rest` elements.
image

I'd also suggest suffixing them with `Element`, so `SplitOnRestElement`, `ExtractRestElement` & `ExcludeRestElement`.

benzaria:
Sure !

source/split-on-rest-element.d.ts
```diff
@@ -0,0 +1,93 @@
...
+export type SplitOnRestElement<Array_ extends UnknownArray, Options extends SplitOnRestElementOptions = {}> = IfNotAnyOrNever<Array_,
+	DeconstructRestArray<Array_, ApplyDefaultOptions<SplitOnRestElementOptions, DefaultSplitOnRestElementOptions, Options>>
+>;
+
+/**
+Deconstructs an array on its rest element and returns the split portions.
+
+See {@link SplitOnRestElement SplitOnRestElement} for details.
+*/
+export type DeconstructRestArray<
```
som-sm: Rename it to `_ SplitOnRestElement`.

source/split-on-rest-element.d.ts
```diff
@@ -0,0 +1,99 @@
...
+*/
+type SplitOnRestElementOptions = {
+	/**
+	Whether to preserve the optional modifier (`?`).
+
+	- When set to `true` (default), the optional modifiers are preserved as-is. For example, `SplitOnRestElement<[number, string?, ...boolean[]], {keepOptionals: true}>` returns `[[number, string?], boolean[], []]`.
+
+	- When set to `false`, optional elements like `T?` are transformed to `T | undefined` or simply `T` depending on the `exactOptionalPropertyTypes` compiler option. For example:
+		- With `exactOptionalPropertyTypes` enabled, `SplitOnRestElement<[number, string?, ...boolean[]], {keepOptionals: false}>` returns `[[number, string], boolean[], []]`.
+		- And, with it disabled, the result is `[[number, string | undefined], boolean[], []]`.
```
som-sm: I've added some more bits here, please verify.

Also, add example.

source/split-on-rest-element.d.ts
```diff
@@ -0,0 +1,99 @@
...
+
+	@default true
+	*/
+	keepOptionals?: boolean;
```
som-sm: Rename it to `preserveOptionalModifier`.

source/split-on-rest-element.d.ts
```diff
@@ -0,0 +1,99 @@
...
+If no `Rest` element exists, the entire array is treated as a static prefix with empty `Rest` element and suffix parts.
+
+By default, The optional modifier (`?`) is preserved. See {@link SplitOnRestElementOptions.keepOptionals keepOptionals} option to change this behavior.
+
+@example
+```ts
+import type {SplitOnRestElement} from 'type-fest';
+
+type T1 = SplitOnRestElement<[number, ...string[], boolean]>;
+//=> [[number], [string], [boolean]]
```
som-sm: IIRC, it previously returned `[[number], string[], [boolean]]`, right?

Umm...I think that made more sense.

source/split-on-rest-element.d.ts
```diff
@@ -0,0 +1,99 @@
...
+	*/
+	keepOptionals?: boolean;
+};
+
+type DefaultSplitOnRestElementOptions = {
+	keepOptionals: true;
+};
+
+/**
+Splits an array on its [`Rest`](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types) element portion into three parts:
```
som-sm: The first sentence should be a complete sentence because that goes into the README as-is.

IMO, this reads better:
```
Splits an array into three parts, where the first part contains all the elements before the rest element, the second part is the rest element itself, and the third part contains all the elements after the rest element.
```

And then you can have a note, like this:
```
Note: If any of the parts are missing, then they will be represented as empty arrays. For example, `SplitOnRestElement<[string, number]>` returns `[[string, number], [], []]`, where parts corresponding to the rest element and elements after it are empty.
```

source/split-on-rest-element.d.ts
```diff
@@ -0,0 +1,99 @@
...
+@category Array
+*/
+export type SplitOnRestElement<Array_ extends UnknownArray, Options extends SplitOnRestElementOptions = {}> = IfNotAnyOrNever<Array_,
+	DeconstructRestArray<Array_, ApplyDefaultOptions<SplitOnRestElementOptions, DefaultSplitOnRestElementOptions, Options>>
+>;
+
+/**
+Deconstructs an array on its rest element and returns the split portions.
+
+See {@link SplitOnRestElement SplitOnRestElement} for details.
```
som-sm: I think the link can be removed here.

source/split-on-rest-element.d.ts
benzaria: True, IDK why i removed it ?

source/split-on-rest-element.d.ts
benzaria: mybe a litle too loog for readme, but sure !

som-sm: Only the first sentence is for the README. The note is just for the type.

<br>

And if you still feel it's too long, feel free to make it concise.

benzaria:
@som-sm I think I resolved all ur comments.

benzaria:
@som-sm @sindresorhus can u guys take a look i need to continue on `Reverse` #1167 and `IndexOf...` #1155

source/extract-rest-element.d.ts
```diff
@@ -0,0 +1,32 @@
...
+
+@see ExcludeRestElement, SplitOnRestElement
+@category Array
+*/
+export type ExtractRestElement<T extends UnknownArray> =
+	SplitOnRestElement<T>[1] extends infer Result extends UnknownArray
+		? Result extends readonly []
+			? never
+			: Result[number]
+		: never;
```
som-sm: Is doing just this _not_ enough?
```ts
export type ExtractRestElement<T extends UnknownArray> = SplitOnRestElement<T>[1][number];
```
Passes all the test cases.

test-d/exclude-rest-element.ts
```diff
@@ -0,0 +1,53 @@
...
+expectType<ExcludeRestElement<[]>>({} as []);
+expectType<ExcludeRestElement<[1]>>({} as [1]);
+expectType<ExcludeRestElement<[1, 2, 3]>>({} as [1, 2, 3]);
+expectType<ExcludeRestElement<readonly ['a', 'b']>>({} as readonly ['a', 'b']);
+
+// Leading Rest element
+expectType<ExcludeRestElement<[...string[], 1]>>({} as [1]);
+expectType<ExcludeRestElement<[...number[], 'a', 'b']>>({} as ['a', 'b']);
+expectType<ExcludeRestElement<[...any[], true]>>({} as [true]);
+expectType<ExcludeRestElement<[...[], 'last']>>({} as ['last']);
```
som-sm: This case can be removed; it doesn't really add any value since `[...[], 'last']` will immediately get simplified to `['last']`.

test-d/exclude-rest-element.ts
```diff
@@ -0,0 +1,53 @@
...
+
+// Optional & Mixed Optional
+expectType<ExcludeRestElement<[string?, boolean?, ...number[]]>>({} as [string?, boolean?]);
+expectType<ExcludeRestElement<[1?, ...string[]]>>({} as [1?]);
+
+// Unions
+expectType<ExcludeRestElement<[1, ...string[]] | [2, ...number[]]>>({} as [1] | [2]);
+expectType<ExcludeRestElement<[...boolean[], 'end'] | ['start', ...string[]]>>({} as ['end'] | ['start']);
+
+// Readonly and Nested
```
som-sm: Remove `Nested` since there are no nested cases under this.

test-d/exclude-rest-element.ts
```diff
@@ -0,0 +1,53 @@
...
+expectType<ExcludeRestElement<[number, ...number[]]>>({} as [number]);
+
+// Only Rest element
+expectType<ExcludeRestElement<string[]>>({} as []);
+expectType<ExcludeRestElement<readonly number[]>>({} as readonly []);
+expectType<ExcludeRestElement<readonly [...boolean[]]>>({} as readonly []);
+expectType<ExcludeRestElement<[...string[]]>>({} as []);
+
+// Optional & Mixed Optional
+expectType<ExcludeRestElement<[string?, boolean?, ...number[]]>>({} as [string?, boolean?]);
```
som-sm, line 35: Add a case containing required, optional and rest.
```ts
expectType<ExcludeRestElement<[string, boolean?, ...number[]]>>({} as [string, boolean?]);
```

test-d/extract-rest-element.ts
```diff
@@ -0,0 +1,58 @@
...
+expectType<ExtractRestElement<['foo', ...number[], true]>>({} as number);
+expectType<ExtractRestElement<['foo', ...boolean[], 'bar']>>({} as boolean);
+expectType<ExtractRestElement<[...Array<'foo'>, unknown]>>({} as 'foo');
+expectType<ExtractRestElement<['bar', ...Array<5>]>>({} as 5);
+
+// Leading Rest element
+expectType<ExtractRestElement<[...string[], 1]>>({} as string);
+expectType<ExtractRestElement<[...number[], 'a', 'b']>>({} as number);
+expectType<ExtractRestElement<[...any[], true]>>({} as any);
+expectType<ExtractRestElement<[...[], 'last']>>({} as never);
```
som-sm: Same as above.

test-d/extract-rest-element.ts
```diff
@@ -0,0 +1,58 @@
+import {expectType} from 'tsd';
+import type {ExtractRestElement} from '../index.d.ts';
+
+// Simple Rest element detection
+expectType<ExtractRestElement<['start', ...string[]]>>({} as string);
+expectType<ExtractRestElement<['foo', ...number[], true]>>({} as number);
+expectType<ExtractRestElement<['foo', ...boolean[], 'bar']>>({} as boolean);
+expectType<ExtractRestElement<[...Array<'foo'>, unknown]>>({} as 'foo');
+expectType<ExtractRestElement<['bar', ...Array<5>]>>({} as 5);
```
som-sm: This can be remove I guess, since we already have the `leading`, `middle` and `trailing` tests. This doesn't cover anything different.

test-d/extract-rest-element.ts
```diff
@@ -0,0 +1,58 @@
...
+expectType<ExtractRestElement<['foo', ...Array<'bar'>]>>({} as 'bar');
+expectType<ExtractRestElement<[number, ...number[]]>>({} as number);
+
+// Rest element Only
+expectType<ExtractRestElement<string[]>>({} as string);
+expectType<ExtractRestElement<readonly number[]>>({} as number);
+expectType<ExtractRestElement<readonly [...boolean[]]>>({} as boolean);
+expectType<ExtractRestElement<[...string[]]>>({} as string);
+
+// Optional & Mixed Optional
```
som-sm: Same as above.

test-d/extract-rest-element.ts
```diff
@@ -0,0 +1,58 @@
...
+expectType<ExtractRestElement<['a', ...string[], 'z']>>({} as string);
+expectType<ExtractRestElement<['x', ...boolean[], true]>>({} as boolean);
+expectType<ExtractRestElement<['x', ...any[], 'y']>>({} as any);
+
+// Trailing Rest element
+expectType<ExtractRestElement<[1, 2, ...string[]]>>({} as string);
+expectType<ExtractRestElement<['foo', ...Array<'bar'>]>>({} as 'bar');
+expectType<ExtractRestElement<[number, ...number[]]>>({} as number);
+
+// Rest element Only
```
som-sm: Try keeping consistent casing with comments as well, looks bad IMO.
```suggestion
// Rest element only
```

test-d/split-on-rest-element.ts
```diff
@@ -0,0 +1,62 @@
+import {expectType} from 'tsd';
+import type {SplitOnRestElement} from '../index.d.ts';
+
+// Fixed Tuples (No Rest element)
+expectType<SplitOnRestElement<[]>>({} as [[], [], []]);
+expectType<SplitOnRestElement<[1]>>({} as [[1], [], []]);
+expectType<SplitOnRestElement<[1, 2, 3]>>({} as [[1, 2, 3], [], []]);
+expectType<SplitOnRestElement<readonly ['a', 'b', 'c']>>({} as [['a', 'b', 'c'], [], []]);
+
+// Finite Rest elements (flattened manually or fixed-length rest element)
```
som-sm: IMO, these cases just create confusion. If you spread a tuple (fixed length), TS immediately flattens it, so there's nothing special about these tests, it just looks different.

test-d/split-on-rest-element.ts
```diff
@@ -0,0 +1,62 @@
...
+expectType<SplitOnRestElement<number[]>>({} as [[], number[], []]);
+expectType<SplitOnRestElement<unknown[]>>({} as [[], unknown[], []]);
+expectType<SplitOnRestElement<any[]>>({} as [[], any[], []]);
+
+// Readonly
+expectType<SplitOnRestElement<readonly []>>({} as [[], readonly [], []]);
+expectType<SplitOnRestElement<readonly [number]>>({} as [[number], [], []]);
+expectType<SplitOnRestElement<readonly [...number[], 2]>>({} as [[], number[], [2]]);
+expectType<SplitOnRestElement<readonly [1, ...string[], 2]>>({} as [[1], string[], [2]]);
+expectType<SplitOnRestElement<readonly [1, 2, 3]>>({} as [[1, 2, 3], [], []]);
```
som-sm: If `Exclude` preserves `readonly`, `Split` should also preserve it.

som-sm: And _not_ sure how common are readonly rest elements, but should they be preserved too?

```ts
SplitOnRestElement<[1, 2, 3, ...readonly string[]]>
//=> [[1, 2, 3], readonly string[], []]
```

test-d/split-on-rest-element.ts
```diff
@@ -0,0 +1,62 @@
...
+
+// Unions
+expectType<SplitOnRestElement<[...Array<string | number>, boolean]>>({} as [[], Array<string | number>, [boolean]]);
+expectType<SplitOnRestElement<[...Array<{id: string}>, number]>>({} as [[], Array<{id: string}>, [number]]);
+expectType<SplitOnRestElement<[1, ...Array<[string, number]>, 2]>>({} as [[1], Array<[string, number]>, [2]]);
+expectType<SplitOnRestElement<[1, 2] | [3, 4]>>({} as [[1, 2], [], []] | [[3, 4], [], []]);
+expectType<SplitOnRestElement<[1, ...number[]] | ['foo', ...string[]]>>({} as [[1], number[], []] | [['foo'], string[], []]);
+
+// Exotic Types
```
som-sm: Remove these, there's nothing exotic about these

test-d/split-on-rest-element.ts
```diff
@@ -0,0 +1,62 @@
...
+// Fixed Tuples (No Rest element)
+expectType<SplitOnRestElement<[]>>({} as [[], [], []]);
+expectType<SplitOnRestElement<[1]>>({} as [[1], [], []]);
+expectType<SplitOnRestElement<[1, 2, 3]>>({} as [[1, 2, 3], [], []]);
+expectType<SplitOnRestElement<readonly ['a', 'b', 'c']>>({} as [['a', 'b', 'c'], [], []]);
+
+// Finite Rest elements (flattened manually or fixed-length rest element)
+expectType<SplitOnRestElement<[...[1, 2], 3]>>({} as [[1, 2, 3], [], []]);
+expectType<SplitOnRestElement<[...[string?, number?]]>>({} as [[string?, number?], [], []]);
+expectType<SplitOnRestElement<[...[string?, number?]], {preserveOptionalModifier: false}>>({} as [[string, number], [], []]);
```
som-sm: Add a separate block of tests for `preserveOptionalModifier` option.

source/exclude-rest-element.d.ts
```diff
@@ -0,0 +1,35 @@
+import type {SplitOnRestElement} from './split-on-rest-element.d.ts';
+import type {IsArrayReadonly} from './internal/array.d.ts';
+import type {UnknownArray} from './unknown-array.d.ts';
+
+/**
+Creates a tuple with the [`Rest`](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types) element removed.
```
som-sm:
```suggestion
Creates a tuple with the [`rest`](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types) element removed.
```

source/extract-rest-element.d.ts
```diff
@@ -0,0 +1,32 @@
+import type {SplitOnRestElement} from './split-on-rest-element.d.ts';
+import type {UnknownArray} from './unknown-array.d.ts';
+
+/**
+Extracts the [`Rest`](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types) element type from an array.
```
som-sm:
```suggestion
Extracts the [`rest`](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types) element type from an array.
```

source/split-on-rest-element.d.ts
```diff
@@ -0,0 +1,101 @@
...
+};
+
+type DefaultSplitOnRestElementOptions = {
+	preserveOptionalModifier: true;
+};
+
+/**
+Splits an array into three parts,
+where the first contains all elements before the rest element,
+the second is the [`Rest`](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types) element itself,
```
som-sm:
```suggestion
the second is the [`rest`](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types) element itself,
```

source/split-on-rest-element.d.ts
```diff
@@ -0,0 +1,101 @@
...
+and the third contains all elements after the rest element.
+
+Note: If any of the parts are missing, then they will be represented as empty arrays.
+For example, `SplitOnRestElement<[string, number]>` returns `[[string, number], [], []]`,
+where parts corresponding to the rest element and elements after it are empty.
+
+By default, The optional modifier (`?`) is preserved.
+See {@link SplitOnRestElementOptions.preserveOptionalModifier `preserveOptionalModifier`} option to change this behavior.
```
som-sm: Just add the link at the bottom.
```suggestion
@see {@link SplitOnRestElementOptions}
```

som-sm (changes requested):

test-d/split-on-rest-element.ts
sindresorhus: > And not sure how common are readonly rest elements, but should they be preserved too?

Yes

test-d/extract-rest-element.ts
sindresorhus: :arrow_up: Applies to all your PRs.

test-d/split-on-rest-element.ts
benzaria: It looks like its not possibe to do that. The spread operator ignores the readonly keyword:
```ts
type A = [...readonly string[]];
//    ^? type A = string[]
type B = readonly [...string[]];
//    ^? type B = readonly string[]

````

File: source/extract-rest-element.d.ts (hunk 12 above)
Comment (benzaria) [reply source/extract-rest-element.d.ts]: OMG i didnt notice this because i copy pasted ExcludeRestElement and modify it

File: test-d/split-on-rest-element.ts (hunk 21 above)
Comment (som-sm) [reply test-d/split-on-rest-element.ts]: Umm...interesting, looks like the `readonly` for the rest element isn't preserved.

```ts
type Test<Array_ extends UnknownArray> = {
	[P in keyof Array_]: Array_[P]
};

type T = Test<[1, 2, 3?, ...readonly string[]]>;
//   ^? type T = [1, 2, 3?, ...string[]]
```

Another example:
```ts
function foo(arg: [number, ...readonly string[]]) {
  const [first, ...rest] = arg;
  //               ^? const rest: string[]
}
```

som-sm: In that case, let's also remove `readonly` for the empty array case. So, `SplitOnRestElement<readonly []>` should return `[[], [], []]`.

benzaria: Sure, so i wanted to use `Writable` for this job but i think there is no need for it a simple `Array_ extends readonly [] ? [] : Array_` will do.

test-d/extract-rest-element.ts
benzaria: I will one by one.

test-d/split-on-rest-element.ts
som-sm: @benzaria If the overall array is `readonly` then we can preserve the `readonly`, it's _only_ when the rest element is `readonly` we cannot preserve it.

So, `SplitOnRestElement<readonly [1, 2, 3, ...string[], 4]>` should still return `readonly [[1, 2, 3], string[], [4]]`. 

If the rest element is `readonly`, then it's fine to not preserve it, like it's okay to return ``[[1, 2, 3], string[], 4]`` for `SplitOnRestElement<[1, 2, 3, ...readonly string[], 4]>`.

<br>

> In that case, let's also remove `readonly` for the empty array case. So, `SplitOnRestElement<readonly []>` should return `[[], [], []]`.

Should return `readonly [[], [], []]`.

benzaria: Ohhh easy!

benzaria: Should I make it an options: true by default or hardcode it into type ?

som-sm: We don't want to give option for preserving array readonly, it should be built-in to the type. Refer [the PR description](https://github.com/sindresorhus/type-fest/pull/1173#issue-3127319334).

benzaria: Done!

benzaria:
@som-sm Ready !

source/exclude-rest-element.d.ts
```diff
@@ -0,0 +1,37 @@
...
+@see ExtractRestElement, SplitOnRestElement
+@category Array
+*/
+export type ExcludeRestElement<Array_ extends UnknownArray> =
+	SplitOnRestElement<Array_> extends infer Result
+		? Result extends readonly UnknownArray[]
+			? IsArrayReadonly<Array_> extends true
+				? Readonly<[...Result[0], ...Result[2]]>
+				: [...Result[0], ...Result[2]]
+			: Result // May be `never` or `any`
```
som-sm: Remove comment, it's not actually correct, incase of `never` the control doesn't reach the else branch, it directly becomes `never`.

```ts
type Test<T> = T;

type Example<T> =
  Test<T> extends infer Result
    ? Result extends never
      ? '...'
      : 'BlahBlahBlah' // Doesn't matter
    : never;

type T1 = Example<never>;
//=> never
```

som-sm: And, for `any` the result is actually the following:
```ts
Readonly<[...any[0], ...any[2]]> | [...any[0], ...any[2]] | any
```

It works in this case because there's an `| any` at the end, but I prefer handling `any` cases separately in cases like these.

source/split-on-rest-element.d.ts
```diff
@@ -0,0 +1,113 @@
...
+					DefaultSplitOnRestElementOptions,
+					Options
+				>
+			>
+		> extends infer Result extends UnknownArray
+			? IsArrayReadonly<Array_> extends true
+				? Readonly<Result>
+				: Result
+			: never // Should never happen
+		: never; // Should never happen
```
som-sm, line 74:
```suggestion
	Array_ extends unknown // For distributing `Array_`
		? IfNotAnyOrNever<Array_, _SplitOnRestElement<
			Array_,
			ApplyDefaultOptions<SplitOnRestElementOptions, DefaultSplitOnRestElementOptions, Options>
		>> extends infer Result extends UnknownArray
			? If<IsArrayReadonly<Array_>, Readonly<Result>, Result>
			: never // Should never happen
		: never; // Should never happen
```

PR #1167, opened 2025-06-05
Add `Reverse` type

benzaria:
- [`Reverse`](source/reverse.d.ts) - Creates a new array type by Reversing the order of each element in the original array.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -161,6 +161,9 @@ Click the type names for complete docs.
 - [`WritableKeysOf`](source/writable-keys-of.d.ts) - Extract all writable (non-readonly) keys from the given type.
 - [`HasWritableKeys`](source/has-writable-keys.d.ts) - Create a `true`/`false` type depending on whether the given type has any writable fields.
 - [`Spread`](source/spread.d.ts) - Mimic the type inferred by TypeScript when merging two objects or two arrays/tuples using the spread syntax.
+- [`SplitOnSpread`](source/split-on-spread.d.ts) - Splits an Array on its spreaded portion. return's `[[...Head], [Spread], [...Tail]]`.
+- [`ExtractSpread`](source/split-on-spread.d.ts) - Return's the spread element type from and array.
+- [`ExcludeSpread`](source/split-on-spread.d.ts) - Create a array with the spread element removed.
 - [`IsEqual`](source/is-equal.d.ts) - Returns a boolean for whether the two given types are equal.
 - [`TaggedUnion`](source/tagged-union.d.ts) - Create a union of types that share a common discriminant property.
 - [`IntRange`](source/int-range.d.ts) - Generate a union of numbers (includes the start and excludes the end).
--- a/readme.md
+++ b/readme.md
@@ -295,6 +295,7 @@ Click the type names for complete docs.
 
 ### Improved built-in
 
+- [`ExtendsStrict`](source/extends-strict.d.ts) - A stricter version of `extends` that Checks Stricily if one type extends another without distribution.
 - [`ExtractStrict`](source/extract-strict.d.ts) - A stricter version of `Extract<T, U>` that ensures every member of `U` can successfully extract something from `T`.
 - [`ExcludeStrict`](source/exclude-strict.d.ts) - A stricter version of `Exclude<T, U>` that ensures every member of `U` can successfully exclude something from `T`.
 
--- a/readme.md
+++ b/readme.md
@@ -239,6 +239,7 @@ Click the type names for complete docs.
 - [`Arrayable`](source/arrayable.d.ts) - Create a type that represents either the value or an array of the value.
 - [`Includes`](source/includes.d.ts) - Returns a boolean for whether the given array includes the given item.
 - [`Join`](source/join.d.ts) - Join an array of strings and/or numbers using the given string as a delimiter.
+- [`Reverse`](source/array-reverse.d.ts) - Creates a new array type by Reversing the order of each element in the original array.
 - [`ArraySlice`](source/array-slice.d.ts) - Returns an array slice of a given range, just like `Array#slice()`.
 - [`LastArrayElement`](source/last-array-element.d.ts) - Extracts the type of the last element of an array.
 - [`FixedLengthArray`](source/fixed-length-array.d.ts) - Create a type that represents an array of the given type and length.
--- a/readme.md
+++ b/readme.md
@@ -239,7 +239,7 @@ Click the type names for complete docs.
 - [`Arrayable`](source/arrayable.d.ts) - Create a type that represents either the value or an array of the value.
 - [`Includes`](source/includes.d.ts) - Returns a boolean for whether the given array includes the given item.
 - [`Join`](source/join.d.ts) - Join an array of strings and/or numbers using the given string as a delimiter.
-- [`Reverse`](source/array-reverse.d.ts) - Creates a new array type by Reversing the order of each element in the original array.
+- [`Reverse`](source/reverse.d.ts) - Creates a new array type by Reversing the order of each element in the original array.
 - [`ArraySlice`](source/array-slice.d.ts) - Returns an array slice of a given range, just like `Array#slice()`.
 - [`LastArrayElement`](source/last-array-element.d.ts) - Extracts the type of the last element of an array.
 - [`FixedLengthArray`](source/fixed-length-array.d.ts) - Create a type that represents an array of the given type and length.
--- a/source/reverse.d.ts
+++ b/source/reverse.d.ts
@@ -1,0 +1,132 @@
+import type {ApplyDefaultOptions} from './internal/object.d.ts';
+import type {IsArrayReadonly} from './internal/array.d.ts';
+import type {SplitOnSpread} from './split-on-spread.d.ts';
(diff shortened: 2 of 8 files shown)


source/reverse.d.ts
```diff
@@ -0,0 +1,112 @@
...
+	Whether to preserve the optional modifier (`?`).
+
+	- When set to `true` and all members are optional, the optional modifiers are preserved as-is. but if one of the members is required the (`?`) is replaced with `| undefined`.
+	- When set to `false`, optional elements like `T?` are transformed to `T | undefined` or simply `T` depending on the `exactOptionalPropertyTypes` compiler option.
+
+	@example
+	```
+	import type {Reverse} from 'type-fest';
+
+	type Ex1 = Reverse<[string, number, boolean]>;
```
sindresorhus: Don't use abbreviations.

source/reverse.d.ts
```diff
@@ -0,0 +1,112 @@
...
+	*/
+	preserveOptionalModifier?: boolean;
+};
+
+type DefaultReverseOptions = {
+	preserveOptionalModifier: false;
+};
+
+/**
+Reverse an Array Items.
```
sindresorhus:
```suggestion
Reverse array/tuple elements.
```

source/reverse.d.ts
```diff
@@ -0,0 +1,112 @@
...
+					? KeepOptional extends false
+						? [If<IsExactOptionalPropertyTypesEnabled, First, First | undefined>] // Add `| undefined` for optional elements, if `exactOptionalPropertyTypes` is disabled.
+						: [First?]
+					: [First],
+				...Tail,
+			]>
+			: never;
+
+/**
+Creates a new array type by Reversing the order of each element in the original array.
```
sindresorhus:
```suggestion
Creates a new array type by reversing the element order of the original array.
```

source/reverse.d.ts
```diff
@@ -0,0 +1,112 @@
...
+/**
+Reverse options.
+
+@see {@link Reverse}
+*/
+type ReverseOptions = {
+	/**
+	Whether to preserve the optional modifier (`?`).
+
+	- When set to `true` and all members are optional, the optional modifiers are preserved as-is. but if one of the members is required the (`?`) is replaced with `| undefined`.
```
sindresorhus:
```suggestion
	- When set to `true` and all members are optional, the optional modifiers are preserved as-is. but if any member is required the (`?`) is replaced with `| undefined`.
```

sindresorhus:
- Export is missing. Add Reverse to index.d.ts
- Needs to be added to readme
- Should it be called `Reverse` or `ReverseTuple`?

benzaria:
Hey @sindresorhus, it's been a while.
I will address the suggested changes and I think the name `Reverse` is fitting cuz its for arrays and tuples so using `ReverseTuple` will be misleading

source/reverse.d.ts
benzaria: sure thing

source/reverse.d.ts
```diff
@@ -0,0 +1,114 @@
...
+/**
+Reverse array/tuple elements.
+*/
+type _Reverse<
+	Array_ extends UnknownArray,
+	KeepOptional extends boolean,
+	Head extends UnknownArray = [],
+	Tail extends UnknownArray = [],
+> =
+	keyof Array_ & `${number}` extends never // Is `Array_` leading a rest element or empty
```
sindresorhus, line 69: Would be useful with a short code comment that describes the algorithm.

test-d/reverse.ts
```diff
@@ -0,0 +1,95 @@
...
+declare const never: never;
+declare const string: string;
+declare const number: number;
+declare const symbol: symbol;
+declare const boolean: boolean;
+
+// Edge cases
+expectType<Reverse<[]>>([]);
+expectType<Reverse<any>>(any);
+expectType<Reverse<never>>(never);
```
sindresorhus:
```suggestion
expectType<Reverse<never>>(never);
expectType<Reverse<[unknown]>>([{} as unknown] as const);
```

sindresorhus:
https://github.com/sindresorhus/type-fest/pull/1166#issuecomment-2972873246 was merged. You mentioned you needed it for this.

test-d/reverse.ts
```diff
@@ -0,0 +1,95 @@
...
+expectType<Reverse<[1, boolean, 'x']>>(['x', boolean, 1] as const);
+expectType<Reverse<[number, string, boolean]>>([boolean, string, number] as const);
+expectType<Reverse<[1, 2, 3, 4]>>([4, 3, 2, 1] as const);
+expectType<Reverse<[0, 1, 2, 3, 4]>>([4, 3, 2, 1, 0] as const);
+expectType<Reverse<['x', 1, true]>>([true, 1, 'x'] as const);
+
+// Optional/undefined
+expectType<Reverse<[1?, 2?, 3?]>>([3, 2, 1] as const);
+expectType<Reverse<[1?, 2?, 3?], {preserveOptionalModifier: true}>>({} as [3?, 2?, 1?]);
+expectType<Reverse<[1, 2?, 3?], {preserveOptionalModifier: true}>>({} as [3 | undefined, 2 | undefined, 1]);
```
som-sm: Behaviour with optional elements is _not_ correct, you can't return `[3 | undefined, 2 | undefined, 1]` for `Reverse<[1, 2?, 3?]>` because at runtime input could be `[1, 2]` and reversing it would produce `[2, 1]` which wouldn't conform to `[3 | undefined, 2 | undefined, 1]`. 

So, the output for cases containing optional elements needs fixing.

benzaria:
> @som-sm
> Yeah, please go ahead, we can close this PR and move ahead with yours if that feels better.

I’ve managed to:

* Incorporate the suggested change by adding `TailAcc |` to the `_Reverse` recursive call.
* Fix the case where all elements are optional.
* Remove `ReverseOptions` and refactor the implementation for clarity.
* Update the JSDoc and tests to reflect the new behavior.

If there are any missing tests or additional scenarios worth covering, feel free to list or add them.

som-sm:
> Fix the case where all elements are optional.

@benzaria Please check https://github.com/sindresorhus/type-fest/pull/1266#issuecomment-3393452518.

benzaria:
> @benzaria Please check [#1266 (comment)](https://github.com/sindresorhus/type-fest/pull/1266#issuecomment-3393452518).

@som-sm I took a note of that and I fixed the problem.

readme.md
```diff
@@ -248,6 +248,7 @@ Click the type names for complete docs.
 - [`Arrayable`](source/arrayable.d.ts) - Create a type that represents either the value or an array of the value.
 - [`Includes`](source/includes.d.ts) - Returns a boolean for whether the given array includes the given item.
 - [`Join`](source/join.d.ts) - Join an array of strings and/or numbers using the given string as a delimiter.
+- [`Reverse`](source/reverse.d.ts) - Reverses the order of elements in a tuple or array type.
```
som-sm: This should match the JSDoc description.

source/reverse.d.ts
```diff
@@ -0,0 +1,83 @@
...
+type T3 = Reverse<[string, number?, ...boolean[]]>;
+//=> [...boolean[], number, string] | [string]
+
+type T4 = Reverse<[string?, number?, ...boolean[]]>;
+//=> [...boolean[], number, string] | [string] | []
+```
+
+@category Array
+*/
+export type Reverse<Array_ extends UnknownArray> =
```
som-sm: Better to name the type `ArrayReverse` as mentioned in #1266.

> I've kept the name as `ArrayReverse` instead of just `Reverse` to match with our existing `Array*` types (like `ArraySlice`, `ArraySplice`).

test-d/reverse.ts
som-sm, line 1: Cases like labelled tuples, non-tuple arrays seem missing. Can you once cross-check from the tests in #1266.

readme.md
benzaria: they are the same, both in JSDoc and readme!

source/array-reverse.d.ts
```diff
@@ -0,0 +1,85 @@
...
+			: [...HeadAcc, ...Array_, ...TailAcc] // Add the rest element between the accumulated elements.
+		: Array_ extends readonly [(infer First)?, ...infer Rest]
+			// Accumulate elements before a rest element in Arrayreverse order.
+			? IsOptionalKeyOf<Array_, 0> extends true
+				? (
+					| TailAcc // Union all possible cases when optional elements exist.
+					| _ArrayReverse<Rest, HeadAcc, [
+						If<IsExactOptionalPropertyTypesEnabled, First, First | undefined>, // Add `| undefined` for optional elements, if `exactOptionalPropertyTypes` is disabled.
+						...TailAcc,
+					]>
```
som-sm, line 81: This breaks tail recursion, the existing long tuple test cases don't catch this because they _only_ contain required elements. If you instantiate with `Partial<TupleOf<50, number>>`, you'll see that the type breaks with a recursion depth error.

Refer #1266 for a tail recursive implementation. You can use the following test case for verification: 
```ts
expectType<TupleOf<IntRange<0, 51>, number>>({} as ArrayReverse<Partial<TupleOf<50, number>>>);
```

PR #1168, opened 2025-06-06
Add custom lint rule to ensure import paths end with `.d.ts` extension

som-sm:
- Adds an auto-fixable solution for #1152, improving DX as mentioned in [this comment](https://github.com/sindresorhus/type-fest/pull/1152#issuecomment-2910311853).
 
  https://github.com/user-attachments/assets/d13cf93a-fe11-40be-a2d2-ba7a53cd899e

  image

- Updates `xo` because there are some bugs in the existing version.

- Enables the `allowJS` flag in `tsconfig.json` to ensure the `xo.config.js` file is checked. This shouldn't be a problem because we have `noEmit` enabled, so it shouldn't matter if `.js` files are part of the program. Also, this doesn't check all `.js` files for errors, only those that have the `@ts-check` directive.
  
- This is kinda better than the `no-restricted-imports` rule, because that restricts certain extensions, whereas this enforces the desired one. 
    And interestingly, it also catches some existing errors that weren't caught earlier. And these errors are not because of any of our recent changes, these were there since the very beginning.
    https://github.com/sindresorhus/type-fest/blob/d71242aee4f6a285ed2b20da8dba03111650d2bd/source/pick-deep.d.ts#L4-L5
    https://github.com/sindresorhus/type-fest/blob/d71242aee4f6a285ed2b20da8dba03111650d2bd/test-d/internal/is-number-like.ts#L2

### DIFF
--- a/lint-rules/import-path.js
+++ b/lint-rules/import-path.js
@@ -1,0 +1,52 @@
+import path from 'node:path';
+
+export const importPathRule = /** @type {const} */ ({
+	meta: {
+		type: 'problem',
+		docs: {
+			description: 'Enforces import paths to end with a \'.d.ts\' extension.',
+		},
+		fixable: 'code',
+		messages: {
+			incorrectImportPath:
+                'Import path \'{{importPath}}\' must end with a \'.d.ts\' extension. Use \'{{fixedImportPath}}\' instead.',
+		},
+		schema: [],
+	},
+	defaultOptions: [],
+	create(context) {
+		return {
+			ImportDeclaration(node) {
+				const importPath = node.source.value;
+
+				// Skip if not relative path
+				if (!(importPath.startsWith('./') || importPath.startsWith('../'))) {
+					return;
+				}
+
+				const filename = path.basename(importPath);
+				const firstDotIndex = filename.indexOf('.');
+				const extension = firstDotIndex === -1 ? '' : filename.slice(firstDotIndex);
+
+				// Skip if the import path already ends with `.d.ts`
+				if (extension === '.d.ts') {
+					return;
+				}
+
+				const importPathWithoutExtension = extension.length > 0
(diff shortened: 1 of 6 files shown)


lint-rules/import-path.js
```diff
@@ -0,0 +1,52 @@
+import path from 'node:path';
+
+export const importPathRule = /** @type {const} */ ({
+	meta: {
+		type: 'problem',
```
som-sm, line 5: The `/** @type {const} */` comment ensures that the inferred type of `meta.type` is `'problem'` and not `string`, because otherwise there would be complaints in `xo.config.js`.

It would have been nice if XO exposed rule creation utilities (like `ESLintUtils.RuleCreator.withoutDocs`) from `typescript-eslint`, that would have ensured proper type safety on this rule. That said, even with that utility, we would have still required `node` types.

som-sm:
> Enables the `allowJS` flag in `tsconfig.json` to ensure the `xo.config.js` file is checked. This shouldn't be a problem because we have `noEmit` enabled, so it shouldn't matter if `.js` files are part of the program. Also, this doesn't check all `.js` files for errors, only those that have the `@ts-check` directive.

And this shouldn't cause slowdowns even when we disable `skipLibCheck`, because `checkJS` is still disabled, so it should only check files that have the `@ts-check` directive.

<br>

Also, on a different note, after the V5 updates, I remember seeing errors when `skipLibCheck` was disabled, but I don't see those errors anymore. Maybe the issue got fixed, so maybe we can disable that back?

sindresorhus:
> Also, on a different note, after the V5 updates, I remember seeing errors when skipLibCheck was disabled, but I don't see those errors anymore. Maybe the issue got fixed, so maybe we can disable that back?

Yeah, I think we can. I also saw errors, but not anymore.

PR #1169, opened 2025-06-06
Disable `skipLibCheck` compiler option

som-sm:
Refer: https://github.com/sindresorhus/type-fest/pull/1168#issuecomment-2948991459


som-sm:
Strange, those errors are back now 🤔

som-sm:
Looks like it now wants `DOM` definitions to be added via the `lib` option.

PR #1171, opened 2025-06-07
Add `UnionToEnum` and `TupleToEnum` types and fix `UnionToTuple` output

### LINKED ISSUE #1084: Union to Enum

When making dynamic enum from string template union:

```typescript
const x = ['create','read']
const y = ['other','foo']
type SomeTemplate = `${x}_${y}` // <-- A string union
const theUnion = x.flatMap(a => y.map(b => `${a}_${b}`))
const theEnum = Object.fromEntries(theUnion.map(a => [a,a]))
```

If we use `Record<SomeTemplate, SomeTemplate>`, the type of every key is `SomeTemplate` instead of the literal. Now I don't know how many will use it, but at least I did so I know that the enum is really just an enum instead of some other type. However, an argument can be made that the `Record<K, K>` is indication enough that it is an enum.

### Type source

```typescript
type UnionToEnum<T extends string> = {
  [k in T]: k;
};
```

benzaria:
Closes #1084 

Add `UnionToEnum<T, Numeric, Options>` utility for converting a union/tuple into an Enum-like object.

 - `string`, `number`, or `symbol` unions or tuples
 - **Numeric Enum** values starting from configurable **Index**, default: `1`

**Example:**
```ts
import type {UnionToEnum} from 'type-fest';

type E1 = UnionToEnum<'A' | 'B' | 'C'>;
//=> { A: 'A'; B: 'B'; C: 'C' }

type E2 = UnionToEnum<'X' | 'Y' | 'Z', {numeric: true}>;
//=> { X: 1; Y: 2; Z: 3 }

type E3 = UnionToEnum<['Play', 'Pause', 'Stop'], {numeric: true; startIndex: 3}>;
//=> { Play: 3; Pause: 4; Stop: 5 }

type E4 = UnionToEnum<['some_key', 'another_key']>;
//=> { 'some_key': 'some_key'; 'another_key': 'another_key' }

type E5 = UnionToEnum<never>;
//=> {}
```

### DIFF
--- a/source/union-to-enum.d.ts
+++ b/source/union-to-enum.d.ts
@@ -1,69 +1,17 @@
-import type {ApplyDefaultOptions} from './internal/object.d.ts';
+import type {ToEnumOptions, TupleToEnum} from './tuple-to-enum.d.ts';
 import type {UnionToTuple} from './union-to-tuple.d.ts';
-import type {UnknownArray} from './unknown-array.d.ts';
-import type {IsLiteral} from './is-literal.d.ts';
-import type {Simplify} from './simplify.d.ts';
-import type {IsNever} from './is-never.d.ts';
-import type {Sum} from './sum.d.ts';
 
 /**
-{@link UnionToEnum} Options.
-*/
-type UnionToEnumOptions = {
-	/**
-	The first numeric value to assign when using numeric indices.
-
-	@default 1
-
-	@example
-	```
-	type E1 = UnionToEnum<['Play', 'Pause', 'Stop'], {numeric: true}>;
-	//=> { Play: 1; Pause: 2; Stop: 3 }
-
-	type E2 = UnionToEnum<['Play', 'Pause', 'Stop'], {numeric: true; startIndex: 3}>;
-	//=> { Play: 3; Pause: 4; Stop: 5 }
-
-	type E3 = UnionToEnum<['Play', 'Pause', 'Stop'], {numeric: true; startIndex: -1}>;
-	//=> { Play: -1; Pause: 0; Stop: 1 }
-	```
-	*/
-	startIndex?: number;
-	/**
-	Whether to use numeric indices as values.
-
-	@default false
-
-	@example
-	```
-	type E1 = UnionToEnum<'X' | 'Y' | 'Z'>;
-	//=> { X: 'X'; Y: 'Y'; Z: 'Z' }
-
-	type E2 = UnionToEnum<'X' | 'Y' | 'Z', {numeric: true}>;
-	//=> { X: 1; Y: 2; Z: 3 }
-
-	type E3 = UnionToEnum<['Play', 'Pause', 'Stop'], {numeric: true; startIndex: 3}>;
-	//=> { Play: 3; Pause: 4; Stop: 5 }
-	```
-	*/
-	numeric?: boolean;
-};
-
-type DefaultUnionToEnumOptions = {
-	startIndex: 1;
-	numeric: false;
-};
-
-/**
-Converts a union or tuple of property keys (string, number, or symbol) into an **Enum**.
+Converts a union of property keys (string, number, or symbol) into an **Enum**.
 
 The keys are preserved, and their values are either:
 
 - Their own literal values (by default)
-- Or numeric indices (`1`, `2`, ...) if {@link UnionToEnumOptions.numeric `numeric`} is `true`. Union ordering is not guaranteed.
+- Or numeric indices (`1`, `2`, ...) if {@link ToEnumOptions.numeric `numeric`} is `true`. Union ordering is not guaranteed.
 
-By default, **numeric Enums** start from **Index `1`**. See {@link UnionToEnumOptions.startIndex `startIndex`} to change this behavior.
+By default, **numeric Enums** start from **Index `1`**. See {@link ToEnumOptions.startIndex `startIndex`} to change this behavior.
 
-This is useful for creating strongly typed enums from a union/tuple of literals.
+This is useful for creating strongly typed enums from a union of literals.
 
 @example
 ```
@@ -75,13 +23,16 @@ type E1 = UnionToEnum<'A' | 'B' | 'C'>;
 type E2 = UnionToEnum<'X' | 'Y' | 'Z', {numeric: true}>;
 //=> { X: 1; Y: 2; Z: 3 }
 
-type E3 = UnionToEnum<['Play', 'Pause', 'Stop'], {numeric: true; startIndex: 3}>;
+type E3 = UnionToEnum<'Play' | 'Pause' | 'Stop', {numeric: true; startIndex: 3}>;
 //=> { Play: 3; Pause: 4; Stop: 5 }
 
-type E4 = UnionToEnum<['some_key', 'another_key']>;
-//=> { 'some_key': 'some_key'; 'another_key': 'another_key' }
+type E4 = UnionToEnum<1 | 2 | 3>;
+//=> { 1: 1; 2: 2; 3: 3 }
 
-type E5 = UnionToEnum<never>;
+type E5 = UnionToEnum<1 | 2 | 3, {numeric: true; startIndex: -1}>;
+//=> { 1: -1; 2: 0; 3: 1 }
+
+type E6 = UnionToEnum<never>;
 //=> {}
 ```
 
@@ -111,41 +62,13 @@ const Template = createEnum(verb, resource);
 // }
 ```
 
-@see UnionToTuple
+@see {@link TupleToEnum}
+@see {@link UnionToTuple}
 @category Object
 */
 export type UnionToEnum<
-	Keys extends (
-		[Keys] extends [PropertyKey]
-			? PropertyKey
-			: readonly PropertyKey[]
-	),
-	Options extends UnionToEnumOptions = {},
-> = IsNever<Keys> extends true ? {}
-	: _UnionToEnum<
-		[Keys] extends [UnknownArray] ? Keys : UnionToTuple<Keys>,
-		ApplyDefaultOptions<UnionToEnumOptions, DefaultUnionToEnumOptions, Options>
-	>;
-
-/**
-Core type for {@link UnionToEnum}.
-*/
-type _UnionToEnum<
-	Keys extends UnknownArray,
-	Options extends Required<UnionToEnumOptions>,
-> = Simplify<{readonly [
-	K in keyof Keys as K extends `${number}`
-		? Keys[K] extends PropertyKey
-			? IsLiteral<Keys[K]> extends true // TODO: update to accept template literals
-				? Keys[K]
-				: never // Not a literal
-			: never // Not a property key
-		: never // Not an index
-	]: Options['numeric'] extends true
-		? K extends `${infer N extends number}`
-			? Sum<N, Options['startIndex']>
-			: never // Not an index
-		: Keys[K]
-}>;
+	Keys extends PropertyKey,
+	Options extends ToEnumOptions = {},
+> = TupleToEnum<UnionToTuple<Keys>, Options>;
 
 export {};
--- a/test-d/union-to-enum.ts
+++ b/test-d/union-to-enum.ts
@@ -1,36 +1,16 @@
-import {expectType} from 'tsd';
+import {expectAssignable, expectType} from 'tsd';
 import type {CamelCasedProperties, UnionToEnum} from '../index.d.ts';
 
 // Union input
+expectType<UnionToEnum<'Only', {numeric: true}>>({Only: 1} as const);
 expectType<UnionToEnum<'b' | 'a' | 'd' | 'c'>>({b: 'b', a: 'a', d: 'd', c: 'c'} as const);
 expectType<UnionToEnum<3 | 2 | 4 | 1>>({1: 1, 2: 2, 3: 3, 4: 4} as const);
-
-// Tuple input
-expectType<UnionToEnum<['One', 'Two']>>({One: 'One', Two: 'Two'} as const);
-expectType<UnionToEnum<['X', 'Y', 'Z'], {numeric: true}>>({X: 1, Y: 2, Z: 3} as const);
-expectType<UnionToEnum<['A', 'B'] | ['C', 'D']>>({} as Readonly<{A: 'A'; B: 'B'} | {C: 'C'; D: 'D'}>);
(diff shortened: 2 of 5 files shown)


benzaria:
@som-sm @sindresorhus Should **CamelCase** be `true` by default ?

sindresorhus:
This type is doing too much. It should not do case transformation. That would be a task for a separate type. It should purely transform union to enum.

benzaria:
> This type is doing too much. It should not do case transformation. That would be a task for a separate type. It should purely transform union to enum.

fair enough i will remove case transformation.

sindresorhus:
I don't think it should do startingIndex either. It's too limiting and could be done by a separate type too.

benzaria:
> I don't think it should do startingIndex either. It's too limiting and could be done by a separate type too.

Should I keep the starting index `1` or `0` ?

sindresorhus:
Hmm, didn't think of that. Maybe we do indeed need the `startIndex`.

benzaria:
> Hmm, didn't think of that. Maybe we do indeed need the `startIndex`.

True, diff users may want diff index for there enum.

benzaria:
@sindresorhus @som-sm Ready !

source/union-to-enum.d.ts
```diff
@@ -0,0 +1,115 @@
...
+//    readFile: 'read_file',
+//    readFolder: 'read_folder',
+//    readLink: 'read_link',
+//    deleteFile: 'delete_file',
+//    deleteFolder: 'delete_folder',
+//    deleteLink: 'delete_link',
+// }
+```
+
+@author benzaria
```
sindresorhus: When we comment something, we expect you to apply it to all your pull requests so we don't have to repeat it.

source/union-to-enum.d.ts
```diff
@@ -0,0 +1,115 @@
...
+@see UnionToTuple
+@category Object
+*/
+export type UnionToEnum<
+	Keys extends PropertyKey | readonly PropertyKey[],
+	Numeric extends boolean = false,
```
sindresorhus: This would be better as an option.

readme.md
```diff
@@ -245,6 +245,7 @@ Click the type names for complete docs.
 - [`ReadonlyTuple`](source/readonly-tuple.d.ts) - Create a type that represents a read-only tuple of the given type and length.
 - [`TupleToUnion`](source/tuple-to-union.d.ts) - Convert a tuple/array into a union type of its elements.
 - [`UnionToTuple`](source/union-to-tuple.d.ts) - Convert a union type into an unordered tuple type of its elements.
+- [`UnionToEnum`](source/union-to-enum.d.ts) - Convert a union or tuple of property keys into an **Enum**-like object.
```
sindresorhus: "enum-like" is a bit vague.

source/union-to-enum.d.ts
```diff
@@ -0,0 +1,115 @@
...
+import type {BuildTuple} from './internal/tuple.d.ts';
+import type {Simplify} from './simplify.d.ts';
+import type {IsNever} from './is-never.d.ts';
+
+/**
+{@link UnionToEnum} Options.
+*/
+type UnionToEnumOptions = {
+	/**
+	The starting **Index** of Numeric enums.
```
sindresorhus:
```suggestion
	The first numeric value to assign when using numeric indices.
```

test-d/union-to-enum.ts
```diff
@@ -0,0 +1,93 @@
...
+	writeFile: 'write_file',
+	writeFolder: 'write_folder',
+	writeLink: 'write_link',
+	readFile: 'read_file',
+	readFolder: 'read_folder',
+	readLink: 'read_link',
+	deleteFile: 'delete_file',
+	deleteFolder: 'delete_folder',
+	deleteLink: 'delete_link',
+} as const);
```
sindresorhus, line 70: Some additional tests (AI generated):

```js
// Empty array cases
expectType<UnionToEnum<[]>>({});
expectType<UnionToEnum<readonly []>>({});

// Edge cases for startIndex
expectType<UnionToEnum<['a', 'b'], true, {startIndex: 0}>>({a: 0, b: 1} as const);
expectType<UnionToEnum<['x'], true, {startIndex: -1}>>({} as const);
expectType<UnionToEnum<['test'], true, {startIndex: 100}>>({test: 100} as const);

// Numeric edge cases
expectType<UnionToEnum<0 | -1 | 42>>({0: 0, [-1]: -1, 42: 42} as const);
expectType<UnionToEnum<[0, -5, 999], true>>({0: 1, [-5]: 2, 999: 3} as const);

// Mixed types with numeric indexing
expectType<UnionToEnum<['str', 42, 'another'], true>>({str: 1, 42: 2, another: 3} as const);

// Single element cases
expectType<UnionToEnum<[0]>>({0: 0} as const);
expectType<UnionToEnum<[0], true, {startIndex: 5}>>({0: 5} as const);

// Verify that readonly behavior is consistent
expectType<UnionToEnum<['a']>>({a: 'a'} as const);
```

source/union-to-enum.d.ts
```diff
@@ -0,0 +1,115 @@
...
+type E5 = UnionToEnum<never>;
+//=> {}
+```
+
+@example
+```
+import type {UnionToEnum, CamelCasedProperties} from 'type-fest';
+
+const verb = ['write', 'read', 'delete'] as const;
+const resrc = ['file', 'folder', 'link'] as const;
```
sindresorhus: Don't use abbreviations

source/union-to-enum.d.ts
benzaria: What is wrong with me. have I a short memory or some 🤦‍♂️

source/union-to-enum.d.ts
benzaria: IDK, that will make it harder to reach. Numeric isn't an option as startIndex is, Numeric will be used quite often because its more of a feature of this type, anyone can make a mapped obj from a union with ease. the Numeric transformation is what make it standout. I think we should keep it with the generic arguments.

readme.md
benzaria: I tried to be clear, maybe to much.

test-d/union-to-enum.ts
benzaria: Sure !

benzaria: The expected value does evaluate, but an error get thrown. Should I prevent this error from showing or keep it in.
```ts
expectType<UnionToEnum<['x'], true, {startIndex: -1}>>({} as const); //=> {}
but throw; Error: Type instantiation is excessively deep and possibly infinite. 
```

source/union-to-enum.d.ts
```diff
@@ -0,0 +1,114 @@
...
+/**
+{@link UnionToEnum} Options.
+*/
+type UnionToEnumOptions = {
+	/**
+	The first numeric value to assign when using numeric indices.
+
+	@default 1
+	*/
+	startIndex?: number;
```
som-sm: Add example.

source/union-to-enum.d.ts
som-sm: Should be an option, `UnionToEnum<'X' | 'Y' | 'Z', {numeric: true}>` is more clearer than `UnionToEnum<'X' | 'Y' | 'Z', true>`. I'd prefer typing in a few extra characters for better readability.

test-d/union-to-enum.ts
som-sm: `-ve` `startIndex` should also work, you can use `Sum` instead of `BuildTuple`.

som-sm (commented):
What should be the behaviour in cases where `UnionToEnum` returns an index signature?

For example, currently invalid index signatures are returned in cases like these:
```ts
type T1 = UnionToEnum<string | number>;
// {readonly [x: string]: string; readonly [x: number]: number}

type T2 = UnionToEnum<[string, 'foo'], true>;
// {readonly [x: string]: 1; readonly foo: 2}
```

image<br>

Maybe we should skip all non-literals.

test-d/union-to-enum.ts
benzaria: Ohh good improvement. [030ce52](https://github.com/sindresorhus/type-fest/pull/1171/commits/030ce52bd2a03be291eb889df79f4ae42517ab87)

benzaria:
> What should be the behaviour in cases where `UnionToEnum` returns an index signature?

Should also skip `foo${string}` !

source/union-to-enum.d.ts
benzaria: moved to Options in [030ce52](https://github.com/sindresorhus/type-fest/pull/1171/commits/030ce52bd2a03be291eb889df79f4ae42517ab87)

source/union-to-enum.d.ts
benzaria: Done in [030ce52](https://github.com/sindresorhus/type-fest/pull/1171/commits/030ce52bd2a03be291eb889df79f4ae42517ab87)

source/union-to-enum.d.ts
```diff
@@ -0,0 +1,145 @@
...
+*/
+type UnionToEnumOptions = {
+	/**
+	The first numeric value to assign when using numeric indices.
+
+	@default 1
+
+	@example
+	```
+	type E2 = UnionToEnum<['Play', 'Pause', 'Stop'], {numeric: true}>;
```
som-sm: Should start with `E1`.

source/union-to-enum.d.ts
```diff
@@ -0,0 +1,145 @@
...
+};
+
+type DefaultUnionToEnumOptions = {
+	startIndex: 1;
+	numeric: false;
+};
+
+/**
+Converts a union or tuple of property keys (string, number, or symbol) into an **Enum**.
+
```
sindresorhus, line 6: It should explicitly say that union ordering is not guaranteed.

.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 yarn.lock
+\!*
```
sindresorhus: Unrelated change

sindresorhus (approved):

source/union-to-enum.d.ts
benzaria: i added it to numeric.

.gitignore
benzaria: im so sorry it was a mistake, i use it to mask some private tests and experimental types. i just forgot about it!

som-sm:
Apologies, haven't been able to re-read this. I'll give it a read sometime next week.

benzaria:
@som-sm, Hey 👋 just following up on this, you mentioned you’d review it next week, but I haven’t seen any updates yet.
No rush if you’re busy, I just wanted to check whether there’s anything else needed from my side to move this forward.

test-d/union-to-enum.ts
```diff
@@ -0,0 +1,111 @@
...
+
+// Edge cases for startIndex
+expectType<UnionToEnum<['x'], {numeric: true; startIndex: -1}>>({x: -1} as const);
+expectType<UnionToEnum<['x', 'y'], {numeric: true; startIndex: -100}>>({x: -100, y: -99} as const);
+expectType<UnionToEnum<['test'], {numeric: true; startIndex: 100}>>({test: 100} as const);
+
+// Numeric edge cases
+expectType<UnionToEnum<0 | -1 | 42>>({0: 0, [-1]: -1, 42: 42} as const);
+expectType<UnionToEnum<[0, -5, 999], {numeric: true}>>({0: 1, [-5]: 2, 999: 3} as const);
```
sindresorhus: Maybe a few more tests:

```ts
// Symbols with numeric indices
expectType<UnionToEnum<[typeof sym1, typeof sym2], {numeric: true}>>({[sym1]: 1, [sym2]: 2} as const);

// Readonly tuple input
expectType<UnionToEnum<readonly ['a', 'b', 'c']>>({a: 'a', b: 'b', c: 'c'} as const);
```

benzaria: Done

benzaria:
@som-sm Could you please take some time to review this PR when you get a chance, so we can move it forward?

source/union-to-enum.d.ts
```diff
@@ -0,0 +1,147 @@
...
+//    deleteFolder: 'delete_folder',
+//    deleteLink: 'delete_link',
+// }
+```
+
+@see UnionToTuple
+@category Object
+*/
+export type UnionToEnum<
+	Keys extends PropertyKey | readonly PropertyKey[],
```
som-sm: This allows `Keys` to be a combination of `PropertyKey` and `readonly PropertyKey[]`, like:

```ts
type Current = UnionToEnum<'A' | ['B', 'C']>;
//=> { readonly A: "A"; }

type Ideal = UnionToEnum<'A' | ['B', 'C']>;
//=> { readonly A: "A"; readonly B: "B"; readonly C: "C"; }
```

source/union-to-enum.d.ts
benzaria: Done!

source/union-to-enum.d.ts
benzaria: Great catch! However, I think it’s better to restrict using both approaches (union and tuple) at the same time rather than allowing it.

The reason is that the behavior becomes unpredictable or effectively useless when `numeric` is `true`, since adding a union element to a tuple input defeats the purpose of numeric indexing.

Done in [7efbf20](https://github.com/sindresorhus/type-fest/pull/1171/commits/7efbf2066a1510ef40755d3615ab63d132f4ab43)

Emiyaaaaa:
This type name and parameter make me confused, maybe it can be split to `UnionToEnum` and `TupleToEnum`? @sindresorhus @som-sm what do you think?

som-sm:
> This type name and parameter make me confused, maybe it can be split to `UnionToEnum` and `TupleToEnum`? @sindresorhus @som-sm what do you think?

@Emiyaaaaa Yeah, pretty valid point, if the type is `UnionToEnum`, it shouldn't accept tuples. I think it makes sense to split it to `UnionToEnum` and `TupleToEnum`.

benzaria:
@som-sm I also have a suggestion, why not just call it `ToEnum`. It's straight forward and eliminates the need for 2 types.

som-sm:
> @som-sm I also have a suggestion, why not just call it `ToEnum`. It's straight forward and eliminates the need for 2 types.

Umm...no, `ToEnum` is quite vague, I think it's fine to split it into two types.

PR #1173, opened 2025-06-07
`ArrayTail`: Enable `preserveReadonly` option by default

som-sm:
Completes point no. 3 of #450.

<br>

IMO, instead of flipping the default value for `preserveReadonly`, we should remove the option altogether. Because, otherwise, this option should ideally be present on all array types. 

And removing `readonly` from an array is pretty straightforward anyway.

```ts
type T = Writable<ArrayTail<readonly [number, string]>>
//=> [number]
```

### DIFF
--- a/source/merge-deep.d.ts
+++ b/source/merge-deep.d.ts
@@ -86,7 +86,7 @@ type OmitRestTypeHelper<
 	Tail extends UnknownArrayOrTuple,
 	Type extends UnknownArrayOrTuple,
 	Result extends UnknownArrayOrTuple = [],
-> = Tail extends []
+> = Tail extends readonly []
 	? Result
 	: OmitRestType<Tail, [...Result, FirstArrayElement<Type>]>;
 
--- a/test-d/array-tail.ts
+++ b/test-d/array-tail.ts
@@ -7,19 +7,19 @@ expectType<[]>(getArrayTail([]));
 expectType<[]>(getArrayTail(['a']));
 expectType<[]>(getArrayTail(['a', 'b', 'c']));
 
-expectType<[]>(getArrayTail([] as const));
-expectType<[]>(getArrayTail(['a'] as const));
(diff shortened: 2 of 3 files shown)


source/merge-deep.d.ts
```diff
@@ -86,7 +86,7 @@ type OmitRestTypeHelper<
 	Tail extends UnknownArrayOrTuple,
 	Type extends UnknownArrayOrTuple,
 	Result extends UnknownArrayOrTuple = [],
-> = Tail extends []
+> = Tail extends readonly []
```
som-sm, line 89: I was getting a bunch of `Type instantiation is excessively deep and possibly infinite.` errors in this file after flipping the default value.

I suspected it was probably an `extends somearray` conditional that needed fixing, and luckily, the very first conditional I updated made all those errors go away 🤷‍♂️

sindresorhus:
> IMO, instead of flipping the default value for preserveReadonly, we should remove the option altogether.

👍

test-d/array-tail.ts
```diff
@@ -1,48 +1,49 @@
...
 
 // All optional elements
 expectType<['b'?]>([] as ArrayTail<['a'?, 'b'?]>);
+expectType<readonly [number?]>({} as ArrayTail<readonly [string?, number?]>);
+
+// Rest element
+expectType<readonly [number, boolean, ...string[]]>({} as ArrayTail<readonly [string, number, boolean, ...string[]]>); // Required & Rest
+expectType<readonly [number?, boolean?, ...string[]]>({} as ArrayTail<readonly [string?, number?, boolean?, ...string[]]>); // Optional & Rest
+expectType<readonly [number, boolean?, ...string[]]>({} as ArrayTail<readonly [string, number, boolean?, ...string[]]>); // Required, Optional & Rest
+// expectType<readonly [...string[], string, number]>({} as ArrayTail<readonly [...string[], string, number]>); // Rest & Required
```
som-sm, line 36: This is a bug, that needs fixing. I'll open a PR shortly.

PR #1175, opened 2025-06-08
`ArrayTail`: Fix behaviour with non-tuple arrays

som-sm:
This fixes the behaviour of `ArrayTail` with non-tuple arrays (e.g., `string[]`) and arrays with a leading rest element (e.g., `[...string[], string]`.

- Non tuple arrays

  ```ts
  type Current = ArrayTail<string[]>;
  //    ^? type Current = []
  
  type Updated = ArrayTail<string[]>;
  //    ^? type Updated = string[]
  ```
  Returning `string[]` for `Array<string[]>` makes more sense than returning `[]` because that better aligns with the runtime behaviour.
  ```diff
    const test = ['foo', 'bar', 'baz'];
    //    ^? const test: string[]
    
    const tail = test.slice(1) as ArrayTail<typeof test>; // ["bar", "baz"]
  - //    ^? const tail: []
  + //    ^? const tail: string[]
  ```

- Array with leading rest element

  ```ts
  type Current = ArrayTail<[...string[], string]>;
  //    ^? type Current = []
  
  type Updated = ArrayTail<[...string[], string]>;
  //    ^? type Updated = [...string[], string]
  ```


source/array-tail.d.ts
```diff
@@ -24,15 +24,18 @@ add3(4);
...
 @category Array
 */
-export type ArrayTail<TArray extends UnknownArray> =
+export type ArrayTail<TArray extends UnknownArray> = IfNotAnyOrNever<TArray,
 	TArray extends UnknownArray // For distributing `TArray`
 		? _ArrayTail<TArray> extends infer Result
 			? If<IsArrayReadonly<TArray>, Readonly<Result>, Result>
 			: never // Should never happen
-		: never; // Should never happen
+		: never
```
som-sm, line 32: The comment here should still be there, but had to remove it because of a bug in `@stylistic/type-generic-spacing`.

image

source/merge-deep.d.ts
```diff
@@ -8,12 +8,20 @@ import type {
...
 import type {SimplifyDeep} from './simplify-deep.d.ts';
 import type {UnknownArray} from './unknown-array.d.ts';
 
+type Mutable<TArray> = {-readonly [T in keyof TArray]: TArray[T]};
+
+type ArrayTail<TArray> = TArray extends UnknownArray
+	? keyof TArray & `${number}` extends never
+		? []
+		: Mutable<_ArrayTail<TArray>>
+	: never;
```
som-sm: Again, there were a bunch of `Type instantiation is excessively deep and possibly infinite` errors in this file.

This type is quite complicated to _really_ understand where the actual issue is, so for now I've just negated all the recent changes made in `ArrayTail`. It'll be interesting to sit down and simplify the implementation of `MergeDeep`, I think there should be some room for simplification.

<br>

Also, I had to use a custom implementation for `Writable` because the current one doesn't work well with all types of arrays. This needs fixing too.

PR #1179, opened 2025-06-09
`ArrayTail`: Fix incorrect JSDoc example

som-sm:
The `curry` implementation currently in the JSDoc is incorrect.
https://github.com/sindresorhus/type-fest/blob/f3aabd8a5dd5efb6e6bc5b414201b9c5e9ddcdf7/source/array-tail.d.ts#L12-L22

The error is evident if the function accepts more than two arguments. For example:
```ts
declare const curry: <Arguments extends unknown[], Return>(
  function_: (...arguments_: Arguments) => Return,
  ...arguments_: ArrayTail<Arguments>
) => (...arguments_: ArrayTail<Arguments>) => Return;

const test = (a: string, b: number, c: boolean) => [a, b, c];

// 1. Expects the last two arguments
const add3 = curry(test, 3, false); 

// 2. Again expects the last two arguments
add3(4, false);
```

This PR fixes the `curry` implementation and also adds a simpler set of examples.

### DIFF
--- a/source/array-tail.d.ts
+++ b/source/array-tail.d.ts
@@ -35,10 +35,10 @@ type G = ArrayTail<readonly [...string[], 1, 2]>;
 ```
 import type {ArrayTail} from 'type-fest';
 
-type Curry<Func> = Func extends (...args_: infer Args) => infer ReturnValue
-	? Args extends readonly []
-		? ReturnValue
-		: (arg: Args[0]) => Curry<(...args: ArrayTail<Args>) => ReturnValue>
+type Curry<Func> = Func extends (...agruments_: infer Arguments) => infer Return
+	? Arguments extends readonly []
+		? Return
+		: (agrument: Arguments[0]) => Curry<(...agruments_: ArrayTail<Arguments>) => Return>
 	: never;
 
 declare function curry<Func extends Function>(fn: Func): Curry<Func>;


source/array-tail.d.ts
```diff
@@ -9,17 +9,44 @@ Extracts the type of an array or tuple minus the first element.
...
+
+@example
+```
+import type {ArrayTail} from 'type-fest';
+
+type Curry<Func> = Func extends (...args_: infer Args) => infer ReturnValue
+	? Args extends readonly []
+		? ReturnValue
+		: (arg: Args[0]) => Curry<(...args: ArrayTail<Args>) => ReturnValue>
+	: never;
```
sindresorhus:
```suggestion
type Curry<Func> = Func extends (...arguments_: infer Arguments) => infer ReturnValue
	? Arguments extends readonly []
		? ReturnValue
		: (argument: Arguments[0]) => Curry<(...arguments_: ArrayTail<Arguments>) => ReturnValue>
	: never;
```

som-sm: @sindresorhus Interesting that `XO` disables the default `unicorn/prevent-abbreviations` replacements. Otherwise, I would have seen errors regarding `Args`.

image

sindresorhus: The goal is to eventually enable everything there, but I had to do it slowly so not to overload users with too many validations.

PR #1180, opened 2025-06-10
Add `IsOptional`, `IsNullable` types

### LINKED ISSUE #900: Add `IsNullable` and `IsOptional`
I'd like to add `IsNullable` and `IsOptional` types to check if they contain `null` and `undefined`, respectively.

```ts
type IsNullable<T> = Extract<T, null> extends never ? false : true
type IsOptional<T> = Extract<T, undefined> extends never ? false : true
```

```ts
type T1 = IsNullable<string> // false
type T2 = IsNullable<string | null> // true
type T3 = IsNullable<string | undefined> // false
type T4 = IsNullable<string | null | undefined> // true

type T5 = IsOptional<string> // false
type T6 = IsOptional<string | null> // false
type T7 = IsOptional<string | undefined> // true
type T8 = IsOptional<string | null | undefined> // true
```

I'm currently using these types to generate a Zod schema based on an interface. They help choose the appropriate type depending on whether a value can be null or undefined.

```ts
type Zodify<T extends Record<string, any>> = {
  [K in keyof T]-?: IsNullable<T[K]> extends true
    ? z.ZodNullable<z.ZodType<T[K]>>
    : IsOptional<T[K]> extends true
      ? z.ZodOptional<z.ZodType<T[K]>>
      : z.ZodType<T[K]>
};

interface Foo {
  bar: string;
  nullable: number | null;
  optional?: string;
  optionalNullable?: string | null;
}

const schema = z.object({
  bar: z.string(),
  nullable: z.number().nullable(),
  optional: z.string().optional(),
  optionalNullable: z.string().optional().nullable(), // should chain optional first
} satisfies Zodify<Foo>);
```

jinyongp:
- `IsOptional` - Returns a boolean for whether the given type includes `undefined`.
- `IsNullable` - Returns a boolean for whether the given type includes `null`.

resolved #900

### DIFF
--- a/source/is-nullable.d.ts
+++ b/source/is-nullable.d.ts
@@ -1,0 +1,24 @@
+/**
+Returns a boolean for whether the given type includes `null`.
+
+@example
+```ts
+import type {IsNullable} from 'type-fest';
+
+type A = IsNullable<string>
+//=> false
+
+type B = IsNullable<string | null>
+//=> true
+
+type C = IsNullable<string | undefined>
+//=> false
+
+type D = IsNullable<string | null | undefined>
+//=> true
+```
+
+@category Type Guard
+@category Utilities
+*/
+export type IsNullable<T> = Extract<T, null> extends never ? false : true
--- a/source/is-optional.d.ts
+++ b/source/is-optional.d.ts
@@ -1,0 +1,24 @@
+/**
+Returns a boolean for whether the given type includes `undefined`.
+
+@example
+```ts
+import type {IsOptional} from 'type-fest';
+
+type A = IsOptional<string>
+//=> false
+
+type B = IsOptional<string | null>
+//=> false
+
+type C = IsOptional<string | undefined>
+//=> true
+
+type D = IsOptional<string | null | undefined>
+//=> true
+```
+
+@category Type Guard
+@category Utilities
+*/
+export type IsOptional<T> = Extract<T, undefined> extends never ? false : true;
--- a/test-d/is-nullable.ts
+++ b/test-d/is-nullable.ts
@@ -1,0 +1,18 @@
+import { expectType } from 'tsd';
+import type { IsNullable } from '../source/is-nullable.js';
+
+expectType<IsNullable<any>>(true);
+expectType<IsNullable<null>>(true);
+expectType<IsNullable<null | undefined>>(true);
+expectType<IsNullable<string | null>>(true);
+expectType<IsNullable<string | null | undefined>>(true);
+
+expectType<IsNullable<string>>(false);
+expectType<IsNullable<string | undefined>>(false);
+
+expectType<IsNullable<void>>(false)
+expectType<IsNullable<undefined>>(false);
+expectType<IsNullable<never>>(false);
+expectType<IsNullable<unknown>>(false);
+expectType<IsNullable<() => void>>(false);
+expectType<IsNullable<string & null>>(false);
--- a/test-d/is-optional.ts
+++ b/test-d/is-optional.ts
@@ -1,0 +1,18 @@
+import { expectType } from 'tsd';
+import type { IsOptional } from '../source/is-optional.js';
+
+expectType<IsOptional<any>>(true);
+expectType<IsOptional<undefined>>(true);
+expectType<IsOptional<null | undefined>>(true);
+expectType<IsOptional<string | undefined>>(true);
+expectType<IsOptional<string | null | undefined>>(true);
+
+expectType<IsOptional<string>>(false);
+expectType<IsOptional<string | null>>(false);
+
+expectType<IsOptional<void>>(false)
+expectType<IsOptional<null>>(false);
+expectType<IsOptional<never>>(false);
+expectType<IsOptional<unknown>>(false);
+expectType<IsOptional<() => void>>(false);
+expectType<IsOptional<string & undefined>>(false);
--- a/index.d.ts
+++ b/index.d.ts
@@ -151,6 +151,7 @@ export type {IsUnion} from './source/is-union.d.ts';
 export type {IsLowercase} from './source/is-lowercase.d.ts';
 export type {IsUppercase} from './source/is-uppercase.d.ts';
 export type {IsOptional} from './source/is-optional.d.ts';
+export type {IsNullable} from './source/is-nullable.d.ts';
 
 // Template literal types
 export type {CamelCase} from './source/camel-case.d.ts';
--- a/index.d.ts
+++ b/index.d.ts
@@ -150,6 +150,7 @@ export type {If} from './source/if.d.ts';
 export type {IsUnion} from './source/is-union.d.ts';
 export type {IsLowercase} from './source/is-lowercase.d.ts';
 export type {IsUppercase} from './source/is-uppercase.d.ts';
+export type {IsOptional} from './source/is-optional.d.ts';
 
 // Template literal types
 export type {CamelCase} from './source/camel-case.d.ts';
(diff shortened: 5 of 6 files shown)


source/is-nullable.d.ts
```diff
@@ -0,0 +1,24 @@
+/**
+Returns a boolean for whether the given type includes `null`.
+
+@example
+```ts
+import type {IsNullable} from 'type-fest';
+
+type A = IsNullable<string>
```
sindresorhus:
```suggestion
type A = IsNullable<string>;
```

source/is-nullable.d.ts
```diff
@@ -0,0 +1,24 @@
...
+//=> false
+
+type D = IsNullable<string | null | undefined>;
+//=> true
+```
+
+@category Type Guard
+@category Utilities
+*/
+export type IsNullable<T> = Extract<T, null> extends never ? false : true;
```
som-sm: Should it be named `IsNull`, because `IsNullable` gives a sense that it checks for both `null` and `undefined`?

source/is-optional.d.ts
```diff
@@ -0,0 +1,24 @@
...
+//=> false
+
+type D = IsOptional<string | null | undefined>;
+//=> true
+```
+
+@category Type Guard
+@category Utilities
+*/
+export type IsOptional<T> = Extract<T, undefined> extends never ? false : true;
```
som-sm: And, this should probably be named `IsUndefined`.

source/is-nullable.d.ts
```diff
@@ -0,0 +1,24 @@
+/**
+Returns a boolean for whether the given type includes `null`.
```
som-sm, line 4: Since this type checks whether the given type **includes** `null` and _not_ whether the given type **is** `null`, it probably makes sense to _not_ prefix it with `Is`. Maybe, `HasNull`?

Because I'd expect `IsNull<string | null>` to return `boolean` instead of `true`. @sindresorhus WDYT?

test-d/is-nullable.ts
```diff
@@ -0,0 +1,17 @@
...
+
+expectType<IsNullable<string>>(false);
+expectType<IsNullable<string | undefined>>(false);
+
+expectType<IsNullable<void>>(false);
+expectType<IsNullable<undefined>>(false);
+expectType<IsNullable<never>>(false);
+expectType<IsNullable<unknown>>(false);
+expectType<IsNullable<() => void>>(false);
+expectType<IsNullable<string & null>>(false);
```
som-sm, line 18: Add case:
```ts
expectType<IsNullable<any>>(true);
```

test-d/is-optional.ts
```diff
@@ -0,0 +1,17 @@
...
+
+expectType<IsOptional<string>>(false);
+expectType<IsOptional<string | null>>(false);
+
+expectType<IsOptional<void>>(false);
+expectType<IsOptional<null>>(false);
+expectType<IsOptional<never>>(false);
+expectType<IsOptional<unknown>>(false);
+expectType<IsOptional<() => void>>(false);
+expectType<IsOptional<string & undefined>>(false);
```
som-sm, line 18: Add case:
```ts
expectType< IsOptional<any>>(true);
```

source/is-nullable.d.ts
jinyongp: The `IsNull` type already exists. [source](https://github.com/sindresorhus/type-fest/blob/main/source/is-null.d.ts)

`IsNull` is used to check if a given type is null, while `IsNullable` checks if a given union type includes null. The same goes for `IsOptional`.

source/is-optional.d.ts
jinyongp: While `IsUndefined` does not exist, if it did, it would be used to check if a given type is `undefined`, similar to `IsNull`. `IsOptional` checks for the inclusion of `undefined`.

source/is-nullable.d.ts
jinyongp: If names like `HasNull` or `HasUndefined` are more appropriate, I will make the adjustments.

test-d/is-nullable.ts
jinyongp: I had previously included that test case, but I removed it due to the following error occurring during the tsd test.

```sh
  test-d/is-nullable.ts:4:0
  ✖  4:0  Parameter type boolean is declared too wide for argument type true.  

  test-d/is-optional.ts:4:0
  ✖  4:0  Parameter type boolean is declared too wide for argument type true.  
```

https://github.com/sindresorhus/type-fest/actions/runs/15548740799/job/43775129730?pr=1180

No typescript error occurred, and it verified as `true`.

test-d/is-optional.ts
jinyongp: As mentioned above.

source/is-optional.d.ts
sindresorhus: I think `IsOptional` is fine. It would also match when we eventually add an `Optional` type: https://github.com/sindresorhus/type-fest/issues/31

source/is-nullable.d.ts
sindresorhus: `null` and `nullable` is not the same. I think `IsNullable` is the correct name for this utility. `HasNull` would work, but is not as good.

som-sm: @sindresorhus Ok, looks like there's some confusion.

There are two things:
1. This type just checks for `null`, not `null | undefined` (i.e. "nullable"). So, `IsNullable<undefined>` is currently `false`. So, either we should fix the implementation _or_ the name.

2. Our `Is*` types currently return `boolean` if the input is a mixed type, like `IsLowercase<"a" | "A">` returns `boolean`. But this type never returns `boolean`, so `IsNullable<string | null>` is currently `true` (not `boolean`). So, I was suggesting to _not_ prefix it with `Is`, or we can update the implementation to return `boolean` (but I don't think returning `boolean` makes sense here).

test-d/is-nullable.ts
som-sm: @jinyongp That's because currently `IsNullable<any>` returns `boolean`, you need to add an extra `IsAny` check in both the types.

jinyongp: I've fixed it. Thank you for letting me know.

source/is-nullable.d.ts
sindresorhus: 1. The problem is the TS type `NonNullable`. It removes both `null` and `undefined`, but that's wrong according to its name. `Nullable` means something that can be `null`. `Nullish` is something that can be either `null` or `undefined`.
2. They're fundamentally different cases. IsNullable checks type structure, so it should return true | false. IsLowercase analyzes string content, so unions make it return boolean.

som-sm: 1. Gotcha, I see where the confusion was, I was treating "nullable" as both `null | undefined` similar to TS's `NonNullable` type. But, I get your point now, apologies for the confusion.

2. Cool, I don't have a strong preference here, it was just a suggestion.

Also @sindresorhus, I missed that we already have a `IsNull` type. Wouldn’t it be confusing to have an `IsNullable` as well, considering “nullable” is a synonym for `null`? The only difference is that `IsNull<string | null>` returns `false`, while `IsNullable<string | null>` returns `true`.

som-sm: > I missed that we already have a `IsNull` type. Wouldn’t it be confusing to have an `IsNullable` as well, considering “nullable” is a synonym for `null`? The only difference is that `IsNull<string | null>` returns `false`, while `IsNullable<string | null>` returns `true`.

Maybe I'm _just_ confused, probably `IsNull` means—is it exactly `null`, and `IsNullable` means—can it be `null`. If so, please ignore my comment, apologies for the noise.

sindresorhus: @jinyongp Can you add a note about how it differs from `NonNullable` and how that one is technically incorrect. Just in case users wonder why it doesn't handle `undefined`.

jinyongp:
```
@description
It checks for null only, while NonNullable removes both null and undefined.
```

Please let me know if you want a different explanation or if it needs to be written in a different format.

PR #1183, opened 2025-06-10
Add `Filter` type

benzaria:
Filters elements from an `Array` based on whether they match the given `Type`.
If `Type` is `Boolean`, it filters out `falsy` values like `Boolean(T)` does.

Strict controls whether strict or loose type comparison is used (defaults to loose).

### DIFF
--- a/source/internal/array.d.ts
+++ b/source/internal/array.d.ts
@@ -1,7 +1,10 @@
 import type {If} from '../if.d.ts';
+import type {IsAny} from '../is-any.d.ts';
 import type {IsNever} from '../is-never.d.ts';
-import type {OptionalKeysOf} from '../optional-keys-of.d.ts';
+import type {IsUnion} from '../is-union.d.ts';
+import type {EmptyObject} from '../empty-object.d.ts';
 import type {UnknownArray} from '../unknown-array.d.ts';
+import type {OptionalKeysOf} from '../optional-keys-of.d.ts';
 import type {IsExactOptionalPropertyTypesEnabled, IfNotAnyOrNever} from './type.d.ts';
 
 /**
@@ -156,3 +159,22 @@ type _CollapseRestElement<
 				>
 				: never // Should never happen, since `[(infer First)?, ...infer Rest]` is a top-type for arrays.
 		: never; // Should never happen
+
+/**
+Represents a empty array, the `readonly? []` value.
+*/
+export type EmptyArray = readonly [] | [];
+
+/**
+Cleans any extra empty arrays/objects from a union
+*/
+type CleanEmpty<T> = IsUnion<T> extends true
+	? T extends EmptyArray | EmptyObject
+		? never
+		: T
+	: T;
+
+type IsLeadingRestElement<T extends UnknownArray> =
+	If<IsAny<T>, false, number extends T['length']
+		? T[number] extends T[0] ? true : false
+		: false>;
--- a/source/internal/array.d.ts
+++ b/source/internal/array.d.ts
@@ -168,13 +168,42 @@ export type EmptyArray = readonly [] | [];
 /**
 Cleans any extra empty arrays/objects from a union
 */
-type CleanEmpty<T> = IsUnion<T> extends true
+export type CleanEmpty<T> = IsUnion<T> extends true
 	? T extends EmptyArray | EmptyObject
 		? never
 		: T
 	: T;
 
-type IsLeadingRestElement<T extends UnknownArray> =
-	If<IsAny<T>, false, number extends T['length']
-		? T[number] extends T[0] ? true : false
-		: false>;
+/**
+Determines whether the first element of a tuple is a rest element (e.g., `...string[]`).
+
+This is useful for identifying tuple types that begin with a variadic segment.
+
+@example
+```
+type A = IsLeadingRestElement<[...string[], number]>;
+//=> true
+
+type B = IsLeadingRestElement<[number, string]>;
+//=> false
+
+type C = IsLeadingRestElement<[...any[]]>;
+//=> true
+
+type D = IsLeadingRestElement<[]>;
+//=> false
+
+type E = IsLeadingRestElement<[undefined, ...number[]]>;
+//=> false
+```
+*/
+// ! This type is not perfect wet but its working fine with `Filter` for now.
+export type IsLeadingRestElement<T extends UnknownArray> =
+	IsAny<T> extends true ? false
+		: number extends T['length']
+			? T[number] extends T[0]
+				? T extends [infer H, ...infer R] // Prevent `[type, ...type[]]` from being `true`
+					? false
+					: true
+				: false
+			: false;
--- a/source/internal/type.d.ts
+++ b/source/internal/type.d.ts
@@ -2,6 +2,7 @@ import type {If} from '../if.d.ts';
 import type {IsAny} from '../is-any.d.ts';
 import type {IsNever} from '../is-never.d.ts';
 import type {Primitive} from '../primitive.d.ts';
+import type {ExtendsStrict} from '../extends-strict.d.ts';
 
 /**
 Matches any primitive, `void`, `Date`, or `RegExp` value.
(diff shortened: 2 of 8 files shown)


source/filter.d.ts
```diff
@@ -0,0 +1,101 @@
...
+			: [
+				...HeadAcc,
+				...IfFilter<FilterType<Array_[number], Type, Strict>, Array_>,
+				...TailAcc,
+			]
+		: Array_ extends readonly [(infer First)?, ...infer Rest]
+			? _Filter<Rest, Type, Strict, [
+				...HeadAcc,
+				...IfFilter<FilterType<First, Type, Strict>,
+					'0' extends OptionalKeysOf<Array_> // TODO: replace with `IsOptionalKeyOf`.
```
sindresorhus: This can be done now.

source/filter.d.ts
```diff
@@ -0,0 +1,103 @@
...
+*/
+type IfFilter<T extends boolean, V extends UnknownArray> = [T] extends [true] ? V : [];
+
+/**
+Filters elements from an array based on whether they extend the given type.
+
+If `Type` is `Boolean`, it filters out `falsy` values like {@link Boolean `Boolean(T)`} does.
+
+Strict controls whether strict or loose type comparison is used (defaults to loose).
+
```
sindresorhus, line 69: Need proper usage examples.

sindresorhus:
Needs to be added to readme and index.d.ts

sindresorhus:
Make sure to go through the pull request yourself and the diff and find mistakes. Etc. AI can be helpful for this.

source/filter.d.ts
benzaria: sure thing!

benzaria:
should the name of `ObjectFilter` type be `FilterObject` instead.

source/filter.d.ts
```diff
@@ -0,0 +1,150 @@
...
+	type T1 = Filter<[1, 2, 3 | 4, 3?, 4?], 3>;
+	//=> [3 | 4, 3?]
+
+	type T2 = Filter<[1, 2, 3 | 4, 3?, 4?], 3, {strict: true}>;
+	//=> [3?]
+
+	type T3 = Filter<{a: 1; b: 2; c: 3 | 4; d: 3?; e: 4?}; 3>;
+	//=> {c: 3 | 4; d: 3?}
+
+	type T4 = Filter<{a: 1; b: 2; c: 3 | 4; d: 3?; e: 4?}; 3, {strict: true}>;
```
benzaria: I will fix this example

sindresorhus:
> should the name of ObjectFilter type be FilterObject instead.

`FilterObject`

source/filter.d.ts
```diff
@@ -0,0 +1,150 @@
...
+	type T1 = Filter<[1, 2, 3 | 4, 3?, 4?], 3>;
+	//=> [3 | 4, 3?]
+
+	type T2 = Filter<[1, 2, 3 | 4, 3?, 4?], 3, {strict: true}>;
+	//=> [3?]
+
+	type T3 = Filter<{a: 1; b: 2; c: 3 | 4; d: 3?; e: 4?}, 3>;
+	//=> {c: 3 | 4; d: 3?}
+
+	type T4 = Filter<{a: 1; b: 2; c: 3 | 4; d: 3?; e: 4?}, 3, {strict: true}>;
```
sindresorhus:
```suggestion
	type T4 = Filter<{a: 1; b: 2; c: 3 | 4; d: 3?; e?: 4}, 3, {strict: true}>;
```

etc

test-d/object-filter.ts
```diff
@@ -0,0 +1,113 @@
...
+
+// Edge cases
+expectType<ObjectFilter<any, never>>({});
+expectType<ObjectFilter<never, any>>({} as never);
+expectType<ObjectFilter<never, never>>({} as never);
+
+expectType<ObjectFilter<{}, number>>({});
+expectType<ObjectFilter<{a: never; b: never}, number>>({});
+expectType<ObjectFilter<{a: never; b: never}, never>>({});
+expectType<ObjectFilter<{a: never; b: never}, never, {strict: true}>>({} as {a: never; b: never});
```
sindresorhus: some more tests:

```ts
// Readonly properties (verifies readonly modifiers are preserved)
expectType<ObjectFilter<{readonly a: 1; b: 2; readonly c: 3}, number>>({} as {readonly a: 1; b: 2; readonly c: 3});
expectType<ObjectFilter<{readonly a: string; b: number; readonly c: boolean}, string>>({} as {readonly a: string});
expectType<ObjectFilter<{readonly a?: string; b?: number; c: boolean}, string>>({} as {readonly a?: string});
```

benzaria: I totally missed to add cases for readonly props and arrays.

source/filter.d.ts
benzaria: I fixed them.

benzaria:
> > should the name of ObjectFilter type be FilterObject instead.
> 
> `FilterObject`

Just pointing out that we also could merge them in one `Filter` type.

sindresorhus:
A single type is tempting. Any downsides?

benzaria:
> A single type is tempting. Any downsides?

should not be any, I guess.

benzaria:
@sindresorhus the merge of both types was flawless with no problems.

source/internal/array.d.ts
```diff
@@ -87,6 +88,42 @@ Returns whether the given array `T` is readonly.
...
+type Pass2 = IsEmptyArray<readonly []>;
+//=> true
+
+type Fail1 = IsEmptyArray<[0]>;
+//=> false
+
+type Fail2 = IsEmptyArray<[0?]>;
+//=> false
+
+type Fail3 = IsEmptyArray<...string[]>;
```
sindresorhus: I don't think this is valid TS. Ensure all examples are correct.

source/internal/type.d.ts
```diff
@@ -129,4 +130,37 @@ export type IsExactOptionalPropertyTypesEnabled = [(string | undefined)?] extend
...
+	false: ExtendsLoose<T, U>;
+}[`${S}`];
+
+/**
+Performs a loose type comparison: checks if any of the members in `T` extends `U`.
+
+This is useful when needing to know if any member of `T` extends `U` without returning `boolean`.
+*/
+export type ExtendsLoose<T, U> = IsNotFalse<T extends U ? true : false>;
+// ? Should this type get exposed publicly
```
sindresorhus: We can discuss that later if there is a need. No need for comment about it on every type.

source/internal/type.d.ts
```diff
@@ -129,4 +130,37 @@ export type IsExactOptionalPropertyTypesEnabled = [(string | undefined)?] extend
...
+
+/**
+Union of `falsy` types in TypeScript. `never` isn't included.
+*/
+export type Falsy = 0 | 0n | '' | false | null | undefined; // `| never`
+
+/**
+Checks if `T` is **not** {@link Falsy `falsy`} similar to `Boolean(T)`.
+*/
+// TODO: expose this type publicly
```
sindresorhus:
```suggestion
```

source/internal/type.d.ts
benzaria: For sure!

source/internal/array.d.ts
benzaria: You're right, I fixed it.

sindresorhus:
Some things:

1. The `Boolean` special-case is currently too broad.

Right now truthy filtering is being selected not only for `Boolean`, but also for wider filter types like `unknown` and `any`. That changes behavior in a surprisng way. Example: filtering `[0, "", false]` by `unknown` ends up as `[]`, when it should keep all values.

2. Union results ar being over-narrowed.

When filtering a union, branches that become empty are removed from the final result. That makes the output unsound for real union inputs.

Examples:
- `Filter<[1] | [2], 1>` resolves to `[1]`, but should be `[1] | []`.
- `Filter<{a: 1} | {b: 2}, 1>` resolves to `{a: 1}`, but should be `{a: 1} | {}`.

3. One strict object example in docs is incorrect/inconsistant with the rest of the examples.

4. Tests are missing coverage for the two cases above.

- filtering by `unknown`/`any` with falsy values,
- union branches where one side filters to an empty result.

PR #1184, opened 2025-06-10
Fix `AbstractClass` link to TS docs

AllenSH12:
There is a docs deprecation warning when visiting the old link, use the new one that describes `AbstractClass`


sindresorhus:
You need to update the link in the type docs too.

AllenSH12:
done thank you for the catch 🙏

PR #1189, opened 2025-06-15
Add `IsOptionalKeyOf`, `IsReadonlyKeyOf`, `IsRequiredKeyOf`, and `IsWritableKeyOf` types

benzaria:
adding object/array keys type guards `Is*KeyOf` to check the state/modifier of key(s) 

> [!NOTE]
> Do they need seperated test ?
> I think that most of the cases are covered in `*KeysOf`, so if they ever fail we know that `Is*KeyOf` failed. 
> the only cases that should be tested is union of keys: `IsOptionalKeyOf<Type, 'a' | 'b'>`

### DIFF
--- a/source/optional-keys-of.d.ts
+++ b/source/optional-keys-of.d.ts
@@ -1,5 +1,5 @@
 /**
-Extract all optional keys from the given type.
+Extract all optional keys from the given `Type`.
 
 This is useful when you want to create a new type that contains different type values for the optional keys only.
 
@@ -45,7 +45,7 @@ export type OptionalKeysOf<Type extends object> =
 /**
 Returns a boolean to whether `Key` is an optional key of `Type`.
 
-This is useful when you want to create a new type that contains different type values for the optional keys only.
+This is useful when writing utility types or schema validators that need to differentiate `optional` keys.
 
 @example
 ```
@@ -72,7 +72,10 @@ type T2 = IsOptionalKeyOf<User, 'name'>
 type T3 = IsOptionalKeyOf<User, 'name' | 'luckyNumber'>
 //=> boolean
 
-type T4 = IsOptionalKeyOf<User | Admin, 'surname'>
+type T4 = IsOptionalKeyOf<User | Admin, 'name'>
+//=> false
+
+type T5 = IsOptionalKeyOf<User | Admin, 'surname'>
 //=> boolean
 ```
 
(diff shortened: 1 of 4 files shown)


source/optional-keys-of.d.ts
```diff
@@ -1,5 +1,5 @@
 /**
-Extract all optional keys from the given type.
+Extract all optional keys from the given `Type`.
```
sindresorhus: I prefer the previous style. The first sentence here should match the one in the readme, and the readme has no context of `Type`.

source/optional-keys-of.d.ts
```diff
@@ -31,9 +31,60 @@ const update2: UpdateOperation<User> = {
...
+			[Key in keyof Type as
+			IsOptionalKeyOf<Type, Key> extends false
+				? never
+				: Key
+			]: never
+		}) & keyof Type // Intersect with `keyof Type` to ensure result of `OptionalKeysOf<Type>` is always assignable to `keyof Type`
+		: never; // Should never happen
+
+/**
+Returns a boolean to whether `Key` is an optional key of `Type`.
```
sindresorhus:
```suggestion
Returns a boolean for whether `Key` is an optional key of `Type`.
```

Please get ChatGPT to review the code for typos and bugs.

source/optional-keys-of.d.ts
```diff
@@ -31,9 +31,60 @@ const update2: UpdateOperation<User> = {
...
+
+	luckyNumber?: number;
+}
+
+interface Admin {
+	name: string;
+	surname?: string
+}
+
+type T1 = IsOptionalKeyOf<User, 'luckyNumber'>
```
sindresorhus:
```suggestion
type T1 = IsOptionalKeyOf<User, 'luckyNumber'>;
```

som-sm (changes requested):
Please create separate files for new types.

som-sm:
Are these internal types, if so they shouldn't be exported. And if not, they should be exported from `index.d.ts`.

som-sm:
All of these new types should also be exported from `index.d.ts`.

benzaria:
Ready !

sindresorhus:
Thanks 👍

Sorry for the delay. Did not have time for reviewing pull requests in the summer.

PR #1191, opened 2025-06-16
Fix: add index signature support for `Paths`

### LINKED ISSUE #1190: Paths should handle Record types with arbitrary string keys
Currently, `Paths<Record<string, { a: number; b: number; }>>;` resolves to `never`. It should instead use template literal types and resolve to `` `${string}` | `${string}.a` | `${string}.b` ``.

benzaria:
Closes #1190 

 **Changes**
 - fix `Paths` accepts index signature types and output all possible paths.
 - refactor: split type into smaller pices and removed unnecessary conditions and repetition in `Paths`.
 - refactor: Updated `LiteralUnion` to return `(string & {})` instead of `(string & Record<never, never>)`.
 - add: Introduced `EmptyArray` and `IsEmptyArray` types (internal).
 - add: Introduced `OwnKeys` types that extract own proprety keys from object/arrays.

 **Tests**
 - add: Added tests for new features and bug fixes.
 - update: Updated existing tests to reflect changes.


source/paths.d.ts
```diff
@@ -135,6 +145,11 @@ type DefaultPathsOptions = {
 	depth: number;
 };
 
+/**
+Represents pathable keys, the `string` or `number` value.
+*/
```
som-sm: I don't think an explanation is required here, it's pretty self explanatory.

source/internal/array.d.ts
```diff
@@ -96,6 +96,32 @@ Returns whether the given array `T` is readonly.
...
+Returns a `boolean` for whether the type is an empty array, the `[]` or `readonly []` value.
+
+@example
+```
+import type {IsEmptyArray} from 'type-fest';
+
+type Pass = IsEmptyArray<[]>; //=> true
+type Pass = IsEmptyArray<readonly []>; //=> true
+type Fail = IsEmptyArray<[0]>; //=> false
+type Fail = IsEmptyArray<[0?]>; //=> false
```
sindresorhus: Code examples should be runnable, so they cannot have duplicate type names.

source/internal/array.d.ts
```diff
@@ -96,6 +96,32 @@ Returns whether the given array `T` is readonly.
 */
 export type IsArrayReadonly<T extends UnknownArray> = If<IsNever<T>, false, T extends unknown[] ? false : true>;
 
+/**
+Represents an empty array, the `[]` or `readonly []` value.
+*/
+export type EmptyArray = readonly [] | []; // The extra `[]` just to prevent TS from expanding the type.
```
sindresorhus:
```suggestion
export type EmptyArray = readonly [] | []; // The extra `[]` is just to prevent TS from expanding the type.
```

sindresorhus:
> refactor: Updated LiteralUnion to return (string & {}) instead of (string & Record<never, never>).

Can you elaborate on why that is better? And it needs a test, if possible.

test-d/empty-object.ts
```diff
@@ -34,3 +34,11 @@ const _a: unknown = bar.id;
...
 const baz: Union = {id: 42};
 expectType<{id: number}>(baz);
+
+// Edge cases
+expectType<IsEmptyObject<never>>(false);
+expectType<IsEmptyObject<any>>({} as boolean);
+expectType<IsEmptyObject<Record<never, never>>>(true);
+
+type Signature = Record<keyof any, never>;
+expectType<IsEmptyObject<Signature>>(true); // ! Should this be fixed ?
```
sindresorhus: There are two possibilities here:

1. Only exact `⁠{}` and `⁠EmptyObject` return ⁠true: predictable but less useful.
2. Any semantically empty type returns ⁠true: more intuitive but less strict.

I would personally go with 2.

However, if we go with 1, then `IsEmptyObject<Record<never, never>>` should be `false` too.

source/internal/array.d.ts
benzaria: Do I also change the examples on `IsEmptyObject` ? I copied them form there.
https://github.com/sindresorhus/type-fest/blob/1205f5366847b377854e2db5e0740137144546a8/source/empty-object.d.ts#L38-L42

test-d/empty-object.ts
benzaria: If u think about it `Record<never, never>` returns `{}`, in the other hand `Record<keyof any, never>` returns an index signature like:
```ts
type T = Record<keyof any, never>
/*  ^? {
    [x: string]: never;
    [x: number]: never;
    [x: symbol]: never;
} */
```
which is not an empty object IMO, and should return false

benzaria:
> > refactor: Updated LiteralUnion to return (string & {}) instead of (string & Record<never, never>).
> 
> Can you elaborate on why that is better? And it needs a test, if possible.

@sindresorhus Its just a visual improvement that does *not* change the functionality of the type.

I test it using this example and some others should I add a test file?
```ts
type LiteralUnion<LiteralType, BaseType extends Primitive> = LiteralType | (BaseType & Simplify<{}>);

// Pass
expectType<LiteralUnion<'foo', string>>({} as (string & {}) | 'foo');
expectType<LiteralUnion<'foo', string>>({} as (string & Record<never, never>) | 'foo');
// Fail
expectType<string>({} as (string & {}) | 'foo');
```

sins921 (approved):

sins921 (approved):

source/internal/array.d.ts
benzaria: Done!

test-d/empty-object.ts
sindresorhus: But it enforces an empty object:

```ts
type T = Record<keyof any, never>;

const a: T = {}; // Pass
const b: T = {x: true}; // Fail
```

benzaria: Ummm..., maybe u are right, no need to fix then.

source/internal/array.d.ts
```diff
@@ -96,6 +96,33 @@ Returns whether the given array `T` is readonly.
...
+
+@example
+```
+import type {IsEmptyArray} from 'type-fest';
+
+type Pass1 = IsEmptyArray<[]>; //=> true
+type Pass2 = IsEmptyArray<readonly []>; //=> true
+type Fail1 = IsEmptyArray<[0]>; //=> false
+type Fail2 = IsEmptyArray<[0?]>; //=> false
+type Fail3 = IsEmptyArray<...string[]>; //=> false
```
som-sm: Keep the result (`//=> ...`) on a separate line.

source/internal/array.d.ts
```diff
@@ -96,6 +96,33 @@ Returns whether the given array `T` is readonly.
 */
 export type IsArrayReadonly<T extends UnknownArray> = If<IsNever<T>, false, T extends unknown[] ? false : true>;
 
+/**
+Represents an empty array, the `[]` or `readonly []` value.
+*/
+export type EmptyArray = readonly [] | []; // The extra `[]` is just to prevent TS from expanding the type.
```
som-sm, line 102: Can an example for this ``// The extra `[]` is just to prevent TS from expanding the type.`` be added?

test-d/empty-object.ts
```diff
@@ -34,3 +34,11 @@ const _a: unknown = bar.id;
 
 const baz: Union = {id: 42};
 expectType<{id: number}>(baz);
+
+// Edge cases
+expectType<IsEmptyObject<never>>(false);
+expectType<IsEmptyObject<any>>({} as boolean);
+expectType<IsEmptyObject<Record<never, never>>>(true);
```
som-sm: I think it's better to add a case for `{}` as well, even if we already have `Record<never, never>`.

source/literal-union.d.ts
```diff
@@ -15,23 +17,25 @@ import type {LiteralUnion} from 'type-fest';
 
 // Before
 
-type Pet = 'dog' | 'cat' | string;
+type Pet1 = 'dog' | 'cat' | string;
+//    ^? type Pet1 = string
```
som-sm: Use `//=>`.

source/literal-union.d.ts
```diff
@@ -15,23 +17,25 @@ import type {LiteralUnion} from 'type-fest';
...
 // After
 
 type Pet2 = LiteralUnion<'dog' | 'cat', string>;
+//    ^? type Pet2 = 'dog' | 'cat' | (string & {})
 
-const pet: Pet2 = '';
-// You **will** get auto-completion for `dog` and `cat` literals.
+const pet2: Pet2 = '';
 ```
+You **will** get auto-completion for `dog` and `cat` literals.
```
som-sm: Not sure why you've made a bunch of changes here, it looks like the only intended change was to add the type for `Pet`.

source/paths.d.ts
```diff
@@ -179,14 +193,13 @@ open('listB.1'); // TypeError. Because listB only has one element.
 export type Paths<T, Options extends PathsOptions = {}> = _Paths<T, ApplyDefaultOptions<PathsOptions, DefaultPathsOptions, Options>>;
 
 type _Paths<T, Options extends Required<PathsOptions>> =
-	T extends NonRecursiveType | ReadonlyMap<unknown, unknown> | ReadonlySet<unknown>
-		? never
-		: IsAny<T> extends true
-			? never
+	T extends NonRecursiveType | ReadonlyMap<unknown, unknown> | ReadonlySet<unknown> ? never
+		: IsAny<T> extends true ? never
```
som-sm: Why change the formatting here?

source/paths.d.ts
```diff
@@ -179,14 +193,13 @@ open('listB.1'); // TypeError. Because listB only has one element.
...
 			: T extends UnknownArray
 				? number extends T['length']
-					// We need to handle the fixed and non-fixed index part of the array separately.
-					? InternalPaths<StaticPartOfArray<T>, Options> | InternalPaths<Array<VariablePartOfArray<T>[number]>, Options>
+					? // We need to handle the fixed and non-fixed index part of the array separately.
+					| InternalPaths<StaticPartOfArray<T>, Options> // ! Should this be updated to use `ExcludeRestElement`.
+					| InternalPaths<Array<VariablePartOfArray<T>[number]>, Options> // ! And this to use `ExtractRestElement`.
```
som-sm: Again, why make formatting changes?

source/paths.d.ts
```diff
@@ -195,62 +208,50 @@ type _Paths<T, Options extends Required<PathsOptions>> =
 type InternalPaths<T, Options extends Required<PathsOptions>> =
 	Options['maxRecursionDepth'] extends infer MaxDepth extends number
 		? Required<T> extends infer T
-			? T extends EmptyObject | readonly []
-				? never
+			? IsNever<keyof T> extends true
+				? never // Ignore empty array/object and not index signature types (e.g, `Record<string, 'foo'>`).
```
som-sm: Umm...`keyof []` is _not_ `never`, so I don't think `IsNever<keyof T>` can replace both `EmptyObject` and `readonly []`.

Also, add a test to ensure that the `readonly []` conditional is present.

source/paths.d.ts
```diff
@@ -195,62 +208,50 @@ type _Paths<T, Options extends Required<PathsOptions>> =
...
-									? (Key | ToString<Key>)
-									: never
-						) extends infer TranformedKey extends string | number ?
+					[Key in keyof T]: Key extends PathableKeys // Limit `Key` to string or number.
+						? (Options['bracketNotation'] extends true
+							? IsNumberLike<Key> extends true
+								? `[${Key}]`
+								: (Key | ToString<Key>)
+							: (Key | ToString<Key>) // If `Key` is a number, return `Key | `${Key}``, because both `array[0]` and `array['0']` work.
+						) extends infer TranformedKey extends PathableKeys ?
```
som-sm: Looks like an irrelevant change.

source/empty-object.d.ts
```diff
@@ -35,12 +37,16 @@ Returns a `boolean` for whether the type is strictly equal to an empty plain obj
...
 ```
 
 @see EmptyObject
 @category Object
 */
-export type IsEmptyObject<T> = T extends EmptyObject ? true : false;
+export type IsEmptyObject<T> =
+	IsNever<T> extends true ? false
+		: T extends EmptyObject ? true
+			: false;
```
som-sm: I don't think fixing `IsEmptyObject`'s behaviour with `never` is related to this PR.

source/paths.d.ts
```diff
@@ -135,6 +144,11 @@ type DefaultPathsOptions = {
 	depth: number;
 };
 
+/**
+Represents pathable keys, the `string` or `number` value.
+*/
+type PathableKeys = string | number;
```
som-sm, line 151: Ideally, this should _not_ be part of this PR. It introduces a bunch of refactoring changes, which makes it harder to review the actual bug-related changes.

source/paths.d.ts
```diff
@@ -195,62 +208,50 @@ type _Paths<T, Options extends Required<PathsOptions>> =
...
-											? (
-												Options['bracketNotation'] extends true
-													? SubPath extends `[${any}]` | `[${any}]${string}`
-														? `${TranformedKey}${SubPath}` // If next node is number key like `[3]`, no need to add `.` before it.
-														: `${TranformedKey}.${SubPath}`
-													: never
-											) | (
-												Options['bracketNotation'] extends false
-													? `${TranformedKey}.${SubPath}`
-													: never
```
som-sm: Why remove `Options['bracketNotation'] extends false` conditional in this PR?

source/paths.d.ts
```diff
@@ -195,62 +208,50 @@ type _Paths<T, Options extends Required<PathsOptions>> =
...
+				? never // Ignore empty array/object and not index signature types (e.g, `Record<string, 'foo'>`).
 				: {
-					[Key in keyof T]:
-					Key extends string | number // Limit `Key` to string or number.
-						? (
-							Options['bracketNotation'] extends true
-								? IsNumberLike<Key> extends true
-									? `[${Key}]`
-									: (Key | ToString<Key>)
-								: Options['bracketNotation'] extends false
```
som-sm: Why remove `Options['bracketNotation'] extends false` conditional in this PR?

som-sm (changes requested):
> Maybe we could return an uncollapsed union like `` (string & {}) | `${string}.a` | `${string}.b` ``, but we’d have to decide in what situations we should return an uncollapsed union.

As mentioned [here](https://github.com/sindresorhus/type-fest/issues/1190#issuecomment-2978142610), just adding a check for `string` wouldn't be sufficient, because there would be other cases as well where the result would get collapsed.

The existing implementation fails in the following cases:

- ```ts
   type T1= Paths<{[x: Uppercase<string>]: {a: string}; C: {a: string}}>;
   //=> Uppercase<string> | `${Uppercase<string>}.a`
   ```
   `C.a` is lost, result should have probably been: ``Uppercase<string> | (`${Uppercase<string>}.a` & {}) | "C.a"``

- ```ts
   type T = Paths<{a: {[x: string]: number} | {b: number}}>;
   //=> "a" | `a.${string & {}}`
   ```
   `a.b` is lost, result should have probably been ``'a' | (`a.${string}` & {}) | 'a.b'``.

There would _likely_ be more cases, hence it needs to be thought out properly.

<br>

Also, it'd be much easier to review if the PR only fixed the bug and _not_ introduce a bunch of refactoring changes along with it.

test-d/empty-object.ts
som-sm: Umm...I think both behaviours have their own use cases, maybe add an option?

source/literal-union.d.ts
```diff
@@ -15,23 +17,25 @@ import type {LiteralUnion} from 'type-fest';
...
 
 @category Type
 */
-export type LiteralUnion<
-	LiteralType,
-	BaseType extends Primitive,
-> = LiteralType | (BaseType & Record<never, never>);
+export type LiteralUnion<LiteralType, PrimitiveType extends Primitive> = LiteralType | (PrimitiveType & Simplify<{}>);
```
som-sm: Is `Simplify` also required here?

benzaria: Yes, removing it will break the type.

sins921: por favor no soy un juego tengo necesidades y mucha hamre

som-sm: In that case,  I'd say `Record<never, never>` was better.

source/internal/array.d.ts
benzaria: @som-sm What I meant is if `type EmptyArray = readonly []` alone, when used in a type it will showup when hovering over it as `readonly []`, but when `[]` is added it prevent that and show as `EmptyArray` instead.
Its just a visual preference and dont change anything.

source/empty-object.d.ts
benzaria: Sure i will reverte it!

source/paths.d.ts
benzaria: Because its so kludged and hard to navigate and read, It must be formatted.

source/paths.d.ts
benzaria: Why is the purpose of it in the first place?

source/literal-union.d.ts
benzaria: Imagine this

old behavior
```ts
type T = Paths<Record<string, {a: number; b: number}>>;
//      ^? (string & Record<never, never>) | (`${string}.b` & Record<never, never>) | (`${string}.a` & Record<never, never>)
```
new behavior
```ts
type T = Paths<Record<string, {a: number; b: number}>>;
//      ^? (string & {}) | (`${string}.b` & {}) | (`${string}.a` & {})
```

So which one do u think is better?

source/literal-union.d.ts
benzaria: For markdown, I put the comments outside the code block so they get rendered.

source/paths.d.ts
benzaria: I removed it, but its a common syntax i use for ternary operator.
```ts
| T extends string ? 'string'
: T extends number ? 'number'
: T extends symbol ? 'symbol'
: ...
```
make it more readable like an `if-elseif...else` statement

benzaria:
@sindresorhus @som-sm Hey guys, so I wanted to discuss the new changes.
I was able to add the intended behavior and covering edge cases that @som-sm mentioned:
```ts
declare const edge1: Paths<{a: {[x: string]: number} | {b: number}}>;
expectType<'a' | 'a.b' | (`a.${string}` & {})>(edge1); // Pass

declare const edge2: Paths<{[x: Uppercase<string>]: {a: string}; C: {a: string}}>;
expectType<'C' | 'C.a' | (Uppercase<string> & {}) | (`${Uppercase<string>}.a` & {})>(edge2); // Pass

declare const stringRecord: Paths<Record<string, {a: number; b: number}>>;
expectType<(string & {}) | `${string}.a` | `${string}.b`>(stringRecord); // Pass

declare const templateRecord: Paths<Record<`on${string}`, {a: number; b: number}>>;
expectType<(`on${string}` & {}) | (`on${string}.a` & {}) | (`on${string}.b` & {})>(templateRecord); // Pass

declare const complexRecord: Paths<Record<number, {a: number; b: number}> & DeepObject>;
expectType<number | `${number}` | 'a' | (`${number}.b` & {}) | (`${number}.a` & {}) | 'a.b' | 'a.b2' | 'a.b3' | 'a.b.c' | ${`a.b2.${number}` & {}} | 'a.b.c.d'>(complexRecord); // Pass
```

Explaination
Now `Paths` will return a `LiteralUnion` for any wide string type like (`string`, `Uppercase<string>`, `on${number}`)
For now test are fine but they will fail bcs `string !== (string & {})` so we need to go over test and make some changes.

At the moment this it the best result so far, but maybe I can find a way to include the `LiteralUnion` only if types extends others but that will be triky to achieve.

> Also, it'd be much easier to review if the PR only fixed the bug and not introduced a bunch of refactoring changes along with it.

I do apologize for the extra refactoring and changes, but I couldn't stand the compact type and all-in-one approach that was originally used. A best practice is to split the types into reusable pices that do one job. I apologize again 🙏for the mess.

test-d/empty-object.ts
benzaria: It does exist !

source/literal-union.d.ts
benzaria: Sure!

source/internal/array.d.ts
som-sm: @benzaria You mean something like this?

```ts
type EmptyArray = readonly [] | [];

type Test = [EmptyArray];
//   ^? type Test = [EmptyArray]
```

```ts
type EmptyArray = readonly [];

type Test = [EmptyArray];
//   ^? type Test = [readonly []]
```

benzaria: Yes !

source/paths.d.ts
benzaria: if its not `true` -> its `false` there is no in bettwen. `bracketNotation extends boolean` !

source/literal-union.d.ts
som-sm: Oh gotcha, are you aware why is `Simplify` required here?

Works fine in this simple case:
```ts
type Test = "foo" | "bar" | (string & {});
//   ^? type Test = (string & {}) | "foo" | "bar"
```

benzaria: True, a hand written `LitreralUnion` like `type Test = "foo" | "bar" | (string & {});` works.
But when used inside a type, TS tries to *simplify* the `& {}`, this behavior can be seen in:

```ts
type NonNullable<T> = T & {}
//=> returns `T` and not `T & {}`

type Simplify<T> = {[K in keyof T]: T[K]} & {}
//=> returns `T` and not `T & {}`
```

What `Record<never, never>` or `Simplify<{}>`  does is to ***Obfuscate*** the raw `& {}` form getting *simplified*.

source/internal/array.d.ts
benzaria: Done!

source/paths.d.ts
benzaria: Oh, I miseed that!
I got it fixed using the `OwnKeys<T>` type.

source/literal-union.d.ts
som-sm: I think the comment was placed next to `''`, to highlight that this is where you'd expect the auto-complete.

source/paths.d.ts
som-sm: This is _not_ the PR to do that.

source/paths.d.ts
som-sm: Why do that in this PR, what's the scope of this PR?

sindresorhus:
> I do apologize for the extra refactoring and changes, but I couldn't stand the compact type and all-in-one approach that was originally used. A best practice is to split the types into reusable pices that do one job.

We are more than happy to have refactoring and readability improvements. What we are trying to tell you is that those changes are better done separately as they make it hard to see the actual changes.

source/literal-union.d.ts
som-sm: Gotcha!

Here's a simple example illustrating the same:
```ts
import { Simplify } from 'type-fest';

type LiteralUnion1<T, Base> = T | (Base & {});
type LiteralUnion2<T, Base> = T | (Base & Simplify<{}>);
type LiteralUnion3<T, Base> = T | (Base & Record<never, never>);

type T1 = LiteralUnion1<'foo' | 'bar', string>;
//   ^? type T1 = string
type T2 = LiteralUnion2<'foo' | 'bar', string>;
//   ^? type T2 = "foo" | "bar" | (string & {})
type T3 = LiteralUnion3<'foo' | 'bar', string>;
//   ^? type T3 = "foo" | "bar" | (string & Record<never, never>)
```
Playground: https://tsplay.dev/wX8bJw

<br>

And looks like the following test would ensure that the union remains uncollapsed:
```ts
expectType<'foo' | 'bar' | (string & {})>({} as LiteralUnion<'foo' | 'bar', string>);
```
The above test passes for `LiteralUnion2` & `LiteralUnion3`, but fails for `LiteralUnion1`.

benzaria:
Okay captain, lets keep it for another PR.
@som-sm I reverted the refactoring changes.

sindresorhus:
https://github.com/sindresorhus/type-fest/pull/1191/files#r2152241210

sindresorhus:
> https://github.com/sindresorhus/type-fest/pull/1191/files#r2152241210

Bump ⬆️

sindresorhus:
Closing for lack of activity.

PR #1192, opened 2025-06-17
refactor `IsNumericLike` and fix `UnionMin`, `UnionMax` types

benzaria:
- refactor: remove unnecessary condition for `IsNumericLike` that cosed an error and add cases for `never`/`any`.
- fix: behavior of `UnionMin` to `UnionMax` returning `0` for `never` , `any` for `number`, erroring for `any` and `Infinity`.
- add: introduce `IsAnyOrNever` type alternative for `IfNotAnyOrNever`. (internal)

Original behavior:
```ts
type A = IsNumericLike<'1.2.3'>
//=> true

type T = UnionMax<never>
//=> 0

type T = UnionMax<any>
//=> any // Error: infinite
```

Current behavior:
```ts
type A = IsNumericLike<'1.2.3'>
//=> false

type T = UnionMax<never>
//=> never

type T = UnionMax<any>
//=> any
```

### DIFF
--- a/source/internal/numeric.d.ts
+++ b/source/internal/numeric.d.ts
@@ -55,10 +55,16 @@ Note: Just supports numbers from 0 to 999.
 
 @example
 ```
-type A = UnionMin<3 | 1 | 2>;
+type A = UnionMin<1 | 3 | 2>;
 //=> 1
 
-type B = UnionMin<never>;
+type B = UnionMin<number>;
+//=> 0
+
+type C = UnionMin<any>;
+//=> any
+
+type D = UnionMin<never>;
 //=> never
 ```
 */
@@ -84,13 +90,20 @@ Note: Just supports numbers from 0 to 999.
 type A = UnionMax<1 | 3 | 2>;
 //=> 3
 
-type B = UnionMax<never>;
+type B = UnionMax<number>;
+//=> number
+
+type C = UnionMax<any>;
+//=> any
+
+type D = UnionMax<never>;
 //=> never
 ```
 */
 export type UnionMax<N extends number> =
 	IsAnyOrNever<N> extends true ? N // Returns `any` or `never`
-		: InternalUnionMax<N>;
+		: [number] extends [N] ? number
+			: InternalUnionMax<N>;
 
 /**
 The actual implementation of `UnionMax`. It's private because it has some arguments that don't need to be exposed.
--- a/source/internal/numeric.d.ts
+++ b/source/internal/numeric.d.ts
@@ -1,5 +1,5 @@
 import type {IsNever} from '../is-never.d.ts';
-import type {NegativeInfinity, PositiveInfinity} from '../numeric.d.ts';
+import type {Finite, NegativeInfinity, PositiveInfinity} from '../numeric.d.ts';
 import type {UnknownArray} from '../unknown-array.d.ts';
 import type {StringToNumber} from './string.d.ts';
 import type {IsAnyOrNever} from './type.d.ts';
@@ -44,9 +44,10 @@ type E = IsNumberLike<'a'>;
 //=> false
 */
 export type IsNumberLike<N> =
-	N extends number | `${number}`
-		? true
-		: false;
+	IsAnyOrNever<N> extends true ? N
+		: N extends number | `${number}`
+			? true
+			: false;
 
 /**
 Returns the minimum number in the given union of numbers.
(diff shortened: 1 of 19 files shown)


source/internal/numeric.d.ts
```diff
@@ -27,25 +27,25 @@ Supports floating-point as a string.
...
-type A = IsNumberLike<'1'>;
+type A = IsNumberLike<1>;
 //=> true
 
-type B = IsNumberLike<'-1.1'>;
+type B = IsNumberLike<'1'>;
 //=> true
 
-type C = IsNumberLike<1>;
+type C = IsNumberLike<'-1.1'>;
```
som-sm: Looks like you've changed the order here, why?

source/internal/numeric.d.ts
```diff
@@ -27,25 +27,25 @@ Supports floating-point as a string.
...
+type E = IsNumberLike<'a'>;
 //=> false
 */
 export type IsNumberLike<N> =
-	N extends number ? true
-		:	N extends `${number}`
-			? true
-			: N extends `${number}.${number}`
-				? true
-				: false;
```
som-sm, line 48: Umm...not sure why was this check added.

source/internal/numeric.d.ts
```diff
@@ -27,25 +27,25 @@ Supports floating-point as a string.
...
 //=> true
 
-type D = IsNumberLike<'a'>;
+type D = IsNumberLike<'5e-20'>;
+//=> true
+
+type E = IsNumberLike<'a'>;
 //=> false
 */
 export type IsNumberLike<N> =
```
som-sm, line 46: Please add test for `'1.2.3'`.

source/internal/numeric.d.ts
```diff
@@ -64,9 +64,11 @@ export type UnionMin<N extends number> = InternalUnionMin<N>;
 The actual implementation of `UnionMin`. It's private because it has some arguments that don't need to be exposed.
 */
 type InternalUnionMin<N extends number, T extends UnknownArray = []> =
-	T['length'] extends N
+	IsNever<N> extends true
```
som-sm: `UnionMin<never>` should return `never`.

Also, add test for this.

som-sm: Also, check behaviour with `any` and `number`. `any` should return `any` (maybe use `IfNotAnyOrNever`) & `number` should return `number`.

source/internal/numeric.d.ts
benzaria: from most obvious to less, number->`number`->float->scientific

source/internal/numeric.d.ts
benzaria: What about `UnionMax` it also returns `0` for `never`

source/internal/numeric.d.ts
som-sm: IMO, the existing order was more obvious as `IsNumberLike` is mostly meant to be used with strings, it just happens to work with numbers as well—it's `IsNumber[Like]` not `IsNumber`.

source/internal/numeric.d.ts
som-sm: Yeah, `UnionMax` should _also_ return `never` for `never`.

source/internal/numeric.d.ts
benzaria: No problem, reverted.

source/internal/numeric.d.ts
benzaria: Done and also added some extra cases

benzaria:
**Reason to introduce `IsAnyOrNever`:**

```ts
type CountTo<T> = count to T;

type Count1<T> = IfNotAnyOrNever<T, CountTo>;
type Count2<T> = IsAnyOrNever<T> extends true ? T : CountTo<T>;

type A = Count1<any>
//       ~~~~~~~~~~
// Error: Type instantiation is excessively deep and possibly infinite.

type B = Count2<any>
//=> any
```

IMO, `IsAnyOrNever` is a better alternative and should replace the current implementation of `IfNotAnyOrNever`. This is because, when using an `If`-style utility, TypeScript always evaluates both branches. That means even the *falsy* or undesired branch is computed, which can result in unnecessary work, potential type errors, or wasted computation despite it ultimately being discarded.

benzaria:
@som-sm should `UnionMax` return `PositiveInfinity` for `number`? I feel its more precise because `number` do include '**+∞**' and its the maximum number. the same for `UnionMin` should return '**-∞**'. WDYT?

sindresorhus:
> should UnionMax return PositiveInfinity for number? I feel its more precise because number do include '+∞' and its the maximum number. the same for UnionMin should return '-∞'. WDYT?

I think it should stay `number`.

PR #1194, opened 2025-06-17
Add `RemovePrefix` type

som-sm:
Adds `RemovePrefix` type as mentioned in #1188.

Let me know if the naming feels off, here are a few alternatives:
- `StripLeading`
- `StripPrefix`
- `TrimPrefix`

### DIFF
--- a/source/remove-prefix.d.ts
+++ b/source/remove-prefix.d.ts
@@ -8,7 +8,7 @@ import type {Or} from './or.d.ts';
 */
 type RemovePrefixOptions = {
 	/**
-	When enabled, instantiations with non-literal prefixes (e.g., `string`, `Uppercase<string>`, `` `on${string}` ``) simply return back `string`, since their precise structure cannot be statically determined.
+	When enabled, instantiations with non-literal prefixes (e.g., `string`, `Uppercase<string>`, `` `on${string}` ``) simply return `string`, since their precise structure cannot be statically determined.
 
 	Note: Disabling this option can produce misleading results that might not reflect the actual runtime behavior.
 	For example, ``RemovePrefix<'on-change', `${string}-`, {strict: false}>`` returns `'change'`, but at runtime, prefix could be `'handle-'` (which satisfies `` `${string}-` ``) and removing `'handle-'` from `'on-change'` would not result in `'change'`.
@@ -65,13 +65,13 @@ type RemovePrefixOptions = {
 	type A = RemovePrefix<`${string}/${number}`, `${string}:`, {strict: true}>;
 	//=> `${string}/${number}`
 
-	type B = RemovePrefix<`${string}/${number}`, `${string}:`, {strict: true}>;
+	type B = RemovePrefix<`${string}/${number}`, `${string}:`, {strict: false}>;
 	//=> `${string}/${number}`
 
 	type C = RemovePrefix<'on-change', `${number}-`, {strict: true}>;
 	//=> 'on-change'
 
-	type D = RemovePrefix<'on-change', `${number}-`, {strict: true}>;
+	type D = RemovePrefix<'on-change', `${number}-`, {strict: false}>;
 	//=> 'on-change'
 	```
 	*/
--- a/source/remove-prefix.d.ts
+++ b/source/remove-prefix.d.ts
@@ -1,0 +1,128 @@
+import type {ApplyDefaultOptions} from './internal/object.d.ts';
+import type {IfNotAnyOrNever, Not} from './internal/type.d.ts';
+import type {IsStringLiteral} from './is-literal.d.ts';
+import type {Or} from './or.d.ts';
+
+/**
+@see {@link RemovePrefix}
+*/
+type RemovePrefixOptions = {
+	/**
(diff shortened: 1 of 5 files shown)


source/remove-prefix.d.ts
```diff
@@ -0,0 +1,128 @@
...
+	//=> `${string}/${number}`
+
+	type C = RemovePrefix<'on-change', `${number}-`, {strict: true}>;
+	//=> 'on-change'
+
+	type D = RemovePrefix<'on-change', `${number}-`, {strict: true}>;
+	//=> 'on-change'
+	```
+	*/
+	strict?: boolean;
```
som-sm, line 78: Let me know if anything is unclear here or if this feels too verbose.

source/remove-prefix.d.ts
```diff
@@ -0,0 +1,128 @@
...
+	Note: If it can be statically determined that the input string can never start with the specified non-literal prefix, then the input string is returned as-is, regardless of the value of this option.
+	For example, ``RemovePrefix<`${string}/${number}`, `${string}:`>`` returns `` `${string}/${number}` ``, since a string of type `` `${string}/${number}` `` can never start with a prefix of type `` `${string}:` ``.
+	```
+	import type {RemovePrefix} from 'type-fest';
+
+	type A = RemovePrefix<`${string}/${number}`, `${string}:`, {strict: true}>;
+	//=> `${string}/${number}`
+
+	type B = RemovePrefix<`${string}/${number}`, `${string}:`, {strict: true}>;
+	//=> `${string}/${number}`
```
sindresorhus: This is exactly the same as `A`

source/remove-prefix.d.ts
```diff
@@ -0,0 +1,128 @@
...
+import type {IfNotAnyOrNever, Not} from './internal/type.d.ts';
+import type {IsStringLiteral} from './is-literal.d.ts';
+import type {Or} from './or.d.ts';
+
+/**
+@see {@link RemovePrefix}
+*/
+type RemovePrefixOptions = {
+	/**
+	When enabled, instantiations with non-literal prefixes (e.g., `string`, `Uppercase<string>`, `` `on${string}` ``) simply return back `string`, since their precise structure cannot be statically determined.
```
sindresorhus:
```suggestion
	When enabled, instantiations with non-literal prefixes (e.g., `string`, `Uppercase<string>`, `` `on${string}` ``) simply return `string`, since their precise structure cannot be statically determined.
```

source/remove-prefix.d.ts
```diff
@@ -0,0 +1,128 @@
...
+	//=> string
+
+	type B = RemovePrefix<'on-change', `${string}-`, {strict: false}>;
+	//=> 'change'
+
+	type C = RemovePrefix<'on-change', string, {strict: true}>;
+	//=> string
+
+	type D = RemovePrefix<'on-change', string, {strict: false}>;
+	//=> 'n-change'
```
sindresorhus, line 32: I think this should return `string` too. I don't think removing one character is useful or expected behavior even with non-strict.

sindresorhus:
Is there a use-case for non-strict?

sindresorhus:
I think RemovePrefix is the correct name.

source/remove-prefix.d.ts
som-sm: Umm...I think we should just leave it as-is. Non-strict mode is primarily meant for type-level use, and in that context, I believe it makes sense to follow TS's default behavior:

```ts
type Test1 = 'foobar' extends `${string}${infer Rest}` ? Rest : never;
//   ^? type Test1 = "oobar"

type Test2 = 'FOOBAR' extends `${Uppercase<string>}${infer Rest}` ? Rest : never;
//   ^? type Test2 = "OOBAR"

type Test3 = 'FOOBAR' extends `${infer First}${infer Rest}` ? [First, Rest] : never;
//   ^? type Test3 = ["F", "OOBAR"]
```

<br>

Moreover, this exact example can be used to replace `RemoveFirstLetter<S>`.
https://github.com/sindresorhus/type-fest/blob/9926e5dba487f9aec27556dc16030e4b7a6b7a08/source/delimiter-case.d.ts#L24-L26
image

som-sm:
> Is there a use-case for non-strict?

Can be useful for type level manipulations, like the one mentioned [here](https://github.com/sindresorhus/type-fest/pull/1194#discussion_r2153143052).

source/remove-prefix.d.ts
som-sm: Added this usecase in [d38a7c2](https://github.com/sindresorhus/type-fest/pull/1194/commits/d38a7c213b1f629c908f8f22097a7621d420d5bc).

source/remove-prefix.d.ts
som-sm: Fixed in [5693114](https://github.com/sindresorhus/type-fest/pull/1194/commits/56931146407b3d4347da648a0860d4f44f56fe27).

source/remove-prefix.d.ts
som-sm: Maybe we can add it to the description that non-strict mode is primarily for type level manipulations, along with the `RemoveFirstLetter` example.

mesqueeb:
@sindresorhus i need this util, when will it be released?

sindresorhus:
https://github.com/sindresorhus/type-fest/issues/450

PR #1198, opened 2025-06-20
`ConditionalKeys`: Fix behaviour with arrays and unions

som-sm:
This PR:

- Fixes behaviour with arrays:
  ```ts
  type Current1 = ConditionalKeys<[string, number, string], string>;
  //   ^? type Current1 = 3 | "0" | "2" | (() => ArrayIterator<"0" | "2">) | { [x: number]: boolean | undefine…
  
  type Current2 = ConditionalKeys<string[], string>;
  //   ^? type Current2 = number | (() => ArrayIterator<number>) | { [x: number]: boolean | undefined; length?…
  
  type Current3 = ConditionalKeys<[string, number, ...string[]], string | number>;
  //   ^? type Current3 = number | "0" | "1" | (() => ArrayIterator<number | "0" | "1">) | { [x: number]: bool…
  
  type Updated1 = ConditionalKeys<[string, number, string], string>;
  //   ^? type Updated1 = "0" | "2"
  
  type Updated2 = ConditionalKeys<string[], string>;
  //   ^? type Updated2 = number
  
  type Updated3 = ConditionalKeys<[string, number, ...string[]], string | number>;
  //   ^? type Updated3 = number | "0" | "1"
  ```
  So, now for arrays we _only_ return back numeric keys.

- Fixes one edge case with unions:

  Currently, `ConditionalKeys` only returns _common matching_ keys when instantiated with unions, for example, `ConditionalKeys<{a: string} | {b: string}, string>` returns `never` and _not_ `'a' | 'b'`. Considering this behaviour is fine, `ConditionalKeys<{a: string} | {a: number}, string>` should return `never` and not `'a'`, because the common key `'a'` should match the condition across all union members.
  
  ```ts
  type Current = ConditionalKeys<{a: string} | {a: number}, string>;
  //   ^? type Current = "a"
  
  type Updated = ConditionalKeys<{a: string} | {a: number}, string>;
  //   ^? type Updated = never
  ```
  Maybe we can simply return _all_ matching keys across union members instead of just common matching keys. But IMO, it's fine to return back just the common keys.
  
<br>  

Here are some other minor behaviour changes:
- Fixes behaviour with primitives:
  ```ts
  type Current1 = ConditionalKeys<string, number>;
  //   ^? type Current1 = string | number | (() => StringIterator<string>) | (() => string) | ((pos: number) =…
  
  type Current2 = ConditionalKeys<number, () => number>;
  //   ^? type Current2 = ((radix?: number) => string) | (() => number) | ((fractionDigits?: number) => string…
  
  type Updated1 = ConditionalKeys<string, number>;
  //   ^? type Updated1 = "length"
  
  type Updated2 = ConditionalKeys<number, () => number>;
  //   ^? type Updated2 = "valueOf"
  ```
  Currently, we always return back all the keys of the primitive, so `ConditionalKeys<SomePrimitive>` returns `keyof SomePrimitive`. This happens because homomorphic mapped types, when instantiated with primitives, return them back as is.
  
- Keys with `never` value can now also be extracted using `any` as the condition.
   ```ts
   type Current = ConditionalKeys<{a: never; b: never}, any>;
   //   ^? type Current = never
   
   type Updated = ConditionalKeys<{a: never; b: never}, any>;
   //   ^? type Updated = 'a' | 'b'
   ```
   
- `string` indexors now return `string | number` instead of just `string`.
   ```ts
   type Current = ConditionalKeys<{[x: string]: boolean}, boolean>;
   //   ^? type Current = string
   
   type Updated = ConditionalKeys<{[x: string]: boolean}, boolean>;
   //   ^? type Updated = string | number
   ```
   
- `any` as input returns `PropertyKey` instead of `string`.
   ```ts
   type Current = ConditionalKeys<any, number>;
   //   ^? type Current = string
   
   type Updated = ConditionalKeys<any, number>;
   //   ^? type Updated = PropertyKey
   ```

### DIFF
--- a/source/conditional-keys.d.ts
+++ b/source/conditional-keys.d.ts
@@ -36,7 +36,7 @@ type NoMatchingKeys = ConditionalKeys<{a?: string}, string>;
 //=> never
 ```
 
-You can also extract array indicies whose value match the specified condition, as shown below:
+You can also extract array indices whose value match the specified condition, as shown below:
 ```
 import type {ConditionalKeys} from 'type-fest';
 
--- a/source/conditional-keys.d.ts
+++ b/source/conditional-keys.d.ts
@@ -1,5 +1,7 @@
-import type {If} from './if.d.ts';
-import type {IsNever} from './is-never.d.ts';
+import type {ExtendsStrict} from './extends-strict.d.ts';
+import type {IfNotAnyOrNever} from './internal/type.d.ts';
+import type {TupleToObject} from './tuple-to-object.d.ts';
+import type {UnknownArray} from './unknown-array.d.ts';
 
 /**
 Extract the keys from a type where the value type of the key extends the given `Condition`.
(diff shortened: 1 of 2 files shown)


sindresorhus:
> Maybe we can simply return all matching keys across union members instead of just common matching keys. But IMO, it's fine to return back just the common keys.

I think returning the common keys is the most correct.

som-sm:
> I think returning the common keys is the most correct.

Yup, agree!

Issue #1203, opened 2025-07-22
Split has a different behavior depending on the position of ${string} in tempalte literal types


jclaveau:
### Bug description

```ts
import {
  type Split,
} from 'type-fest'

type Parts1 = Split<`${string}aa`, ''> // => [string, "a", "a"]
type Parts2 = Split<`a${string}a`, ''> // => ["a", string, "a"]
type Parts3 = Split<`aa${string}`, ''> // => ["a", "a"]


type Parts11 = Split<`${string}.a.a`, '.'> // => [string, "a", "a"]
type Parts21 = Split<`a.${string}.a`, '.'> // => ["a", string, "a"]
type Parts31 = Split<`a.a.${string}`, '.'> // => ["a", "a", string]
```

### Repro

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbwFBzjAnmApnAymAG2BgBokBfOAMyghDgHINsBaKrAZxgaSWZwAKAQ1gcAjHAC8eQsQA8AAwAkCLlGAA7AObkhQhSUYMAfHAD0ZqaYDaazVsMAiIY6cuAun0yCRMDgCYpGSIYRSEVO21dAyNTCys4a2dXOEiHOGTPfjhhUQBmIPwQsPDVGHUomIYTc0tJG2S3R09ebNy-MQlpIvllMoqdADohYarBmvj6xLSmpqzvHN8AruDekYjy+3JRwwZxuLqGl0MZjI8vbEX8lZ7QhRH1-q2xicPExrOUtM8gA

PR #1204, opened 2025-07-22
`Split`: Fix behaviour with template strings ending with interpolated types

jclaveau:
Fixes https://github.com/sindresorhus/type-fest/issues/1203

Hoping I didn't miss any case

### DIFF
--- a/test-d/split.ts
+++ b/test-d/split.ts
@@ -40,7 +40,8 @@ expectType<['a', 'b', 'c'] | ['a,b,c'] | ['x', 'y', 'z'] | ['x:y:z']>({} as Spli
 // which, when split by `.`, would result in `['a', 'b', 'c', 'd']`, which wouldn't conform to the output type of `['a', 'b', string]`.
 expectType<['a', 'b', string]>({} as Split<`a.b.${string}`, '.', {strictLiteralChecks: false}>);
 
-// Ensure the last dynamic "string" is not droped when split by '' https://github.com/sindresorhus/type-fest/issues/1203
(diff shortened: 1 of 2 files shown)


test-d/split.ts
```diff
@@ -40,6 +40,9 @@ expectType<['a', 'b', 'c'] | ['a,b,c'] | ['x', 'y', 'z'] | ['x:y:z']>({} as Spli
 // which, when split by `.`, would result in `['a', 'b', 'c', 'd']`, which wouldn't conform to the output type of `['a', 'b', string]`.
 expectType<['a', 'b', string]>({} as Split<`a.b.${string}`, '.', {strictLiteralChecks: false}>);
 
+// Ensure the last dynamic "string" is not droped when split by '' https://github.com/sindresorhus/type-fest/issues/1203
+expectType<['a', 'b', string]>({} as Split<`ab${string}`, '', {strictLiteralChecks: false}>);
+
```
som-sm, line 46: Add case:

```ts
expectType<['a', 'b', string]>({} as Split<`a.b.${string}`, '.', {strictLiteralChecks: false}>);
```

jclaveau: I see no difference with https://github.com/sindresorhus/type-fest/blob/5ad3b1c256462b09ba5db1371a8d3ec66ba2e1b6/test-d/split.ts#L41

Could you elaborate plz?

som-sm: Oh okay, didn't realise it was already there.

PR #1206, opened 2025-07-28
`Jsonify`: handle unknown as `JsonValue`

### LINKED ISSUE #1195: `Jsonify<unknown>` should be either `unknown` or `JsonValue`
Assuming you're serializing a non-strictly typed JSON data, all of its descendents were serialized too, thus `JsonValue` seems like a reasonable assumption. But resolving to `unknown` would be enough too.

Currently:
```ts
type U = Jsonify<unknown> // never
type V = Jsonify<{foo: unknown}> // {}
type W = Jsonify<[unknown]> // [null]
```

I am OK to submit a PR if you're OK with it

GerkinDev:
Closes #1195

Had a bit of struggles with the test

```ts
declare const tupleStringJson: Jsonify<string[] & ['some value']>;
expectType<['some value']>(tupleStringJson);
```

because

```ts
type JsonifyList<T extends UnknownArray> = T extends readonly [infer F, ...infer R]
```

infered `{F: 'some value', R: unknown[]}`, thus the dirty hack that is probably not ideal. Yet, it passes all the tests and it somewhat makes sense.

This problem looks like a known issue in TS tuples handling & intersection with arrays, but depending on the POV, it may be more of a feature than a bug.

### DIFF
--- a/source/internal/keys.d.ts
+++ b/source/internal/keys.d.ts
@@ -1,5 +1,6 @@
 import type {IsAny} from '../is-any.d.ts';
 import type {IsLiteral} from '../is-literal.d.ts';
+import type {IsUnknown} from '../is-unknown.d.ts';
 import type {ToString} from './string.d.ts';
 
 // Returns `never` if the key or property is not jsonable without testing whether the property is required or optional otherwise return the key.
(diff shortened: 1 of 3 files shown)


source/internal/keys.d.ts
```diff
@@ -25,7 +26,7 @@ export type FilterDefinedKeys<T extends object> = Exclude<
 	{
 		[Key in keyof T]: IsAny<T[Key]> extends true
 			? Key
-			: undefined extends T[Key]
+			: IsUnknown<T[Key]> extends true ? Key : undefined extends T[Key]
```
GerkinDev, line 29: Given how `any` is handled I suppose it makes sense

sindresorhus:
> If requested, I can preserve unknown as mentioned in the issue

Let’s not preserve unknown. Jsonify’s job is to produce something assignable to JsonValue, so mapping unknown to JsonValue keeps the contract clear and makes it more
ergonomic without sacrificing soundness.

PR #1234, opened 2025-09-18
Add `StrictEventTarget`

### LINKED ISSUE #1230: Strict `EventTarget`

We would like to have a way to add generics to `EventTarget`.

```ts
const emitter = EventTarget<{
  a: void;
  b: number[];
}>();
```

means `a` event with no payload, and `b` event with `number[]` payload.

Maybe should actually provide `Event` object instead as values.

### Type source

https://github.com/Richienb/strict-event-target/tree/main

I think it should be some `interface extends EventTarget`, and we override `add`/`remove` - `EventListener`

TypeScript issue: https://github.com/microsoft/TypeScript/issues/43477

Richienb:
Fixes #1230


Richienb:
> There are no errors here:

Yeah so I wanna fix it here, and deprecate that

sindresorhus:
Closing as this is not moving forward. I don't want to have PRs open for too long.

PR #1237, opened 2025-09-20
Add lint rule to catch cases of missing exports in index.d.ts

sindresorhus:
To prevent https://github.com/sindresorhus/type-fest/commit/4f9c248b5729e1d1d2735dae5ec7922d1c70ba3a in the future.

For types that we export from one file and is internally in othre files, we just prefix with underscore to make it immediately clear it's an internal type.


sindresorhus:
I have no idea why this got closed and I definitely did not delete the branch like GitHub is saying...

PR #1238, opened 2025-09-20
Add lint rule to catch cases of missing exports in index.d.ts

sindresorhus:
To prevent https://github.com/sindresorhus/type-fest/commit/4f9c248b5729e1d1d2735dae5ec7922d1c70ba3a in the future.

For types that we export from one file and is internally in othre files, we just prefix with underscore to make it immediately clear it's an internal type.

### DIFF
--- a/lint-rules/require-exported-types.js
+++ b/lint-rules/require-exported-types.js
@@ -1,0 +1,160 @@
+import path from 'node:path';
+import fs from 'node:fs';
+
+/**
+Cache for index exports to avoid repeated lookups
+*/
+const indexCache = new Map();
+
+export const requireExportedTypesRule = /** @type {const} */ ({
+	meta: {
+		type: 'suggestion',
+		docs: {
+			description: 'Enforce that exported types are also exported from index.d.ts',
+			category: 'Best Practices',
+			recommended: true,
+		},
+		messages: {
+			missingExport: 'Type `{{typeName}}` is exported from this file but not from index.d.ts. Add it to index.d.ts or use `// eslint-disable-next-line type-fest/require-exported-types` to ignore.',
+			noTypeInfo: 'Rule requires TypeScript type information. Configure `parserOptions.project` in your ESLint config.',
+		},
+		schema: [
+			{
+				type: 'object',
+				properties: {
+					indexFile: {
+						type: 'string',
+						description: 'Path to the index file (default: "index.d.ts")',
+					},
+				},
+				additionalProperties: false,
+			},
+		],
+	},
+	defaultOptions: [],
+	create(context) {
+		// Only run on TypeScript declaration files in source directory
+		const filename = context.filename ?? context.getFilename?.() ?? '';
+		if (!filename.includes('/source/') || !filename.endsWith('.d.ts')) {
+			return {};
+		}
+
+		// Skip internal files
+		if (filename.includes('/internal/')) {
+			return {};
+		}
+
+		const options = context.options?.[0] ?? {};
+		const indexFileName = options.indexFile ?? 'index.d.ts';
+
+		// Get TypeScript type-aware services
+		const parserServices = context.sourceCode?.parserServices;
+
+		// Type information is required for this rule
+		if (!parserServices?.program || !parserServices?.esTreeNodeToTSNodeMap) {
+			// Report once per file that type information is required
+			return {
+				'Program'(node) {
+					context.report({
+						node,
+						messageId: 'noTypeInfo',
+					});
+				},
+			};
+		}
+
+		const {program} = parserServices;
+		const checker = program.getTypeChecker();
+
+		// Find project root
+		let projectRoot = path.dirname(filename);
+		while (projectRoot !== path.dirname(projectRoot)) {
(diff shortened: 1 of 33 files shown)


som-sm:
@sindresorhus The error doesn't go away even after exporting the type from `index.d.ts`. Probably happens because once the exports from `index.d.ts` are cached, the cache is never refreshed — or maybe I am missing something.

https://github.com/user-attachments/assets/9c1f8c5b-917f-4a67-93e0-0b49620482bc

sindresorhus:
Should work better now.

som-sm:
> Should work better now.

I see that you removed the caching completely. I was wondering if we could have done something like this instead:
```js
Program() {
	if (path.basename(filename) === 'index.d.ts') {
		// Burst cache
	}
},
```

lint-rules/require-exported-types.js
```diff
@@ -0,0 +1,142 @@
...
+				},
+				additionalProperties: false,
+			},
+		],
+	},
+	defaultOptions: [],
+	create(context) {
+		// Only run on TypeScript declaration files in source directory
+		// Convert to forward slashes for consistent path checking across platforms
+		const filename = (context.filename ?? context.getFilename?.() ?? '').replaceAll('\\', '/');
```
som-sm:
```suggestion
		const filename = context.filename.replaceAll('\\', '/');
```

lint-rules/require-exported-types.js
```diff
@@ -0,0 +1,142 @@
...
+		// Skip internal files
+		if (filename.includes('/internal/')) {
+			return {};
+		}
+
+		const options = context.options?.[0] ?? {};
+		const indexFileName = options.indexFile ?? 'index.d.ts';
+
+		// Get TypeScript type-aware services
+		const parserServices = context.sourceCode?.parserServices;
```
som-sm:
```suggestion
		const parserServices = context.sourceCode.parserServices;
```

lint-rules/require-exported-types.js
```diff
@@ -0,0 +1,142 @@
...
+
+		// Type information is required for this rule
+		if (!parserServices?.program || !parserServices?.esTreeNodeToTSNodeMap) {
```
som-sm, line 51:
```suggestion
		if (!parserServices?.program || !parserServices.esTreeNodeToTSNodeMap) {
```

lint-rules/require-exported-types.js
```diff
@@ -0,0 +1,142 @@
...
+
+			// Skip types starting with underscore (internal/private convention)
+			if (typeName.startsWith('_')) {
+				return;
+			}
+
+			// Skip if inside declare namespace
+			const ancestors = context.sourceCode.getAncestors(node);
+			const isInsideDeclareNamespace = ancestors.some(ancestor =>
+				ancestor.type === 'TSModuleDeclaration' && ancestor.declare === true,
```
som-sm:
```suggestion
				ancestor.type === 'TSModuleDeclaration' && ancestor.declare,
```

som-sm:
Also, just realised that there are some types that do not export their options, but their options are still being exported from `index.d.ts`. It seems to work, but is kinda weird.

https://github.com/sindresorhus/type-fest/blob/5af60a1330c304eae03765299281ce08710aa323/source/replace.d.ts#L3
https://github.com/sindresorhus/type-fest/blob/5af60a1330c304eae03765299281ce08710aa323/index.d.ts#L182

sindresorhus:
> I see that you removed the caching completely.

I plan to look into caching later. It's not important enough to do right now.

sindresorhus:
> Also, just realised that there are some types that do not export their options, but their options are still being exported from index.d.ts. It seems to work, but is kinda weird.

This is really something we should report to TypeScript. It should not allow using unexported types when in ESM mode.

som-sm:
> This is really something we should report to TypeScript. It should not allow using unexported types when in ESM mode.

Yeah, true, we have bumped into this a couple of times now.

som-sm (approved):

som-sm:
I was trying to extend this lint rule to also check if all the exports from `index.d.ts` are actually being exported from their respective source files as well. While doing so, I realised that the `getExportsOfModule` API returns non-exports as well, for example:

```ts
// remove-prefix.d.ts
type RemovePrefix = ...;
export type RemovePrefix = ...;
```

If we call `getExportsOfModule` on `remove-prefix.d.ts`, it returns back both `RemovePrefixOptions` and `RemovePrefix`, even though `RemovePrefixOptions` is not exported. So, looks like it's not treating `remove-prefix.d.ts` as a module.

<hr>

So, I simply added a `export {}` at the end of `remove-prefix.d.ts`, and it seemed to fix the issue. Now, if we try to import something that isn't being exported, we get an error.
image

@sindresorhus Is this information of any use?

som-sm:
> This is really something we should report to TypeScript. It should not allow using unexported types when in ESM mode.

Looks like this is the intended behaviour, refer https://github.com/microsoft/TypeScript/issues/57764.

sindresorhus:
https://github.com/sindresorhus/type-fest/pull/1244

Issue #880, opened 2024-05-15
`PickDeep` strips `null`


relsunkaev:
If you have following types
```ts
type Owner = {
  name: string;
  age: number;
};
type Pet = {
  name: string;
  age: number;
  owner: Owner | null;
};
```
the resulting type of `PickDeep<Pet, "name" | "owner.name">` is
```ts
type Result = {
    name: string;
    // `owner` is no longer nullable
    owner: {
        name: string;
    };
}
```

Emiyaaaaa:
Thanks report, but I don't think this is a bug

Let's look some cases:

## case 1
```ts
type Pet = {
 owner: { name: 'name1' }
      | { name: 'name2' ; foo: '' };
};
type Result = PickDeep<Pet, 'owner.name'>;
```
`PickDeep` should pick every union type, so result should indeed be
```ts
type Result = {
 owner: { name: 'name1' } | { name: 'name2' }
}
```

## case 2
```ts
type Pet = {
 owner: { name: 'name1' }
      | { foo: '' };
};
type Result = PickDeep<Pet, 'owner.name'>;
```
`PickDeep` can not find key "name" in `{ foo: '' }`, so result is
```ts
type Result = {
 owner: { name: 'name1' }
}
```

## case 3
```ts
type Pet = {
 owner: { name: 'name1' }
      | null;
};
type Result = PickDeep<Pet, 'owner.name'>;
```
so according to case 2, result should be
```ts
type Result = {
 owner: { name: 'name1' }
}
```

sindresorhus:
What would "case 2" become if so?

Emiyaaaaa:
> What would "case 2" become if so?

that will be 
```ts
type Result = {
 owner: { name: 'name1' } | { foo: "" }
}
```

It's look like confused, I am beginning to waver on whether need do this for "type safety"

Issue #1223, opened 2025-09-07
PickDeep merge picked arrays instead of picked array items


Wendystraite:
### Bug description

This can either be considered a bug or an improvement to `PickDeep` :

Instead of merging `(A & B)[]` it is merging `A[] & B[]` : this behavior causes typings errors in resulting code.

[...]

PR #1241, opened 2025-09-24
`pick-deep`: Fix the case when the key or element is an union type (#1224)

### LINKED ISSUE #1224: PickDeep does not pick types unioned to object or arrays
### Bug description

`PickDeep` does not pick types unioned to object or arrays.

The bug in code ([playground](https://www.typescriptlang.org/play/#code/JYWwDg9gTgLgBDAnmApnA3nACsAxgawBEUUw4BfOAMyghDgHIlUBaKlAZxgYG4BYAFCDmaQgEMYYuAF4MguArgQARgCsAXHC5RgAOwDmcAD4Y4Yzdr36ecZZt0BXEMpRQbuTcogQANijG6FMZwjj4+wQ66ACYoVHooUfwCimZQUJqY5lowOgY2diFOLm5wHrbefgEK5ADaALoR0bHxiYLkSYIA9ABUgmK4MA5i4erCyGgAYt4A8moycsmKKhoLKSkFjs6uSSmUAGSrawplXr7+ujuK7W2CgigAHqgDCXCjAiJwUxCzqvPo8ks1BYclZgv9FkcNkVtgCjicKudLtVgqFwiZIjE4roEklyIJup0xqhPjM5rIcARiKQADziSQAGjgAANlgA6AAk6AYygYwQYuAY5CZAD4OgIen0BkMRkTJt4AIJpP6w1LpQ6Q+zQtwq2oNA7go7HTwIgJIij1XG3AQPJ4wF5vD5fRVQZUQ1UZFXrTVbbVulLws6mnX1RqYlqWgQE2UkiDO+YUogkMC0iRiRlMsRpDnoTbFcjZ7m8kz8wUipJAA)) : 

```ts
// actual   : type FooObj = { obj: { b: number } & { c: boolean } }
// expected : type FooObj = { obj: string | { b: number; c: boolean } | null | undefined }
type FooObj = PickDeep<
  { obj: string | { a: string; b: number; c: boolean } | null | undefined },
  `obj.${'b' | 'c'}`
>;
```

.. and you can't pick these values by `PickDeep<..., 'obj' | 'obj.b'>` or your resulting type contain the full object defeating the purpose of the type.

### Repro

https://www.typescriptlang.org/play/#code/JYWwDg9gTgLgBDAnmApnA3nACsAxgawBEUUw4BfOAMyghDgHIlUBaKlAZxgYG4BYAFCDmaQgEMYYuAF4MguArgQARgCsAXHC5RgAOwDmcAD4Y4Yzdr36ecZZt0BXEMpRQbuTcogQANijG6FMZwjj4+wQ66ACYoVHooUfwCimZQUJqY5lowOgY2diFOLm5wHrbefgEK5ADaALoR0bHxiYLkSYIA9ABUgmK4MA5i4erCyGgAYt4A8moycsmKKhoLKSkFjs6uSSmUAGSrawplXr7+ujuK7W2CgigAHqgDCXCjAiJwUxCzqvPo8ks1BYclZgv9FkcNkVtgCjicKudLtVgqFwiZIjE4roEklyIJup0xqhPjM5rIcARiKQADziSQAGjgAANlgA6AAk6AYygYwQYuAY5CZAD4OgIen0BkMRkTJt4AIJpP6w1LpQ6Q+zQtwq2oNA7go7HTwIgJIij1XG3AQPJ4wF5vD5fRVQZUQ1UZFXrTVbbVulLws6mnX1RqYlqWgQE2UkiDO+YUogkMC0iRiRlMsRpDnoTbFcjZ7m8kz8wUipJAA

taiyakihitotsu:
This PR fixes #1224.  
(I’m opening this because I previously sent a message about the issue, but it seems that no one has started working on it.)

## Current Status
https://github.com/sindresorhus/type-fest/pull/1241#issuecomment-3389196769 -> https://github.com/sindresorhus/type-fest/pull/1241#issuecomment-3756828299 -> https://github.com/sindresorhus/type-fest/pull/1241#issuecomment-3764800851

## Approach

The issue can be broken down into two separate bugs:

 - `PickDeep` fails to pick values when the object’s properties are of a union type. (2025/09/29 ... ok)
 - `PickDeep` doesn't properly merge results when the path is a union type. ~~(2025/09/29 ... it has a [bug](https://github.com/sindresorhus/type-fest/pull/1241#issuecomment-3341992076))~~ (2025/10/01 ... [ok](https://github.com/sindresorhus/type-fest/pull/1241#issuecomment-3355791883))

The first was fixed by applying distribution over the union type in the object.

~~For the second issue, the fix involves evaluating each path in the union individually, obtaining the result as a union of objects, and then merging them using `MergeNarrow`.~~ (2025/10/01 ... [See](https://github.com/sindresorhus/type-fest/pull/1241#issuecomment-3355791883))

## related issue

#1224, #1223, #880 ([See](https://github.com/sindresorhus/type-fest/pull/1241#issuecomment-3756828299))

~~I'm not sure whether the expected value shown in #1223 is truly correct according to the `PickDeep` specification, so I’ve decided to leave that issue for now.~~ (2026/01/16 ... [See](https://github.com/sindresorhus/type-fest/pull/1241#issuecomment-3756828299))

~~(In the current commit, the result for that case would be `{arr: []}`. I think this is caused by `MergeNarrow`.)~~ (2025/10/01 ... [See](https://github.com/sindresorhus/type-fest/pull/1241#issuecomment-3355791883))

~~If I should include a fix for #1223 in this PR as well, please let me know.~~ (2026/01/16 ... [See](https://github.com/sindresorhus/type-fest/pull/1241#issuecomment-3756828299))

## note
I rewrote the tests to be more strict, using the following pattern:

```typescript
expectType<true>({} as IsEqual<Actual, Expected>)
```

This is because some previous tests using `expectType<T>(value)` were passing unintentionally.


som-sm:
@taiyakihitotsu Thanks for the PR! I’m a bit occupied with festivities, I’ll try to review it sometime next week.

taiyakihitotsu:
I've added a fix commit for #1223.

@sindresorhus
Thanks!

@som-sm
Understood, I’ll wait. Enjoy!

taiyakihitotsu:
I found a bug in this case and will fix it as well.

```typescript
type unionKeyUnionObject_Actual = PickDeep<{obj: {a: string; b: number; c: boolean} | {d: 0}}, `obj.${'b' | 'c' | 'd'}`>; // {obj: {b: number; d: 0; c: boolean}}
```

taiyakihitotsu:
And `LastOfUnion` is also defined in

https://github.com/sindresorhus/type-fest/blob/ee29ef7b9d6280126562baae95795d4ed97ff3c8/source/union-to-tuple.d.ts#L13

I'll move it to internal if ok to combine them into one PR.

taiyakihitotsu:
I’ve got an idea for the fix, but not sure when I can get back to it, so I leave this as a draft for now.
Maybe I’ll pick it up again in a week or so.

taiyakihitotsu:
@som-sm

I fixed the issues (#1224, #1223) and [a bug in the comment I posted earlier](https://github.com/sindresorhus/type-fest/pull/1241#issuecomment-3341992076).
Sorry about the size of the diff; I wrote it nearly the whole structure.

### diff
I defined a converter that transforms template literals like `a.${'b' | 'c'} | 'x'` into a tree structure like `{a: {b: ''; c: ''}; x: ''}`.
This makes it easier to get a union type using `keyof` at the same path level, which helps preserve the shape when merging.

(When the implementation of Path relies on Union type distribution, the original structure can be lost. For example, in an extreme case like `{a: 0, b: 0} | {a: 1, b: 1}`, if they are not merged at the same point in time, it may end up being split into `{a: 0} | {a: 1} | {b: 0} | {b: 1}`.)

I've also defined some internal helper functions for certain getter functions, which ignore the distinction between `string` and `number` types for numeric keys.
(Since the path does not differentiate between the two, there's no meaningful reason for the source object to do so either.)

### ~~`IsNever` bug~~
**Outdated. see [this](https://github.com/sindresorhus/type-fest/pull/1241#issuecomment-3356824883)**
(Check the edit history if you want to see)

### move to internal
It seems reasonable to mark some of the functions as internal.
(Example: `IsKeyOf`, `LastOfUnion`, `_IsEqual` (rename to -> `InternalIsEqual` if `IsNever` uses it for a fix) 
Would it be acceptable to include that change in this PR as well?

som-sm:
> The current `IsNever` implementation has a bug:

Is there a simpler reproduction of where `IsNever` fails?

taiyakihitotsu:
**This comment is outdated, see [this new comment](https://github.com/sindresorhus/type-fest/pull/1241#issuecomment-3356824883)**

---

@som-sm

It may only fail when using recursion with `LastOfUnion<T>` and `IsNever`. This is the same issue we’re currently facing with `IsEqual`.

## **[Current `IsEqual`](https://github.com/sindresorhus/type-fest/commit/785549f36465e3f3d99a08832784b603261f74f2)**
### Fixed
 - This fixed the [three cases we discussed](https://github.com/sindresorhus/type-fest/pull/1231#issuecomment-3313480326)
 - `never` and `unknown` pattern fixed. https://github.com/sindresorhus/type-fest/pull/1251
### Failed
 - `never` returned `LastOfUnion`.

## Internal `_IsEqual`
```typescript
type _IsEqual<A, B> =
		? true
		: false;
 ```   
This has bugs above the `fixed` part.
But it can detect well of `never` of `LastOfUnion` if defining like:

```typescript
export type CurrentIsNever<T> = _IsEqual<T, never>
```

### whole code
I'm researching it in `source/is-equal.d.ts`.

```typescript
export type IsEqual<A, B> =
	[A, B] extends [infer AA, infer BB]
		? [AA] extends [never]
			? [BB] extends [never]
				? true
				: false
			: [BB] extends [never]
				? false
				: _IsEqual<AA, BB>
		: false;

import type {UnionToIntersection} from './union-to-intersection.d.ts'
type LastOfUnion<T> = UnionToIntersection<T extends any ? () => T : never> extends () => (infer R)	? R : never;

export type PreviousIsNever<T> = [T] extends [never] ? true : false;
export type IsNever<T> = IsEqual<T, never>
export type CurrentIsNever<T> = _IsEqual<T, never>

type PreviousUnionRecursion<T, R = {}> = LastOfUnion<T> extends infer L ? PreviousIsNever<L> extends false ? PreviousUnionRecursion<Exclude<T, L>, R | L> : R : never;
type UnionRecursion<T, R = {}> = LastOfUnion<T> extends infer L ? IsNever<L> extends false ? UnionRecursion<Exclude<T, L>, R | L> : R : never;
type CurrentUnionRecursion<T, R = {}> = LastOfUnion<T> extends infer L ? CurrentIsNever<L> extends false ? CurrentUnionRecursion<Exclude<T, L>, R | L> : R : never;

// @ts-expect-error
type PrevTest = PreviousUnionRecursion<string | number>;
// Type instantiation is excessively deep and possibly infinite.

// @ts-expect-error
type FixedTest = UnionRecursion<string | number>; 
// Type instantiation is excessively deep and possibly infinite.

type CurrentFixedTest = CurrentUnionRecursion<string | number>; // string | number | {}

// This version fails the `equalWrappedTupleIntersectionToBeNeverAndNeverExpanded` test in `test-d/is-equal.ts`.
type _IsEqual<A, B> =
		? true
		: false;
```

I have two plans.

One is to redefine `IsNever` as shown above in `CurrentIsNever`, and at the same time make `_IsEqual` internal and rename like `InternalIsEqual`. This solution has already been discovered, is simple to implement, and maintains consistency in terms of `never` checking. However, it sacrifices some theoretical rigor. 
(Users may select `IsEqual<L, never>` instead of `IsNever<L>` when using `LastOfUnion<T> extends infer L`, then it fails.)

The other plan is to fix `IsEqual` and then redefine `IsNever` using the revised `IsEqual`. So far, this approach has not been successful.

Either way, I feel it's better to make the PR separate.

taiyakihitotsu:
~~Ah, maybe this isn't IsEqual's job — it might be LastOfUnion's responsibility.~~

```typescript
type LastOfUnion<T> =
  UnionToIntersection<
    T extends infer A
      ? () => A
      : never> extends () => (infer R)
    ? R extends T ? R : never
    : never;
```

~~This fixes all above cases.
I will research a lot.~~

This works well only in the test cases above, but it cannot replace the existing code.

taiyakihitotsu:
@som-sm

```typescript
type LastOfUnion<T> =
  UnionToIntersection<
    T extends infer A
      ? () => A
      : never> extends () => (infer R)
    ? R
    : never;

type a = LastOfUnion<never>; // unknown
```

OK I see that this returns `unknown`, and `_IsEqual<never, unknown>` returns `true` unintentionally.
So `LastOfUnion<U> extends infer L` and `L extends Something` works as expected, `IsNever` using `_IsEqual` works fine.
But the current `IsEqual` returns `false` for `L` even if it's expected to be `never` (`L` is `unknown` in this case though).

The current (at [590efae8223ed1ac61855840dff114360b7c6807](https://github.com/sindresorhus/type-fest/commit/590efae8223ed1ac61855840dff114360b7c6807)) definition of `IsEqual` and `IsNever` aren't incorrect.
It's not a bug in `IsNever`.

```typescript
type test = LastOfUnion<Exclude<string, string>> extends infer U ? PreviousIsNever<U> : 'false';
// ^ false

export type PreviousIsNever<T> = [T] extends [never] ? true : false;
```

Apologies for my misunderstanding.

In any case, `LastOfUnion` is already used in [source/union-to-tuple.d.ts](https://github.com/sindresorhus/type-fest/blob/590efae8223ed1ac61855840dff114360b7c6807/source/union-to-tuple.d.ts#L13), and this commit defines it redundantly.
So I want to fix and move it to internal.

Should I include it in this PR? Or not?

sindresorhus:
> Should I include it in this PR? Or not?

In this PR, yes.

taiyakihitotsu:
@sindresorhus @som-sm

Thanks for your response!
I've added a commit moving `LastOfUnion` to `source/internal/type.d.ts`.

I wrote a comment and provided two examples of how to use it for recursion.

And I made a small change to `UnionToTuple`, hiding the second argument.

## The latest status of this PR

 - This PR fixes both: #1223, #1224.
 - `LastOfUnion` now returns `never` if `never` passed (it previously returned `unknown`).
 - `LastOfUnion` moved to `source/internal/type.d.ts`.
 - No bugs have been found in `IsNever` and `IsEqual`.
 - Refactored `UnionToTuple`.
 - Added a test case for `UnionToTuple<never>` (`[]`).

taiyakihitotsu:
@som-sm

Sorry for leaving this hanging.

 - Merged the latest main branch.
 - Switched to the internal `StringToNumber`.

The functionality hasn’t changed since https://github.com/sindresorhus/type-fest/pull/1241#issuecomment-3389196769.

---

I just wanted to share the update.
Feel free to take a look whenever you have a moment.

taiyakihitotsu:
I think this issue (https://github.com/sindresorhus/type-fest/issues/880) is related as well.

https://github.com/sindresorhus/type-fest/pull/1241/files#diff-3104c46023d65f634fe4bbb58d8a42358c369d3967ca7f29c62eea3c94388910R168

```typescript
type unionElement1_Actual = PickDeep<{obj: string | {a: string; b: number; c: {d: 'result'}} | null | undefined}, 'obj.b'>;
type unionElement1_Expected = {obj: string | null | undefined | {b: number}};
expectType<true>({} as IsEqual<unionElement1_Actual, unionElement1_Expected>);
```

test-d/pick-deep.ts
```diff
@@ -59,70 +60,150 @@ interface DeepInterface extends DeepType {
...
+type unionKeyObjectArrayArray_Expected = {arr: Array<Array<{b: number; c: boolean}>>};
+expectType<true>({} as IsEqual<unionKeyObjectArrayArray_Actual, unionKeyObjectArrayArray_Expected>);
+
+type unionSameKeysObject_Actual = PickDeep<{a: string | {b: 1 | true; c: 2; d: {g: {f: 9; h: 10}}} | {b: '1'; c: '2'}; x: 10 | 11; y: [[0, 1], 2, 3]}, `a.${'b' | 'c'}` | 'x'>;
+type unionSameKeysObject_Expected = {x: 10 | 11; a: string | {b: 1 | true; c: 2} | {b: '1'; c: '2'}};
+expectType<true>({} as IsEqual<unionSameKeysObject_Actual, unionSameKeysObject_Expected>);
+
+type unionTupleTuple_Actual = PickDeep<[0, string | [1, [2]]], '1.1'>;
+type unionTupleTuple_Expected = [unknown, string | [unknown, [2]]];
+expectType<true>({} as IsEqual<unionTupleTuple_Actual, unionTupleTuple_Expected>);
```
sindresorhus, line 214: Some more tests:


```suggestion
expectType<true>({} as IsEqual<unionTupleTuple_Actual, unionTupleTuple_Expected>);

// Test: readonly array should preserve readonly modifier when picking into elements
type ReadonlyArrayContainer = {
	items: ReadonlyArray<{a: 1; b: 2}>;
};
type readonlyArrayPick_Actual = PickDeep<ReadonlyArrayContainer, `items.${number}.a`>;
type readonlyArrayPick_Expected = {items: ReadonlyArray<{a: 1}>};
expectType<true>({} as IsEqual<readonlyArrayPick_Actual, readonlyArrayPick_Expected>);

// Test: mutable array should stay mutable
type MutableArrayContainer = {
	items: Array<{a: 1; b: 2}>;
};
type mutableArrayPick_Actual = PickDeep<MutableArrayContainer, `items.${number}.a`>;
type mutableArrayPick_Expected = {items: Array<{a: 1}>};
expectType<true>({} as IsEqual<mutableArrayPick_Actual, mutableArrayPick_Expected>);
```

source/pick-deep.d.ts
```diff
@@ -1,10 +1,159 @@
...
+	Key extends keyof T
+		? T[Key]
+		: Key extends string | number
+			? StringToNumber<`${Key}`> extends infer Number_ extends number
+				? IsNever<Number_> extends false
+					? `${Key}` extends `${keyof T extends (string | number) ? keyof T : never}`
+						? T[(`${Key}` extends infer K extends keyof T ? K : never)] | T[(Number_ extends infer K extends keyof T ? K : never)]
+						/* `${number}` case */
+						: `${Key}` extends `${infer N extends number}`
+							? [IsEqual<N, number>, T] extends [true, Array<infer ArrayType>]
```
sindresorhus:
```suggestion
							? [IsEqual<N, number>, T] extends [true, ReadonlyArray <infer ArrayType>]
```

source/pick-deep.d.ts
```diff
@@ -1,10 +1,159 @@
...
+
+/* `BuildObject` supporting `${number}`. */
+type Build<K extends PropertyKey, L, M extends object | UnknownArray> =
+	`${K extends string ? K : never}` extends `${infer N extends number}`
+		? IsEqual<N, number> extends true
+			? L[]
+			: M extends UnknownArray
+				? [...TupleOf<N>, L]
+				: BuildObject<K, L, M>
+		: BuildObject<K, L, M>;
```
sindresorhus, line 167: Needs to preserve readonly modifier. Maybe:


```suggestion
type Build<K extends PropertyKey, L, M extends object | UnknownArray> =
	`${K extends string ? K : never}` extends `${infer N extends number}`
		? IsEqual<N, number> extends true
			/* Preserve readonly for arrays */
			? M extends unknown[]
				? L[]
				: M extends readonly unknown[]
					? readonly L[]
					: never
			: M extends unknown[]
				? [...TupleOf<N>, L]
				: M extends readonly unknown[]
					? readonly [...TupleOf<N>, L]
					: BuildObject<K, L, M>
		: BuildObject<K, L, M>;
```

source/pick-deep.d.ts
```diff
@@ -1,10 +1,159 @@
...
+
+/* Just rename */
+type As<Source, T> = Extract<Source, T>;
```
sindresorhus: Not useful. Just use `Extract` directly.

source/pick-deep.d.ts
```diff
@@ -77,76 +226,109 @@ type Street = PickDeep<Configuration, 'userConfig.address.1.street2'>;
...
-type PickDeepObject<RecordType extends object, P extends string | number> =
-	P extends `${infer RecordKeyInPath}.${infer SubPath}`
-		? ObjectValue<RecordType, RecordKeyInPath> extends infer ObjectV
-			? IsNever<ObjectV> extends false
-				? BuildObject<RecordKeyInPath, InternalPickDeep<NonNullable<ObjectV>, SubPath>, RecordType>
+type RecursionPickDeep<NextParent, NextPathTree extends PathTreeType> =
+	NextParent extends infer NextParentArray extends UnknownArray
+		/* NextParent: array */
+		? IsEqual<IsTuple<NextParentArray>, false> extends true
+			? NextParentArray extends Array<infer _>
```
sindresorhus:
```suggestion
			? NextParentArray extends readonly unknown[]
```

source/pick-deep.d.ts
```diff
@@ -77,76 +226,109 @@ type Street = PickDeep<Configuration, 'userConfig.address.1.street2'>;
...
+					? IsEqual<`${number}`, `${CoerceKeyof<NextPathTree>}`> extends true
+						/* `leadingSpreadArray2_Actual` in `test-d/pick-deep.ts` */
+						? NextParent
+						/* `tailingSpreadArray1_Actual` in `test-d/pick-deep.ts` */
+						: [...TupleOf<StringToNumber<`${CoerceKeyof<NextPathTree>}`>>, PickOrSelf<NextParent, CoerceKeyof<NextPathTree>>]
+					/* Not end */
+					: InternalPickDeep<PickOrSelf<NextParent, CoerceKeyof<NextPathTree>>, As<ForceGet<NextPathTree, CoerceKeyof<NextPathTree>>, PathTreeType>> extends infer Result
+						? IsEqual<`${number}`, `${CoerceKeyof<NextPathTree>}`> extends true
+							/* `leadingSpreadArray1_Actual` in `test-d/pick-deep.ts` */
+							? Result[]
```
sindresorhus:
```suggestion
							? NextParentArray extends unknown[]
								? Result[]
								: readonly Result[]
```

source/pick-deep.d.ts
```diff
@@ -77,76 +226,109 @@ type Street = PickDeep<Configuration, 'userConfig.address.1.street2'>;
...
+			? L extends number | string
+				? IsEqual<true, IsKeyOf<Parent, L>> extends true
+					/* Detect an end of path. */
+					? IsEqual<LeafMarker, PathTree[L]> extends true
+						? PickOrSelf<Parent, L> extends infer PickResult
+							? PickResult extends UnknownArray
+								? _PickDeep<Parent, PathTree, Exclude<K, L>, U | Simplify<[...TupleOf<StringToNumber<`${L}`>>, PickResult]>>
+								: _PickDeep<Parent, PathTree, Exclude<K, L>, U | Simplify<PickResult>>
+							: never
+						: _PickDeep<Parent, PathTree, Exclude<K, L>, U | Simplify<Build<L, RecursionPickDeep<ForceGet<Parent, L>, As<ForceGet<PathTree, L>, PathTreeType>>, As<Parent, object>>>>
```
sindresorhus:
```suggestion
						/* Not a leaf - recurse deeper */
						: ForceGet<PathTree, L> extends infer SubTree
							? _PickDeep<Parent, PathTree, Exclude<K, L>, U | Simplify<Build<L, RecursionPickDeep<ForceGet<Parent, L>, Extract<SubTree, PathTreeType>>, Extract<Parent, object>>>>
							: never
```

sindresorhus (changes requested):

taiyakihitotsu:
@sindresorhus

Thanks for your review!
I apologize for missing some cases regarding `readonly` properties and leaving trivial mistakes in the code.
I’ve addressed all of your suggestions.

---

### my note

This test case which you suggested had not been passed at first (I pasted it).

```typescript
// Test: readonly array should preserve readonly modifier when picking into elements
type ReadonlyArrayContainer = {
	items: ReadonlyArray<{a: 1; b: 2}>;
};
type readonlyArrayPick_Actual = PickDeep<ReadonlyArrayContainer, `items.${number}.a`>;
type readonlyArrayPick_Expected = {items: ReadonlyArray<{a: 1}>};
expectType<true>({} as IsEqual<readonlyArrayPick_Actual, readonlyArrayPick_Expected>);
```

This review's code (https://github.com/sindresorhus/type-fest/pull/1241#discussion_r2695598791) fixed this tests (https://github.com/sindresorhus/type-fest/pull/1241#discussion_r2695593864):

---

This PR closes #1224 (and #1223 tested in https://github.com/sindresorhus/type-fest/pull/1241/files#diff-3104c46023d65f634fe4bbb58d8a42358c369d3967ca7f29c62eea3c94388910R195).
It may also potentially close #880, but there is an ongoing discussion that has been suspended (https://github.com/sindresorhus/type-fest/issues/880#issuecomment-2121625777).

To clarify this point, I’ve added a comment and a test case (in this commit https://github.com/sindresorhus/type-fest/pull/1241/commits/fdcb3df19d46e9eac8c2be74e524f01cf80e65b0) to ensure that `null` and `undefined` are preserved for this scenario, as previously discussed here: https://github.com/sindresorhus/type-fest/issues/880#issuecomment-2115864014.
At least for this use case, it seems to work without issues.

taiyakihitotsu:
I've fixed `IsKeyOf` because `IsKeyOf<Coll, never>` returned `never` incorrectly.
It causes a problem that `IsKeyOf<Coll, never>` could satisfy `extends true` unintentionally.

The other status has not been changed from https://github.com/sindresorhus/type-fest/pull/1241#issuecomment-3756828299.

taiyakihitotsu:
Drafting until https://github.com/sindresorhus/type-fest/pull/1349 is merged

taiyakihitotsu:
Close this because there are too many conflicts.
I'll open a new PR with this patch.

PR #1244, opened 2025-09-25
Add ESLint rule to ensure all source files have empty exports

sindresorhus:
This prevents TypeScript from treating .d.ts files as ambient/global declarations.

See: https://github.com/microsoft/TypeScript/issues/57764


lint-rules/require-export.js
```diff
@@ -0,0 +1,46 @@
...
+			recommended: true,
+		},
+		messages: {
+			noEmptyExport: 'File must have `export {}` to ensure it is treated as a module.',
+		},
+		fixable: 'code',
+	},
+	defaultOptions: [],
+	create(context) {
+		const filename = (context.filename ?? context.getFilename?.() ?? '').replaceAll('\\', '/');
```
som-sm:
```suggestion
		const filename = context.filename.replaceAll('\\', '/');
```

lint-rules/require-export.js
```diff
@@ -0,0 +1,46 @@
...
+		if (!filename.includes('/source/') || !filename.endsWith('.d.ts')) {
+			return {};
+		}
+
+		let hasEmptyExport = false;
+
+		return {
+			ExportNamedDeclaration(node) {
+				// Check if this is specifically `export {}`
+				if (!node.declaration && (!node.specifiers || node.specifiers.length === 0) && !node.source) {
```
som-sm: `node.specifiers` seems to always be an array, so `!node.specifiers` check can be removed.

```suggestion
				if (!node.declaration && node.specifiers.length === 0 && !node.source) {
```

lint-rules/require-export.test.js
```diff
@@ -0,0 +1,68 @@
...
+	assert.equal(errors.length, 1);
+});
+
+test('auto-fix adds export {}', () => {
+	const errors = runRule('/source/foo.d.ts');
+	const fix = errors[0].fix({
+		insertTextAfter: (node, text) => ({node, text}),
+	});
+	assert.equal(fix.text, '\nexport {};\n');
+});
```
som-sm: IMO, it makes more sense to just accept a `node` in `runRule`, because all the handlers should be called with the same `node`. This also slightly simplifies the tests.

```suggestion
const runRule = (filename, node) => {
	const errors = [];
	const context = createContext(filename, errors);
	const handlers = requireExportRule.create(context);

	for (const handler of Object.keys(handlers)) {
		handlers[handler]?.(node);
	}

	return errors;
};

test('ignores non-.d.ts files', () => {
	const handlers = requireExportRule.create(createContext('/source/foo.ts'));
	assert.deepEqual(handlers, {});
});

test('ignores files outside source', () => {
	const handlers = requireExportRule.create(createContext('/test/foo.d.ts'));
	assert.deepEqual(handlers, {});
});

test('processes source .d.ts files', () => {
	const handlers = requireExportRule.create(createContext('/source/foo.d.ts'));
	assert.ok(handlers['Program:exit']);
});

test('passes with export {}', () => {
	const errors = runRule('/source/foo.d.ts', {declaration: null, specifiers: [], source: null});
	assert.equal(errors.length, 0);
});

test('fails without export {}', () => {
	const errors = runRule('/source/foo.d.ts', {declaration: null, specifiers: [{type: 'ExportSpecifier'}]});
	assert.equal(errors.length, 1);
});

test('auto-fix adds export {}', () => {
	const node = {declaration: null, specifiers: [{type: 'ExportSpecifier'}]};
	const errors = runRule('/source/foo.d.ts', node);
	const fix = errors[0].fix({
		insertTextAfter: (node, text) => ({node, text}),
	});
	assert.equal(fix.node, node);
	assert.equal(fix.text, '\nexport {};\n');
});
```

som-sm: Also, `'fails without export {}'` and `'fails with non-empty export'` tests are essentially the same, so _just_ removed one.

Issue #1136, opened 2025-05-12
v6


sindresorhus:
- Remove deprecated `If*` types

ig-raiderler:
Impressive project structure! The architecture really demonstrates solid design principles.

PR #1246, opened 2025-09-25
`FixedLengthArray`: Fix element type

### LINKED ISSUE #1245: can't destructure FixedLengthArray correctly
### Bug description

```
import type {FixedLengthArray} from "type-fest";

declare const arr: FixedLengthArray<string, 2>;

const [a, b] = arr;
//        ^? string|undefined
```

Using `noUncheckedIndexedAccess: true`.

### Repro

https://www.typescriptlang.org/play/?noUncheckedIndexedAccess=true&exactOptionalPropertyTypes=true#code/JYWwDg9gTgLgBDAnmApnA3gMWADxQEwBkUA7AcxgAsBBKKAQ0QF84AzKCEOAIiVQFpWKAM4xuAbgBQk-CgDGAG3pQ0ciCVFxlUAFxxseIqQo06jADyiowcgBo4AJgB8UyWo3wA2vXsAjALpwALxadFIA9OFw0THRAHoA-JJAA

som-sm:
Fixes #1245 

Updated the implementation of `FixedLengthArray`. The updated implementation simply calls `BuildTuple` and then removes keys like `'pop'`, `'push'` from it.

Notes:
- Removed the third type argument `ArrayPrototype` from `FixedLengthArray`. This was used for internal computation and shouldn't have been exposed, so it's more of a bug than a breaking change, and since it had a default type, it's safe to remove it.
- Added `number` to `ArrayLengthMutationKeys`, so that out-of-bounds access is prevented. I can't think of a case where this would break something.

### DIFF
--- a/source/fixed-length-array.d.ts
+++ b/source/fixed-length-array.d.ts
@@ -38,7 +38,6 @@ color.pop();
 
 Use-cases:
 - Declaring fixed-length tuples or arrays with a large number of items.
-- Creating a range union (for example, `0 | 1 | 2 | 3 | 4` from the keys of such a type) without having to resort to recursive types.
 - Creating an array of coordinates with a static length, for example, length of 3 for a 3D vector.
 
 @example


som-sm:
Also, IMO, it doesn't feel very useful if we can _only_ generate tuples having the same type in all positions. Maybe in the next major version we should change the API to simply accept a tuple, like:
```ts
export type FixedLengthArray<TArray> = Except<TArray, ArrayLengthMutationKeys>;
```
If I understand correctly, the only purpose of this type is to remove methods like `'push'`, `'pop'` from a tuple, right?

sindresorhus:
Should these tests pass?

```ts
expectAssignable<string[]>({} as FixedLengthArray<string, 3>);
```

```ts
declare const a: FixedLengthArray<string, 3>;
declare const i: number;
expectType<string>(a[i]);
```

sindresorhus:
A few more tests to add:

```ts
expectType<3>({} as FixedLengthArray<string, 3>['length']);
expectType<string>({} as FixedLengthArray<string, 3>[number]);
```

sindresorhus:
> Also, IMO, it doesn't feel very useful if we can only generate tuples having the same type in all positions. Maybe in the next major version we should change the API to simply accept a tuple, like:

That sounds more like a different type. This type is homogeneous while your suggestion is heterogeneous. Could maybe be:

```ts
export type FixedTuple<T extends unknown[]> = Except<T, ArrayLengthMutationKeys>;
```

?

som-sm:
> That sounds more like a different type. This type is homogeneous while your suggestion is heterogeneous. Could maybe be:

True, but now that we have `BuildTuple` available, we can create homogeneous tuples simply by using `FixedTuple` and `BuildTuple`, like:
```ts
type RGB = FixedTuple<BuildTuple<3, number>>
```
So, does it _still_ make sense to have two separate types?

som-sm:
> Should these tests pass?
> 
> ```ts
> expectAssignable<string[]>({} as FixedLengthArray<string, 3>);
> ```

This shouldn't, because `string[]` has methods like `'push'`, `'pop'`, while `FixedLengthArray<string, 3>` does not. This is similar to why `readonly string[]` is not assignable to `string[]`. Added the following test case instead:
```ts
expectAssignable<readonly string[]>({} as FixedLengthArray<string, 3>);
```

<hr>

> ```ts
> declare const a: FixedLengthArray<string, 3>;
> declare const i: number;
> expectType<string>(a[i]);
> ```

This one is related to the following point:

> * Added `number` to `ArrayLengthMutationKeys`, so that out-of-bounds access is prevented. I can't think of a case where this would break something.

You can't index the array using the non-literal `number` type, because `number` could be something out-of-bounds. Like, in your example, `i` could hold a value of `10` at runtime.

<br>

> A few more tests to add:
> 
> ```ts
> expectType<string>({} as FixedLengthArray<string, 3>[number]);
> ```

This test is similar to the above. LMK if you see _any_ issues with removing `number`.

sindresorhus:
> So, does it still make sense to have two separate types?

Yeah, discoverability. `FixedTuple` is more of a building block.

som-sm:
> * Added `number` to `ArrayLengthMutationKeys`, so that out-of-bounds access is prevented. I can't think of a case where this would break something.

I'll add a note regarding this in the doc.

sindresorhus:
How about this?

```ts
expectType<string | undefined>({} as FixedLengthArray<string, 3>[number]);
```

som-sm:
> How about this?
> 
> ```ts
> expectType<string | undefined>({} as FixedLengthArray<string, 3>[number]);
> ```

Yeah, I was also thinking about this. I think this is better. So, instead of removing the `number` index signature, let's make it `readonly`. This way, we could still read from out-of-bounds indices but not write to them.

This also helps when we create non-tuples using `FixedLengthArray`, because if we remove the `number` index signature, then we would not be able to access any index from something like `FixedLengthArray<string, number>`.

som-sm:
> Yeah, I was also thinking about this. I think this is better. So, instead of removing the `number` index signature, let's make it `readonly`. This way, we could still read from out-of-bounds indices but not write to them.

Turns out it's not that straightforward. 

1. If we add the `number` index signature with an extra `| undefined`, like:
    ```ts
    export type FixedLengthArray<Element, Length extends number> =
	    Except<BuildTuple<Length, Element>, ArrayLengthMutationKeys | number>
	    & {readonly [x: number]: Element | undefined};
    ```
    Then the following test cases break:
    ```ts
    expectAssignable<readonly [string, string, string]>({} as FixedToThreeStrings);
    expectAssignable<readonly string[]>({} as FixedToThreeStrings);
    ```
    image

2. If we just make the `number` index signature `readonly` w/o adding the extra `| undefined`, like:
    ```ts
    export type FixedLengthArray<Element, Length extends number> =
	    Except<BuildTuple<Length, Element>, ArrayLengthMutationKeys | number>
	    & {readonly [x: number]: Element};
    ```
    Then out-of-bounds accesses have the extra `| undefined` _only_ when `noUncheckedIndexedAccess` is enabled.
    ```ts
    // @noUncheckedIndexedAccess: true
    const outOfBounds = fixedToThreeStrings[3];
    //=> string | undefined
    ```
    ```ts
    // @noUncheckedIndexedAccess: false
    const outOfBounds = fixedToThreeStrings[3];
    //=> string
    ```
    And also, if we directly index from the type, the result never has the extra `| undefined` irrespective of the `noUncheckedIndexedAccess` setting. This seems to be a TS quirk, I was expecting `noUncheckedIndexedAccess` to affect type-level accesses as well.
    ```ts
    type OutOfBounds = FixedToThreeStrings[3];
    //=> string
    ```

So, I have come up with a hybrid approach wherein we don't add the `number` index signature for tuples, so out-of-bounds accesses simply error, and for non-tuples (which should be a rarer use case) we add the `number` index signature similar to point no. 2.

source/fixed-length-array.d.ts
```diff
@@ -1,45 +1,122 @@
...
+
+const color: FixedLengthArray<number, 3> = [255, 128, 64];
+
+color.pop();
+//=> Error: Property 'pop' does not exist on type 'FixedLengthArray<number, 3>'.
+```
 
 Use-cases:
 - Declaring fixed-length tuples or arrays with a large number of items.
 - Creating a range union (for example, `0 | 1 | 2 | 3 | 4` from the keys of such a type) without having to resort to recursive types.
```
som-sm: > Creating a range union (for example, `0 | 1 | 2 | 3 | 4` from the keys of such a type) without having to resort to recursive types.

Not sure what this means, should we remove this? 

In fact feels like we can completely remove the use-cases section, it doesn't add enough value IMO.

> Declaring fixed-length tuples or arrays with a large number of items.

Feels like the primary use-case is to create arrays with "large number of items", which isn't true.

som-sm:
@sindresorhus Made significant changes to this PR, also updated the JSDoc completely, please review.

test-d/fixed-length-array.ts
```diff
@@ -1,11 +1,87 @@
...
+expectAssignable<readonly string[]>({} as FixedLength);
+
+// Reading
+// Note: The extra `undefined` is only present when `noUncheckedIndexedAccess` is enabled.
+expectType<string | undefined>(fixedLength[0]);
+expectType<string | undefined>(fixedLength[100]);
+// Note: Reading directly from the type doesn't include `undefined`.
+expectType<string>({} as FixedLength[100]);
+
+// Writing
```
sindresorhus, line 60: This is a breaking change. Fixed length does not imply immutable.

source/fixed-length-array.d.ts
sindresorhus: I think it was there to suggest it could be used for something like this:

```ts
type IndexRange<L extends number> =
	Exclude<keyof FixedLengthArray<unknown, L>, keyof any[]>;
```

but we can drop it. not that useful.

test-d/fixed-length-array.ts
som-sm: @sindresorhus Sorry, I didn't get this. Is it the name that's causing confusing, `FixedLength` is _just_ a bad name I chose, should have been something like `VariableLengthImmutableArray`.

> This is a breaking change.

Umm...here's the diff from the current version:

1. `string[]` is no longer assignable to `FixedLength`, i.e., this case fails:
    
    ```ts
    expectAssignable<FixedLength>({} as string[]);
    ```

    This happens because an extraneous `0` slips into the resulting type because of how it's implemented. Specifically, `ArrayPrototype` has a `0` index which doesn't get excluded properly.

    https://github.com/sindresorhus/type-fest/blob/7fb2f75f3eef405036186d8505f684a376d55b95/source/fixed-length-array.d.ts#L36-L39

    Image

2. Similarly, accessing the `0`th index never has the `| undefined`, i.e., this case fails:
    ```ts
    expectType<string | undefined>(fixedLength[0]);
    ```

3. Lastly, writing is allowed, i.e., these cases fail:
    ```ts
    // Writing
    // @ts-expect-error
    fixedLength[0] = 'a';
    // @ts-expect-error
    fixedLength[100] = 'b';
    ```

All of the above three are bugs, rather than breaking changes. Let me know if I am missing something.

<hr>

Here's a breakdown of how the type currently works:

So, if we create a variable of type `FixedArrayLength<..., number>`, it will allow us to assign an array of _any_ length to it, like:
```ts
let twoLength: FixedLengthArray<string, number> = ['a', 'b']; // ✅ OK
let threeLength: FixedLengthArray<string, number> = ['a', 'b', 'c']; // ✅ OK
let fourLength: FixedLengthArray<string, number> = ['a', 'b', 'c', 'd']; // ✅ OK
let fiveLength: FixedLengthArray<string, number> = ['a', 'b', 'c', 'd', 'e']; // ✅ OK
```

But, once assigned, we should _not_ be able to mutate the array because we'd neither be able to write to it nor be able to invoke mutating methods on it.
```ts
// @ts-expect-error
twoLength[0] = 'c';

// @ts-expect-error
twoLength.push('c');
```

We can, though, reassign the variable with a new array (of different length) altogether, like:
```ts
let twoLength: FixedLengthArray<string, number> = ['a', 'b'];

twoLength = ['a', 'b', 'c', 'd', 'e']; // ✅ OK
```

sindresorhus: Sorry, I should have been clearer. Assigning to an index worked before (in `main`), so it's a breaking change. And also in general a "fixed-length array" supports reassigning elements, just not resizing. It's not immutable.

sindresorhus: > All of the above three are bugs, rather than breaking changes. Let me know if I am missing something.

3 is not a bug, but intentional behavior, but only if the index is known.

sindresorhus: Basically, this is the behavior I think would be correct:

```ts
declare const fixture1: FixedLengthArray<string, 2>;
fixture1[0] = 'a';

declare const fixture2: FixedLengthArray<string, 2>;
// @ts-expect-error
fixture2[3] = 'a';

declare const fixture3: FixedLengthArray<string, number>;
// @ts-expect-error
fixture3[0] = 'a';
```

sindresorhus: > `declare const fixture3: FixedLengthArray<string, number>;`

Should we even allow this? The type is not really useful without knowing the exact size.

som-sm: > And also in general a "fixed-length array" supports reassigning elements, just not resizing. It's not immutable.

Yeahhh, that makes sense!!

But, how do we prevent writing out-of-bounds?

```ts
const someArr: FixedLengthArray<string, number> = ['a', 'b'];

someArr[0] = 'x'; // ✅ Should be allowed, and this makes sense

someArr[2] = 'z'; // ❌ But this shouldn't be allowed, because it's essentially `push`
```

But I guess that's fine, because I think writing out-of-bounds is a very rare scenario.

som-sm: > Sorry, I should have been clearer. Assigning to an index worked before (in `main`), so it's a breaking change. And also in general a "fixed-length array" supports reassigning elements, just not resizing. It's not immutable.

True, makes sense! However, when length is a `number` (not a literal), how do we differentiate between writes within bounds and writes outside bounds, like:
```ts
let someArr: FixedLengthArray<string, number> = ['a', 'b'];

someArr[0] = 'x'; // ✅ Should be allowed, makes sense!

someArr[2] = 'z'; // ❌ But, this shouldn't be allowed, because it's essentially `push`. 

// But later if `someArr` is reassigned to a longer array
someArr = ['a', 'b', 'c'];

// Then assigning to the 2nd index should be allowed
someArr[2] = 'z'; // ✅ Should be allowed now
```
But, we can’t do all of this statically, so we either allow assigning to all indices or prevent assigning to any.

<hr>

> Basically, this is the behavior I think would be correct:
> 
> ```ts
> declare const fixture1: FixedLengthArray<string, 2>;
> fixture1[0] = 'a';
> 
> declare const fixture2: FixedLengthArray<string, 2>;
> // @ts-expect-error
> fixture2[3] = 'a';
> 
> declare const fixture3: FixedLengthArray<string, number>;
> // @ts-expect-error
> fixture3[0] = 'a';
> ```

This is exactly how it's working right now. If the length is known, we can only write to known indices (and not to out-of-bounds indices), but if the length is not known (i.e. `number`), we cannot write to _any_ index.

1. Case where length is known:
    https://github.com/sindresorhus/type-fest/blob/55d7d399ad22a8061b9504d9d6e24433d20508da/test-d/fixed-length-array.ts#L20-L27
2. Case where length is not known (i.e. `number`):
    https://github.com/sindresorhus/type-fest/blob/55d7d399ad22a8061b9504d9d6e24433d20508da/test-d/fixed-length-array.ts#L60-L64
    But this deviates from the existing behaviour, as mentioned in the 3rd point above.

som-sm: > Should we even allow this? The type is not really useful without knowing the exact size.

Yeah, I was thinking _exactly_  the same, because even while writing the documentation, I couldn't really think of a practical use case for this.

IG, we can drop support for this.

som-sm: > IG, we can drop support for this.

But that could potentially be a breaking change, not sure if people are actually using this type with `number` length.

sindresorhus: > But that could potentially be a breaking change, not sure if people are actually using this type with number length.

Probably not worth the risk. I have added it for v6: https://github.com/sindresorhus/type-fest/issues/1136

som-sm: > But, we can’t do all of this statically, so we either allow assigning to all indices or prevent assigning to any.

And what about this?

sindresorhus: > And what about this?

It's kinda risky to change this too. I say we leave it (to be allowed). Disallowing `number` in v6 will make this moot anyway.

som-sm:
Updated the PR, here's a quick summary:

<table>
  <thead>
    <tr>
      <th rowspan="2" align="left">Length</th>
      <th rowspan="2" align="left">Behaviour</th>
      <th colspan="2" align="center">Reading</th>
      <th colspan="2" align="center">Writing</th>
    </tr>
    <tr>
      <th align="center">Within bounds</th>
      <th align="center">Out of bounds</th>
      <th align="center">Within bounds</th>
      <th align="center">Out of bounds</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td rowspan="2" valign="middle" align="left"><strong>Literal</strong></td>
      <td align="left">Updated</td>
      <td align="center">Allowed</td>
      <td align="center">Errors</td>
      <td align="center">Allowed</td>
      <td align="center">Errors</td>
    </tr>
    <tr>
      <td align="left">Current</td>
      <td align="center"><div>Allowed</div><i>(🐛 Adds <code>| undefined</code> if <code>noUncheckedIndexedAccess</code> is ON)</i></td>
      <td align="center"><div>🐛 Allowed</div><i>(Adds <code>| undefined</code> depending on <code>noUncheckedIndexedAccess</code>)</i></td>
      <td align="center">Allowed</td>
      <td align="center">🐛 Allowed</td>
    </tr>
    <tr>
      <td rowspan="2" valign="middle" align="left"><strong>Non Literal</strong></td>
      <td align="left">Updated</td>
      <td align="center"><div>Allowed</div><i>(Adds <code>| undefined</code> depending on <code>noUncheckedIndexedAccess</code>)</i></td>
      <td align="center"><div>Allowed</div><i>(Adds <code>| undefined</code> depending on <code>noUncheckedIndexedAccess</code>)</i></td>
      <td align="center">Allowed</td>
      <td align="center">Allowed</td>
    </tr>
    <tr>
      <td align="left">Current</td>
      <td align="center"><div>Allowed</div><i>(Adds <code>| undefined</code> depending on <code>noUncheckedIndexedAccess</code>)</i></td>
      <td align="center"><div>Allowed</div><i>(Adds <code>| undefined</code> depending on <code>noUncheckedIndexedAccess</code>)</i></td>
      <td align="center">Allowed</td>
      <td align="center">Allowed</td>
    </tr>
  </tbody>
  <tr>
    <td colspan="6" align="right">🐛 Represents bugs</td>
  </tr>
</table>

PR #1247, opened 2025-09-25
Add `TupleOf` type

som-sm:
Exposes the internal `BuildTuple` type.


sindresorhus:
How about we name it just `Tuple`?

Alternatively `TupleOf` or `FilledTuple`.

som-sm:
> How about we name it just `Tuple`?
> 
> Alternatively `TupleOf` or `FilledTuple`.

I think `TupleOf` makes more sense as opposed to _just_ `Tuple` because you can only create homogeneous tuples with this type.

sindresorhus:
Ok. Let's go with TupleOf

PR #1251, opened 2025-09-28
`IsEqual`: Fix behaviour when instantiated with `never` and `unknown`

taiyakihitotsu:
This issue had already existed I think, and I missed it in a previous PR #1231.
Sorry about that.

```typescript
type test_IsEqual = _IsEqual<never, unknown> // true
```


som-sm:
Fails if the order the reversed, i.e., `IsEqual<unknown, never>` returns `true` (not `false`).

taiyakihitotsu:
@som-sm

Thanks, I fixed it.

(Using `[a & b, a | b]` to determine which type is `unknown` and which is `never` failed in some "intersection with infer" test cases, so I wrote it in a more direct way.)

source/is-equal.d.ts
```diff
@@ -26,8 +26,15 @@ type Includes<Value extends readonly any[], Item> =
...
-		? _IsEqual<A, B>
+	[A, B] extends [infer a, infer b]
+		// If using a and b instead of [a] and [b] as the left terms, `IsEqual<any, any>` returns `false`.
+		? [a] extends [never]
+			? [b] extends [never]
+				? true
+				: false
+			: [b] extends [never]
+				? false
+				: _IsEqual<a, b>
```
som-sm:
```suggestion
	[A, B] extends [infer AA, infer BB]
		? [AA] extends [never]
			? [BB] extends [never]
				? true
				: false
			: [BB] extends [never]
				? false
				: _IsEqual<AA, BB>
```

Use captial letters for types, and the comment can be removed.

taiyakihitotsu: @som-sm
Thanks, Refactoring done.

som-sm (approved):

PR #1252, opened 2025-09-28
remaster `Includes` type

benzaria:
- Fully supports optional elements (?) → returns `boolean` if the match is optional.
- Correctly handles required undefined separately from optional (?).

Feature | Old (Recursive) | New (Mapped)
-- | -- | --
Optional element support | ❌ | ✅
Required undefined distinction | ❌ | ✅
Result precision | true/false | true/false/boolean
Edge case coverage | limited | complete

### DIFF
--- a/test-d/includes.ts
+++ b/test-d/includes.ts
@@ -1,57 +1,66 @@
 import {expectType} from 'tsd';
-import type {Includes} from '../index.d.ts';
+import type {Includes} from '../source/includes.d.ts';
 
-const includesEmptyArray: Includes<[], 'abc'> = false;
-expectType<false>(includesEmptyArray);
+declare const boolean: boolean;
 
-const includesSingleItemArray: Includes<['colors'], 'colors'> = true;
-expectType<true>(includesSingleItemArray);
+// ---------- Basic definite cases ----------
+expectType<Includes<[1, 2, 3], 1>>(true);
+expectType<Includes<[1, 2, 3], 3>>(true);
+expectType<Includes<[1, 2, 3], 4>>(false);
 
-const readonlyArray = ['a', 'b', 'c'] as const;
-const includesReadonlyArray: Includes<typeof readonlyArray, 'a'> = true;
-expectType<true>(includesReadonlyArray);
+// ---------- Optional elements ----------
+expectType<Includes<[a: string, b?: number], string>>(true);
+expectType<Includes<[a: string, b?: number], number>>(boolean);
+expectType<Includes<[a: string, b?: number], undefined>>(false);
 
-const includesComplexMultiTypeArray: Includes<[
-	{
-		prop: 'value';
-		num: 5;
-		anotherArr: [1, '5', false];
-	},
-	true,
-	null,
-	'abcd',
-], 'abc'> = false;
-expectType<false>(includesComplexMultiTypeArray);
+// ---------- Explicit undefined ----------
+expectType<Includes<[a: string, b: undefined], undefined>>(true);
+expectType<Includes<[a: string, b: undefined], string>>(true);
+expectType<Includes<[a: string, b: undefined], number>>(false);
 
-const noExtendsProblem: Includes<[boolean], true> = false;
-expectType<false>(noExtendsProblem);
+// ---------- Empty tuple ----------
+expectType<Includes<[], any>>(false);
+expectType<Includes<[], never>>(false);
+expectType<Includes<[], undefined>>(false);
 
-const objectIncludes: Includes<[{}], {a: 1}> = false;
-expectType<false>(objectIncludes);
+// ---------- Homogeneous arrays ----------
+expectType<Includes<string[], string>>(true);
+expectType<Includes<string[], number>>(false);
+expectType<Includes<readonly number[], number>>(true);
+expectType<Includes<readonly number[], string>>(false);
 
-const objectIncludesPass: Includes<[{a: 1}], {a: 1}> = true;
-expectType<true>(objectIncludesPass);
+// ---------- Nested ----------
+expectType<Includes<[[1, 2], [3, 4]], [1, 2]>>(true);
+expectType<Includes<[[1, 2], [3, 4]], [2, 3]>>(false);
+expectType<Includes<[{a: 1; b: 2}, {a: 3; b: 4}], {a: 1; b: 2}>>(true);
+expectType<Includes<[{a: 1; b: 2}, {c: 3; d: 4}], {c: 3; d: 4}>>(true);
 
-const nullIncludesUndefined: Includes<[null], undefined> = false;
-expectType<false>(nullIncludesUndefined);
+// ---------- Optional + required mix ----------
+expectType<Includes<[1, 3, 3?], 3>>(true);
+expectType<Includes<[1, 2?, 3?], 1>>(true);
+expectType<Includes<[1, 2?, 3?], 2>>(boolean);
+expectType<Includes<[1, 2?, 3?], 3>>(boolean);
 
-const nullIncludesNullPass: Includes<[null], null> = true;
-expectType<true>(nullIncludesNullPass);
+// ---------- Unions ----------
+expectType<Includes<[1 | 2, 3], 1>>(false);
+expectType<Includes<[1 | 2, 3], 1 | 2>>(true);
+expectType<Includes<[1, 3] | [2, 3], 1>>(true);
 
-// Verify that incorrect usage of `Includes` produces an error.
+// ---------- Literal cases ----------
+declare const sym: unique symbol;
+expectType<Includes<[typeof sym], typeof sym>>(true);
+expectType<Includes<[typeof sym], symbol>>(false);
 
-// Missing all generic parameters.
-// @ts-expect-error
-type A0 = Includes;
+// ---------- Mixed optional ----------
+expectType<Includes<[1?, 2?], 1>>(boolean);
+expectType<Includes<[1?, 2?], 2>>(boolean);
+expectType<Includes<[1?, 2?], 3>>(false);
+expectType<Includes<[undefined?], undefined>>(boolean);
+expectType<Includes<[undefined], undefined>>(true);
+expectType<Includes<[undefined?], number>>(false);
 
-// Missing `Item` generic parameter.
-// @ts-expect-error
-type A1 = Includes<['my', 'array', 'has', 'stuff']>;
-
-// Value generic parameter is a string not an array.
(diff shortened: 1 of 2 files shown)


benzaria:
After consideration a recursive types is better to cover the rest element cases.
if there is any trick that can be used in the first maped object type to handle the rest element, whoud that be better ?

sindresorhus:
@claude

sindresorhus:
Suggestions for some more tests:

```ts
expectType<Includes<readonly [1, 2?, 3?], 2>>(boolean);

expectType<Includes<[1?, 2?, 3?], 1>>(boolean);
expectType<Includes<[1?, 2?, 3?], 4>>(false);

expectType<Includes<[unknown], unknown>>(true);
expectType<Includes<[unknown], string>>(false);
expectType<Includes<[1, unknown], unknown>>(true);

expectType<Includes<[null], null>>(true);
expectType<Includes<[null], undefined>>(false);

expectType<Includes<[true, false], true>>(true);
expectType<Includes<[true, false], boolean>>(false);
expectType<Includes<[boolean], boolean>>(true);
```

som-sm:
@sindresorhus 
> ```ts
> expectType<Includes<[true, false], boolean>>(false);
> ```

Umm, this should be `true`, because:

```
Includes<[true, false], boolean>
=> Includes<[true, false], true> | Includes<[true, false], false>
=> true | true
=> true
```

<br>

> ```ts
> expectType<Includes<[boolean], boolean>>(true);
> ```

And, this should be `boolean`, because:
```
Includes<[boolean], boolean>
=> Includes<[true], true> | Includes<[true], false> | Includes<[false], true> | Includes<[false], false>
=> true | false | false | true
=> boolean
```

test-d/includes.ts
```diff
@@ -1,57 +1,65 @@
...
+expectType<Includes<[{a: 1; b: 2}, {a: 3; b: 4}], {a: 1; b: 2}>>(true);
+expectType<Includes<[{a: 1; b: 2}, {c: 3; d: 4}], {c: 3; d: 4}>>(true);
 
-const objectIncludesPass: Includes<[{a: 1}], {a: 1}> = true;
-expectType<true>(objectIncludesPass);
+// ---------- Optional/Rest ----------
+expectType<Includes<[1, 3, 3?], 3>>(true);
+expectType<Includes<[1, 2?, 3?], 1>>(true);
+expectType<Includes<[1, 2?, 3?], 2>>(boolean);
+expectType<Includes<[1, ...Array<2>, 3], 3>>(true);
```
som-sm, line 41: This should be `boolean`.

test-d/includes.ts
```diff
@@ -1,57 +1,65 @@
...
-expectType<false>(includesComplexMultiTypeArray);
+// ---------- Empty tuple ----------
+expectType<Includes<[], any>>(false);
+expectType<Includes<[], never>>(false);
+expectType<Includes<[], undefined>>(false);
 
-const noExtendsProblem: Includes<[boolean], true> = false;
-expectType<false>(noExtendsProblem);
+// ---------- Homogeneous arrays ----------
+expectType<Includes<string[], string>>(boolean);
```
som-sm, line 22: Add:
```ts
expectType<Includes<string[], "a">>(boolean);
```

test-d/includes.ts
```diff
@@ -1,57 +1,65 @@
...
+expectType<Includes<[1, 3?, ...Array<2>], 2>>(boolean);
+expectType<Includes<[1, 3?, ...Array<2>], 3>>(boolean);
+expectType<Includes<[1, 2, ...Array<2>], 2>>(true);
+expectType<Includes<[1, ...Array<2>, 2], 2>>(true);
+expectType<Includes<[1, 2?, ...Array<2>], 2>>(boolean);
 
-const nullIncludesUndefined: Includes<[null], undefined> = false;
-expectType<false>(nullIncludesUndefined);
+// ---------- Unions ----------
+expectType<Includes<[1 | 2, 3], 1>>(false);
```
som-sm: Should be `boolean`, right?

test-d/includes.ts
```diff
@@ -1,57 +1,65 @@
...
+expectType<Includes<[1 | 2, 3], 1 | 2>>(true);
```
som-sm: This should also be `boolean`.

test-d/includes.ts
```diff
@@ -1,57 +1,65 @@
...
+expectType<Includes<[1, 3] | [2, 3], 3>>(true);
+expectType<Includes<[1, 3] | [2, 3], 1>>(boolean);
```
som-sm: Add:
```ts
expectType<Includes<[1, 3] | [2, 3], 5>>(false);
```

test-d/includes.ts
```diff
@@ -1,57 +1,65 @@
...
-const nullIncludesNullPass: Includes<[null], null> = true;
-expectType<true>(nullIncludesNullPass);
+// ---------- Literal cases ----------
+declare const sym: unique symbol;
+expectType<Includes<[typeof sym], typeof sym>>(true);
+expectType<Includes<[typeof sym], symbol>>(false);
 
-// Verify that incorrect usage of `Includes` produces an error.
+// ---------- Undefined ----------
+expectType<Includes<[1?, 2?], undefined>>(false);
```
som-sm, line 54: Does this change with `exactOptionalPropertyTypes`?

sindresorhus:
@som-sm `Includes` here is exact-equality (`IsEqual`) based. No assignability, no distribution over `Item`. So I feel like this is correct:

- `Includes<[true, false], boolean>` ⇒ false*
- `Includes<[boolean], boolean>` ⇒ true
- It only returns `boolean` when uncertainty is introduced by optionals/rest.

som-sm:
> @som-sm `Includes` here is exact-equality (`IsEqual`) based. No assignability, no distribution over `Item`. So I feel like this is correct:

@sindresorhus Umm...ok gotcha, but in my experience distributing is _usually_ a good idea. I'm not sure of good use cases for this type, but here's a quick, contrived example:

```ts
declare function includes<
	const T extends readonly unknown[],
	const Target extends Includes<T, Target> extends false ? never : unknown,
>(array: T, target: Target): void;

includes([1, 2, 3] as [1, 2, 3 | 4], 4); // Errors, but shouldn't
// Argument of type '4' is not assignable to parameter of type 'never'
```

<br>

@benzaria Could you share some practical, real-world examples for this type?

benzaria:
I intentionally designed `Includes` to be strict following the original behavior.

But @som-sm had a good point of adding distribution, which can be useful.

I can add an option, that triggers it and make it false by default.

@sindresorhus WDYT

sindresorhus:
Yeah, I think a `{distributeItem}` option could be useful, but it should be off by default.

test-d/includes.ts
benzaria: which one? all of them are right!

som-sm: Looks like I commented at the wrong place, all of these look fine.

```ts
expectType<Includes<[1, 3, 3?], 3>>(true);
expectType<Includes<[1, 2?, 3?], 1>>(true);
expectType<Includes<[1, 2?, 3?], 2>>(boolean);
expectType<Includes<[1, ...Array<2>, 3], 3>>(true);
```

test-d/includes.ts
benzaria: I also managed to add support for this:

```ts
expectType<Includes<string[], 'a'>>(boolean);
expectType<Includes<[string, number], 'a'>>(boolean);
expectType<Includes<[1, 2, 3], number>>(boolean);
```

test-d/includes.ts
benzaria: no it does **not**

sindresorhus (approved):

sindresorhus:
Missed this, it should be `distributeItem` singular, not plural. We are distributing the "Item" type.

benzaria:
> Missed this, it should be `distributeItem` singular, not plural. We are distributing the "Item" type.

I did notice that, but in this case, we’re actually distributing both the Item type and the array element types — that’s why I chose the plural form.
Would you still prefer keeping it singular for consistency?

sindresorhus:
I still prefer `distributeItem` (singular). While array elements do get distributed, that's an implementation detail of the comparison logic. The option specifically controls whether to distribute the `Item` parameter. The name should describe what the user is controlling (distribution of the `Item` type), not all the internal mechanics.

sindresorhus:
Some more things:

1. `distributeItem: true` currently over-resolves some uncertain `any` cases to `true` instead of `boolean`.

Because the comparison path uses a strict-extends helper that returns `true` when either side is `any`, cases like:
- `Includes<[any], unknown, {distributeItem: true}>`
- `Includes<[1], any, {distributeItem: true}>`
resolve to `true`.

That feels too strong for uncertainty semantics. I’d expect `boolean` there.

2. Type-error coverage was dropped from the tests.

The previous tests checked invalid generic usage (missing generics, non-array input, etc.). Those checks are now gone, so regressions in input constraints could slip through.

3. Minor docs polish:
- The option description is a bit ambiguous about what is actually distributed.
- There is a small typo: “eachother” should be “each other”.

Opn question: is the `any` behavior in distribute mode intentional? If yes, I think it should be documented explicitly because it differs from other uncertainty cases.

test-d/includes.ts
```diff
@@ -1,57 +1,97 @@
...
+// ---------- Edge cases ----------
+expectType<Includes<[never], any>>(false);
+expectType<Includes<[never], never>>(true);
+expectType<Includes<[never], any, {distributeItem: true}>>(false);
+expectType<Includes<[never], never, {distributeItem: true}>>(true);
+expectType<Includes<[unknown], unknown>>(true);
+expectType<Includes<[unknown], string>>(false);
+expectType<Includes<[1, unknown], unknown>>(true);
+expectType<Includes<any, any>>({} as any);
+expectType<Includes<never, never>>({} as never);
```
sindresorhus, line 97:
```suggestion
expectType<Includes<never, never>>({} as never);

// ---------- Invalid usage ----------
// @ts-expect-error
type IncludesMissingGenerics = Includes;
// @ts-expect-error
type IncludesMissingItem = Includes<[1, 2, 3]>;
// @ts-expect-error
type IncludesWithNonArrayInput = Includes<'abc', 'a'>;
// @ts-expect-error
type IncludesWithObjectInput = Includes<{key: 'value'}, 7>;
// @ts-expect-error
type IncludesWithInvalidOptions = Includes<[1], 1, {distributeItem: 1}>;
```

PR #1253, opened 2025-09-28
add: `IndexOf`, `LastIndexOf` and `IndicesOf` types

benzaria:
this is a remake of the #1155

- **`IndexOf`** Returns the index of the first occurrence of a value in an array, or `never` if it is not present.
- **`LastIndexOf`** Returns the index of the last occurrence of a value in an array, or `never` if it is not present.
- **`IndicesOf`** Return an array of indexes of all the occurrences of a value in an array, or `[]` if it is not present.

this version added support for arrays (rest elements) so it returns a negative index if the match is behind the rest element, which will come perfect with the next plan of introducing `ArrayAt` type.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -158,6 +158,9 @@ export type {IsUppercase} from './source/is-uppercase.d.ts';
 export type {IsOptional} from './source/is-optional.d.ts';
 export type {IsNullable} from './source/is-nullable.d.ts';
 export type {TupleOf} from './source/tuple-of.d.ts';
+export type {IndexOf} from './source/index-of.d.ts';
+export type {LastIndexOf} from './source/last-index-of.d.ts';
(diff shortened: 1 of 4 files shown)


sindresorhus:
1. `IndexOf` currently leaks tuple accumulator types into the result for variable-length arrays. In some cases it returns a tuple-like type inside `LiteralUnion` instead of a numeric indx.

2. `IndicesOf` can return element value types where it should return index types. For variable arrays, the fallback branch is building from `Array_[number]` instead of numeric indices.

3. `LastIndexOf` `fromIndex` behavior looks incorrect and inconsistent. It appears to treat `fromIndex` as a reverse offset in one path, and another path ignores `fromIndex` entirely. A simple tuple repro gives `0` where `2` is expected.

4. Needs tests

PR #1156, opened 2025-05-26
Add `Xor` and `XorAll` types

benzaria:
**External**
 - Xor Exclusive Or
 - XorAll Return true when the specified number of trues present, default 1 <= x <= length - 1, 
   can be specified using {strict: number} option

> [!NOTE]
> This PR depends on #1155 !

PR #1254, opened 2025-09-29
Add `Xor` type

benzaria:
adding `Xor` type (remake #1156 )

### DIFF
--- a/source/xor.d.ts
+++ b/source/xor.d.ts
@@ -1,5 +1,6 @@
-import type {If} from './if.d.ts';
-import type {IsNever} from './is-never.d.ts';
+import type {Not} from './internal/type.d.ts';
+import type {And} from './and.d.ts';
+import type {Or} from './or.d.ts';
 
 /**
 Returns a boolean for whether only one of two given types is true.
@@ -77,23 +78,6 @@ type G = Xor<never, never>;
 @see {@link And}
 @see {@link Or}
 */
-export type Xor<A extends boolean, B extends boolean> =
-	_Xor<If<IsNever<A>, false, A>, If<IsNever<B>, false, B>>; // `never` is treated as `false`
-
-/**
-Core type for `Xor` equivalent to:
-
-```
-type Xor<A extends boolean, B extends boolean> = And<Or<A, B>, Not<And<A, B>>>;
-```
-*/
-type _Xor<A extends boolean, B extends boolean> =
-	A extends true
-		? B extends true
-			? false
-			: true
-		: B extends true
-			? true
-			: false;
+export type Xor<A extends boolean, B extends boolean> = And<Or<A, B>, Not<And<A, B>>>;
 
 export {};


source/xor.d.ts
```diff
@@ -0,0 +1,93 @@
+
+/**
+Returns `false` if the giving type is `never`, else return it.
+*/
+type IfNever<T> = [T] extends [never] ? false : T;
```
sindresorhus: We already have a `IsNever` type. You could use it with `If` type.

source/xor.d.ts
```diff
@@ -0,0 +1,93 @@
...
+//=> boolean
+
+type F = Xor<never, boolean>;
+//=> boolean
+
+type G = Xor<never, never>;
+//=> false
+```
+
+@see And, Or
```
sindresorhus:
```suggestion
@see {@link And}
@see {@link Or}
```

sindresorhus (changes requested):

source/xor.d.ts
```diff
@@ -0,0 +1,93 @@
...
+*/
+export type Xor<A extends boolean, B extends boolean> = _Xor<IfNever<A>, IfNever<B>>; // `never` is treated as `false`
+
+export type _Xor<A extends boolean, B extends boolean> =
```
som-sm: Remove `export`

```suggestion
type _Xor<A extends boolean, B extends boolean> =
```

readme.md
```diff
@@ -179,7 +179,8 @@ Click the type names for complete docs.
 - [`DistributedOmit`](source/distributed-omit.d.ts) - Omits keys from a type, distributing the operation over a union.
 - [`DistributedPick`](source/distributed-pick.d.ts) - Picks keys from a type, distributing the operation over a union.
 - [`And`](source/and.d.ts) - Returns a boolean for whether two given types are both true.
-- [`Or`](source/or.d.ts) - Returns a boolean for whether either of two given types are true.
+- [`Or`](source/or.d.ts) - Returns a boolean for whether either of two given types is true.
```
som-sm: Why change the description for `Or`? It now mismatches from the description of `Or` in it's JSDoc.

source/xor.d.ts
```diff
@@ -0,0 +1,93 @@
+
+/**
+Returns `false` if the giving type is `never`, else return it.
+*/
+type IfNever<T> = [T] extends [never] ? false : T;
+// TODO: use to simplify `And`, `Or` types
```
som-sm: Do you mean rewrite `And` and `Or` with `Xor`? I feels it makes more sense to write `Xor` using `And` and `Or`, like:

```ts
type Xor<A extends boolean, B extends boolean> = Or<And<Not<A>, B>, And<A, Not<B>>>;
```

But this would currently not work because `Not` doesn't handle `never` properly. So, your implementation  is fine, just remove this comment.

som-sm (changes requested):

source/xor.d.ts
benzaria: i missed that!

source/xor.d.ts
benzaria: Ohh no, the comment point to `IfNever` that i want to use to simplify this pattern

```ts
type T = _Or<If<IsNever<A>, false, A>, If<IsNever<B>, false, B>>
```
which is used in `Or` and repeats multiple unnecessary checks trough the process and `IfNever` is a custom private types between `Xor` and `Or` which directly infer the result:

```ts
type T = _Or<IfNever<A>, IfNever<B>>
```

benzaria: I was going to use `And` and `Or` for this, but the extra conditions didn’t make much sense to me — why prettify things and lose performance?

A working `Xor` would look like this:

```ts
type Xor<A extends boolean, B extends boolean> = And<Or<A, B>, Not<And<A, B>>>;
```

readme.md
benzaria: that sliped from another branch. I'll remove it.

source/xor.d.ts
benzaria: the `@link` don't work unless the type refrenced is imported! it just shows as text.

source/xor.d.ts
benzaria: its simplify the condition https://github.com/sindresorhus/type-fest/pull/1254#discussion_r2404608413
ofc this is just for preformance, i can reuse the same pattern as `Or`.

source/xor.d.ts
som-sm: > Ohh no, the comment point to `IfNever` that i want to use to simplify this pattern

No, please remove this abstraction and use `<If<IsNever<T>, false, T>`, similar to how it’s done in `Or`:
https://github.com/sindresorhus/type-fest/blob/3bd9de60d5384543c82c3de2a1461bdecc0da876/source/or.d.ts#L78-L79

It's quite simple and easy to understand and doesn't need further abstraction.

som-sm: > I was going to use `And` and `Or` for this, but the extra conditions didn’t make much sense to me — why prettify things and lose performance?

There's NO performance loss as such, and it's actually better doing it with `And` and `Or` because it helps validate that `And` and `Or` themselves are implemented correctly. But, it's fine for now.

benzaria: Okay, I'll use ur seggestion for `IfNever`.
For the `And` and `Or` version of `Xor` it does work perfectly, I test it. If it's better to make it this way tell me.

som-sm: > For the `And` and `Or` version of `Xor` it does work perfectly, I test it. If it's better to make it this way tell me.

Yeah, let's do it using `And`, `Or` & `Not`.

<br>

Either definition works:

1. > A working `Xor` would look like this:
    > 
    > ```ts
    > type Xor<A extends boolean, B extends boolean> = And<Or<A, B>, Not<And<A, B>>>;
    > ```

2. > I feels it makes more sense to write `Xor` using `And` and `Or`, like:
    > 
    > ```ts
    > type Xor<A extends boolean, B extends boolean> = Or<And<Not<A>, B>, And<A, Not<B>>>;
    > ```

som-sm (approved):

PR #1256, opened 2025-09-30
`ReadonlyTuple`: Deprecate in favour of `TupleOf`

som-sm:
Now that we have [`TupleOf`](https://github.com/sindresorhus/type-fest/pull/1250) exposed, I don't think we need a separate `ReadonlyTuple` type because it simply wraps the result of `TupleOf` in `Readonly`.

This PR also refactors the implementation of `ReadonlyTuple` to use `TupleOf`, which in turn fixes the following bug wherein the type wasn't distributing over the `Length` argument.
```ts
type Current = ReadonlyTuple<string, 3 | 4>;
//=> readonly [string, string, string]

type Fixed = ReadonlyTuple<string, 3 | 4>;
//=> readonly [string, string, string] | readonly [string, string, string, string]
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -345,6 +345,7 @@ Click the type names for complete docs.
 - `HomomorphicOmit` - See [`Except`](source/except.d.ts)
 - `IfAny`, `IfNever`, `If*` - See [`If`](source/if.d.ts)
 - `MaybePromise` - See [`Promisable`](source/promisable.d.ts)
+- `ReadonlyTuple` - See [`TupleOf`](source/tuple-of.d.ts)
 
 ## Tips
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
Maybe worth adding it to https://github.com/sindresorhus/type-fest?tab=readme-ov-file#alternative-type-names too?

som-sm:
> Maybe worth adding it to https://github.com/sindresorhus/type-fest?tab=readme-ov-file#alternative-type-names too?

Yeah, I thought this too, but do you wanna add this now or after we have removed the type?

sindresorhus:
We can just do it now. When I type is deprecated, we can just pretend it doesn't exist.

PR #1261, opened 2025-10-06
`Or`/`And`: Fix documentation errors and add link to `Xor`

benzaria:
#### 💡 Reasoning

* **Over-dependence on a complex external utility (`AllExtend`)**
  The previous version of `And` relied on `AllExtend`, which is recursive and includes rest-element collapsing, `never`/`any` handling, and option merging — all of which are **massive overkill** for a simple 2-input boolean gate.

* **Performance & readability**
  `AllExtend` triggers multiple recursive and distributive evaluations. This adds **compiler overhead** and **reduces clarity**, turning a straightforward logical operation into a deep generic resolution that can slow down large TypeScript projects.

* **Loss of boolean simplicity**
  A boolean combinator like `And` should always evaluate to a plain `true` or `false`. Using `AllExtend` introduced unnecessary conditional complexity and made the result harder to reason about.

### DIFF
--- a/source/or.d.ts
+++ b/source/or.d.ts
@@ -80,11 +80,10 @@ type G = Or<never, never>;
 export type Or<A extends boolean, B extends boolean> =
 	_Or<If<IsNever<A>, false, A>, If<IsNever<B>, false, B>>; // `never` is treated as `false`
 
-type _Or<A extends boolean, B extends boolean> =
-	A extends true
+export type _Or<A extends boolean, B extends boolean> = A extends true
+	? true
+	: B extends true
 		? true
-		: B extends true
-			? true
-			: false;
+		: false;
 
 export {};


som-sm:
> #### 💡 Reasoning
> * **Over-dependence on a complex external utility (`AllExtend`)**
>   The previous version of `And` relied on `AllExtend`, which is recursive and includes rest-element collapsing, `never`/`any` handling, and option merging — all of which are **massive overkill** for a simple 2-input boolean gate.
> * **Performance & readability**
>   `AllExtend` triggers multiple recursive and distributive evaluations. This adds **compiler overhead** and **reduces clarity**, turning a straightforward logical operation into a deep generic resolution that can slow down large TypeScript projects.
> * **Loss of boolean simplicity**
>   A boolean combinator like `And` should always evaluate to a plain `true` or `false`. Using `AllExtend` introduced unnecessary conditional complexity and made the result harder to reason about.

The reasoning makes _no_ sense to me, feels like some AI-generated fluff.

The JSDoc corrections and the refactoring in tests are fine, but it doesn't make sense to change the implementation.

benzaria:
I’d like to clarify that this change isn’t arbitrary, it’s based on measurable compile-time performance differences.
You can verify it yourself:

```ts
// avg: 22.1 ms, max: 24.6 ms
type AndOld<A extends boolean, B extends boolean> = AllExtend<[A, B], true>;

// avg: 16.3 ms, max: 16.8 ms
type AndNew<A extends boolean, B extends boolean> =
	A extends true
		? B extends true
			? true
			: false
		: false;
```

When designing a library, performance should always be considered, if not as the top priority, then at least as an important factor.
While the difference may seem small, reducing type-checking complexity and avoiding unnecessary recursion or conditional depth directly improves compile-time responsiveness, especially in larger projects.

The goal here isn’t micro-optimization for its own sake, but rather keeping the library lightweight, predictable, and easy for the TypeScript compiler to reason about.

@sindresorhus could you please share your thoughts on this change?

sindresorhus:
> When designing a library, performance should always be considered, if not as the top priority, then at least as an important factor.

I agree in principle, but it's not as easy as just measuring the performance of a single function in isolation: https://sindresorhus.com/blog/micro-benchmark-fallacy

22ms vs 16ms micro-timings without a reproducible harness across real projects (and TS versions) is noise. If there’s a meaningful compile-time win, prove it on a large repo with a scriptable `tsc --generateTrace` (or similar) and show >3% end-to-end, not a micro test.

sindresorhus:
One argument for this change is that it makes it easier to copy-paste the type, but it already has another import anyway, so it's still not just select and copy.

sindresorhus:
So in summary, I don't think we want to remove using `AllExtend`.

benzaria:
> I agree in principle, but it's not as easy as just measuring the performance of a single function in isolation: https://sindresorhus.com/blog/micro-benchmark-fallacy

Good point, the article clarified a lot and helped me realize where my reasoning was off. I’ll discard the changes to `And` for now and revert it to its previous implementation.

benzaria:
@sindresorhus is this PR is ready to merge !?

Issue #585, opened 2023-03-30
`OverloadParameters` and `OverloadReturnType`


roydukkey:
Basically native `Parameter` and `ReturnType` [do not handle overloaded interfaces](https://github.com/microsoft/TypeScript/issues/14107).

Here is a workaround: https://github.com/microsoft/TypeScript/issues/14107#issuecomment-1146738780

sindresorhus:
Looks useful. I have had this need before too.

xusiyuan841028:
```ts
// support 5 times overload at most
type OverloadedReturnType<T> = 
    T extends { (...args: any[]) : infer R; (...args: any[]) : infer R; (...args: any[]) : infer R ; (...args: any[]) : infer R; (...args: any[]) : infer R } ? R  :
    T extends { (...args: any[]) : infer R; (...args: any[]) : infer R; (...args: any[]) : infer R ; (...args: any[]) : infer R } ? R  :
    T extends { (...args: any[]) : infer R; (...args: any[]) : infer R; (...args: any[]) : infer R } ? R  :
    T extends { (...args: any[]) : infer R; (...args: any[]) : infer R } ? R  :
    T extends (...args: any[]) => infer R ? R : any;

function fn5(a: boolean): boolean;
function fn5(a: string): string;
function fn5(a: number): number;
function fn5(a: null): null;
function fn5(a: { a: string }): { a: string };
function fn5(a: string | number | boolean | null | { a: string } | { a: number }) : {a: string } | string | number | boolean | null | { a: number } {
  return a;
}

// OK
type RFn5 = OverloadedReturnType<typeof fn5>;

function fn6(a: boolean): boolean;
function fn6(a: string): string;
function fn6(a: number): number;
function fn6(a: null): null;
function fn6(a: { a: string }): { a: string };
function fn6(a: { a: number }): { a: number };
function fn6(a: string | number | boolean | null | { a: string } | { a: number }) : {a: string } | string | number | boolean | null | { a: number } {
  return a;
}

// boolean type is lost
type RFn6 = OverloadedReturnType<typeof fn6>;
```ts

xusiyuan841028:
But  can't implement `OverloadedParameters` ...

pls refer to 
https://github.com/microsoft/TypeScript/issues/14107 
https://github.com/microsoft/TypeScript/issues/32164
[...]

PR #1264, opened 2025-10-09
Add `FunctionOverloads` type

### LINKED ISSUE #868: A type util to produce an alternation of function overloads.
Example situation: 

```typescript
type F = {
    (thing: string): void;
    (things: string[]): void;
}

type Z = Parameters<F>[0]
```

Type Z here resolves to `string[]`, not `string | string[]`, because overloads just get the last declared one matched. 

I am curious to know if this is possible to implement. As far as I can tell this also affects the standard function overload approach, e.g.: 

```typescript
function fun(thing: string): void; 
function fun(stuff: string[]): void;
// function fun(thing: string | string[]); // uncomment me to allow type Y to match `string`.
function fun(thing: string | string[]) { console.log(Array.isArray(thing) ? thing.join('; ') : thing); }

type F = {
    (thing: string): void;
    (things: string[]): void;
}

type Z = Parameters<F>[0]
type Y = Parameters<typeof fun>[0];

let zz:Z = 'abc';
let yy:Y = 'abc';
```

To be clear, what I'm asking for is a util e.g. `OverloadParameters` which in this case would produce for type `OverloadParameters<F>` a value of `[string] | [string[]]`. ~~I'm actually currently unsure of how to unwrap the parameter-array situation there but I'm sure it's manageable.~~ As expected, the type `([string] | [string[]])[0]` is `string | string[]`.

andreww2012:
Resolves https://github.com/sindresorhus/type-fest/issues/868 and resolves partially https://github.com/sindresorhus/type-fest/issues/585


test-d/function-overloads.ts
```diff
@@ -0,0 +1,100 @@
...
+	| Function7
+	| Function8
+	| Function9
+	| Function10
+	| Function11
+	| Function12
+>(manyOverloads);
+
+declare const noOveloads: FunctionOverloads<{}>;
+expectType<never>(noOveloads);
```
sindresorhus: Maybe also test some edge-cases:

```ts
declare const anyOverload: FunctionOverloads<any>;
expectType<(...arguments_: readonly unknown[]) => unknown>(anyOverload);

declare const neverOverload: FunctionOverloads<never>;
expectType<never>(neverOverload);

declare const unknownOverload: FunctionOverloads<unknown>;
expectType<never>(unknownOverload);
```

sindresorhus:
I think https://github.com/microsoft/TypeScript/issues/14107 may be relevant. Maybe link to it in the doc comment.

source/function-overloads.d.ts
```diff
@@ -0,0 +1,79 @@
+import type {FunctionWithMaybeThisParameter} from './internal/function.d.ts';
+import type {IsEqual} from './is-equal.d.ts';
+
+/**
+Create a union of all the function's overloads.
+
```
sindresorhus, line 7:
```suggestion
TypeScript's built-in utility types like `Parameters` and `ReturnType` only work with the last overload signature. This type extracts all overload signatures as a union, allowing you to work with each overload individually.

Use-cases:
- Extract parameter types from specific overloads using `Extract` and `Parameters`
- Analyze all possible function signatures in type-level code
- Extract event handler signatures from framework APIs

```

source/function-overloads.d.ts
```diff
@@ -0,0 +1,79 @@
...
+// Extract the type of `submit` event like this:
+
+import type {ArrayTail, FunctionOverloads} from 'type-fest';
+import HelloWorld from './HelloWorld.vue';
+
+type SubmitEventType = ArrayTail<Parameters<Extract<FunctionOverloads<InstanceType<typeof HelloWorld>['$emit']>, (event: 'submit', ...args: any[]) => void>>>;
+//=> [formData: FormData]
+```
+
```
sindresorhus, line 62: Should mention limitations with this code:

- Functions that have identical parameters but different `this` types or return types will only extract one overload
- Generic type parameters are lost and inferred as `unknown`

andreww2012:
Thank you for the review!

> I think [microsoft/TypeScript#14107](https://github.com/microsoft/TypeScript/issues/14107) may be relevant. Maybe link to it in the doc comment.

I'm not sure how I'd link this issue. I think https://github.com/microsoft/TypeScript/issues/32164 is actually more relevant (and the author of the trick I've used also left the same comment about the method of getting union of overloads), but I don't see additional value of including it the docs too.

sindresorhus:
> I think https://github.com/microsoft/TypeScript/issues/32164 is actually more relevant (and the author of the trick I've used also left the same comment about the method of getting union of overloads), but I don't see additional value of including it the docs too.

It can be useful as context to the limitation.

source/function-overloads.d.ts
```diff
@@ -0,0 +1,97 @@
+import type {FunctionWithMaybeThisParameter} from './internal/function.d.ts';
+import type {IsEqual} from './is-equal.d.ts';
+
+/**
+Create a union of all the function's overloads.
+
+TypeScript's built-in utility types like `Parameters` and `ReturnType` only work with the last overload signature. This type extracts all overload signatures as a union, allowing you to work with each overload individually.
```
sindresorhus:
```suggestion
TypeScript's built-in utility types like `Parameters` and `ReturnType` only work with the last overload signature, [by design](https://github.com/microsoft/TypeScript/issues/32164). This type extracts all overload signatures as a union, allowing you to work with each overload individually.
```

maybe something like this

andreww2012:
Mentioned the issue that this type addresses.

Also removed the limitation of losing `readonly` modifier if other parameters are present from the docs because actually this combination is completely useless in practice:

```ts
declare function foo(bar: string, ...baz: readonly number[]): void;
declare const array: number[];
foo('', ...array); // Allowed despite `array` not being readonly
```

sindresorhus (approved):

source/function-overloads.d.ts
```diff
@@ -0,0 +1,96 @@
...
+						// makes compiler infer a last overload that do not equal one of the concatenated ones.
+						// Thus, we're ending up iterating over all the overloads from bottom to top.
+						// Credits: https://github.com/microsoft/TypeScript/issues/32164#issuecomment-1146737709
+						CheckedOverloads & AllOverloads,
+						CheckedOverloads & ((this: ThisType, ...arguments_: ParametersType) => ReturnType),
+						MustStopIfParametersAreEqual extends true ? false : true,
+						ParametersType
+			>
+			| FunctionWithMaybeThisParameter<ThisType, ParametersType, ReturnType>
+	: never;
```
benzaria, line 99: why only checking the `Parameters` and not `This` and `ReturnType` as well.
this code below is working fine and passed all tests!

```suggestion
type FunctionOverloadsInternal<
	AllOverloads,
	CheckedOverloads = {},
	MustStopIfOverloadsAreEqual extends boolean = true,
	LastOverload extends UnknownArray = [],
> = AllOverloads extends (
	this: infer ThisType,
	...arguments_: infer ParametersType extends UnknownArray
) => infer ReturnType
	? // This simultaneously checks if the last and the current parameters are equal and `MustStopIfParametersAreEqual` flag is true
	IsEqual<
		[...LastOverload, MustStopIfOverloadsAreEqual],
		[ThisType, ParametersType, ReturnType, true]
	> extends true
		? never
		:
			| FunctionOverloadsInternal<
						// Normally, in `(FunctionType extends (...args: infer P) => infer R)`, compiler infers
						// `P` and `R` from the last function overload.
						// This trick (intersecting one of the function signatures with the full signature)
						// makes compiler infer a last overload that do not equal one of the concatenated ones.
						// Thus, we're ending up iterating over all the overloads from bottom to top.
						// Credits: https://github.com/microsoft/TypeScript/issues/32164#issuecomment-1146737709
						CheckedOverloads & AllOverloads,
						CheckedOverloads & FunctionWithMaybeThisParameter<ThisType, ParametersType, ReturnType>,
						Not<MustStopIfOverloadsAreEqual>,
						[ThisType, ParametersType, ReturnType]
			>
			| FunctionWithMaybeThisParameter<ThisType, ParametersType, ReturnType>
	: never;
```

this will eliminate the current limitation of `this` and `returnType` overloads.

test-d/function-overloads.ts
```diff
@@ -0,0 +1,109 @@
...
+expectType<never>(noOverloads);
+
+declare const anyOverload: FunctionOverloads<any>;
+expectType<(...arguments_: readonly unknown[]) => unknown>(anyOverload);
+
+declare const neverOverload: FunctionOverloads<never>;
+expectType<never>(neverOverload);
+
+declare const unknownOverload: FunctionOverloads<unknown>;
+expectType<never>(unknownOverload);
```
benzaria, line 109: some test for `This` and `ReturnType` overloads

```suggestion
expectType<never>(unknownOverload);

type ThisFunction1 = (this: {a: 1}, foo: string) => void;
type ThisFunction2 = (this: {b: 2}, foo: string) => void;

declare const thisOverloads: FunctionOverloads<ThisFunction1 & ThisFunction2>;
expectType<ThisFunction1 | ThisFunction2>(thisOverloads);

type ReturnFunction1 = (foo: string) => string;
type ReturnFunction2 = (foo: string) => number;

declare const returnOverloads: FunctionOverloads<ReturnFunction1 & ReturnFunction2>;
expectType<ReturnFunction1 | ReturnFunction2>(returnOverloads);
```

source/function-overloads.d.ts
```diff
@@ -0,0 +1,96 @@
+import type {FunctionWithMaybeThisParameter} from './internal/function.d.ts';
+import type {IsEqual} from './is-equal.d.ts';
```
benzaria, line 2: use `UnknownArray` type
https://github.com/andreww2012/type-fest/blob/f91847cd73034ed07ace99d24e18aacfb20c2254/source/function-overloads.d.ts#L69

use `Not` type
https://github.com/andreww2012/type-fest/blob/f91847cd73034ed07ace99d24e18aacfb20c2254/source/function-overloads.d.ts#L90

```suggestion
import type {FunctionWithMaybeThisParameter} from './internal/function.d.ts';
import type {UnknownArray} from './unknown-array.d.ts';
import type {Not} from './internal/type.d.ts';
import type {IsEqual} from './is-equal.d.ts';
```

benzaria (changes requested):

source/internal/function.d.ts
```diff
@@ -0,0 +1,18 @@
+import type {IsUnknown} from '../is-unknown.d.ts';
+
+/**
+Constructs a function type with an optional `this` parameter.
```
som-sm, line 4: Please add some examples.

source/function-overloads.d.ts
```diff
@@ -0,0 +1,96 @@
...
+TypeScript's built-in utility types like `Parameters` and `ReturnType` only work with the last overload signature, [by design](https://github.com/microsoft/TypeScript/issues/32164). This type extracts all overload signatures as a union, allowing you to work with each overload individually.
+
+Use-cases:
+- Extract parameter types from specific overloads using `Extract` and `Parameters`
+- Analyze all possible function signatures in type-level code
+- Extract event handler signatures from framework APIs
+
+Known limitions:
+- Functions that have identical parameters but different `this` types or return types will only extract one overload (the last one)
+- Generic type parameters are lost and inferred as `unknown`
```
som-sm: Add at least one example for each case.

source/function-overloads.d.ts
```diff
@@ -0,0 +1,96 @@
...
+	cancel: [];
+}>();
+
+// Extract the parameter types of the `submit` event:
+import type {ArrayTail, FunctionOverloads} from 'type-fest';
+import HelloWorld from './HelloWorld.vue';
+
+type SubmitEventType = ArrayTail<Parameters<Extract<FunctionOverloads<InstanceType<typeof HelloWorld>['$emit']>, (event: 'submit', ...arguments_: readonly any[]) => void>>>;
+//=> [formData: FormData]
+```
```
som-sm, line 61: Examples should _not_ contain arbitrary code, they should be copy pasteable to the playground.

source/set-return-type.d.ts
```diff
@@ -19,13 +19,9 @@ type MyWrappedFunction = SetReturnType<MyFunctionThatCanThrow, SomeOtherType | u
...
-		// We want to detect this situation just to display a friendlier type upon hovering on an IntelliSense-powered IDE.
-		IsUnknown<ThisArgument> extends true ? (...arguments_: Arguments) => TypeToReturn : (this: ThisArgument, ...arguments_: Arguments) => TypeToReturn
-	) : (
-		// This part should be unreachable, but we make it meaningful just in case…
-		(...arguments_: Parameters<Function_>) => TypeToReturn
-	);
+	Function_ extends (this: infer ThisArgument, ...arguments_: infer Arguments) => any
+		? FunctionWithMaybeThisParameter<ThisArgument, Arguments, TypeToReturn>
+		: // This part should be unreachable, but we make it meaningful just in case…
+		(...arguments_: Parameters<Function_>) => TypeToReturn;
```
som-sm: Looks better if `ThisParameterType` is used:

```suggestion
export type SetReturnType<Function_ extends (...arguments_: any[]) => any, TypeToReturn> =
	FunctionWithMaybeThisParameter<ThisParameterType<Function_>, Parameters<Function_>, TypeToReturn>;
```

source/set-parameter-type.d.ts
```diff
@@ -104,16 +103,17 @@ type HandleLog2 = SetParameterType<HandleMessage, {2: string}>;
...
-			IsUnknown<ThisArgument> extends true
-				? (...arguments_: MergeObjectToArray<Arguments, P>) => ReturnType<Function_>
-				: (this: ThisArgument, ...arguments_: MergeObjectToArray<Arguments, P>) => ReturnType<Function_>
-		)
+		? FunctionWithMaybeThisParameter<
+			ThisArgument,
+			MergeObjectToArray<Arguments, P>,
+			ReturnType<Function_>
+		>
 		: Function_;	// This part should be unreachable
```
som-sm: Looks better if `ThisParameterType` is used:


```suggestion
export type SetParameterType<
	Function_ extends (...arguments_: any[]) => unknown,
	P extends Record<number, unknown>,
> =
	FunctionWithMaybeThisParameter<ThisParameterType<Function_>, MergeObjectToArray<Parameters<Function_>, P>, ReturnType<Function_>>;
```

source/function-overloads.d.ts
andreww2012: When parameters are the same up to `this` or return type, TypeScript would use only one overload. So the fact that `FunctionOverloads` would only return a single overload actually matches the "real" compiler behavior.

With `this` types being different, it chooses the last overload:

```ts
declare const foo: ((this: string) => void) & ((this: number) => string);
foo.call(''); // ts(2345): Argument of type 'string' is not assignable to parameter of type 'number'.
```

With return types being different, it chooses the first overload for some reason:

```ts
declare const foo: (() => string) & (() => number);
const bar = foo(); // string
```

And in this case, `FunctionOverloads` lies about the actual return type:

```ts
type Test = FunctionOverloads<(() => string) & (() => number)>; // () => number
```

So this *might* need to be fixed, but currently this also matches the built-in `ReturnType` behavior:

```ts
type Test = ReturnType<(() => string) & (() => number)>; // number
```

som-sm (changes requested):

source/function-overloads.d.ts
benzaria: I understand your concern about how overload resolution works in TypeScript’s compile-time enforcement, but it’s important to clarify: the **exception behavior you point out** applies to **overload selection during type checking**, not to the **type-level utility** we’re building.

TypeScript will resolve function calls to the first matching overload signature, ignoring differences in return type and sometimes in `this` context. However, this only affects how function calls are checked, not how overloads are declared or represented at the type level.

```ts
declare const foo: ((this: string) => void) & ((this: number) => string);
type T = typeof foo
//   ^? type T = ((this: string) => void) & ((this: number) => string)
// overloads still preserved
```

Our **type utility** is meant for **static analysis and broader use cases**, not just for expressing runtime call behavior. Users might want to use `FunctionOverloads<T>` to:

* Inspect all declared overloads for documentation or introspection
* Metaprogramming and type-level transformations
* Build stricter API transformations or mappings

These use cases go **beyond how the runtime implementation behaves**. So while matching TypeScript’s call resolution is a good baseline, we shouldn’t constrain the type utility only to what the type checker “chooses” for calls.

In short: the overload resolution rules are a valid observation, but they don’t limit what our type utilities **can** represent. If anything, the distinction shows the **need** for a richer type-level union so developers can see all overload signatures, even those that TypeScript’s call resolution might ignore during checking.

@sindresorhus , @som-sm could you share your thoughts on this?

andreww2012: I understand your point too, but IMHO matching the actual compiler behavior is more practical and will be sufficient in 95% of situations. If some of the declared overloads get discarded by the compiler, it's likely the sign of the programmer not knowing about this TypeScript peculiarity.

If anything, we can introduce a type utility option to control whether these "phantom" overloads will be present in the resulting type or not, but I'm not sure what practical example would I use in this option's docs.

source/function-overloads.d.ts
andreww2012: Do you mean the TypeScript playground?

som-sm: Yeah, like the `import HelloWorld from './HelloWorld.vue';` import statement would error if pasted to playground.

test-d/function-overloads.ts
sindresorhus, line 109: Some more tests:
```suggestion
expectType<never>(unknownOverload);

declare function declaredOverload(input: string): {kind: 'string'};
declare function declaredOverload(input: number, flag: boolean): {kind: 'number'};

declare const declaredOverloadUnion: FunctionOverloads<typeof declaredOverload>;
expectType<((input: string) => {kind: 'string'}) | ((input: number, flag: boolean) => {kind: 'number'})>(declaredOverloadUnion);

type ThisOverload1 = (this: Date, foo: string) => void;
type ThisOverload2 = (this: URL, foo: number) => void;
declare const thisOverloadUnion: FunctionOverloads<ThisOverload1 & ThisOverload2>;
expectType<ThisOverload1 | ThisOverload2>(thisOverloadUnion);

type SameParametersDifferentReturn1 = (foo: string) => number;
type SameParametersDifferentReturn2 = (foo: string) => string;
declare const sameParametersDifferentReturn: FunctionOverloads<SameParametersDifferentReturn1 & SameParametersDifferentReturn2>;
expectType<SameParametersDifferentReturn2>(sameParametersDifferentReturn);

declare function genericOverload<T>(input: T): T;
declare function genericOverload(input: string): string;
declare const genericOverloadUnion: FunctionOverloads<typeof genericOverload>;
expectType<((input: unknown) => unknown) | ((input: string) => string)>(genericOverloadUnion);

type InterfaceOverload = {
	(input: string): 1;
	(input: number): 2;
};

declare const interfaceOverloadUnion: FunctionOverloads<InterfaceOverload>;
expectType<((input: string) => 1) | ((input: number) => 2)>(interfaceOverloadUnion);

type SameParametersDifferentThis1 = (this: Date, foo: string) => number;
type SameParametersDifferentThis2 = (this: URL, foo: string) => number;
declare const sameParametersDifferentThis: FunctionOverloads<SameParametersDifferentThis1 & SameParametersDifferentThis2>;
expectType<SameParametersDifferentThis2>(sameParametersDifferentThis);
```

PR #1265, opened 2025-10-10
Add lint rule to validate JSDoc codeblocks using TS compiler

som-sm:
This PR introduces a lint rule that extracts example code blocks from JSDoc comments and validates them using `@typescript/vfs`. It then reports any errors at their corresponding locations.

The errors and their locations are identical to normal TS errors, so the experience feels just like editing a regular TS file.

<br>
image
image

https://github.com/user-attachments/assets/8aef7a11-c7a5-4bca-93d0-1b0df94740de

---

## Working
Linting a single file involves the following steps:

1. **Setting up the environment**
    A virtual TypeScript environment with a single virtual file is created on top of the existing file system. And the virtual environment is created such that it has access to the files on disk. 

    **Location of virtual files**
    While setting up the virtual environment, we specify the root directory. The virtual environment then creates a virtual directory called `vfs` within the root directory, and all virtual files are created relative to this `vfs` directory.

    For example, the following snippet:
    ```ts
    // type-fest/lint-rules/validate-jsdoc-codeblocks.js
    import ts from "typescript";
    import {createFSBackedSystem, createVirtualTypeScriptEnvironment} from "@typescript/vfs";
    
    const FILENAME = 'example-codeblock.ts';
    
    const compilerOptions = {...};
    
    const virtualFsMap = new Map();
    virtualFsMap.set(FILENAME, 'Some code here...');
    
    const rootDir = path.join(import.meta.dirname, '..'); // We go back one level from `lint-rules` dir
    const system = createFSBackedSystem(virtualFsMap, rootDir, ts);
    const env = createVirtualTypeScriptEnvironment(system, [FILENAME], ts, compilerOptions);
    ```
    creates the following directory structure:
    ```
    type-fest/
    ├── package.json
    ├── source/
    │   ├── all-extend.d.ts
    │   ├── ...
    │   └── xor.d.ts
    └── vfs/ (virtual)
        └── example-codeblock.ts  (virtual file)
    ```

    So, if we want to `import` the `Or` type from our `example-codeblock.ts` virtual file, we'd have to use `'../source/or.d.ts'` as the import path, like:
    ```ts
    import type {Or} from '../source/or.d.ts';
    
    const tt: Or<true, true> = true;
    const tf: Or<true, false> = true;
    const ft: Or<false, true> = true;
    const ff: Or<false, false> = false;
    ```

    The location of the virtual file might seem problematic, but in practice it’s not, because our JSDoc example codeblocks never use relative imports. Instead, they always import from `'type-fest'`:
    ```ts
    import type {Or} from 'type-fest';
    
    const tt: Or<true, true> = true;
    const tf: Or<true, false> = true;
    const ft: Or<false, true> = true;
    const ff: Or<false, false> = false;
    ```
    **Note**: Importing directly from `'type-fest'` works fine because Node.js allows [self referencing a package using its name](https://nodejs.org/api/packages.html#self-referencing-a-package-using-its-name).

2. **Running the lint rule on each exported type node**
    - The rule extracts the JSDoc comment corresponding to that node and then extracts the codeblocks within the JSDoc.
    - It iterates over the codeblocks, replaces the contents of the virtual file with the codeblock, and runs TypeScript diagnostics on the virtual file.
    - Finally, it collects any diagnostic errors and reports them back at their exact locations.

---

The lint rule surfaced around ~250 errors (on publicly accessible docs), so this PR also fixes/improves multiple JSDoc code examples. Some common errors that it catches:

- Missing `import` statements
    image

- Floating examples
    image

- References to hypothetical functions/variables
    image

- Duplicate identifiers
    image

- Syntax errors
    image

- Typos
    image
    image

### DIFF
--- a/source/split.d.ts
+++ b/source/split.d.ts
@@ -16,7 +16,7 @@ export type SplitOptions = {
 
 	@example
 	```ts
-	import type {Split} from "type-fest";
+	import type {Split} from 'type-fest';
 
 	type Example1 = Split<`foo.${string}.bar`, '.', {strictLiteralChecks: false}>;
 	//=> ['foo', string, 'bar']


som-sm:
@sindresorhus This PR is still a WIP, because there are still a lot of code blocks that have errors in them, but would appreciate if you could give an early feedback on this.

Couple of questions:
- At several places we have code examples for internal types, like:
  https://github.com/sindresorhus/type-fest/blob/72f491faf221f0fa55a41a1f462f7e7a7809e6f5/source/get.d.ts#L82-L93
  Now, in such cases, we can't have an `import` statement because internal types are not publicly available. 
  
  So, we can either make this rule work only for exported types, or inline the examples like I've done in this change:
  image
  
  Or, maybe we can do something even better, wdyt?

- At times we highlight errors in our codeblocks, but with this lint rule in place, we can't have errors in codeblocks, so I guess it's fine to add an `@ts-expect-error` directive in such cases. This would silence the error and also clearly state the intent that an error is indeed expected. Refer this change:

    
  github com_sindresorhus_type-fest_pull_1265_files (1)


- I think this rule makes sense and is useful, but unfortunately it takes around min and a half to run, which might be a problem. Although, the in editor DX is pretty fast, it's just when it runs on all files together.

  https://github.com/user-attachments/assets/8aef7a11-c7a5-4bca-93d0-1b0df94740de
  
  <br>
  
  > **Performance Consideration**: Line 68 creates a new virtual TypeScript environment for **each** code block. For files with many examples, this could be slow. Consider reusing the environment or batching:
  > 
  > ```js
  > // Current approach creates new env per codeblock
  > const env = createVirtualTypeScriptEnvironment(system, [FILENAME], ts, compilerOptions);
  > ```
  > 
  
  > **Suggestion**: Create one environment per file and update it with new content, rather than recreating for each block.
  
  ~This suggestion isn't possible because you can't update the virtual file once the virtual environment is created.~

  I'm looking into ways for improving this, but let me know if you have some thoughts around this.

som-sm:
> * but unfortunately it takes around min and a half to run, which might be a problem.

Should be fairly fast now, the env creation happens _only_ once now!

Now, the only task left is to fix the remaining 214 errors 😪

som-sm:
> the env creation happens _only_ once now!

No, it _can't_ run just once. 

<br>

> * but unfortunately it takes around min and a half to run, which might be a problem.

The env creation now happens per file instead of the initial per file per code block setup. So, it now takes around a min. Refer [#d974efa](https://github.com/sindresorhus/type-fest/pull/1265/commits/d974efa2b750b100a29137ff1be4d4ec66f89b2d).

sindresorhus:
This is a very useful rule 🙌

sindresorhus:
> Now, in such cases, we can't have an import statement because internal types are not publicly available.

I think we should just ignore the internal types. Many will eventually be exposed and then they can be fixed then.

> At times we highlight errors in our codeblocks, but with this lint rule in place, we can't have errors in codeblocks, so I guess it's fine to add an @ts-expect-error directive in such cases. This would silence the error and also clearly state the intent that an error is indeed expected. Refer this change:

:+1:

> I think this rule makes sense and is useful, but unfortunately it takes around min and a half to run, which might be a problem. Although, the in editor DX is pretty fast, it's just when it runs on all files together.

Doesn't ESLint cache things? Maybe something is not working with the cache.

Alternatively, we could only run this rule in CI.

source/asyncify.d.ts
```diff
@@ -9,22 +9,13 @@ Use-case: You have two functions, one synchronous and one asynchronous that do t
...
-}
-
-type AsyncifiedFooGetter = Asyncify<typeof getFooSync>;
-//=> type AsyncifiedFooGetter = (someArg: SomeType) => Promise<Foo>;
-
-// Same as `getFooSync` but asynchronous.
-const getFooAsync: AsyncifiedFooGetter = (someArg) => {
-	// TypeScript now knows that `someArg` is `SomeType` automatically.
-	// It also knows that this function must return `Promise<Foo>`.
-	// If you have `@typescript-eslint/promise-function-async` linter rule enabled, it will even report that "Functions that return promises must be async.".
```
som-sm, line 24: Changed this example because these three points have nothing to do with the type.

source/is-never.d.ts
```diff
@@ -11,29 +11,41 @@ Useful in type utilities, such as checking if something does not occur.
 ```
 import type {IsNever, And} from 'type-fest';
 
-// https://github.com/andnp/SimplyTyped/blob/master/src/types/strings.ts
-type AreStringsEqual<A extends string, B extends string> =
-	And<
-		IsNever<Exclude<A, B>> extends true ? true : false,
-		IsNever<Exclude<B, A>> extends true ? true : false
```
som-sm, line 18: Removed this example because the primary focus here wasn't on `IsNever`. Added a better example (see below).

source/literal-to-primitive-deep.d.ts
som-sm, line 1: Replaced this example because it had multiple issues, but the updated example conveys a similar idea as the original example.

source/union-to-intersection.d.ts
```diff
@@ -13,34 +13,6 @@ type Intersection = UnionToIntersection<Union>;
...
-		b2: (argB: string) => undefined,
-	}
-}
-
-const union = [new CommandOne(), new CommandTwo()].map(instance => instance.commands);
-type Union = typeof union;
-//=> {a1(): void; b1(): void} | {a2(argA: string): void; b2(argB: string): void}
-
-type Intersection = UnionToIntersection<Union>;
-//=> {a1(): void; b1(): void; a2(argA: string): void; b2(argB: string): void}
```
som-sm, line 41: This example is almost similar to the previous one for this type, so simply removed it instead of replacing it with an alternative.

source/value-of.d.ts
```diff
@@ -5,36 +5,16 @@ Please upvote [this issue](https://github.com/microsoft/TypeScript/issues/31438)
...
 
-onlyBar('foo');
-//=> TypeError ...
+type B = ValueOf<{id: number; name: string; active: boolean}, 'name'>;
+//=> string
 
-onlyBar('bar');
-//=> 2
+type C = ValueOf<{id: number; name: string; active: boolean}, 'id' | 'name'>;
+//=> number | string
```
som-sm, line 17: The existing example had issues, and a simpler example seemed better in this case.

lint-rules/test-utils.js
```diff
@@ -98,3 +100,168 @@ export const createTypeAwareRuleTester = (fixtureFiles, options = {}) => {
...
+
+jsdoc('@example', '```\ntype A = string;\n```', '@category Test');
+// Returns;
+// /**
+// @example
+// ```
+// type A = string;
+// ```
+// @category Test
+// *​/
```
som-sm, line 156: There's a zero-width space character b/w `*` and `/` to prevent `*/` from being interpreted as the end of the JSDoc comment. Because of which there's a `/* eslint-disable no-irregular-whitespace */` comment at the start of this file.

Umm...not sure if there's a better way of achieving this, I tried a couple different solutions but nothing seemed to work.

lint-rules/validate-jsdoc-codeblocks.test.js
```diff
@@ -0,0 +1,596 @@
...
+		exportTypeAndOption(jsdoc(fence(code1, 'ts'))),
+		exportTypeAndOption(jsdoc(fence(code1, 'typescript'))),
+
+		// Multiple code blocks
+		exportTypeAndOption(
+			jsdoc('@example', fence(code1, 'ts'), '\nSome text in between.\n', '@example', fence(code2)),
+		),
+
+		// Multiple exports and multiple properties
+		exportTypeAndOption(jsdoc(fence(code1)), jsdoc(fence(code2))),
```
som-sm, line 119: These utils help make the test cases terse, otherwise this already lengthy file would have been even lengthier.

For example, the following 
```ts
exportTypeAndOption(jsdoc(fence(code1)), jsdoc(fence(code2)))
```

evaluates to:
````ts
/**
```
type A = string;
```
*/
export type T0 = string;

/**
```
type B = number;
```
*/
export type T1 = string;

export type TOptions = {
        /**
        ```
        type A = string;
        ```
        */
        p0: string;

        /**
        ```
        type B = number;
        ```
        */
        p1: string;
};
````

File: lint-rules/validate-jsdoc-codeblocks.test.js [hunk 7]
```diff
@@ -0,0 +1,596 @@
...
+					/**
+					Some description.
+					\`\`\`ts
+					type A = Subtract<1, 2>;
+					\`\`\`
+					@category Test
+					*/
+					p0: string;
+				};
+			`,
```
som-sm, line 169: For invalid tests, the code is _not_ generated via some utility, as it is for valid cases, because here we want to validate the exact line and column numbers where the errors appear, and if the code were obfuscated behind some utility, then those numbers would feel magical/arbitrary, reducing the overall readability and maintainability of the test cases.

lint-rules/validate-jsdoc-codeblocks.test.js
```diff
@@ -0,0 +1,596 @@
...
+			errors: [
+				errorAt({line: 4, textBeforeStart: 'type A = ', target: 'Subtract'}),
+				errorAt({line: 14, textBeforeStart: '\ttype A = ', target: 'Subtract'}),
```
som-sm, line 172: For errors, we use a utility because manually typing the `column` and `endColumn` values is error-prone and also less readable.

For example, in the following test
````ts
{
	code: dedenter`
		/**
		@example
		\`\`\`ts
		type NoImport = Add<1, 2>;
		\`\`\`
		*/
		export type Add = number;
	`,
	errors: [
		{messageId: 'invalidCodeblock', line: 4, column: 17, endLine: 4, endColumn: 20},
	],
},
````
the numbers `17` and `20` feel magical/arbitrary and it's very easy to mess up those numbers and end up with false-positive tests.

It's much more readable if the `column` and `endColumn` values are generated like this:
```ts
errors: [
	{
		messageId: 'invalidCodeblock',
		line: 4,
		column: 'type NoImport = '.length + 1,
		endLine: 4,
		endColumn: 'type NoImport = '.length + 1 + 'Add'.length,
	},
],
```
But, this is still repetitive and has scope for simplification.

And, that's what the `errorAt` utility simplifies:
```ts
errors: [
	errorAt({
		line: 4,
		textBeforeStart: 'type NoImport = ',
		target: 'Add',
	}),
],
```

**Note**: The `line` and `endLine` values are manually entered because those are pretty straightforward. And `errorAt` uses the `line` value if an `endLine` is not explicitly passed, which also reduces the repetition a bit.

som-sm (commented):
I've added comments wherever I felt it needed some explanation. Hope it makes the review process a tad bit easier.

source/is-unknown.d.ts
```diff
@@ -11,34 +11,23 @@ Useful in type utilities, such as when dealing with unknown data from API calls.
 ```
 import type {IsUnknown} from 'type-fest';
 
-// https://github.com/pajecawav/tiny-global-store/blob/master/src/index.ts
```
som-sm, line 14: The existing example had several issues, and it wasn’t really instantiating `IsUnknown` with `unknown`. Replaced it with a simpler example.

som-sm:
@sindresorhus This PR is finally ready for review. It took some effort to get it to the finish line, but I’m happy with how it turned out. I’ve also updated the PR description to include a detailed explanation of how this lint rule works.

The only changes within the `source` directory are inside JSDoc example code blocks. I’ve fixed those examples using the following approach:

1. If an example was easily fixable (e.g., missing import, typo, or floating snippet), I simply made the fix without checking the logic or relevance of the example.

2. If an example contained errors that weren’t easily fixable, I improved it from the ground up. I’ve added comments wherever I made such changes.

som-sm:
> * At times we highlight errors in our codeblocks, but with this lint rule in place, we can't have errors in codeblocks, so I guess it's fine to add an `@ts-expect-error` directive in such cases. This would silence the error and also clearly state the intent that an error is indeed expected. Refer this change:

**NOTE**: Using the `@` symbol (`@ts-expect-error`) inside JSDoc codeblocks breaks rendering. There's already an [open issue](https://github.com/microsoft/TypeScript/issues/47679#issuecomment-3495389667) for the same.

image

lint-rules/test-utils.js
som-sm: > Because of which there's a `/* eslint-disable no-irregular-whitespace */` comment at the start of this file.

Now, the rule is disabled _only_ for comments.
https://github.com/sindresorhus/type-fest/blob/657dd4dd8b54d32d6cc0ca75d9ec6a492c725d59/xo.config.js#L88-L98

sindresorhus:
This is super nice! 🎉 And seeing all the changes to code examples, it was really needed. It's hard to manually ensure code examples are correct.

PR #1030, opened 2025-01-02
`SetRequired`: Fix support for removal of optional modifiers from tuples

som-sm:
This PR:
1. Fixes `SetRequired`, enabling it to remove optional modifiers from tuples, for eg.,

     ```ts
     SetRequired<[number?, number?], 0> //=> [number, number?]
     ```
     This fix will allow us to solve #1026.
2. Fixes minor issues with `IsArrayReadonly` and adds tests for it.
3. Adds an internal `IfArrayReadonly` type.

PR #1266, opened 2025-10-11
Add `ArrayReverse` type

som-sm:
This PR fixes [this issue](https://github.com/sindresorhus/type-fest/pull/1167#discussion_r2422533832) in #1167. I was playing around with the fix and had done most of the work in the process, so I thought I’d open a separate PR.

Credits to @benzaria for suggesting this type.

`Reverse` works normally when the input array doesn't contain optional elements, but if the array includes an optional element, the result splits into a union of separate tuples, like:
https://github.com/sindresorhus/type-fest/blob/1c2e057790fd851e339ead90e72710f6be27e4a7/source/array-reverse.d.ts#L30-L50

---

I've kept the name as `ArrayReverse` instead of just `Reverse` to match with our existing `Array*` types (like `ArraySlice`, `ArraySplice`).

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,9 +25,9 @@
 		"node": ">=20"
 	},
 	"scripts": {
-		"test:tsc": "node --max-old-space-size=5120 ./node_modules/.bin/tsc",
-		"test:tsd": "tsd",
-		"test:xo": "xo",
+		"test:tsc": "node --max-old-space-size=6144 ./node_modules/.bin/tsc",
+		"test:tsd": "node --max-old-space-size=6144 ./node_modules/.bin/tsd",
+		"test:xo": "node --max-old-space-size=6144 ./node_modules/.bin/xo",
 		"test:linter": "node --test",
 		"test": "run-p test:*"
 	},
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,4 +42,4 @@ jobs:
           node-version: 24
       - run: npm install
       - run: npm install typescript@${{ matrix.typescript-version }}
-      - run: NODE_OPTIONS="--max-old-space-size=5120" npx tsc
+      - run: NODE_OPTIONS="--max-old-space-size=6144" npx tsc
--- a/.github/workflows/ts-canary.yml
+++ b/.github/workflows/ts-canary.yml
@@ -23,4 +23,4 @@ jobs:
       - run: npm install typescript@${{ matrix.typescript-version }}
       - name: show installed typescript version
         run: npm list typescript --depth=0
-      - run: NODE_OPTIONS="--max-old-space-size=5120" npx tsc
+      - run: NODE_OPTIONS="--max-old-space-size=6144" npx tsc


som-sm:
For inputs containing only optional elements, I originally wasn't splitting the result, for example, `ArrayReverse<[1?, 2?, 3?]>`  was returning `[3?, 2?, 1?]`. While this type of output is accurate, it's not precise, because `[3?, 2?, 1?]` could be `[3]` | `[3, 2]` but reversing `[1?, 2?, 3?]` would never yield `[3]` or `[3, 2]`.

som-sm:
@benzaria Would appreciate if you could share your thoughts on this!

source/array-reverse.d.ts
```diff
@@ -0,0 +1,84 @@
...
+		? TArray extends readonly [...infer Rest, infer Last]
+			? _ArrayReverse<Rest, BeforeRestAcc, [...AfterRestAcc, Last], Result> // Accumulate elements that are present after the rest element in reverse order.
+			: Result | [...AfterRestAcc, ...TArray, ...BeforeRestAcc] // Add the rest element between the accumulated elements.
+		: TArray extends readonly [(infer First)?, ...infer Rest]
+			? If<IsOptionalKeyOf<TArray, '0'>,
+				_ArrayReverse<
+					Rest,
+					[First | (If<IsExactOptionalPropertyTypesEnabled, never, undefined>), ...BeforeRestAcc], // Add `| undefined` for optional elements, if `exactOptionalPropertyTypes` is disabled.
+					AfterRestAcc,
+					Result | BeforeRestAcc
```
benzaria, line 79: Thanks for taking the time to revisit #1167.

I took a close look at the changes, and from what I can see, the only substantial additions are the `Result | BeforeRestAcc` and `Result | [...AfterRestAcc, ...TArray, ...BeforeRestAcc]` parts. These seem to slightly extend the existing behavior rather than alter the core logic.

My initial PR already established the recursive reversal mechanism and accumulator handling correctly, so this new version mostly builds directly on top of that implementation.

If the goal is simply to add these additional result variants, it would make more sense to integrate those improvements into the existing PR instead of discarding it. That way, we preserve authorship and continuity while still merging the refinements.

I’m happy to update my PR to include these tweaks if that helps move it forward.

Also worth mentioning — my implementation had the `| undefined` part because of the `preserveOptionalModifier` option, which adds `?` to elements even when they follow a required one. That was an oversight on my part, but it can easily be adjusted to include the behavior you suggested.

test-d/array-reverse.ts
```diff
@@ -0,0 +1,68 @@
...
+expectType<[6, 5, [3, 4], [1, 2]]>({} as ArrayReverse<[[1, 2], [3, 4], 5, 6]>);
+expectType<[Array<{y: string}>, Array<{x: string}>]>({} as ArrayReverse<[Array<{x: string}>, Array<{y: string}>]>);
+expectType<[3, [1, 2]] | [[1, 2]] | []>({} as ArrayReverse<[[1, 2]?, 3?]>);
+
+// Unions
+expectType<[3, 2, 1] | readonly ['b', 'a']>({} as ArrayReverse<[1, 2, 3] | readonly ['a', 'b']>);
+expectType<readonly [number, string] | number[]>({} as ArrayReverse<readonly [string, number] | number[]>);
+expectType<readonly [3, 2, 1] | readonly [2, 1] | readonly [1] | ['b', 'a'] | ['a'] | [] | [number, string]>(
+	{} as ArrayReverse<readonly [1, 2?, 3?] | ['a'?, 'b'?] | [string, number]>,
+);
```
benzaria, line 54: The current implementation has a few issues in how it handles optional elements and rest parameters:

```ts
ArrayReverse<[1?, 2?]> => [2, 1] | [1] | []
// This could have been [2?, 1?], which is simpler.

ArrayReverse<[1?, 2?, 3?, ...number[]]> => [...number[], 3, 2, 1] | [2, 1] | [1] | []
// It should be [...number[], 3, 2, 1] | [3, 2, 1] | [2, 1] | [1] | [].

ArrayReverse<[number, string?, ...string[]]> => readonly [...string[], string, number] | readonly [number]
// It should be readonly [...string[], string, number] | readonly [string, number] | readonly [number].

// It seems that when a rest element follows an optional element, valid cases are missing.
```

source/array-reverse.d.ts
som-sm: All recent array types are mostly similar to #1030, so they all look the same.

> I’m happy to update my PR to include these tweaks if that helps move it forward.

Yeah, please go ahead, we can close this PR and move ahead with yours if that feels better. But, please make sure the tests, the docs, and everything are done properly. I had already done all of that, so it felt like it'd be faster to open a separate PR.

test-d/array-reverse.ts
som-sm: > ```ts
> ArrayReverse<[1?, 2?]> => [2, 1] | [1] | []
> // This could have been [2?, 1?], which is simpler.
> ```

Refer https://github.com/sindresorhus/type-fest/pull/1266#issuecomment-3393452518.

<br>

> ```ts
> ArrayReverse<[1?, 2?, 3?, ...number[]]> => [...number[], 3, 2, 1] | [2, 1] | [1] | []
> // It should be [...number[], 3, 2, 1] | [3, 2, 1] | [2, 1] | [1] | [].
> ```

`[...number[], 3, 2, 1] | [3, 2, 1]` can be merged to _just_ `[...number[], 3, 2, 1]`. `[3, 2, 1]` is essentially a subset of `[...number[], 3, 2, 1]`.

<br>

> ```ts
> ArrayReverse<[number, string?, ...string[]]> => readonly [...string[], string, number] | readonly [number]
> // It should be readonly [...string[], string, number] | readonly [string, number] | readonly [number].
> ```

Same here.

source/array-reverse.d.ts
som-sm: > That way, we preserve authorship

I would have anyways added you as a co-author in this PR, so you didn't have to worry about authorship.

<br>

Added in [701b412](https://github.com/sindresorhus/type-fest/pull/1266/commits/701b412cdf0c9d647ac60d34a685fd2739b8a8bb).

som-sm:
@sindresorhus While adding test for long tuples, I realised my implementation was not tail recursive because I had simplified a conditional using the `If` type. While this mistake is obvious, it might get missed. Refer [#6cc3e53](https://github.com/sindresorhus/type-fest/pull/1266/commits/6cc3e53877ab7d9b4af03766bd8323703d615a42).

So, does it make sense to add a note like the following to the `If` type or is it too trivial?

Note: Sometimes using the `If` type can make an implementation non tail recursive, which would affect the performance of the implementation. In such cases, it's better to use a conditional directly. Refer the following example:

```ts
import type {If, IsEqual, StringRepeat} from 'type-fest';

// The following implementation is not tail recursive
type Includes<S extends string, Char extends string> =
	S extends `${infer First}${infer Rest}`
		? If<IsEqual<First, Char>,
			'found',
			Includes<Rest, Char>>
		: 'not found';

// Hence, instantiations with long strings will fail
type HundredZeroes = StringRepeat<'0', 100>;

// @ts-expect-error
type Fails = Includes<HundredZeroes, '1'>;
//           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
// Error: Type instantiation is excessively deep and possibly infinite.

// However, if we use a simple conditional instead of `If`, the implementation becomes tail-recursive
type IncludesWithoutIf<S extends string, Char extends string> =
	S extends `${infer First}${infer Rest}`
		? IsEqual<First, Char> extends true
			? 'found'
			: IncludesWithoutIf<Rest, Char>
		: 'not found';

// Now, instantiations with long strings will work
type Works = IncludesWithoutIf<HundredZeroes, '1'>;
//=> 'not found'
```

sindresorhus:
> So, does it make sense to add a note like the following to the If type or is it too trivial?

Definitely makes sense 👍

package.json
```diff
@@ -25,9 +25,9 @@
 		"node": ">=20"
 	},
 	"scripts": {
-		"test:tsc": "node --max-old-space-size=5120 ./node_modules/.bin/tsc",
-		"test:tsd": "tsd",
-		"test:xo": "xo",
+		"test:tsc": "node --max-old-space-size=6144 ./node_modules/.bin/tsc",
+		"test:tsd": "node --max-old-space-size=6144 ./node_modules/.bin/tsd",
+		"test:xo": "node --max-old-space-size=6144 ./node_modules/.bin/xo",
```
som-sm, line 30: So, the `test:tsd` script now requires some bump in memory. `test:xo` works fine without a bump when run independently, but when run in parallel through `run-p`, apparently it also needs some extra memory.

sindresorhus:
Is this ready?

som-sm:
> Is this ready?

@sindresorhus Yeah, it's ready!

I think we can proceed with this PR and close out #1167, since it still has some unresolved issues. This PR already credits it, so it should be fine to move forward.

PR #1178, opened 2025-06-08
Union support for `Simplify`

benzaria:
Add `Simplify` support for unions

 - fix: `Simplify` return `{}` for functions.
 - feat: refactor `ConditionalSimplify` and `ConditionalSimplifyDeep` types for Union support and improved clarity.
 - feat: add new tests covering these changes.
 - feat: remove unnecessary interface in `ConditionalSimplifyDeep` test file (TODO)
 - doc: improve jsDoc for `Simplify` and `SimplifyDeep`

PR #1268, opened 2025-10-12
add `SimplifyUnion` type

### LINKED ISSUE #1174: `SimplifyUnion`

A type that simplify a union. When merging 2 of more unions TS hide the underline types.
use case:
```ts
type U1 = 5 | ['b'] | symbol
type U2 = null | '2' | {foo: 'bar'}

type U3 = U1 | U2
//	  ^? U3 = U1 | U2 

type U4 = SimplifyUnion<U3>
//	  ^? U4 = symbol | 5 | "2" | ["b"] | {foo: "bar"} | null
```

### Type source

```ts
type SimplifyUnion<T> = {[K in T as string]: K}[string]
```

benzaria:
closes #1174

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -147,7 +147,7 @@ Click the type names for complete docs.
 - [`Entries`](source/entries.d.ts) - Create a type that represents the type of the entries of a collection.
 - [`SetReturnType`](source/set-return-type.d.ts) - Create a function type with a return type of your choice and the same parameters as the given function type.
 - [`SetParameterType`](source/set-parameter-type.d.ts) - Create a function that replaces some parameters with the given parameters.
-- [`Simplify`](source/simplify.d.ts) - Flattens a complex object type to make it more readable and improve editor IntelliSense. And also to transform an interface into a type to aide with assignability.
+- [`Simplify`](source/simplify.d.ts) - Flattens a complex object type to make it more readable and improve editor IntelliSense. And also to transform an interface into a type to aid with assignability.
 - [`SimplifyDeep`](source/simplify-deep.d.ts) - Deeply simplifies an object type.
 - [`SimplifyUnion`](source/simplify-union.d.ts) - Flattens a complex union type to make it more readable and improve editor IntelliSense.
 - [`Get`](source/get.d.ts) - Get a deeply-nested property from an object using a key path, like [Lodash's `.get()`](https://lodash.com/docs/latest#get) function.
--- a/source/simplify.d.ts
+++ b/source/simplify.d.ts
@@ -2,7 +2,7 @@ import type {ConditionalSimplify} from './conditional-simplify.d.ts';
 import type {NonRecursiveType} from './internal/type.d.ts';
 
 /**
-Flattens a complex object type to make it more readable and improve editor IntelliSense. And also to transform an interface into a type to aide with assignability.
+Flattens a complex object type to make it more readable and improve editor IntelliSense. And also to transform an interface into a type to aid with assignability.
 
 @example
 ```


test-d/simplify-union.ts
```diff
@@ -0,0 +1,14 @@
...
+type U2 = (() => void) | {[x: string]: number; b: number};
+
+type UnSimplifiedUnion = U1 | U2;
+//=> U1 | U2 <- obfuscated, hovering over the type doesn't show the underline types in the union.
+
+type SimplifiedUnion = SimplifyUnion<UnSimplifiedUnion>;
+//=> 0 | "a" | null | ['b', 1n] | (() => void) | {[x: string]: number; b: number}
+
+expectType<SimplifiedUnion>({} as UnSimplifiedUnion);
+// Checking all types are preserved.
```
sindresorhus: Weird comment placement.

source/simplify-union.d.ts
```diff
@@ -0,0 +1,27 @@
...
+//=> U1 | U2 <- obfuscated, hovering over the type doesn't show the underlying types in the union.
+
+type SimplifiedUnion = SimplifyUnion<UnSimplifiedUnion>;
+//=> 0 | "a" | null | ['b', 1n] | (() => void) | {[x: string]: number; b: number}
+```
+
+@see {@link Simplify}
+@category Union
+*/
+export type SimplifyUnion<Union> = {[Member in Union as typeof tag]: Member}[typeof tag];
```
som-sm: Doesn't seem to work if you pass the union directly instead of storing it in a type alias first.

```ts
type U1 = {a: 1};
type U2 = {b: 2};

type UnSimplifiedUnion = U1 | U2;
//=> U1 | U2 <- Not simplified

type SimplifiedUnion = SimplifyUnion<U1 | U2>;
//=> U1 | U2 <- Still, not simplified
```

benzaria: weird, i will look into that!

benzaria: @som-sm It seems the issue is that `U1` and `U2` aren’t unions themselves — they’re being recognized as standalone types.

```ts
type U1 = {a: 1};
declare const u1: U1;
// => Displays as `const u1: U1` instead of `const u1: {a: 1}`
```

To properly demonstrate the effect, the example should use actual unions, as shown below:

```ts
type U1 = {a: 1} | {c: 3};
type U2 = {b: 2} | 'a';

type UnSimplifiedUnion = U1 | U2;
// => U1 | U2 (not simplified)

type SimplifiedUnion = SimplifyUnion<U1 | U2>;
// => {a: 1} | {c: 3} | {b: 2} | "a"
```

som-sm: > It seems the issue is that `U1` and `U2` aren’t unions themselves

But `U1 | U2` is a union, so it should work ideally.

The existing `Simplify` type that we have works:
```ts
type U1 = {a: 1};
type U2 = {b: 2};

type UnSimplifiedUnion = U1 | U2;
//=> U1 | U2 <- Not simplified

type SimplifiedUnion = Simplify<U1 | U2>;
//=> {a: 1} | {b: 2} <- Simplified
```

<br>

@benzaria What problem does `SimplifyUnion` solve that `Simplify` doesn't? Is it just that the former handles functions and latter doesn't, if so, this should work:
```ts
type Simplify<T> = T extends Function ? T : {[P in keyof T]: T[P]};
```

benzaria: Actually, my first attempt was the one below, which I originally made for #1178:

```ts
type Simplify<Type> = ConditionalSimplify<
	Type,
	NonRecursiveType | Set<unknown> | Map<unknown, unknown>,
	object
>;
```

It effectively does the same as your suggestion. However, what ended up happening is that some types depending on `Simplify` started to break — such as `MergeDeep`, `ApplyDefaultOptions`, and other types that rely on it.

That led me down a bit of a rabbit hole trying to fix `MergeDeep` and `ApplyDefaultOptions`. I was able to get `ApplyDefaultOptions` working again, but `MergeDeep` still needs a refactor, since it uses a fairly complex merge strategy for options instead of leveraging `ApplyDefaultOptions`.

My suggestion for now is to introduce both an internal and external version of `Simplify`, for better separation of use cases. This way, we can address the current limitation in `Simplify` without breaking half of the library:

```ts
export type _Simplify<T> = { [K in keyof T]: T[K] } & {};

export type Simplify<Type> = ConditionalSimplify<
	Type,
	NonRecursiveType | Set<unknown> | Map<unknown, unknown>,
	object
>;
```
@sindresorhus @som-sm What are your thoughts on this approach?

PR #1274, opened 2025-10-17
`ExcludeRestElement`: Fix generic assignability with arrays

som-sm:
Currently, instantiations of `ExcludeRestElement` with generic types are _not_ assignable to `readonly unknown[]`, this PR fixes this issue. 

This issue is also related to https://github.com/sindresorhus/type-fest/pull/1166#discussion_r2421413812. In general, it's better to handle `any` and `never` cases explicitly.

### DIFF
--- a/test-d/extract-rest-element.ts
+++ b/test-d/extract-rest-element.ts
@@ -1,5 +1,5 @@
 import {expectType} from 'tsd';
-import type {ExtractRestElement} from '../index.d.ts';
+import type {ExtractRestElement, TupleOf} from '../index.d.ts';
 
 // Leading rest element
 expectType<ExtractRestElement<[...string[], 1]>>({} as string);
@@ -49,3 +49,10 @@ expectType<ExtractRestElement<[['a'], ...boolean[], ['z']]>>({} as boolean);
 // Edge: `never` / `any`
 expectType<ExtractRestElement<any>>({} as any);
 expectType<ExtractRestElement<never>>({} as never);
+
+// Long tuples
+type FiftyZeroes = TupleOf<50, '0'>;
+expectType<ExtractRestElement<[...FiftyZeroes, ...number[]]>>({} as number);
+
+type NineHundredNinetyNineZeroes = TupleOf<999, '0'>;
+expectType<ExtractRestElement<[...NineHundredNinetyNineZeroes, ...number[]]>>({} as number);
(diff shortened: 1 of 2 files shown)


source/exclude-rest-element.d.ts
```diff
@@ -26,13 +27,14 @@ type T4 = ExcludeRestElement<[number, string]>;
...
-export type ExcludeRestElement<Array_ extends UnknownArray> =
+export type ExcludeRestElement<Array_ extends UnknownArray> = IfNotAnyOrNever<Array_,
 	SplitOnRestElement<Array_> extends infer Result
 		? Result extends readonly UnknownArray[]
 			? IsArrayReadonly<Array_> extends true
 				? Readonly<[...Result[0], ...Result[2]]>
 				: [...Result[0], ...Result[2]]
-			: Result
-		: never;
+			: never
```
som-sm, line 36: If this is `Result` (and not `never`), then the compiler thinks the type produces something more than _just_ arrays.

source/exclude-rest-element.d.ts
```diff
@@ -26,13 +27,14 @@ type T4 = ExcludeRestElement<[number, string]>;
...
+		: never
+>;
```
benzaria, line 38: Please consider using `IsAnyOrNever` instead, as it provides more reliable handling for recursive types.
Refer my [#1192 comment](https://github.com/sindresorhus/type-fest/pull/1192#issuecomment-2984276375), and your [#1266 comment](https://github.com/sindresorhus/type-fest/pull/1266#issuecomment-3394512153) regarding `If*` types.

```suggestion
export type ExcludeRestElement<Array_ extends UnknownArray> =
	IsAnyOrNever<Array_> extends true ? Array_
		: SplitOnRestElement<Array_> extends infer Result
			? Result extends readonly UnknownArray[]
				? IsArrayReadonly<Array_> extends true
					? Readonly<[...Result[0], ...Result[2]]>
					: [...Result[0], ...Result[2]]
				: never
			: never;
```

som-sm: `IfNotAnyOrNever` works perfectly fine in this case because `ExcludeRestElement` is _not_ a recursive type. Added test cases for long tuples in [#0aed831](https://github.com/sindresorhus/type-fest/pull/1274/commits/0aed831557ca7882123bcee1bc67c8bae13d5686).

<br>

> your [#1266 comment](https://github.com/sindresorhus/type-fest/pull/1266#issuecomment-3394512153) regarding `If*` types.

This was _only_ for recursive types.

som-sm: And, if used properly, `IfNotAnyOrNever` can even be used with recursive types without any problems. Added a note in #1276.

PR #1276, opened 2025-10-17
`IfNotAnyOrNever`: Add note regarding tail recursion

som-sm:
Adds a note regarding how to properly use the `IfNotAnyOrNever` type with recursive types.


source/if.d.ts
```diff
@@ -69,8 +71,6 @@ type Includes<S extends string, Char extends string> =
 		: 'not found';
 
 // Hence, instantiations with long strings will fail
-type HundredZeroes = StringRepeat<'0', 100>;
```
som-sm, line 72: This felt like the `// Hence, instantiations with long strings will fail` comment applies to `type HundredZeroes = StringRepeat<'0', 100>;
`.

PR #1277, opened 2025-10-18
improve `Simplify` type

benzaria:
- Enhanced `Simplify` type to improve flattening and support of union/function/classes
- Added `ExtractCallSignature` internal type to extract call signatures from function types.
- Moved `HasMultipleCallSignatures` internal type to `internal/function.d.ts`.
- Updated `SetOptional`, `SetReadonly`, and `SetRequired` types to utilize `ExtractCallSignature` and `Simplify`.
- Updated tests to reflect changes in type behavior and ensure correctness.

### DIFF
--- a/source/required-deep.d.ts
+++ b/source/required-deep.d.ts
@@ -43,8 +43,8 @@ Note that types containing overloaded functions are not made deeply required due
 @category Set
 @category Map
 */
-export type RequiredDeep<T> =
-T extends BuiltIns ? T
+export type RequiredDeep<T> = T extends BuiltIns
+	? T
 	: T extends Map<infer KeyType, infer ValueType>
 		? Map<RequiredDeep<KeyType>, RequiredDeep<ValueType>>
 		: T extends Set<infer ItemType>


benzaria:
One thing to note is that when passing a function, the resulting type becomes `function & {}`. Any attempt to refine or eliminate the `& {}` intersection causes type errors in `MergeDeep` and `ApplyDefaultOptions`, as discussed [here](https://github.com/sindresorhus/type-fest/pull/1268#discussion_r2427378129).

This seems to be the best solution so far, at least until we find a proper fix for `MergeDeep`.

source/partial-deep.d.ts
```diff
@@ -97,8 +98,8 @@ partialShape.dimensions = [15]; // OK
 export type PartialDeep<T, Options extends PartialDeepOptions = {}> =
 	_PartialDeep<T, ApplyDefaultOptions<PartialDeepOptions, DefaultPartialDeepOptions, Options>>;
 
-type _PartialDeep<T, Options extends Required<PartialDeepOptions>> = T extends BuiltIns | ((new (...arguments_: any[]) => unknown))
-	? T
+type _PartialDeep<T, Options extends Required<PartialDeepOptions>> =
+T extends BuiltIns | ((new (...arguments_: any[]) => unknown)) ? T
```
som-sm: There's no change here, just formatting change, remove this.

source/required-deep.d.ts
```diff
@@ -42,8 +43,8 @@ Note that types containing overloaded functions are not made deeply required due
 @category Set
 @category Map
 */
-export type RequiredDeep<T> = T extends BuiltIns
-	? T
+export type RequiredDeep<T> =
+T extends BuiltIns ? T
```
som-sm: Same here, no change, remove this.

source/internal/function.d.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+Extract the call signature of an object type.
+*/
+export type ExtractCallSignature<Type> =
+	Type extends (this: infer This, ...args: infer Parameters) => infer Return
+		? IsUnknown<This> extends true // TODO: replace with `FunctionWithMaybeThisParameter`
+			? (...args: Parameters) => Return
+			: (this: This, ...args: Parameters) => Return
+		: unknown;
```
som-sm, line 28: Returning `unknown` for non-functions doesn't seem correct in general. It shouldn't be designed just for the specific use case in `Simplify` (`ExtractCallSignature<Type> & _Simplify<Type>`).

sindresorhus:
Bump :)

PR #1105, opened 2025-04-13
Add `ArrayAt` type

giao66:
closed #369

PR #1279, opened 2025-10-19
Add `ArrayAt` type

### LINKED ISSUE #369: Add `ArrayAt` type for stricter `Array.prototype.at()` with tuples
See https://github.com/sindresorhus/ts-extras/issues/39 and https://github.com/microsoft/TypeScript/issues/47660.

```ts
const tuple = ['abc', 123, true] as const;

type First = ArrayAt<typeof tuple, 0>; // 'abc'
type Last = ArrayAt<typeof tuple, -1>; // true
type SecondToLast = ArrayAt<typeof tuple, -2>; // 123
type ThirdToLast = ArrayAt<typeof tuple, -3>; // 'abc'
type OutOfBounds = ArrayAt<typeof tuple, 999>; // should not compile since 999 does not extend 0 | 1 | 2 | -1 | -2 | -3
```

```ts
const array = ['abc', 123, true];

type First = ArrayAt<typeof tuple, 0>; // string | number | true
type Last = ArrayAt<typeof tuple, -1>; // string | number | true
type OutOfBounds = ArrayAt<typeof tuple, 999>; // string | number | true
```

som-sm:
This PR proposes a significantly simpler implementation for #1105. Closes #369.

Opened a new PR as I didn’t want to overwrite the existing one, since it introduces several helper types that might be useful in general. We could potentially repurpose that PR to expose some of those utility types.

### DIFF
--- a/source/array-at.d.ts
+++ b/source/array-at.d.ts
@@ -1,15 +1,13 @@
-import type {ExcludeRestElement} from './exclude-rest-element.d.ts';
-import type {ExtractRestElement} from './extract-rest-element.d.ts';
 import type {If} from './if.d.ts';
 import type {IntRange} from './int-range.d.ts';
+import type {NumberAbsolute} from './internal/numeric.d.ts';
 import type {IsExactOptionalPropertyTypesEnabled} from './internal/type.d.ts';
-import type {IsOptionalKeyOf} from './is-optional-key-of.d.ts';
+import type {LessThanOrEqual} from './less-than-or-equal.d.ts';
 import type {LessThan} from './less-than.d.ts';
 import type {IsNegative} from './numeric.d.ts';
 import type {SplitOnRestElement} from './split-on-rest-element.d.ts';
 import type {Subtract} from './subtract.d.ts';
 import type {Sum} from './sum.d.ts';
-import type {TupleOf} from './tuple-of.d.ts';
 import type {UnknownArray} from './unknown-array.d.ts';
 
 /**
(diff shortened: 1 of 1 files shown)


sindresorhus:
Can you summarize how it's simpler? Any downsides?

som-sm:
> Can you summarize how it's simpler? Any downsides?

Yeah yeah, I'll add all the details once I make the PR "Ready for review".

andr3medeiros:
I'm a bit confused by the documentation: It shows some results as union, which it would not be the real translation of JS Array.at right?

Is the below example something similar to what you want to achieve?

```ts
import { ArraySlice, GreaterThan, IsNegative, LastArrayElement } from 'type-fest';

export type IsPositive<N extends number> = IsNegative<N> extends true ? false : true;
type MakePositive<N extends number> =
	IsPositive<N> extends true ? N : `${N}` extends `${infer _}${infer R extends number}` ? R : N;
	
type ArrayAt<TArray extends readonly unknown[], TIndex extends number> =
	IsPositive<TIndex> extends true
		? TArray[TIndex]
		: GreaterThan<number & MakePositive<TIndex>, TArray['length']> extends true
			? undefined
			: ArraySlice<TArray, TIndex>[0];

type Arr = [1, 2 | undefined, 3];

type Test = ArrayAt<Arr, 1>; // 2 | undefined
type Test2 = ArrayAt<Arr, -4>; // undefined
type Test3 = ArrayAt<Arr, -3>; // 1
type Test4 = ArrayAt<Arr, -2>; // 2 | undefined
type Test5 = ArrayAt<Arr, -1>; // 3
type Test6 = ArrayAt<Arr, 4>; // undefined
```

som-sm:
> I'm a bit confused by the documentation: It shows some results as union, which it would not be the real translation of JS Array.at right?

@andr3medeiros For array types that contain optional elements or rest element, the answer could be a union because the type could hold different values at runtime. For example, the array type `['a', 'b', 'c', ...number[], 'd']` could hold values like the following at runtime:
```ts
['a', 'b', 'c', 'd']
['a', 'b', 'c', 1, 'd']
['a', 'b', 'c', 1, 2, 'd']
['a', 'b', 'c', 1, 2, 3, 'd']
```

Now, if you index `-3` from `['a', 'b', 'c', ...number[], 'd']`, the result has to be a union of all possible outcomes, like:

```ts
['a', 'b', 'c', 'd'].at(-3) // 'b'
['a', 'b', 'c', 1, 'd'].at(-1) // 'c'
['a', 'b', 'c', 1, 2, 'd'].at(-1) // 1
['a', 'b', 'c', 1, 2, 3, 'd'].at(-1) // 2
```

So, therefore the final result of `ArrayAt<['a', 'b', 'c', ...number[], 'd'], -3>` is `'b' | 'c' | number`. The type makes sure it covers all possible outcomes.

andr3medeiros:
@som-sm you're right about the array spread problem. But it got me thinking on this outcome thing. We're talking about touples, not arrays right? Array items are unpredictable. Given that I believe `ArrayAt<['a', 'b', 'c', ...number[], 'd'], -1>` should indeed be "d", `ArrayAt<[1, 2 | undefined], 1>` should be `2 | undefined`.

som-sm:
> We're talking about tuples, not arrays right?

@andr3medeiros Both arrays and tuples. Moreover, you can argue that something like `['a', 'b', ...number[]]` is _also_ a tuple. And, surely something like `['a', 'b', 'c'?, 'd'?]` is a tuple, so `ArrayAt<['a', 'b', 'c'?, 'd'?], -2>` has to be a union of all possible outcomes, i.e., `'a' | 'b' | 'c'` (or `'a' | 'b' | 'c' | undefined`).

<br>

> Given that I believe `ArrayAt<['a', 'b', 'c', ...number[], 'd'], -1>` should indeed be "d", `ArrayAt<[1, 2 | undefined], 1>` should be `2 | undefined`.

Yeah, correct!

This PR is still a WIP, but you can go through the test cases to better understand the different scenarios.

PR #1280, opened 2025-10-19
`GreaterThanOrEqual` / `LessThan`: Fix behaviour with operands like `N` and `N | N + >0`

som-sm:
`GreaterThanOrEqual` currently doesn’t distribute `A` and `B` before checking whether `A extends B`.
https://github.com/sindresorhus/type-fest/blob/5612f648228512eb4e682e2f416858ff7d6d8835/source/greater-than-or-equal.d.ts#L20-L22

As a result, if `A` is `1` and `B` is `1 | 2`, the condition `A extends B` satisfies, causing the type to return `true` and completely ignore the `| 2` part in `B`.

This PR fixes the behavior by distributing `A` and `B` before performing the check.

```ts
type Current = GreaterThanOrEqual<1, 1 | 2>;
//=> true

type Fixed = GreaterThanOrEqual<1, 1 | 2>;
//=> boolean
```

Similar issue happens with `LessThan` because it's dependent on `GreaterThanOrEqual`.

```ts
type Current = LessThan<1, 1 | 2>;
//=> false

type Fixed = LessThan<1, 1 | 2>;
//=> boolean
```

### DIFF
--- a/test-d/less-than.ts
+++ b/test-d/less-than.ts
@@ -26,6 +26,7 @@ expectType<LessThan<-10, -90 | 90>>({} as boolean);
 expectType<LessThan<-16 | 16, 0>>({} as boolean);
 expectType<LessThan<-4 | 45, 20 | 30>>({} as boolean);
 expectType<LessThan<1 | -1 | 3, 0 | 2>>({} as boolean);
+expectType<LessThan<1, 1 | 2>>({} as boolean);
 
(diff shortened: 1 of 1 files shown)


som-sm:
Also, currently `GreaterThan<number, 10>` returns `never` which seems incorrect, should have been `boolean` instead. Similar happens with all the other similar types. Will open a separate PR for this.

PR #1284, opened 2025-10-21
feat: improve `UnionMax`, `UnionMin`, `ReverseSign`, `Negative`, `ToNumber` and add `ToBigint`, `SplitFloat` and `SeperateNegatives` internal types

benzaria:
Changes:

- add negative numbers support for `UnionMax` and `UnionMin`.
- add `SeperateNegatives` internal type to split a number union to `[positives, negatives]`.
- add `bigint` support for `ReverseSign`.
- fix `Negative` type returning `never` for `NegativeInfinity`.
- rename and improve `StringToNumber -> ToNumber`  to include `bigint` and `boolean` similare to `Number(T)`.
- add `ToBigint` internal type to transforme supported types to `bigint`.
- add `SplitFloat` internal type to use as a simple `Round` type at the moment.

### DIFF
--- a/source/internal/numeric.d.ts
+++ b/source/internal/numeric.d.ts
@@ -161,10 +161,10 @@ type T3 = ReverseSign<-1n>;
 type T4 = ReverseSign<1n>;
 //=> -1n
 
-type T5 = ReverseSign<NegativeInfinity>
+type T5 = ReverseSign<NegativeInfinity>;
 //=> Infinity
 
-type T6 = ReverseSign<PositiveInfinity>
+type T6 = ReverseSign<PositiveInfinity>;
 //=> -Infinity
 ```
 */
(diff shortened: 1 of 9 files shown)


source/internal/numeric.d.ts
```diff
@@ -56,33 +59,43 @@ Note: Just supports numbers from 0 to 999.
 
 @example
 ```
-type A = UnionMin<1 | 3 | 2>;
+type T1 = UnionMax<1 | 2 | 3>;
```
sindresorhus:
```suggestion
type T1 = UnionMin<1 | 2 | 3>;
```

source/internal/numeric.d.ts
```diff
@@ -91,60 +104,178 @@ Note: Just supports numbers from 0 to 999.
...
+						: `-${N}` extends `${infer R}` ? R // Handle positive numbers
+							: never
+				) extends infer Result extends string
+					? N extends bigint
+						? ToBigint<Result>
+						: ToNumber<Result>
+					: never;
+
+/**
+Split a union of positive and negative number into a tuple in the form of `[Negatives, Positives]`
```
sindresorhus:
```suggestion
Split a union of positive and negative number into a tuple in the form of `[Positives, Negatives]`
```

source/internal/numeric.d.ts
sindresorhus: No longer correct.

sindresorhus (changes requested):

source/internal/numeric.d.ts
benzaria: ahh yes, i changed the output last minute and forgot it.

source/internal/numeric.d.ts
```diff
@@ -52,99 +55,251 @@ export type IsNumberLike<N> =
...
 ```
 */
 export type UnionMin<N extends number> =
-	IsAnyOrNever<N> extends true ? N
-		: number extends N ? number
-			: NegativeInfinity extends N ? NegativeInfinity
-				: [N] extends [PositiveInfinity] ? PositiveInfinity
-					: InternalUnionMin<Finite<N>>;
+	IsAnyOrNever<N> extends true ? N // Handles `any/never`
+		: number extends N ? number // If `N` is the wide `number` type, we cannot infer a literal manimum
```
sindresorhus:
```suggestion
		: number extends N ? number // If `N` is the wide `number` type, we cannot infer a literal minimum
```

Please run AI over changes to catch typos like this.

som-sm:
@benzaria Are all of these changes related and need to be part of the same PR?

alex-touza:
Why was this closed? This PR has very important fixes for negative numbers support.

PR #1287, opened 2025-10-26
Add 'overrides' property to package.json types

SBoudrias:
Add 'overrides' property to package.json type definition.

Per https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides

### DIFF
--- a/source/package-json.d.ts
+++ b/source/package-json.d.ts
@@ -212,7 +212,7 @@ export namespace PackageJson {
 	Recursive map describing selective dependency version overrides supported by npm.
 	*/
 	type DependencyOverrides = {
-		[packageName: string]: string | DependencyOverrides;
+		[packageName: string]: string | undefined | DependencyOverrides;
 	};
 
 	/**
--- a/test-d/package-json.ts
+++ b/test-d/package-json.ts
@@ -40,6 +40,8 @@ expectType<PackageJson.WorkspaceConfig | string[] | undefined>(packageJson.works
 expectAssignable<PackageJson['overrides']>({foo: '1.0.0'});
 expectAssignable<PackageJson['overrides']>({foo: {'.': '1.0.0', bar: '1.0.0'}});
 expectAssignable<PackageJson['overrides']>({baz: {bar: {foo: '1.0.0'}}});
+expectAssignable<PackageJson['overrides']>({foo: undefined});
+expectAssignable<PackageJson['overrides']>({foo: {bar: undefined}});
 expectAssignable<PackageJson.DevEngineDependency>({
 	name: 'unicorn',
 	version: '>= 1.0.0',


source/package-json.d.ts
```diff
@@ -510,6 +510,11 @@ export namespace PackageJson {
 		*/
 		bundleDependencies?: string[];
 
+		/**
+		Selective version overrides. Allows the definition of custom package versions inside dependencies without manual edits in the `package-lock.json` file.
```
som-sm: I think this description should match with the one that shows up in `package.json` files.

```suggestion
		Overrides is used to support selective version overrides using npm, which lets you define custom package versions or ranges inside your dependencies.
```

source/package-json.d.ts
```diff
@@ -510,6 +510,11 @@ export namespace PackageJson {
 		*/
 		bundleDependencies?: string[];
 
+		/**
+		npm dependency override map
+		*/
+		overrides?: Dependency;
```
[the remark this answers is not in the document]
som-sm: > **P1 Badge Model overrides as nested maps, not plain string dependencies**
> 
> The new `overrides` property is typed as `Dependency`, which resolves to `Record<string, string>`. npm’s `overrides` field accepts nested override objects so you can target transitive dependencies (e.g. `{ "bar": { "foo": "1.2.3" } }` or keyed by `"bar@1.0.0"`). With the current type, any nested structure is a type error even though it is valid JSON supported by npm, so consumers cannot express common override scenarios without casting to `any`. This should be a recursive type allowing string values or nested override maps instead of reusing `Dependency`.
> 
> Useful? React with 👍 / 👎.

As suggested, type for `overrides` should be something like:
```ts
type Overrides = {
    [package: string]: string | Overrides;
};
```

source/package-json.d.ts
som-sm, line 1: Add tests:

```ts
expectAssignable<PackageJson['overrides']>({'foo': '1.0.0'});
expectAssignable<PackageJson['overrides']>({'foo': {'.': '1.0.0', 'bar': '1.0.0'}});
expectAssignable<PackageJson['overrides']>({'baz': {'bar': {'foo': '1.0.0'}}});
```

SBoudrias:
@som-sm addressed your review comments 🤞🏻

source/package-json.d.ts
```diff
@@ -208,6 +208,13 @@ export namespace PackageJson {
 	*/
 	type Dependency = Partial<Record<string, string>>;
 
+	/**
+	Recursive map describing selective dependency version overrides supported by npm.
+	*/
+	type DependencyOverrides = {
+		[packageName: string]: string | DependencyOverrides;
```
som-sm: Add `undefined` as well, we have it `Dependency` as well.

```suggestion
		[packageName: string]: string | undefined | DependencyOverrides;
```

test-d/package-json.ts
```diff
@@ -37,6 +37,9 @@ expectType<string[] | undefined>(packageJson.bundleDependencies);
 expectType<string[] | undefined>(packageJson.bundledDependencies);
 expectType<PackageJson.Dependency | undefined>(packageJson.resolutions);
 expectType<PackageJson.WorkspaceConfig | string[] | undefined>(packageJson.workspaces);
+expectAssignable<PackageJson['overrides']>({foo: '1.0.0'});
+expectAssignable<PackageJson['overrides']>({foo: {'.': '1.0.0', bar: '1.0.0'}});
+expectAssignable<PackageJson['overrides']>({baz: {bar: {foo: '1.0.0'}}});
```
som-sm, line 42: `undefined` tests:
```ts
expectAssignable<PackageJson['overrides']>({foo: undefined});
expectAssignable<PackageJson['overrides']>({foo: {bar: undefined}});
```

som-sm (commented):
LGTM, just add `undefined` as a valid value.

SBoudrias:
Updated!

PR #1290, opened 2025-11-02
`CamelCase`: Add support for `splitOnNumbers` option

sindresorhus:


package.json
```diff
@@ -66,5 +66,10 @@
 		"compilerOptions": {
 			"noUnusedLocals": false
 		}
+	},
+	"xo": {
+		"rules": {
+			"@typescript-eslint/unified-signatures": "off"
+		}
```
som-sm: Looks like this overwrites our `xo.config.js` file because of which `test:xo` fails.

Also, looks like this change is not required at all, tests pass even with this enabled.

sindresorhus: Accidental leftover.

PR #1291, opened 2025-11-02
`ArraySlice` / `StringSlice`: Fix behaviour with unions

### LINKED ISSUE #1289: `StringSlice` is incorrect for union of string literals of different lengths
### Bug description

It ends up cutting all strings down to the length of the shortest string

### Repro

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbzgZRlYA7A5sgNsAYwFM4BfOAMyghDgHIYBPMIgWgqIGcY6AoXpixT5iAEzgBeFGkw4RRADx0AhgDoAFkVy4IdOAB96agO4RjdADRwATAD5ecR44D0zuAD0A-A6dxerlHUIAFdccQAjEjpNbV0DelNzIA

som-sm:
Fixes behaviour of `ArraySlice` with unions.

```ts
type Current = ArraySlice<[0, 1, 2, 3], 1, 2 | 3>;
//=> [1]

type Fixed = ArraySlice<[0, 1, 2, 3], 1, 2 | 3>;
//=> [1] | [2]
```

```ts
type Current1= ArraySlice<[0, 1, 2, 3], 2 | -3, 3>;
//=> []

type Updated = ArraySlice<[0, 1, 2, 3], 2 | -3, 3>;
//=> [2] | [1, 2]
```

<br>

This also fixes behaviour of `StringSlice` (Fixes #1289) with unions as it relies on `ArraySlice`, just needed the default type parameters to be fixed.

### DIFF
--- a/test-d/string-slice.ts
+++ b/test-d/string-slice.ts
@@ -18,26 +18,37 @@ expectType<StringSlice<string, 1, 2>>(null! as string);
 
 // Unions
 // String is union
-expectType<StringSlice<'abcd' | '0123456789'>>({} as 'abcd' | '0123456789');
-expectType<StringSlice<'abcd' | '0123456789', 0, 2>>({} as 'ab' | '01');
+expectType<StringSlice<'012' | 'abcd', 0>>({} as '012' | 'abcd'); // Positive start, no end
+expectType<StringSlice<'012' | 'abcd', -2>>({} as '12' | 'cd'); // Negative start, no end
+expectType<StringSlice<'012' | 'abcd', 0, 2>>({} as '01' | 'ab'); // Positive start, positive end
+expectType<StringSlice<'012' | 'abcd', -2, -1>>({} as '1' | 'c'); // Negative start, negative end
+expectType<StringSlice<'012' | 'abcd', -3, 2>>({} as '01' | 'b'); // Negative start, positive end
+expectType<StringSlice<'012' | 'abcd', 1, -1>>({} as '1' | 'bc'); // Positive start, negative end
 
 // Start is union
-expectType<StringSlice<'abcd', 1 | 3>>({} as 'bcd' | 'd');
-expectType<StringSlice<'abcd', 0 | 2, 3>>({} as 'abc' | 'c');
+expectType<StringSlice<'0123', 1 | -2>>({} as '123' | '23'); // Positive/Negative start, no end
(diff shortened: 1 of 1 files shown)


source/array-slice.d.ts
```diff
@@ -60,13 +61,33 @@ export type ArraySlice<
...
+				: never // Never happens
+		: IsNever<End> extends true
+			? Start extends unknown // To distribute `Start`
+				? _ArraySlice<Array_, Start, End>
+				: never // Never happens
+			: Start extends unknown // To distribute `Start`
+				? End extends unknown // To distribute `End`
+					? _ArraySlice<Array_, Start, End>
+					: never // Never happens
+				: never // Never happens
```
som-sm, line 78: This is complicated/verbose because the default value of `Start`/`End` is `never`, so we have to explicitly check for `never` before distributing, which introduces several branching cases.

<br>

I guess the reason for having the default value as `never` was that if `End` were simply `Array_["length"]`, then union instantiations of `End` would give incorrect results. 

For example,
`ArraySlice<[0, 1, 2] | [0, 1], 0>` would become
`ArraySlice<[0, 1, 2] | [0, 1], 0, 3 | 2>`,
which would then distribute to
`ArraySlice<[0, 1, 2], 3 | 2> | ArraySlice<[0, 1], 3 | 2>`,
and finally to
`ArraySlice<[0, 1, 2], 3> | ArraySlice<[0, 1, 2], 2> | ArraySlice<[0, 1], 3> | ArraySlice<[0, 1], 2>`,
which would be incorrect.

However, if we set the default for `End` to `UnionMax<Array_["length"]>`, that would fix this issue. The only thing is that explicitly passing `never` would no longer behave the same as omitting the argument. But that’s probably acceptable because `never` shouldn’t really be a valid value for `Start`/`End` anyway. In such cases, we can just return something like `readonly unknown[]` or `[]`.

som-sm: > However, if we set the default for `End` to `UnionMax<Array_["length"]>`, that would fix this issue. The only thing is that explicitly passing `never` would no longer behave the same as omitting the argument. But that’s probably acceptable because `never` shouldn’t really be a valid value for `Start`/`End` anyway. In such cases, we can just return something like `readonly unknown[]` or `[]`.

@sindresorhus If this is fine, then I'll open a separate PR for this.

sindresorhus: Yes, seems fine.

PR #1296, opened 2025-11-14
Add `UnwrapPartial` type

### LINKED ISSUE #1294: Consider `UnwrapPartial<T>` as a way to revert the `Partial<T>` modifier

Let’s consider the following case:

```ts
type MyKeyType = 'foo' | 'bar' | 'baz';
type MyObjectType = Record<MyKeyType, number>;
```

When composing the result from scratch, `MyObjectType` must be wrapped with `Partial`.

```ts
const result: Partial<MyObjectType> = {};

for (const key of [
	'foo',
	'bar',
	'baz',
] as const satisfies ReadonlyArray<MyKeyType>) {
	result[key] = Math.random();
}
```

Now that we’ve determined the result, let’s revert the `Partial` for a clean export.

```ts
export const data = result as Required<typeof result>;
//=> const data: Required<Partial<MyObjectType>>
```

It’s not ideal—we’d expect `MyObjectType` instead of the lingering type modifiers. 

Here’s the same export with `Complete`, though:

```ts
export const data = result as Complete<typeof result>;
//=> const data: MyObjectType
```

`Complete` improves DX the same way `Simplify` does. If there’s consensus around including it, I’d be happy to submit a proper PR.

### Type source

```ts
type Complete<T> = T extends Partial<infer P> ? P : T;
```

porada:
Closes #1294.

### DIFF
--- a/source/unwrap-partial.d.ts
+++ b/source/unwrap-partial.d.ts
@@ -14,10 +14,10 @@ type Config = Partial<{
 }>;
 
 type InitializedConfig = UnwrapPartial<Config>;
-//=> { port: number; host: string; secure?: boolean }
+//=> {port: number; host: string; secure?: boolean}
 ```
 
-If the provided type isn’t of `Partial<T>`, `UnwrapPartial` has no effect on the original type.
+Note: If the provided type isn’t of `Partial<T>`, `UnwrapPartial` has no effect on the original type.
 
 @category Object
 */


porada:
Questions that popped up to my mind while working on this:

1. Should we be more strict about array types? Technically `UnwrapPartial<string[]>` will pass as `string[]`, but it doesn’t make much sense (neither does defining `Partial<string[]>`). We could reject types that aren’t based on plain objects, but are we keeping the promise of unwrapping then?
2. I feel like I overdid the tests. @som-sm @sindresorhus — would any of you mind taking a look before we proceed with this?

porada:
Aside from the points mentioned above, this PR is ready for review.

Once these changes are in, I’ll be happy to follow up with [another](https://github.com/sindresorhus/type-fest/issues/1294#issuecomment-3518038463) `Unwrap` type.

test-d/unwrap-partial.ts
som-sm, line 1: Looks like the test cases could be simplified, there are so many test cases of the form `UnwrapPartial<Partial<Something>>`, with different values for `Something`. I think we don't need that many because the value of `Something` doesn't really matter, so we could just keep a few and remove the others.

Tests that don't have `Partial` are useful, like `UnwrapPartial<{a?: string; b: number}>`, `UnwrapPartial<{a?: string; b?: number}>`.

test-d/unwrap-partial.ts
```diff
@@ -0,0 +1,93 @@
...
+expectNever({} as UnwrapPartial<TestType | AnotherTestType>);
+
+// `UnwrapPartial` works with index signatures
+type ArrayLikeTestType = {
+	[index: number]: string;
+};
+
+expectType<ArrayLikeTestType>({} as UnwrapPartial<Partial<ArrayLikeTestType>>);
+expectType<Record<number, string>>({} as UnwrapPartial<Partial<Record<number, string>>>);
+expectType<string[]>({} as UnwrapPartial<Partial<string[]>>);
```
som-sm, line 67: Also add,
```ts
expectType<never>({} as UnwrapPartial<string[]>);
expectType<Array<string | undefined>>({} as UnwrapPartial<Array<string | undefined>>);
```

test-d/unwrap-partial.ts
```diff
@@ -0,0 +1,93 @@
...
+
+expectType<TestType | AnotherTestType>({} as UnwrapPartial<PartialUnionType>);
+expectType<TestType>({} as UnwrapPartial<{a?: string; b?: number} | {c: boolean; d: 'literal'}>);
+expectNever({} as UnwrapPartial<TestType | AnotherTestType>);
+
+// `UnwrapPartial` works with index signatures
+type ArrayLikeTestType = {
+	[index: number]: string;
+};
+
```
som-sm, line 59: Add these:
```ts
expectType<never>({} as UnwrapPartial<{[p: string]: number}>);
expectType<UnwrapPartial<{[p: string]: number | undefined}>>({} as UnwrapPartial<{[p: string]: number | undefined}>);
```

source/unwrap-partial.d.ts
```diff
@@ -0,0 +1,33 @@
...
+@category Object
+*/
+export type UnwrapPartial<PartialObjectType> =
+	PartialObjectType extends Partial<infer ObjectType>
+		? (
+			Partial<ObjectType> extends PartialObjectType
+				? ObjectType
+				: never
+		)
+		: never;
```
som-sm: Should these be `never` or `PartialObjectType`, what would be better?

test-d/unwrap-partial.ts
porada: Hey Som, thanks for the review.

I use `UnwrapPartial<Partial<Something>>` to ensure that the original `Something` type gets preserved. And there is a good number of types to consider.

Cases like `UnwrapPartial<{a?: string; b: number}>` and `UnwrapPartial<{a?: string; b?: number}>` are [already covered](https://github.com/porada/type-fest/blob/de00d904043526a8bdf0404f5d489e90dc55c2f6/test-d/unwrap-partial.ts#L36-L39) (under `` `UnwrapPartial` preserves optional properties ``).

I’ll add any missing cases you pointed out, but feel free to prune things on your end as well.

source/unwrap-partial.d.ts
porada: The way I see this, you shouldn’t unwrap a modifier on a type that doesn’t have one in the first place (goes for all `Unwrap*` types). `UnwrapPartial<number>` should never resolve to `number`.

I was also [wondering earlier](https://github.com/sindresorhus/type-fest/pull/1296#issuecomment-3533816425):

> Should we be more strict about array types? Technically `UnwrapPartial<string[]>` will pass as `string[]`, but it doesn’t make much sense (neither does defining `Partial<string[]>`). 

After thinking it over, I’m leaning towards not giving arrays any special treatment. Let’s assume the user knows what they’re doing.

test-d/unwrap-partial.ts
som-sm: > Cases like `UnwrapPartial<{a?: string; b: number}>` and `UnwrapPartial<{a?: string; b?: number}>` are [already covered](https://github.com/porada/type-fest/blob/de00d904043526a8bdf0404f5d489e90dc55c2f6/test-d/unwrap-partial.ts#L36-L39) (under `` `UnwrapPartial` preserves optional properties ``).

Yeah, I meant all of those are useful, don't remove them.

porada: @som-sm — Unless you want to make any other specific changes, I think this PR is ready to merge.

source/unwrap-partial.d.ts
som-sm: > After thinking it over, I’m leaning towards not giving arrays any special treatment. Let’s assume the user knows what they’re doing.

Yeah, that's fine.

> The way I see this, you shouldn’t unwrap a modifier on a type that doesn’t have one in the first place (goes for all `Unwrap*` types). `UnwrapPartial<number>` should never resolve to `number`.

I brought this up because with unions, IMO, it feels better if `UnwrapPartial<Partial<{x: number, y: number}> | [number, number]>>` returned `{x: number, y: number} | [number, number]` as opposed to just `{x: number; y: number}`. WDYT?

porada: A union with a partial isn’t technically a partial, and handling special cases is a can of worms we probably don’t want to open 🙂

som-sm: > A union with a partial isn’t technically a partial

No, I did not mean that `Partial<{x: number, y: number}> | [number, number]` is a partial. I meant that instantiating `UnwrapPartial` with a union that contains both partial and non-partial items would cause the non-partial items to be removed, which might not be desirable.

> handling special cases is a can of worms we probably don’t want to open

No, we'd obviously not handle unions separately, I meant replacing `never` with `PartialObjectType`. I think that'd be better.

porada: `Writable` (technically `UnwrapReadonly`) works that way already, so I’m fine with returning the original type for consistency. Just made the necessary edits.

:shipit:

som-sm: > `Writable` (technically `UnwrapReadonly`)

They're not the same, right? Just like `Required` and `UnwrapPartial`:
```ts
type Config = Partial<{port: number; host: string; secure?: boolean}>;
//=> {port?: number; host?: string; secure?: boolean}

type T1 = UnwrapPartial<Config>;
//=> {port: number; host: string; secure?: boolean}

type T2 = Required<T1>;
//=> {port: number; host: string; secure: boolean}
```

<br>

> so I’m fine with returning the original type for consistency. Just made the necessary edits.

👍

porada:
All good. Thanks for the passing review!

@som-sm @sindresorhus — Are you open to having a conversation about adding more `Unwrap` types in? I’d be happy to create a new issue for this.

som-sm:
> @som-sm @sindresorhus — Are you open to having a conversation about adding more `Unwrap` types in? I’d be happy to create a new issue for this.

Yeah sure! And feel free to open PRs for other `Unwrap` types.

PR #1300, opened 2025-11-20
Add custom processor to lint JSDoc codeblocks

som-sm:
A common task when reviewing PRs is manually checking JSDoc codeblocks for issues like missing semicolons, broken indentation, use of double quotes etc. This PR addresses that by adding linting to JSDoc codeblocks. They will now be linted just like our regular TS files, so any lint errors in JSDoc codeblocks will be caught automatically, and auto-fixable errors can be fixed just like regular auto-fixable errors.

https://github.com/user-attachments/assets/925e3532-1b4b-4996-a5a3-5d7c1b11c377

---

Common lint errors:

- Use of `interface` instead of `type`
   image

- Incosistent spacing
   image

- Incorrect quotes
   image

- Use of `T[]` for non-simple types.

   image

- Missing semicolon
   image

- Use of spaces instead of tabs

   image

These are just a few examples. Otherwise, all the rules (except the ones explicitly disabled) that apply to our TS source files apply to JSDoc codeblocks as well.

### DIFF
--- a/.github/workflows/ts-canary.yml
+++ b/.github/workflows/ts-canary.yml
@@ -23,4 +23,4 @@ jobs:
       - run: npm install typescript@${{ matrix.typescript-version }}
       - name: show installed typescript version
         run: npm list typescript --depth=0
-      - run: npx tsc
+      - run: NODE_OPTIONS="--max-old-space-size=5120" npx tsc


.github/workflows/main.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           node-version: 24
       - run: npm install
       - run: npm install typescript@${{ matrix.typescript-version }}
-      - run: npx tsc
+      - run: NODE_OPTIONS="--max-old-space-size=5120" npx tsc
```
som-sm, line 45: This bump is needed because the number of files in the TS program has increased from `792` to `839`. The default limit seems to be 4GB, and increasing it by 1GB works fine.

The extra files come from the new `typescript-eslint` package added in this PR.

```bash
# main
npx tsc --listFilesOnly | wc -l # 792
```

```bash
# feat/lint-jsdoc-codeblocks
npx tsc --listFilesOnly | wc -l # 839
```

If this is a problem, we can disable `allowJS` in `tsconfig.json` for now, which will bring down the files to `514` and then come back to this separately.
```bash
# "allowJS": false
npx tsc --listFilesOnly | wc -l # 514
```

lint-processors/fixtures/eslint.config.js
som-sm, line 1: This config file is used to lint the test cases. And, this is present inside the repo instead of the temp directory because it needs certain imports.

lint-processors/jsdoc-codeblocks.test.js
```diff
@@ -0,0 +1,1481 @@
...
+				// Manual loop because `assert.partialDeepStrictEqual` isn't available in Node 20
+				for (const [index, expected] of errors.entries()) {
+					const actual = results[0].messages[index];
+					const actualSubset = Object.fromEntries(Object.keys(expected).map(key => [key, actual[key]]));
+
+					t.assert.deepStrictEqual(actualSubset, expected);
+				}
+
+				const resultsFixed = await eslintFixed.lintFiles([fileName]);
+				t.assert.strictEqual(resultsFixed[0].output, output);
```
som-sm, line 1477: It would have been cleaner if this were done using snapshot testing, it would have moved the `output`s to a separate snapshot file. But `t.assert.snapshot` is _not_ available in Node 20, and we currently run the tests on Node 20.

som-sm (commented):
Added comments wherever I felt it needed some explanation.

som-sm:
@sindresorhus This PR is ready for review now!

- All the changes within the `source` directory are inside JSDoc example codeblocks only. Most of these changes were auto-fixable, and the ones that weren't are captured separately in [this commit](https://github.com/sindresorhus/type-fest/pull/1300/commits/523017c71023ec1110b3c7e85bcd2b9b5de22242).
- Unlike the `validate-codeblocks` lint rule, this processor applies the lint rules to all JSDoc codeblocks, not just the exported ones. This seems fine and makes sense, since it's nice to catch lint errors even in non-exported JSDoc codeblocks.

sindresorhus:
Yay! So nice to have this automatically enforced too 🎉

Issue #1305, opened 2025-11-30
Issue with the new `IsEqual`


yishayhaz:
### Bug description

In the example below, the onChange function should accept `true` as the second argument. With the old `IsEqual`, it does; with the new type, it does not. This is an extremely abstracted example, so let me know if you need more context.

```ts
type FieldProps<T1> = {
  onChange: <T2>(value: T2, shouldBeTrue: IsEqual<T1, T2>) => void;
};

const TextField = <T>(value: T) => {
  ({} as FieldProps<T>).onChange(value, true);
};
```

### Repro

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBDAnmApnA3gSQM4FECOArgIYA2AvnAGZQQhwDkSqAtFStjAwNwCwAUAOZoA8qQAmOAiVIAeAIIAaOACEAfHAC8AgJAAKWQHE1egJRaNhuCgAeMFADtx2OPLgAyOFYA+XuAH44AEY4AC44ACZzW3snbF0DYzMLPxjHZ1UPP18rQJDwqN0dQJgoQhQi8KoybBQ+QX4Aeka4ADkUAHc4KSIyOBY4cQgOOAcIeA7oAGsBIWQ0ADFgFAkABVowbFkAFSCNTQwBODgIBwBhAAtiBwBzFHCdiJMANzJy8O2I5WwLiEIJFQobZle7dPC9OS7ZSfNTmTQaZ4QYDierkeoCADGp04cG2MSWK3EWjgO0ULzeoO2cI06COcD06EoxBcBLWGy221hADpTpdrnc9K9SOVlKVyqZUeimi0xESBpMoFMXMzrDZUBj7OJZvxhHBZazxOsIJsdntibT+MdeVdbqDHuThZSvnAfn8AUCQeFZT0ZKboU9qXBEcjJdqsQ4cbK8XYDcTSQ73rjAxbjgymS59cs2caOdzrfyUIKKaKQRKBGigA

taiyakihitotsu:
Thanks !

This seems to be a problem caused by the inference of generics being deferred by distribution.

I will send a PR later.

### sample

```typescript
export type IsEqual_0<A,B> = A extends B ? true : A extends B ? true : false

type FieldProps_0<T1> = {
  onChange: <T2>(value: T2, shouldBeTrue: IsEqual_0<T1, T2>) => void;
};

const TextField_0 = <T,>(value: T) => {
[...]

PR #1306, opened 2025-11-30
`IsEqual`: Fix generic inference

taiyakihitotsu:
This PR fixes issue #1305.

This approach is similar to the fix previously implemented in https://github.com/sindresorhus/type-fest/commit/5af60a1330c304eae03765299281ce08710aa323.

Note:
I have intentionally ignored the `@typescript-eslint/consistent-type-assertions` rule in the test file (`test-d/is-equal.ts`) to accurately reflect and test the structure of the original reported issue code.

```terminal
  test-d/is-equal.ts:94:3
  ✖  94:3   Always prefer const x: T = { ... }. 
  @typescript-eslint/consistent-type-assertions
```

### DIFF
--- a/test-d/is-equal.ts
+++ b/test-d/is-equal.ts
@@ -89,4 +89,3 @@ expectType<true>(equalTupleIntersectionAndTuple);
 // Test for Issue https://github.com/sindresorhus/type-fest/issues/1305
 type Assignability<T, U, _V extends IsEqual<T, U>> = any;
 type TestAssignability<T> = Assignability<T, T, true>;
-


test-d/is-equal.ts
```diff
@@ -85,3 +85,11 @@ expectType<true>(equalTupleIntersectionToBeNeverAndNeverExpanded);
...
 declare const equalTupleIntersectionAndTuple: IsEqual<[{a: 1}] & [{a: 1}], [{a: 1}]>; // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
 expectType<true>(equalTupleIntersectionAndTuple);
+
+// Test for Issue https://github.com/sindresorhus/type-fest/issues/1305
+type FieldProps<T1> = {
+	onChange: <T2>(value: T2, shouldBeTrue: IsEqual<T1, T2>) => void;
+};
+const _TextField = <T>(value: T) => {
+	({} as FieldProps<T>).onChange(value, true); // eslint-disable-line @typescript-eslint/consistent-type-assertions
+};
```
som-sm: Use the following test instead:

```suggestion
type Assignability<T, U, _V extends IsEqual<T, U>> = any;
type TestAssignability<T> = Assignability<T, T, true>;
```

This is easier to understand and aligns with other generic assignability tests that we have.

taiyakihitotsu: Thanks, addressed.

PR #1309, opened 2025-12-05
Add linting to validate types specified via twoslash (`//=>`) inside JSDoc codeblocks

som-sm:
This PR validates the types specified via **twoslash arrow comments (`//=>`)** inside JSDoc codeblocks using the `getQuickInfoAtPosition` API, with automatic fixes for incorrect types.

image<br>

<br>

It also improves the DX, since we no longer have to manually write the twoslash types, we can simply start the twoslash comment and rely on autofix to fill in the type.

https://github.com/user-attachments/assets/944ef869-e397-412a-a9a7-632f267612a7

<br>
Surprisingly, there were only a few types that were incorrect:

1. Should be `boolean` instead of `false`.
    https://github.com/sindresorhus/type-fest/blob/b3ecd07b5a88378a60ae3c3ef9c356ea06378211/source/all-extend.d.ts#L87-L89
    
    
2. Should be `true` instead of `unknown`.

     image

3. Return type is not `number`.
    https://github.com/sindresorhus/type-fest/blob/b3ecd07b5a88378a60ae3c3ef9c356ea06378211/source/is-literal.d.ts#L244-L245

4. The resulting type is `Infinity` and not `PositiveInfinity` or `NegativeInfinity`.
    
    image
    
    image

---

## Working:

1. Twoslash arrow comments (`//=>`) inside JSDoc codeblocks are identified, and the first position on the previous line where quickinfo is available is retrieved using the `getQuickInfoAtPosition` API.

2. The type portion is extracted from the retrieved quickinfo. For example, if the quickinfo is `type T1 = string | number;`, the extracted type is `string | number`.

3. The extracted type is then normalized to match our linting style. This includes converting double-quoted string literals to single-quoted ones and replacing four-space indentation with tabs.

4. If the normalized type does not match the specified twoslash comment, an autofixable lint error is reported at the appropriate location.

---
## Notes:
1. Adds support for overwriting default compiler options, like
    https://github.com/sindresorhus/type-fest/blob/71dde372b3692128f707c79dde6097c5060026d1/source/all-extend.d.ts#L80-L98
    
    This is similar to how we'd do it in TS playground.
    
2. Catches formatting issues like use of spaces instead of tabs, use of double quotes etc.

     image

3. For object types, type annotations were previously specified inconsistently, sometimes on a single line and sometimes across multiple lines. This PR makes this behaviour deterministic. If the quickinfo type is within a defined length threshold, it must be written in a single line, otherwise, it must be written just as it appears in the quickinfo.
    
    Screenshot 2025-12-20 at 2 27 26 PM
    
    Screenshot 2025-12-20 at 2 36 33 PM

4. Twoslash arrow comments (`//=>`) should now _only_ be used to specify types. All non-type usages (such as error comments) have been converted to plain comments.
   
   image
   
   image

---

## Future improvements:

~The order of unions in quickinfo is sometimes messed up, which leads to changes like these:~

image

image
    
image<br>

~I'll look into this separately if this can be improved.~

Fixed by #1320.

### DIFF
--- a/source/array-reverse.d.ts
+++ b/source/array-reverse.d.ts
@@ -18,7 +18,7 @@ type B = ArrayReverse<readonly [string, number, ...boolean[]]>;
 //=> readonly [...boolean[], number, string]
 
 type C = ArrayReverse<['foo', 'bar'] | readonly [1, 2, 3]>;
-//=> ["bar", "foo"] | readonly [3, 2, 1]
+//=> ['bar', 'foo'] | readonly [3, 2, 1]
 
 type D = ArrayReverse<string[]>;
 //=> string[]
@@ -34,19 +34,19 @@ Note: If the tuple contains optional elements, the result will be a union of tup
 import type {ArrayReverse} from 'type-fest';
 
 type A = ArrayReverse<[string, number, boolean?]>;
-//=> [boolean, number, string] | [number, string]
+//=> [number, string] | [boolean, number, string]
 
 type B = ArrayReverse<[string, number?, boolean?]>;
-//=> [boolean, number, string] | [number, string] | [string]
+//=> [string] | [number, string] | [boolean, number, string]
 
 type C = ArrayReverse<[string?, number?, boolean?]>;
-//=> [boolean, number, string] | [number, string] | [string] | []
+//=> [] | [string] | [number, string] | [boolean, number, string]
 
 type D = ArrayReverse<[string, number?, ...boolean[]]>;
-//=> [...boolean[], number, string] | [string]
+//=> [string] | [...boolean[], number, string]
 
 type E = ArrayReverse<[string?, number?, ...boolean[]]>;
-//=> [...boolean[], number, string] | [string] | []
+//=> [] | [string] | [...boolean[], number, string]
 ```
 
 @category Array


som-sm:
@sindresorhus This PR is ready now. All the changes within the `source` directory are inside JSDoc example codeblocks only, except for the following two changes where the response had to be wrapped in `Simplify`.

1. https://github.com/sindresorhus/type-fest/pull/1309/changes#diff-448563d9be18bc39d64f82b5d6f88cc3389a4bb7574970601a4274da2df325d4

2. https://github.com/sindresorhus/type-fest/pull/1309/changes#diff-e67f2f77b5599dfb3fa2f2ce80230a63116d79e1f7117ec79065fc844fda6f59

source/numeric.d.ts
```diff
@@ -74,13 +74,13 @@ type Float = Integer<1.5>;
...
-//=> 0o10
+//=> 8
 
 type BinaryInteger = Integer<0b10>;
-//=> 0b10
+//=> 2
 
 type HexadecimalInteger = Integer<0x10>;
-//=> 0x10
+//=> 16
```
som-sm, line 83: Looks like non-decimal base types are automatically converted into decimal base.
    
image

sindresorhus:
> This PR is ready now. All the changes within the source directory are inside JSDoc example codeblocks only, except for the following two changes where the response had to be wrapped in Simplify.


Maybe add a short code comment on why `Simplify` is needed in those places.

source/int-closed-range.d.ts
```diff
@@ -17,11 +17,11 @@ Use-cases:
 ```
 import type {IntClosedRange} from 'type-fest';
 
-type Age = IntClosedRange<0, 120>;
-//=> 0 | 1 | 2 | ... | 119 | 120
+type Age = IntClosedRange<0, 20>;
+//=> 0 | 1 | 20 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12 | 13 | 14 | 15 | 16 | 17 | 18 | 19
```
sindresorhus: This is not ideal, and I bet users will wonder about this and open issues.

Some ideas:
- Sort the output
- Some kind of code comment to disable linting for the code block or certain lines.

som-sm:
> why `Simplify` is needed in those places.

1. The first `Simplify`, used in `RequiredDeep`, ensures that the internal helper type `RequiredObjectDeep` doesn't leak into the output type. 

    Without `Simplify`, `RequiredObjectDeep` appears in the output, which is confusing because it is an internal implementation detail that users are not expected to be aware of.
    image<br>
    
    And, this is flagged by this lint rule because, in this case, the specified twoslash type doesn't include `RequiredObjectDeep`, whereas the quickinfo type does.
    https://github.com/sindresorhus/type-fest/blob/b3ecd07b5a88378a60ae3c3ef9c356ea06378211/source/required-deep.d.ts#L25-L34
    
    With `Simplify` in place, there's no `RequiredObjectDeep` in the output.
    image

2. In the second case, it's not an internal type that shows up in the output, but the built-in `Omit` type. In my opinion, it is acceptable for built-in utility types to appear in the output because they have a standard, well-understood meaning. 

    This is flagged by this lint rule because, in this case, the specified twoslash type doesn't include `Omit`, so one possible fix here would have been to update the twoslash type itself, for example:
    ```diff
	    type Info2 = {
		    address: [
			    {
				    street: string;
			    },
			    {
				    street2: string;
				    foo: string;
			    },
		    ];
	    };
    
	    type AddressInfo = OmitDeep<Info2, 'address.1.foo'>;
    -	//=> {address: [{street: string}, {street2: string}]}
    +	//=> {
    +	// 	address: [{
    +	// 		street: string;
    +	// 	}, Omit<{
    +	// 		street2: string;
    +	// 		foo: string;
    +	// 	}, 'foo'>];
    +	// }
    ```
    
    However, I felt it was better to apply `Simplify` to the output, as it makes the output more readable.

---

> Maybe add a short code comment on 

Umm...not sure if a code comment is _really_ necessary here, since `Simplify` is already used in several places in the codebase to address similar issues. That said, if you still think a comment would be helpful, I’ll add a link to this comment.

source/int-closed-range.d.ts
som-sm: Ok, let me look into this then, I had implemented a sorting logic earlier, but it was fairly complicated, so I removed it. I’ll try to add it back with proper tests.

sindresorhus:
> Umm...not sure if a code comment is really necessary here, since Simplify is already used in several places in the codebase to address similar issues. That said, if you still think a comment would be helpful, I’ll add a link to this comment.

Up to you. Not important. Just thought something like:

> We use `Simplify` to prevent `RequiredObjectDeep` from appearing in the type hint.

sindresorhus:
Looks great! Really nice to be able to enforce this. Merry Xmas!

som-sm:
Merry Christmas and happy holidays!

source/subtract.d.ts
```diff
@@ -26,7 +26,7 @@ type D = Subtract<18, 96>;
 //=> -78
 
 type E = Subtract<PositiveInfinity, 9999>;
-//=> PositiveInfinity
+//=> Infinity
```
mrazauskas, line 29: Nice improvements. Interestingly it seems like there is no `Infinity` type. Only a detail that might sound misleading.

Reference:

https://github.com/sindresorhus/type-fest/blob/fe88ad8e4cee79dcb64f1c26f648319bfbcc4308/source/numeric.d.ts#L8-L19

som-sm: @mrazauskas Yeah, but I think showing `PositiveInfinity` is not great either because the quickinfo tooltip doesn't show that.

image

So, I think neither option is perfect, I just went with the one that actually shows up in the tooltip. Not sure why quickinfo decides to show `Infinity` when it doesn’t exist.

mrazauskas: Thanks for clarifying. ~~Now I think this could be reported as bug to TypeScript. Only a detail, but it is rather odd to see quickinfo with types that do not exist. Feels like the internals are leaking here.~~ Seems like this should be flag as a limitation of TypeScript.

PR #1209, opened 2025-07-31
WIP: fix inconsistent merge onto Destination's optional entries and better handling of undefined vs optional entries 

jclaveau:
As show here https://github.com/sindresorhus/type-fest/issues/1208: MergeDeep overwrites Destination's nested entries when they are optional

This PR fixes and test this merging issue.

Moreover, optional entries notion is deeply entangled with undefined as `{ entry?: string | undefined }` is redundant but `{ entry: string | undefined }` is not.

So this PR also implements the tests showing the issues the TS behavior generates.

I'm presently trying to implement a way to merge `optional entries VS possible undefined values` the way the most intuitive I can, but I still wonder if a configuration variable with optional merging strategies will be required:

"Currently" in the following code means, including the commit https://github.com/jclaveau/type-fest/commit/4bafb03dd2d3113683b0318cb17d69d6f319b275

```ts
type OptionalAndUndefinedNestedTest = {
	Left: {
		nested?: {
			in_left: string;
		} | undefined; // Redundancy is allowed and preserved by TS
	};
	Right: {
		nested: {
			in_right: string;
		};
	};
};

type OptionalAndUndefinedNestedRightIntoLeft = MergeDeep<
	OptionalAndUndefinedNestedTest['Left'],
	OptionalAndUndefinedNestedTest['Right']
>;
expectTypeOf<OptionalAndUndefinedNestedRightIntoLeft>().toEqualTypeOf<{
	nested: {
		in_left: string;
		in_right: string;
	} | undefined; // Currently, this " | undefined" is removed and IMHO shouldn't as it's on the value part
}>();

type OptionalAndUndefinedNestedRightIntoRight = MergeDeep<
	OptionalAndUndefinedNestedTest['Right'],
	OptionalAndUndefinedNestedTest['Left']
>;
expectTypeOf<OptionalAndUndefinedNestedRightIntoRight>().toEqualTypeOf<{
	nested?: {
		in_left: string;
		in_right: string;
	} | undefined;  // Currently this undefined is not kept but, IMHO, it should be (in case of further merge for example)
}>();
```

As a conclusion, I think the most intuitive strategy is to 
- Overwrite keys: Overwrite Destination optional keys when they are not optional in the Source
- Merge values: Merge union containing `| undefined`

I would love to have someone's else insight about it.

Presently, my main technical issue is that `Source[Key]` with only the key being optional (not the value) is converted to `value | undefined` at some point I struggle to identify in the codebase.

Thanks in advance !

PR #1319, opened 2025-12-23
`MergeDeep`: Remove extra `undefined` from optional properties

### LINKED ISSUE #1314: merge-deep: wrong types inferred when using partial interfaces
### Bug description

**the correct operation of this utility depends on the order of types!**

Let's say there is `interface Pokemon` like this:

```typescript
import type {MergeDeep} from 'type-fest';

interface Pokemon {
      name: string;
      type: string;
    }
```

**valid** :
```typescript
    const testThroughAssert: Pokemon = {} as MergeDeep<
      Partial<Pokemon>,
      Pokemon
    >;
```
=> `success`

**invalid**:
```typescript
    const testThroughAssert: Pokemon = {} as MergeDeep<
      Pokemon,
      Partial<Pokemon>
    >;
```
=> `error TS2322: Type '{ name?: string | undefined; type?: string | undefined; }' is not assignable to type 'Pokemon'.`

### Repro

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBDAnmApnA3gWRVA5igERRTAF84AzKCEOAciVQFoKUBnGOgbgFgAofsAB2MHBQCGAYzQAFCAGsUICEIz84GzUPEgUALjgcow3Lz6bNjfYZjGhp9ZtL9+klRwTsYAFQAW1AFdcXwBBNjYcGAM5RWVVAF4McnE2OGw8QmIwAB5HCxilFQAaPM0ZcVhgcQAbbIK4gD5Shq4gA

taiyakihitotsu:
Close #1314

(Updated the entire comment.)

This PR is a small change, but it consists of two parts.

## Part 1
```typescript
import type {MergeDeep} from 'type-fest';

interface Pokemon {
      name: string;
      type: string;
    }
```

`MergeDeep<Pokemon, Partial<Pokemon>>` should be equivalent to `Partial<Pokemon>` according to this comment in `merge-deep.d.ts`:
https://github.com/sindresorhus/type-fest/blob/c671aa2e9dbbfb502fde09859286d0caf499eb7a/source/merge-deep.d.ts#L386

However, before this PR, the result type was:
```typescript
{
  name?: string | undefined;
  type?: string | undefined;
}
```

which causes the following error (in the issue):
```
error TS2322: Type '{ name?: string | undefined; type?: string | undefined; }' is not assignable to type 'Pokemon'.
```
(Note: The above is a direct quote from an issue comment. In this case, it seems appropriate to type it as `Partial<Pokemon>` instead of `Pokemon`.)

This extra `| undefined` is unnecessary here.
This PR removes that redundancy with the following change:
https://github.com/sindresorhus/type-fest/pull/1319/files#diff-5ce63d84bd0497ef6dd72ef89147e7bf80f464a1185b1a34139c6c1cb6c37ee5R36

## Part 2
After the change in Part 1, this test doesn't pass:
https://github.com/sindresorhus/type-fest/blob/c671aa2e9dbbfb502fde09859286d0caf499eb7a/test-d/merge-deep.ts#L325

`Required` is added to fix this:
https://github.com/sindresorhus/type-fest/pull/1319/files#diff-5ce63d84bd0497ef6dd72ef89147e7bf80f464a1185b1a34139c6c1cb6c37ee5R61

## Note
The issue described above may be related to PR #1209, which has been suspended since August 31.
I intentionally did not include any changes related to that PR in order to keep this change minimal.

If these changes are considered inappropriate due to the context of PR #1209, I’m fine with closing this PR.

### DIFF
--- a/test-d/merge-deep.ts
+++ b/test-d/merge-deep.ts
@@ -1,5 +1,5 @@
 import {expectType} from 'tsd';
-import type {MergeDeep, MergeDeepOptions, IsEqual} from '../index.d.ts';
+import type {MergeDeep, MergeDeepOptions} from '../index.d.ts';
 
 // Test helper.
 declare function mergeDeep<


test-d/merge-deep.ts
```diff
@@ -165,7 +165,7 @@ expectType<{
 		fooBar: boolean;
 		items: number[];
 	} | undefined;
-}>(fooBarWithOptional);
+}>);
```
som-sm: Is this change necessary?

test-d/merge-deep.ts
```diff
@@ -309,3 +309,35 @@ expectType<[Bar, FooBarReplace, ...FooBarReplace[]]>(tupleIntoTupleWithVariadicR
 
 declare const tupleIntoTupleWithVariadicReplaceReversed: MergeDeep<[Foo, ...Foo[]], [number, Bar, ...Bar[]], {arrayMergeMode: 'replace'; recurseIntoArrays: true}>;
 expectType<[number, FooBarReplace, ...FooBarReplace[]]>(tupleIntoTupleWithVariadicReplaceReversed);
+
+// Test for https://github.com/sindresorhus/type-fest/issues/1314
```
som-sm: Remove this comment, these tests are unrelated to #1314 (refer https://github.com/sindresorhus/type-fest/issues/1314#issuecomment-3744593004).

test-d/merge-deep.ts
```diff
@@ -309,3 +309,35 @@ expectType<[Bar, FooBarReplace, ...FooBarReplace[]]>(tupleIntoTupleWithVariadicR
...
+	type: string;
+	b: {bb?: string; cc: string};
+};
+type RecordPartial = Partial<RecordNotPartial>;
+
+type testOptionalRecordReplacedByRecord = MergeDeep<RecordPartial, RecordNotPartial>;
+expectType<true>({} as IsEqual<RecordNotPartial, testOptionalRecordReplacedByRecord>);
+
+type testRecordReplacedByOptionalRecord = MergeDeep<RecordNotPartial, RecordPartial>;
+expectType<true>({} as IsEqual<RecordPartial, testRecordReplacedByOptionalRecord>);
```
som-sm:
```suggestion
expectType<RecordNotPartial>({} as MergeDeep<RecordPartial, RecordNotPartial>);
expectType<RecordPartial>({} as MergeDeep<RecordNotPartial, RecordPartial>);
```

test-d/merge-deep.ts
```diff
@@ -309,3 +309,35 @@ expectType<[Bar, FooBarReplace, ...FooBarReplace[]]>(tupleIntoTupleWithVariadicR
...
+	name: string;
+	type: string;
+	b: {bb: string; cc: string | undefined};
+};
+
+type testExpectedNotOptional = MergeDeep<NodeOptional, NodeNotOptional>;
+expectType<true>({} as IsEqual<NodeNotOptional, testExpectedNotOptional>);
+
+type testExpectedOptional = MergeDeep<NodeNotOptional, NodeOptional>;
+expectType<true>({} as IsEqual<NodeOptional, testExpectedOptional>);
```
som-sm: If I understand the need for these cases correctly, I think the following is better and simpler:

```suggestion
expectType<{a: string | undefined; b?: number; c?: boolean | undefined}>(
	{} as MergeDeep<{a: string; b: number; c: boolean}, {a: string | undefined; b?: number; c?: boolean | undefined}>,
);
```

In any case, we need a case that ensures explicitly specified `undefined`s (`{a?: string | undefined}`) in optional properties are preserved.

som-sm (commented):
@taiyakihitotsu Added few comments!

test-d/merge-deep.ts
taiyakihitotsu: Addressed.

test-d/merge-deep.ts
taiyakihitotsu: Addressed.

test-d/merge-deep.ts
taiyakihitotsu: Sorry, No.

In my previous environment, this caused a compile-time error that the LSP did not catch.
Now that the issue is resolved, I’ve reverted to the original one.

test-d/merge-deep.ts
taiyakihitotsu: Thanks, I missed this case!
(And I've added the reversed case.)

taiyakihitotsu:
@som-sm
Thanks for your review!
I’ve addressed all of your comments.

PR #1320, opened 2025-12-23
Sort numbers in unions while validating twoslash (`//=>`) types in codeblocks

som-sm:
This is an intermediate PR on #1309 that addresses https://github.com/sindresorhus/type-fest/pull/1309#discussion_r2638284054.


lint-rules/validate-jsdoc-codeblocks.js
som-sm, line 1: The conversion of double quotes to single quotes and spaces to tabs now happens as part of union sorting, not during quickinfo type extraction, since formatting changes made earlier would otherwise be overwritten by the sorting step.

som-sm:
@sindresorhus I've created an intermediate PR so that it's easy to review and refer back to these changes.

PR #1324, opened 2026-01-01
Add `ObjectMerge` type

som-sm:
This PR introduces a new `ObjectMerge` type that merges two object types in a way that aligns with runtime behaviour.

For example, merging `{a: string}` into `{a: number}` (`ObjectMerge<{a: number}, {a: string}>`) is a straightforward replacement, resulting in `{a: string}`. However, merging `{a?: string}` into `{a: number}` (`ObjectMerge<{a: number}, {a?: string}>`) is not as simple as producing `{a?: string}`. In this case, the correct result is `{a: number | string}` (or `{a: number | string | undefined}` when `exactOptionalPropertyTypes` is disabled).

`ObjectMerge` correctly handles such scenarios, and many others, which are explained in detail below.

---

## Why not fix `Merge`?

I initially intended this PR to fix the behaviour of our existing `Merge` type, but it looks like the `Merge` type is _only_ meant to be used at the type level, because its JSDoc example itself doesn't align with runtime behaviour.
https://github.com/sindresorhus/type-fest/blob/fe88ad8e4cee79dcb64f1c26f648319bfbcc4308/source/merge.d.ts#L13-L40

For example, `Merge<{[x: number]: unknown}, {[x: number]: number}>` cannot be `{[x: number]: number}` if it needs to align with runtime behaviour:
```ts
import type {Merge} from 'type-fest';

const left: {[x: number]: unknown} = {0: '1'};
const right: {[x: number]: number} = {1: 1};

const merged = {...left, ...right} as Merge<typeof left, typeof right>;
//=> {[x: number]: number}

if (0 in merged) {
  merged[0].toFixed(); // 💥 No compile time error, but fails at runtime.
}
```

So, `Merge` solves a different use case where we want properties from the second object to always override properties from the first object without considering runtime implications.

---

### Simple cases

In simple cases, `ObjectMerge` behaves exactly like TS's default inference, where overlapping properties are typed according to the second object.

```ts
declare const left: {a: string; b?: string};
declare const right: {a: number; c?: number};

const inferred = {...left, ...right};
//=> {a: number; c?: number; b?: string}

declare const objectMerge: ObjectMerge<typeof left, typeof right>;
//=> {a: number; c?: number; b?: string}
```

<details><summary>All cases</summary>
<p>

https://github.com/sindresorhus/type-fest/blob/55409ab82321ae9e6dc34d65d8830df679b01563/test-d/object-merge.ts#L6-L13

</p>
</details> 

### Optional properties
If the overlapping property is optional in the second object, the behaviour of `ObjectMerge` depends on the `exactOptionalPropertyTypes` compiler setting.

1. With `exactOptionalPropertyTypes` enabled, the behaviour is identical to TS's default inference.
    
    ```ts
    declare const left: {a: string; b?: string; c?: string};
    declare const right: {a?: number; b?: number; c: number};
    
    const inferred = {...left, ...right};
    //=> {a: string | number; b?: string | number; c: number}
    
    declare const objectMerge: ObjectMerge<typeof left, typeof right>;
    //=> {a: string | number; b?: string | number; c: number}
    ```
    
    <details><summary>All cases</summary>
    <p>

    https://github.com/sindresorhus/type-fest/blob/55409ab82321ae9e6dc34d65d8830df679b01563/test-d/object-merge.ts#L15-L38

    </p>
    </details> 

2. With `exactOptionalPropertyTypes` disabled, the default inference doesn't account for the fact that optional properties can also be set as `undefined`, which can lead to uncaught runtime errors.
    
    ```ts
    
    const left: {a: string} = {a: '0'};
    const right: {a?: number} = {a: undefined};
    
    const inferred = {...left, ...right};
    //=> {a: string | number}
    
    inferred.a.toString(); // 💥 No compile time error, but fails at runtime.
    
    declare const objectMerge: ObjectMerge<typeof left, typeof right>;
    //=> {a: string | number | undefined}
    
    // @ts-expect-error
    objectMerge.a.toString(); // ✅ Correctly errors at compile time.
    ```
    
    In this case, `ObjectMerge` accounts for the possibility of `undefined` by including it in the resulting type.

### Readonly properties
The behaviour of `ObjectMerge` is exactly like TS's default inference, i.e., the `readonly` modifier is removed from all properties across both objects.

```ts
declare const left: {a: string; readonly b: string; readonly c: string};
declare const right: {readonly a: number; readonly b: number; c: number};

const inferred = {...left, ...right};
//=> {a: number; b: number; c: number}

declare const objectMerge: ObjectMerge<typeof left, typeof right>;
//=> {a: number; b: number; c: number}
```

<details><summary>All cases</summary>
<p>

https://github.com/sindresorhus/type-fest/blob/55409ab82321ae9e6dc34d65d8830df679b01563/test-d/object-merge.ts#L40-L75

</p>
</details> 

### Number-like string keys
`ObjectMerge` recognises that number `0` and string `'0'` are the same thing, just like TS's default inference.
```ts
declare const left: {0: string; '1': string};
declare const right: {'0': number; 1: number};

const inferred = {...left, ...right};
//=> {'0': number; 1: number}

declare const objectMerge: ObjectMerge<typeof left, typeof right>;
//=> {'0': number; 1: number}
```

<details><summary>All cases</summary>
<p>

https://github.com/sindresorhus/type-fest/blob/55409ab82321ae9e6dc34d65d8830df679b01563/test-d/object-merge.ts#L195-L246

</p>
</details> 

### Index Signatures
1. In simple cases, `ObjectMerge` behaves exactly like TS's default inference.
    ```ts
    declare const left: {[x: string]: string};
    declare const right: {[x: string]: number};
    
    const inferred = {...left, ...right};
    //=> {[x: string]: string | number}
    
    declare const objectMerge: ObjectMerge<typeof left, typeof right>;
    //=> {[x: string]: string | number}
    ```

2. In more complicated cases, the default inference completely collapses and this is where `ObjectMerge` really shines.
    ```ts
    const left: {a: string} = {a: '1'};
    const right: {[x: string]: number} = {a: 1};
    
    const inferred = {...left, ...right};
    //=> {a: string}
    
    inferred.a.toUpperCase(); // 💥 No compile time error, but fails at runtime.
    
    declare const objectMerge: ObjectMerge<typeof left, typeof right>;
    //=> {[x: string]: string | number; a: string | number}
    
    // @ts-expect-error
    objectMerge.a.toUpperCase(); // ✅ Correctly errors at compile time.
    ```
    In this example, `ObjectMerge` correctly recognises that the `string` index signature from the second object has the potential to overwrite the `a` property from the first object, therefore it adjusts the resulting type accordingly.
    
    <br>
    
    ```ts
    const left: {[x: string]: string} = {0: 'foo', 1: 'bar'};
    const right: {[x: number]: number} = {0: 111, 3: 999};
    
    const inferred = {...left, ...right};
    //=> {}
    
    declare const objectMerge: ObjectMerge<typeof left, typeof right>;
    //=> {[x: number]: string | number; [x: string]: string | number}
    ```
    In this example, the inferred type is `{}`, which is not very useful. The type produced by `ObjectMerge`, however, remains accurate and usable.
    
    <br>
    
    ```ts
    const left: {[x: string]: number} = {a: 1};
    const right: {a?: string} = {};
    
    const inferred = {...left, ...right};
    //=> {a?: string}
    
    inferred.a?.toUpperCase(); // 💥 No compile time error, but fails at runtime.
    
    declare const objectMerge: ObjectMerge<typeof left, typeof right>;
    //=> {[x: string]: string | number; a?: string | number}
    
    // @ts-expect-error
    objectMerge.a?.toUpperCase(); // ✅ Correctly errors at compile time.
    ```
    In this example, the default inference ignores the possibility that `a` could be a `number` at runtime, whereas `ObjectMerge` correctly accounts for it.

    There are open issues in TS related to this, refer https://github.com/microsoft/TypeScript/issues/27273, https://github.com/microsoft/TypeScript/issues/56431.
    
    <details><summary>All cases</summary>
    <p>

    https://github.com/sindresorhus/type-fest/blob/55409ab82321ae9e6dc34d65d8830df679b01563/test-d/object-merge.ts#L91-L193

    </p>
    </details> 

There are many more cases that `ObjectMerge` handles beyond the ones shown here. There’s a fairly comprehensive test suite, and I’ve tried to make sure that all kinds of scenarios are covered and that `ObjectMerge` always stays in line with the actual runtime behaviour.

### DIFF
--- a/test-d/internal/normalized-keys.ts
+++ b/test-d/internal/normalized-keys.ts
@@ -19,3 +19,5 @@ expectType<never>({} as NormalizedKeys<never>);
 type T1 = NormalizedKeys<boolean>;
 // @ts-expect-error
 type T2 = NormalizedKeys<string[]>;
+// @ts-expect-error
+type T3 = NormalizedKeys<unknown>;
--- a/test-d/object-merge.ts
+++ b/test-d/object-merge.ts
@@ -267,6 +267,10 @@ expectType<{a: string}>({} as ObjectMerge<{(): string; a: number}, {a: string}>)
 type T1 = ObjectMerge<string, {a: string}>;
 // @ts-expect-error
 type T2 = ObjectMerge<{a: string}, number>;
+// @ts-expect-error
+type T3 = ObjectMerge<{a: string}, unknown>;
+// @ts-expect-error
+type T4 = ObjectMerge<unknown, {a: string}>;
 expectType<unknown>({} as ObjectMerge<string[], {0: string}>);
 expectType<unknown>({} as ObjectMerge<{0: string}, string[]>);
 expectType<unknown>({} as ObjectMerge<string[], number[]>);


test-d/internal/normalized-keys.ts
```diff
@@ -0,0 +1,21 @@
...
+
+// Boundary cases
+expectType<any>({} as NormalizedKeys<any>);
+expectType<never>({} as NormalizedKeys<never>);
+
+// Can only be `PropertyKey`
+// @ts-expect-error
+type T1 = NormalizedKeys<boolean>;
+// @ts-expect-error
+type T2 = NormalizedKeys<string[]>;
```
sindresorhus, line 21:
```suggestion
type T2 = NormalizedKeys<string[]>;
// @ts-expect-error
type T3 = NormalizedKeys<unknown>;
```

test-d/object-merge.ts
```diff
@@ -0,0 +1,288 @@
...
+expectType<{a: string}>({} as ObjectMerge<{a: string}, (a1: string, a2: number) => void>);
+expectType<{a: number}>({} as ObjectMerge<{a: string}, {(a1: string, a2: number): void; a: number}>);
+expectType<{a: string}>({} as ObjectMerge<() => string, {a: string}>);
+expectType<{a: string}>({} as ObjectMerge<{(): string; a: number}, {a: string}>);
+
+// === Non-objects ===
+// @ts-expect-error
+type T1 = ObjectMerge<string, {a: string}>;
+// @ts-expect-error
+type T2 = ObjectMerge<{a: string}, number>;
```
sindresorhus, line 269:
```suggestion
type T2 = ObjectMerge<{a: string}, number>;
// @ts-expect-error
type T3 = ObjectMerge<unknown, {a: string}>;
// @ts-expect-error
type T4 = ObjectMerge<{a: string}, unknown>;
```

sindresorhus:
Looks good and your arguments for why it should exist in addition to `Merge` makes sense to me.

PR #1329, opened 2026-01-22
`Get`: Fix union distribution for partial properties (#1205)

### LINKED ISSUE #1205: Possible incorrect behavior of 'GET<>'
### Bug description

source code of type: https://github.com/sindresorhus/type-fest/blob/main/source/get.d.ts

example:
```typescript
type TestType = {
  data:
    | {
        type: 'type1';
        someVal: number;
      }
    | {
        type: 'type2';
        someVal: string;
      }
    | {
        type: 'type3';
      };
};
/*
   Initial type:
"type1" | "type2" | "type3" - OK */
type Get1Type = Get<TestType, 'data.type'>;
/*
   Initial type:
   unknown - FAILED!
Expected: number | string | undefined */
type Get2Type = Get<TestType, 'data.someVal'>;
```

nhs04047:
Fixes #1205

  ## Problem

  When using `Get` with discriminated unions where some members don't have a property, it returns
   `unknown` instead of `value | undefined`.

  ```typescript
  type Test = {
    data:
      | {type: 'type1'; someVal: number}
      | {type: 'type2'; someVal: string}
      | {type: 'type3'}; // `someVal` doesn't exist here
  };

  // Before: unknown
  // After: number | string | undefined
  type Result = Get<Test, 'data.someVal'>;
```

  Solution

  - Added union distribution trigger (BaseType extends BaseType) in PropertyOf type
  - Changed fallback return type from unknown to undefined for non-existent properties

  Note

  The original code had [a comment](https://github.com/sindresorhus/type-fest/blob/fa55f482694f080ad051297bbb53f42bbbe4ae4b/test-d/get.ts#L30-L31) suggesting unknown was intentional due to structural typing.
  However, returning undefined is more practical for discriminated union use cases. Please let me
   know if this change doesn't align with the library's goals.

### DIFF
--- a/test-d/get.ts
+++ b/test-d/get.ts
@@ -27,8 +27,8 @@ expectTypeOf(get(apiResponse, 'hits.hits.0._source.name')).toEqualTypeOf<Array<{
 
 expectTypeOf(get(apiResponse, 'hits.hits[0]._source.name[0].given[0]')).toBeString();
 
-// TypeScript is structurally typed. It's *possible* this value exists even though it's not on the parent interface, so the type is `unknown`.
-expectTypeOf(get(apiResponse, 'hits.someNonsense.notTheRightPath')).toBeUnknown();
+// When accessing a non-existent property, returns `undefined` (changed from `unknown` to support union distribution).
+expectTypeOf(get(apiResponse, 'hits.someNonsense.notTheRightPath')).toEqualTypeOf<undefined>();
 
 type WithDictionary = {
 	foo: Record<string, {
@@ -64,7 +64,8 @@ expectTypeOf<Get<WithTuples, 'foo[0].bar', NonStrict>>().toBeNumber();
 expectTypeOf<Get<WithTuples, 'foo.0.bar', NonStrict>>().toBeNumber();
 
 expectTypeOf<Get<WithTuples, 'foo[1].baz', NonStrict>>().toBeBoolean();
-expectTypeOf<Get<WithTuples, 'foo[1].bar', NonStrict>>().toBeUnknown();
+// Property 'bar' doesn't exist in foo[1], returns undefined
+expectTypeOf<Get<WithTuples, 'foo[1].bar', NonStrict>>().toEqualTypeOf<undefined>();
 
 expectTypeOf<Get<WithTuples, 'foo[-1]', NonStrict>>().toBeUnknown();
 expectTypeOf<Get<WithTuples, 'foo[999]', NonStrict>>().toBeUnknown();
@@ -87,8 +88,9 @@ type WithNumberKeys = {
 expectTypeOf<Get<WithNumberKeys, 'foo[1].bar', NonStrict>>().toBeNumber();
 expectTypeOf<Get<WithNumberKeys, 'foo.1.bar', NonStrict>>().toBeNumber();
 
-expectTypeOf<Get<WithNumberKeys, 'foo[2].bar', NonStrict>>().toBeUnknown();
-expectTypeOf<Get<WithNumberKeys, 'foo.2.bar', NonStrict>>().toBeUnknown();
+// Key '2' doesn't exist in foo, returns undefined
+expectTypeOf<Get<WithNumberKeys, 'foo[2].bar', NonStrict>>().toEqualTypeOf<undefined>();
+expectTypeOf<Get<WithNumberKeys, 'foo.2.bar', NonStrict>>().toEqualTypeOf<undefined>();
 
 // Test `readonly`, `ReadonlyArray`, optional properties, and unions with null.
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by sindresorhus** [comment]
Thanks for working on this!

The `unknown` return type was an intentional design decision from the original `Get` implementation. TypeScript uses structural typing, so an object typd as `{a: number}` could actually be `{a: number, b: string}` at runtime. Accessing `b` is not necessarily `undefined`, it is genuinely unknown.

Returning `undefined` would be a lie. We cannot guarantee the property does not exist, only that we do not know about it.

For the discriminated union case in #1205, I think the real issue is that `Get` does not distribute over unions. The fix should add union distribution (the `BaseType extends BaseType` part of your PR is correct), but keep `unknown` as the fallback for properties that do not exist in *any* union member.

For properties that exist in *some* members but not others, the result should be `value | undefined` (which your distribution logic achieves). But for properties that exist in *no* members, `unknown` remains the correct answer.

File: test-d/get.ts [hunk 1]
```diff
@@ -152,3 +156,48 @@ expectTypeOf<WithDictionary>().toEqualTypeOf<Get<WithDictionary, readonly []>>()
...
+// Test union at root level
+type TestRootUnion =
+	| {a: number; common: string}
+	| {b: boolean; common: string};
+
+declare const testRootUnion: TestRootUnion;
+
+expectTypeOf(get(testRootUnion, ['common'] as const)).toEqualTypeOf<string>();
+expectTypeOf(get(testRootUnion, ['a'] as const)).toEqualTypeOf<number | undefined>();
+expectTypeOf(get(testRootUnion, ['b'] as const)).toEqualTypeOf<boolean | undefined>();
```
sindresorhus, line 203: Some more tests:

```ts
// Test any/never/unknown base types
expectTypeOf<Get<any, 'foo'>>()...
expectTypeOf<Get<never, 'foo'>>()...
expectTypeOf<Get<unknown, 'foo'>>()...

// Union where property exists in NO members (should be unknown)
type AllLacking = {a: 1} | {b: 2};
expectTypeOf<Get<AllLacking, 'c'>>().toBeUnknown();

// Union where property exists in SOME members (should be value | undefined)
type SomeHave = {a: 1; c: string} | {b: 2};
expectTypeOf<Get<SomeHave, 'c'>>().toEqualTypeOf<string | undefined>();
```

som-sm:
> The `unknown` return type was an intentional design decision from the original `Get` implementation. TypeScript uses structural typing, so an object typd as `{a: number}` could actually be `{a: number, b: string}` at runtime. Accessing `b` is not necessarily `undefined`, it is genuinely unknown.
> 
> Returning `undefined` would be a lie. We cannot guarantee the property does not exist, only that we do not know about it.

Completely agree!

<hr>

> I think the real issue is that `Get` does not distribute over unions.

`Get` already distributes over unions. The distribution happens at the following line:
https://github.com/sindresorhus/type-fest/blob/fa55f482694f080ad051297bbb53f42bbbe4ae4b/source/get.d.ts#L131

<br>

> The fix should add union distribution (the `BaseType extends BaseType` part of your PR is correct)

`BaseType` is already distributed, so doing `BaseType extends BaseType` is actually _not_ necessary.

<br>

> For properties that exist in _some_ members but not others, the result should be `value | undefined` (which your distribution logic achieves). But for properties that exist in _no_ members, `unknown` remains the correct answer.

I think this would be confusing. For example, if `Get<{a: string}, "b">` returns `unknown`, then `Get<{a: string} | {b: number}, "b">` should return `unknown` (`unknown | number`) and _not_ `number | undefined`. 

---

IMO, the existing behaviour of returning `unknown` is fine. And anyways, at runtime, you would _not_ be able to safely access a property that does not exist on all union members.

Here's an example that shows why returning `undefined` could be problematic. It's a bit contrived but it explains the problem.
```ts
type Test =
  | { type: 'type1'; someVal: number }
  | { type: 'type2'; someVal: number }
  | { type: 'type3' }; // `someVal` doesn't exist here

const type3 = { type: 'type3', someVal: true } as const;
const test: Test = type3;

const someVal = (test as any).someVal as number | undefined; // Notice you can't safely access `someVal` from `test`
someVal?.toFixed(); // No compile error, but fails at runtime
```
Playground: https://tsplay.dev/NrAG0w

@nhs04047 Do you have a real-world use case where returning `undefined` would be a better choice?

<br>

IMO, if this is _really_ required, this should be behind a flag, something like `returnUndefinedForMissingProperties`. @sindresorhus Thoughts?

sindresorhus:
> I think this would be confusing. For example, if Get<{a: string}, "b"> returns unknown, then Get<{a: string} | {b: number}, "b"> should return unknown (unknown | number) and not number | undefined.

Good point 👍

sindresorhus:
> IMO, if this is really required, this should be behind a flag, something like returnUndefinedForMissingProperties. @sindresorhus Thoughts?

I would just close this, but lets wait for response from OP.

nhs04047:
Thanks @sindresorhus and @som-sm for the thoughtful review!

I agree with your points.
Structural typing means we can't guarantee a property is `undefined` just because it's not defined in the type — it's genuinely `unknown` at runtime. And as @som-sm pointed out, treating unions differently from single types would break consistency.

Closing this PR!

PR #1332, opened 2026-01-24
Add `ArrayExclude` type

jericirenej:
## Type description and examples
This helper can improve DX by offering a straightforward way of restricting the type of a (readonly) array or a tuple on the basis of a provided type / discriminative union. 

```ts
import type {ArrayExclude} from 'type-fest';

type MixedTuple = [1, 2, string, 3, 4, string, null, undefined];
type NumbersOnlyTuple = ArrayExclude<MixedTuple, string|null|undefined>
//=> [1, 2, 3, 4]

type GradesTuple = [1, 2, 3, 1, 2, 3, 4, 5];
type ExcludedOddsTuple = ArrayExclude<GradesTuple, 1|3|5>;
//=> [2, 2, 4]

type GradesReadonly = readonly [1, 2, 3, 1, 2, 3, 4, 5];
type ExcludedOddsReadonly = ArrayExclude<GradesTuple, 1|3|5>;
//=> readonly [2, 2, 4]

type OrdinaryArray = (string|number|boolean)[];
type NumbersOnlyTuple = ArrayExclude<OrdinaryArray, string|boolean>;
//=>number[];

```

Exclusion types work for primitive and non-primitive values. 

There is a caveat when we are dealing with tuples or readonly arrays that contain elements that are typed as a `boolean` - i.e. not specifically as `true` or `false` - during type inference this results in a discriminative union type return. This issue does not appear for regular arrays.

I read the contribution guidelines, but wasn't sure if I was supposed to raise an issue/proposal first. However, since I already built the type, I created the PR directly. Hopefully that's OK, but I can create a proposal if that is required.

- [X] Tried my best to look for it.

### DIFF
--- a/test-d/array-exclude.ts
+++ b/test-d/array-exclude.ts
@@ -120,6 +120,23 @@ expectType<ExpectedTupleOmitSeveral>(tupleOmitSeveral);
 expectType<never[]>(tupleNever);
 expectType<readonly never[]>(readonlyNever);
 
+// UNION TYPES
+declare const arrayUnionType: ArrayExclude<Array<1 | 2 | 3>, 1>;
+declare const tupleUnionType: ArrayExclude<[1 | 2, 3 | 4], 1 | 4>;
+declare const readonlyUnionType: ArrayExclude<readonly [1 | 2, 3 | 4], 1 | 4>;
+expectType<Array<2 | 3>>(arrayUnionType);
+expectType<[2, 3]>(tupleUnionType);
+expectType<readonly [2, 3]>(readonlyUnionType);
+
+// BOOLEANS
+declare const arrayBoolean: ArrayExclude<Array<number | boolean | string>, string>;
+declare const tupleBoolean: ArrayExclude<[number, boolean], number>;
+declare const readonlyBoolean: ArrayExclude<readonly [number, boolean], number>;
+/* Declare const readonlyUnionType: ArrayExclude<readonly [1 | 2, 3 | 4], 1 | 4>; */
+expectType<Array<number | boolean>>(arrayBoolean);
+expectType<[boolean]>(tupleBoolean);
+expectType<readonly [boolean]>(readonlyBoolean);
+
 // INVALID TYPES
 // @ts-expect-error
 declare const invalidPrimitiveSource: ArrayExclude<string, 'something'>;
@@ -130,15 +147,6 @@ declare const invalidObjectSource: ArrayExclude<
 >;
 expectType<never>(invalidPrimitiveSource);
 expectType<never>(invalidObjectSource);
-
-// BOOLEAN EDGE CASE
-
-// If the tuple has a boolean type specification for one of its elements
-// this results in discriminative union return
-declare const tupleBoolean: ArrayExclude<[number, boolean], number>;
-declare const readonlyBoolean: ArrayExclude<readonly [number, boolean], number>;
-expectType<[false] | [true]>(tupleBoolean);
-expectType<readonly [false] | readonly [true]>(readonlyBoolean);
 
 // DOCUMENTATION EXAMPLES
 declare const docsOrdinaryArray: ArrayExclude<Array<string | number | boolean>, string | boolean>;
--- a/source/array-exclude.d.ts
+++ b/source/array-exclude.d.ts
@@ -1,6 +1,12 @@
 import type {IsArrayReadonly, SetArrayAccess} from './internal/array.d.ts';
 import type {UnknownArray} from './unknown-array.d.ts';
 
+type FilteredElement<Element, ExcludeConditions> = Element extends ExcludeConditions
+	? never
+	: Exclude<Element, ExcludeConditions> extends never
+		? never
+		: Exclude<Element, ExcludeConditions>;
+
 /**
 Exclude types from the source array based on the supplied exclude conditions.
 The source array can be an ordinary array, a readonly array or a tuple.
@@ -21,10 +27,6 @@ type Tuple = ArrayExclude<['literalValue', 1, 2, 3, {prop: 'prop'}, true], strin
 Notes:
 	- If the provided conditions filter out all elements from the source array, a `never[]` type is returned.
 	- If `never` is provided as the exclude condition, then the original type will be returned.
-	- If a tuple or readonly array has elements that are defined as `boolean`
-	(i.e. not narrowed as `true` or `false`) this will result in a discriminative
-	union return. For example, an `ArrayExclude<[number, boolean], number>` will result in
-	a `[true]|[false]` type.
 
 @category Array
  */
@@ -38,14 +40,13 @@ export type ArrayExclude<
 | [infer First, ...infer Rest]
 	? Rest extends []
 		? SetArrayAccess<
-			First extends ExcludeConditions
+			FilteredElement<First, ExcludeConditions> extends never
 				? FilteredTuple extends [] ? never[] : FilteredTuple
-				: [...FilteredTuple, First],
+				: [...FilteredTuple, FilteredElement<First, ExcludeConditions>],
 			IsReadonly
 		>
 		:
-		ArrayExclude<Rest, ExcludeConditions, First extends ExcludeConditions ? FilteredTuple : [...FilteredTuple, First], IsReadonly>
-
+		ArrayExclude<Rest, ExcludeConditions, FilteredElement<First, ExcludeConditions> extends never ? FilteredTuple : [...FilteredTuple, FilteredElement<First, ExcludeConditions>], IsReadonly>
 	: ArrayLike extends Array<infer Union> | ReadonlyArray<infer Union>
 		? SetArrayAccess<Array<Exclude<Union, ExcludeConditions>>, IsReadonly>
 		: never;


sindresorhus:
I think this could be simplified by using the existing IsArrayReadonly and SetArrayAccess helpers from source/internal/array.d.ts

test-d/array-exclude.ts
```diff
@@ -0,0 +1,150 @@
...
+expectType<readonly [false] | readonly [true]>(readonlyBoolean);
+
+// DOCUMENTATION EXAMPLES
+declare const docsOrdinaryArray: ArrayExclude<Array<string | number | boolean>, string | boolean>;
+declare const docsReadonlyArray: ArrayExclude<readonly ['literalValue', 1, 2, 3, {readonly prop: 'prop'}, true], string | 2 | Record<string, unknown>>;
+declare const docsTuple: ArrayExclude<['literalValue', 1, 2, 3, {prop: 'prop'}, true], string | 2 | Record<string, unknown>>;
+
+expectType<number[]>(docsOrdinaryArray);
+expectType<readonly [1, 3, true]>(docsReadonlyArray);
+expectType<[1, 3, true]>(docsTuple);
```
sindresorhus, line 158: Some more tests:


```suggestion
expectType<[1, 3, true]>(docsTuple);

// readonly non-tuple arrays
declare const readonlyArrayExclude: ArrayExclude<readonly (string | number | boolean)[], string | boolean>;
declare const readonlyArrayNoOp: ArrayExclude<readonly string[], number>;
declare const readonlyArrayNever: ArrayExclude<readonly string[], string>;
expectType<readonly number[]>(readonlyArrayExclude);
expectType<readonly string[]>(readonlyArrayNoOp);
expectType<readonly never[]>(readonlyArrayNever);

// `never` input type
declare const neverArray: ArrayExclude<never, string>;
expectType<never>(neverArray);
```

jericirenej: Thanks for the quick response and input, I haven't noticed those internal helpers at all.  I'll try and update the PR and add those tests as well! :+1:

jericirenej:
It was a good thing those extra tests were added, the handling of non-tuple readonly arrays still needed some fixes :+1:

jericirenej:
> Usually, for complicated types like this, it's better to first discuss the different cases before jumping into the implementation, it just saves time and effort.

Thanks for clarifying this, I'll keep it in mind. Initially it seemed like a simple type, but I should have known better that wen things seem simple, that is always a huge red flag :smile: 

Thanks for pointing out the other variants that need to be addressed. I'll take a look!

som-sm:
> Probably `ArrayExclude<[string, number?], number]>` should return `[string, never?]`, and to remove the entire optional element, we need to add `| undefined`, like `ArrayExclude<[string, number?], number | undefined]>` should return `[string]`.

Umm...now I feel like `ArrayExclude<[string, number?], number]>` should return `[string]` i.e., it should completely remove the optional element because if `exactOptionalPropertyTypes` is disabled, `undefined` is not allowed for optional elements. If `exactOptionalPropertyTypes` is enabled, then it makes sense to return `[string, never?]` (which is the same as `[string, undefined?]` in this case).

Also, if `exactOptionalPropertyTypes` is disabled, I guess `ArrayExclude<[string, number?], undefined>` should do nothing, the result shouldn't be `[string, number]` (optional modifier removed). Thoughts?

jericirenej:
I was just thinking about this as well. For me, it seems better that the type is removed completely in that case, because otherwise the optional signature should not impose more constraints on the type definition than a non-optional one.

In other words: If we dont' want the number element to be present, its optional signature should not interfere with that. Also, what if we have `[string, number?, boolean]` and we want to remove the number? Having `[string, boolean]` seems easier to work with than `[string, never?, boolean]`.

I'll push the discriminative union solution to the branch, then I can look into the other two issues you mentioned. I'll try looking in your most recent proposed direction for the optional types. It might also be easier to implement than the initial approach.

EDIT: have no idea why the CI job is failing, locally test runs fine. The complain in the job seems unrelated to current changes.

som-sm:
> Also, what if we have `[string, number?, boolean]`

`[string, number?, boolean]` is _not_ a valid tuple type, required elements cannot follow optional elements.

> I was just thinking about this as well. For me, it seems better that the type is removed completely in that case, because otherwise the optional signature should not impose more constraints on the type definition than a non-optional one.

You mean remove the optional element completely even when `exactOptionalPropertyTypes` is disabled? Like `ArrayExclude<[string, number?], number>` should _also_ return `[string]`? I don't think that should be the case, because with that setting disabled, `undefined` is a valid value for optional elements, so it should be explicitly excluded.

jericirenej:
> `[string, number?, boolean]` is _not_ a valid tuple type, required elements cannot follow optional elements.

Sorry about that invalid tuple type, you are absolutely correct about this - but let's say -- if we went with the strategy of element removal --  that another optional tuple element is after the one we want to exclude, for example `[string, number?, boolean?]`. In this case we might want to have it return `[string, boolean?]`.

> You mean remove the optional element completely even when exactOptionalPropertyTypes is disabled?

Haven't thought about that, you provide a valid point. I think a safe approach would be to leave it as `undefined` and to only remove it when explicitly listed in the generic.
This also resolves the discussion about the optional element removal, as it will be taken care by specifying `undefined` among the exclusion criteria?

EDIT:Formatting

som-sm:
> EDIT: have no idea why the CI job is failing, locally test runs fine. The complain in the job seems unrelated to current changes.

Please ignore that, that's unrelated! I reran the tests, and nothings failing now. It's a flaky test somewhere else.

jericirenej:
### Tuple handling
I haven't been able to figure out a nice way of handling optional tuple elements. 

While I can check for the optional status by comparing lengths differences for the normal array vs. when the first element is explicitly set as required, this only works for that single optional element - but the inference of the rest of the array does not make it easy to preserve those optional props.

For example, when using `ArraySlice` to infer the rest of the arguments:

```ts
import { ArraySlice } from './array-slice.js';
import { IsEqual } from './is-equal.js';
import { SetRequired } from './set-required.js';

type TransformOptional<ArrayLike extends UnknownArray> = 
   IsEqual<SetRequired<ArrayLike, 0>['length'], ArrayLike['length']> extends false 
     ? [SetRequired<ArrayLike, 0>[0] | undefined, ...ArraySlice<ArrayLike, 1>] 
     : ArrayLike;

type B = IsOptionalElement<[number?, string?, boolean?]>;
//=> [number | undefined]


```

I'm not seeing any nice solution here for now :disappointed: 

### Rest params
I haven't been able to investigate the tuple inference yet. I agree that currently it's not optimal, as using rest parameters at the beginning falls back to a non-tuple inference, while using them at the end, can result in a `never[]` type or an otherwise incorrect type being inferred.

```ts
type RestAtBeginning = ArrayExclude<[...string[], number, number, boolean], string>;
//=> (number|boolean)[]

type RestAtEnd1 = ArrayExclude<[number, number, boolean, ...string[]], string>; 
//=> never[]
type RestAtEnd1 = ArrayExclude<[number, number, boolean, ...string[]], number> 
//=> string[]
```

I hope I'll have some time during the week to investigate this, but any ideas are welcome :)

som-sm:
@jericirenej You can use the following implementation, it should handle _all_ different scenarios:

```ts
import type {And} from './and.d.ts';
import type {If} from './if.d.ts';
import type {IsArrayReadonly} from './internal/array.d.ts';
import type {IsExactOptionalPropertyTypesEnabled, Not} from './internal/type.d.ts';
import type {IsNever} from './is-never.d.ts';
import type {IsOptionalKeyOf} from './is-optional-key-of.d.ts';
import type {IsTuple} from './is-tuple.d.ts';

export type ArrayExclude<TArray extends UnknownArray, TExclude> =
	TArray extends UnknownArray // For distributing `TArray`
		? If<
			IsTuple<TArray, {fixedLengthOnly: false}>,
			TupleExclude<TArray, TExclude>,
			Array<Exclude<TArray[number], TExclude>>
		> extends infer Result
			? If<IsArrayReadonly<TArray>, Readonly<Result>, Result>
			: never
		: never; // Should never happen

type TupleExclude<
	TTuple extends UnknownArray,
	TExclude,
	ForwardAccumulator extends UnknownArray = [],
	BackwardAccumulator extends UnknownArray = [],
> =
	TTuple extends UnknownArray // For distributing `TArray`
		? keyof TTuple & `${number}` extends never
			// Enters this branch, if `TArray` is empty (e.g., `[]`),
			// or `TArray` contains no non-rest elements preceding the rest element (e.g., `[...string[]]` or `[...string[], string]`).
			? TTuple extends readonly [...infer Rest, infer Last]
				? TupleExclude<
					Rest,
					TExclude,
					ForwardAccumulator,
					Exclude<Last, TExclude> extends infer ExcludedLast
						? If<IsNever<ExcludedLast>, BackwardAccumulator, [ExcludedLast, ...BackwardAccumulator]>
						: never
				>
				: Exclude<TTuple[number], TExclude> extends infer ExcludedRestElement
					? If<
						IsNever<ExcludedRestElement>,
						[...ForwardAccumulator, ...BackwardAccumulator],
						[...ForwardAccumulator, ...ExcludedRestElement[], ...BackwardAccumulator]
					>
			: TTuple extends readonly [(infer First)?, ...infer Rest]
				? TupleExclude<
					Rest,
					TExclude,
					IsOptionalKeyOf<TTuple, '0'> extends infer IsFirstOptional extends boolean
						? Exclude<
							First | (If<And<IsFirstOptional, Not<IsExactOptionalPropertyTypesEnabled>>, undefined, never>), // Add `| undefined` for optional elements, if `exactOptionalPropertyTypes` is disabled.
							TExclude
						> extends infer ExcludedFirst
							? If<
								IsNever<ExcludedFirst>,
								ForwardAccumulator,
								[...ForwardAccumulator, ...(If<IsFirstOptional, [ExcludedFirst?], [ExcludedFirst]>)]
							>
						: never, // Should never happen
					BackwardAccumulator
				>
				: never // Should never happen, since `[(infer First)?, ...infer Rest]` is a top-type for arrays.
		: never; // Should never happen

export {};
```

<br>

It's tested using the following cases:

```ts
import type {ArrayExclude} from '../index.d.ts';

// All required elements
expectType<[1, 3, 4]>({} as ArrayExclude<[1, 2, 3, 4], 2>);

// All optional elements
expectType<[1?, 3?, 4?]>({} as ArrayExclude<[1?, 2?, 3?, 4?], 2>);

// Mixed required and optional elements
expectType<[1, 3?, 4?]>({} as ArrayExclude<[1, 2, 3?, 4?], 2>);
expectType<[1, 3?, 4?]>({} as ArrayExclude<[1, 2?, 3?, 4?], 2>);

// Explicit undefined in optional elements
expectType<[1, undefined?, 3?]>({} as ArrayExclude<[1, (2 | undefined)?, 3?], 2>);

// Union elements
expectType<[1, 3, 4 | 5]>({} as ArrayExclude<[1, 2 | 3, 4 | 5], 2>);

// Excluding unions
expectType<[1, 3, 5]>({} as ArrayExclude<[1, 2 | 3, 4 | 5], 2 | 4>);

// Required and rest element
// Trailing rest element
expectType<[2, ...string[]]>({} as ArrayExclude<[1, 2, ...string[]], 1>);
expectType<string[]>({} as ArrayExclude<[1, 2, ...string[]], 1 | 2>);
// Leading rest element
expectType<[...string[], 2]>({} as ArrayExclude<[...string[], 1, 2], 1>);
expectType<string[]>({} as ArrayExclude<[...string[], 1, 2], 1 | 2>);
// Rest element in the middle
expectType<[1, ...string[], 3]>({} as ArrayExclude<[1, 2, ...string[], 3, 4], 2 | 4>);
expectType<string[]>({} as ArrayExclude<[1, 2, ...string[], 3, 4], 1 | 2 | 3 | 4>);

// Optional and rest element
expectType<[1?, ...string[]]>({} as ArrayExclude<[1?, 2?, ...string[]], 2>);
expectType<string[]>({} as ArrayExclude<[1?, 2?, ...string[]], 1 | 2>);

// Required, optional, and rest element
expectType<[1, 4?, ...string[]]>({} as ArrayExclude<[1, 2, 3?, 4?, ...string[]], 2 | 3>);

// Excluding rest element
// Trailing rest element
expectType<[1, 2]>({} as ArrayExclude<[1, 2, ...string[]], string>);
expectType<[1?, 2?]>({} as ArrayExclude<[1?, 2?, ...string[]], string>);
expectType<[1, 2, 3?, 4?]>({} as ArrayExclude<[1, 2, 3?, 4?, ...string[]], string>);
expectType<[1, 2, ...number[]]>({} as ArrayExclude<[1, 2, ...Array<string | number>], string>);
expectType<[1?, 2?, ...number[]]>({} as ArrayExclude<[1?, 2?, ...Array<string | number>], string>);
expectType<[1, 2, 3?, 4?, ...number[]]>({} as ArrayExclude<[1, 2, 3?, 4?, ...Array<string | number>], string>);
// Leading rest element
expectType<[1, 2]>({} as ArrayExclude<[...string[], 1, 2], string>);
expectType<[...number[], 1, 2]>({} as ArrayExclude<[...Array<string | number>, 1, 2], string>);
// Rest element in the middle
expectType<[1, 2, 3]>({} as ArrayExclude<[1, ...string[], 2, 3], string>);
expectType<[1, ...number[], 2, 3]>({} as ArrayExclude<[1, ...Array<string | number>, 2, 3], string>);

// Excluding both rest element and non-rest elements
expectType<[1, 3?, 5?, ...number[]]>({} as ArrayExclude<[1, 2, 3?, (4 | 5)?, ...Array<string | number>], 2 | 4 | string>);

// All elements excluded
expectType<[]>({} as ArrayExclude<[1, 2, 3], 1 | 2 | 3>);
expectType<[]>({} as ArrayExclude<[string, number], string | number>);
expectType<[]>({} as ArrayExclude<[1, 2?, 3?, ...string[]], any>);
expectType<[]>({} as ArrayExclude<[1, 2?, 3?, ...string[]], unknown>);

// No elements excluded
expectType<[1, 2, 3]>({} as ArrayExclude<[1, 2, 3], 4>);
expectType<[1, 2?, 3?, ...string[]]>({} as ArrayExclude<[1, 2?, 3?, ...string[]], never>);

// Non-tuple arrays
expectType<string[]>({} as ArrayExclude<Array<string | number>, number>);
expectType<never[]>({} as ArrayExclude<Array<string | number>, string | number>);
```

Note: This is _just_ a quick set of test cases I used for validating the implementation, we'd need more cases before getting this merged.

jericirenej:
Wow, what a nice and thorough implementation! 

Didn't have that much experience with certain combinations of using `extends` to point to inferred type, but it sure does help, as does wrapping the infer operator within brackets to allow for an optional modifier. Test structuring is also cleaner now, with object assertions being much more concise than declaring constants that are only ever used as types. :+1: 

@som-sm Since you re-implemented this from scratch - instead of me comitting this, would it perhaps be more appropriate for me to close this PR and you open your own? It doesn't seem right for me to take the credit for this :grin:

som-sm:
> Since you re-implemented this from scratch - instead of me comitting this, would it perhaps be more appropriate for me to close this PR and you open your own? It doesn't seem right for me to take the credit for this 😁

@jericirenej I can commit changes to this PR itself, or you can close this and I'll open a separate PR, whatever you're comfortable with. I'm fine with anything!

jericirenej:
Well, by this stage, the only valid input left from my side is in suggesting this feature, so I'm closing this PR because credit should go where credit is due: @som-sm  :wink: 

Thanks for the brainstorming, this will be a great helper type!

PR #1338, opened 2026-01-27
fix: `IsEqual`, `{a: t, b: s}` and `{a: t} & {b: s}` are equal

taiyakihitotsu:
Current: https://github.com/sindresorhus/type-fest/pull/1338#issuecomment-3806872235 -> https://github.com/sindresorhus/type-fest/pull/1338#issuecomment-3814008065 -> https://github.com/sindresorhus/type-fest/pull/1338#issuecomment-3842033116 -> https://github.com/sindresorhus/type-fest/pull/1338#issuecomment-3854491753 -> https://github.com/sindresorhus/type-fest/pull/1338#issuecomment-4103309284

---
 
This PR is separated from #1336.

(See https://github.com/sindresorhus/type-fest/pull/1336#issuecomment-3804951968)

## Major Changes
The previous (meaning before this PR) `IsEqual<{a: 0} & {b: 0}, {a: 0; b: 0}>` returns `false`.
This PR prevents it to return `true` via ~~`Simplify`~~`SimplifyDeep` if both are `extends object`.

(intersection of objects, it's deprecated though.
https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#intersection-types)

## Minor Changes
Refactor: delete unused `import`


taiyakihitotsu:
~~I found this definition is incorrect, so made this PR to draft now.~~

~~I reopen it after fixing it.~~

Reopened.

taiyakihitotsu:
@som-sm 

I've finally fixed it.

Now `IsEqual<A, B>` returns `true` correctly if `A` and `B` are objects and mutually assignable.
This means `IsEqual` now treats `{a: t, b: u}` as equal to `{a: t} & {b: u}` recursively.

I also removed the test case (commited in https://github.com/sindresorhus/type-fest/pull/1338/changes/dabe872cccfa18527525409fdbab5447a070c2ee) below because the four new test cases in this PR already cover it. (Please let me know if I missed anything.)

```typescript
// Distinct whether an object is merged by `&` or via `Simplify`, to ensure Branded Types are handled strictly.
export type IntersectionMerge<tuple extends readonly unknown[]> = Except<tuple, 'length'> & {__brand: 'tag'};
type SampleTuple = [0, 1, 2];

expectType<true>({} as IsEqual<{a: 0} & {b: 0}, {a: 0; b: 0}>);
expectType<true>({} as IsEqual<Simplify<IntersectionMerge<SampleTuple>>, Merge<IntersectionMerge<SampleTuple>, IntersectionMerge<SampleTuple>>>);
```

Thanks!

taiyakihitotsu:
I found this case failed even if just using `SimplifyDeep` in `IsEqual`.

```typescript
expectType<true>({} as IsEqual<{aa: {a: {x: 0} & ({y: 0} | {y: 0})} & {b: 0}}, {aa: {a: {x: 0; y: 0}; b: 0}}>); // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
```

```typescript
export type IsEqual<A, B> =
	[A, B] extends [B, A]
		? [A, B] extends [object, object]
			? _IsEqual<SimplifyDeep<A>, SimplifyDeep<B>>
			: _IsEqual<A, B>
		: false;
```

For example, this is a simple case.
https://github.com/taiyakihitotsu/type-fest/blob/fb14a98d577d543ba3eebab454f28fda5a0b5b5f/source/internal/type.d.ts#L218

```typescript
type NT = _IsEqual<{z: {a: 0}}, {z: {a: 0} | {a: 0}}>; // => false
```

Using `SimplifyDeep` can eliminate intersections of `A & A` which should be equal to `A`, and `{a: t} & {b: u}` which should be equal to `{a: t, b: u}`.
But it cannot remove unions like `A | A` which should be equal to `A`.

To remove duplicated types in union, I defined `UniqueUnionDeep` in `internal/type.d.ts`.
(I don't know this helper will be needed for the other definitions though.)

---

Please let me know if I've missed any cases.

taiyakihitotsu:
After a sleep, I've found another error in the definition of `UniqueUnionDeep`.

```typescript

export type UniqueUnion<U> = _UniqueUnion<U, never>;
type _UniqueUnion<U, R> =
	LastOfUnion<U> extends infer K
		? [K] extends [never]
			? R
			: _UniqueUnion<Exclude<U, K>, K extends R ? R : R | K>
		: never;
```

`extends` is completely wrong here.

```typescript
 _UniqueUnion<Exclude<U, K>, K extends R ? R : R | K>
```

~~I will reopen after fixing it.~~

Reopened.

taiyakihitotsu:
## Status
I've added 5 helper type functions in `source/internal/type.d.ts`.

 - `MatchOrNever`: used to define `ExcludeExactly`
 - `ExcludeExactly`: used to define `UniqueUnion`, which would "fix" `Exclude` / `ExcludeStrict` as below.
 - `LastOfUnion`: used to define `UniqueUnion`
 - `UniqueUnion`: used to define `UniqueUnionDeep`
 - `UniqueUnionDeep`: used to define `IsEqual`

This PR passes those test cases (which fails before this PR):

```typescript
// Ensure `{a: t; b: s}` === `{a: t} & {b: s}`, not equal to `{a: u} & {b: v}` if `u` !== `t` or `v` !== `s`.
expectType<true>({} as IsEqual<{a: 0} & {b: 0}, {a: 0; b: 0}>);
expectType<true>({} as IsEqual<{aa: {a: {x: 0} & {y: 0}} & {b: 0}}, {aa: {a: {x: 0; y: 0}; b: 0}}>);
// Ensure `{a: t} | {a: t}` === `{a: t}`
expectType<true>({} as IsEqual<{a: 0} & ({b: 0} | {b: 0}), {a: 0; b: 0}>); // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
expectType<true>({} as IsEqual<{aa: {a: {x: 0} & ({y: 0} | {y: 0})} & {b: 0}}, {aa: {a: {x: 0; y: 0}; b: 0}}>); // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
```

I've submitted this to fix `IsEqual`, ready for review.
And I need a discussion about `ExcludeExactly`.

## Background
I found the bug described above, where unions inside recursive objects cannot be compared directly by the previous `IsEqual` definition which used `SimplifyDeep`.

To resolve this, I implemented `UniqueUnion` to deduplicate types within unions.

```typescript
export type UniqueUnion<U> = _UniqueUnion<U, never>;
type _UniqueUnion<U, R> =
	LastOfUnion<U> extends infer K
		? [K] extends [never]
			? R
			: _UniqueUnion<
				Exclude<U, K>, // (1)
				(<G>() => G extends K & G | G ? 1 : 2) extends
				(<G>() => G extends R & G | G ? 1 : 2)
					? [R, unknown] extends [never, K]
						? K
						: R
					: R | K>
		: never;
```

I think this is the idiomatic way to iterate over union types, but it didn't work as expected.
I finally found that `Exclude` (at (1)) doesn't distinguish between `{readonly a: 0}` and `{a: 0}` for the following reason:: https://www.typescriptlang.org/docs/handbook/utility-types.html#excludeuniontype-excludedmembers

> Constructs a type by excluding from UnionType all union members that are **assignable** to ExcludedMembers.

`ExcludeStrict` in `type-fest` is also the same.
(I commented those details in the `ExcludeExactly` section in `source/internal/type.d.ts`.)

So I decided to define `ExcludeExactly` to satisfy requirement `(1)` which must distinguish `readonly` and `optional` avoid incorrectly removing `{a: 0}` when compared to `{readonly a: 0}`.

And I added the helper function: `MatchOrNever`.
This `MatchOrNever` has a similar construct to `_IsEqual`, so it has the same limitations as noted in the comments.

```typescript
export type MatchOrNever<A, B> =
	[unknown, B] extends [A, never]
		? A
		// This equality code base below doesn't work if `A` is `unknown` and `B` is `never` case.
		// So this branch should be wrapped to take care of this.
		: (<G>() => G extends A & G | G ? 1 : 2) extends (<G>() => G extends B & G | G ? 1 : 2)
			? never
			: A;
```

I've finished creating the strict `IsEqual`.

```typescript
export type IsEqual<A, B> =
	[A, B] extends [B, A]
		? [A, B] extends [object, object]
			? _IsEqual<SimplifyDeep<UniqueUnionDeep<A>>, SimplifyDeep<UniqueUnionDeep<B>>>
			: _IsEqual<A, B>
		: false;
```

---

@som-sm

### PR for `IsEqual` here
Now `IsEqual` is able to judge the four cases described above correctly.

But this is involving many diffs in internal.
I want to point them out, e.g., in case I missed some test cases.

### PR for `ExcludeExactly`
Should I separate this PR and create another branch to fix `ExcludeStrict`?
I think this is not necessarily "a bug" because the `ExcludeStrict` comment refers to assignability [here](https://github.com/sindresorhus/type-fest/blob/main/source/extends-strict.d.ts#L5):

> A stricter, non-distributive version of `extends` for checking whether one type is **assignable** to another.

But I also think that another version of `Exclude` is needed to maintain strict types.

If there is no need to touch `ExcludeStrict`, I would like to propose adding `ExcludeExactly` as a new type.

Is that okay?
Or should it be kept as a local helper function in this PR?

Thanks for reading!

taiyakihitotsu:
~~Note:~~
~~Open to draft because of https://github.com/sindresorhus/type-fest/pull/1345~~

Reopen.

source/paths.d.ts
```diff
@@ -77,10 +77,10 @@ export type PathsOptions = {
 	};
 
 	type AllPaths = Paths<Post, {leavesOnly: false}>;
-	//=> 'id' | 'author' | 'author.id' | 'author.name' | 'author.name.first' | 'author.name.last'
+	//=> 'id' | 'author' | 'author.name' | 'author.name.first' | 'author.name.last' | 'author.id'
 
 	type LeafPaths = Paths<Post, {leavesOnly: true}>;
-	//=> 'id' | 'author.id' | 'author.name.first' | 'author.name.last'
+	//=> 'id' | 'author.name.first' | 'author.name.last' | 'author.id'
```
som-sm: @taiyakihitotsu In this case, I didn't mean we should fix the JSDoc. Why is the order in `Paths` changing, how are the changes in this PR affecting the order of `Paths`? We need to understand the root cause first, ideally the order here shouldn't be changing.

Also, if you pull in the changes from #1346, the order might get fixed. But I'd suggest not taking a pull until it's clear why this is happening.

Hope that makes sense!

taiyakihitotsu: @som-sm 
https://github.com/microsoft/TypeScript/issues/13298#issuecomment-468375328
I think we essentially cannot "fix" this...

(This "union order changed" issue persists even after reverting all changes of `IsEqual`.)

I apologize if I have misunderstood your point.

som-sm: > (This "union order changed" issue persists even after reverting all changes of `IsEqual`.)

Umm...but there's no such error in `main`.

---

With this definition of `IsEqual` there's no error, right?
```ts
export type IsEqual<A, B> =
	[A] extends [B]
		? [B] extends [A]
			? _IsEqual<A, B>
			: false
		: false;

// This version fails the `equalWrappedTupleIntersectionToBeNeverAndNeverExpanded` test in `test-d/is-equal.ts`.
type _IsEqual<A, B> =
		? true
		: false;
```

taiyakihitotsu: No, in my environment...

What I've done is:

 - ~`git reset --soft HEAD~1` for https://github.com/sindresorhus/type-fest/pull/1338/changes/2f1d991cc40a6525755c4644bd4328397f500f8e~ `git reset --soft HEAD~1` -> `git stash` for https://github.com/sindresorhus/type-fest/pull/1338/changes/2f1d991cc40a6525755c4644bd4328397f500f8e
 - Revert `IsEqual` (the same as you suggested).

And this error:

```terminal
  source/paths.d.ts:80:2
  ✖   80:2   Expected twoslash comment to be: //=> id | author | author.name | author.name.first | author.name.last | author.id, but found: //=> id | author | author.id | author.name | author.name.first | author.name.last  type-fest/validate-jsdoc-codeblocks
  ✖   83:2   Expected twoslash comment to be: //=> id | author.name.first | author.name.last | author.id, but found: //=> id | author.id | author.name.first | author.name.last                                                type-fest/validate-jsdoc-codeblocks
  ✖  130:2   Expected twoslash comment to be: //=> author.name | author.id, but found: //=> author.id | author.name                                                                                                            type-fest/validate-jsdoc-codeblocks

  15 warnings
  3 errors
```

And after `git checkout main` then `npm test`, the same error happens "now".
(My current HEAD of main is 051325acc22f044863e52d872eef23a79e170bcb)

taiyakihitotsu: Sorry, forgot to write.
`git reset --soft HEAD~1` -> `git stash`

taiyakihitotsu:
```terminal
git diff HEAD~1
```

I've tested in this.

```terminal
diff --git a/source/is-equal.d.ts b/source/is-equal.d.ts
index 2a562c0..2e8f76f 100644
--- a/source/is-equal.d.ts
+++ b/source/is-equal.d.ts
@@ -28,10 +28,10 @@ type Includes<Value extends readonly any[], Item> =
 @category Utilities
 */
 export type IsEqual<A, B> =
-       [A, B] extends [B, A]
-               ? [A, B] extends [object, object]
-                       ? _IsEqual<SimplifyDeep<UniqueUnionDeep<A>>, SimplifyDeep<UniqueUnionDeep<B
>>>
-                       : _IsEqual<A, B>
+       [A] extends [B]
+               ? [B] extends [A]
+                       ? _IsEqual<A, B>
+                       : false
                : false;
```

som-sm: @taiyakihitotsu That's weird! 

I reverted the changes in [d02b353](https://github.com/sindresorhus/type-fest/pull/1338/commits/d02b3532d0079524975fa44d5a3ba52e9a48dab0) and all the tests pass.

taiyakihitotsu: @som-sm
Even after a `git pull`, `npm test` is still failing...
(Of course HEAD is d02b3532d0079524975fa44d5a3ba52e9a48dab0)

I think we shouldn't fight with it, because ts essentially doesn't guarantee union order.
We can write arrow test cases via `IsEqual` when expecting it to return union type.

Does this approach seem okay to you?

taiyakihitotsu: Ahh...

I tried using `IsEqual` and then `git stash`.
In this case, `npm test` passed (even though nothing changed).

However, I still think it would be worthwhile to consider using `IsEqual` in arrow docs, because the test passing in this case was purely coincidental.

som-sm: > I tried using `IsEqual` and then `git stash`.
> In this case, `npm test` passed (even though nothing changed).

Umm...WDYM?

Please ensure there's nothing wrong with your local setup, this shouldn't be happening.

Ensure you're on the correct branch, have the latest changes pulled in, don't have any local changes. It should be working then!

taiyakihitotsu: @som-sm

I did `rm -rf node_modules` -> `npm i` -> `npm test`, all tests passed.
Then, I deleted d02b3532d0079524975fa44d5a3ba52e9a48dab0 (by `git reset --soft HEAD~1` -> `git stash`), `npm test` was also fine.

After I finished up to this point, lastly I realized that my `typescript` is up to date but `npm` wasn't the latest.
Sorry about my oversight.

---

I've updated `npm` then added a re-reverted d02b3532d0079524975fa44d5a3ba52e9a48dab0 for `source/is-equal.d.ts` and `test-d/is-equal.ts`.
Then `npm test` (after `rm -rf node_modules` -> `npm i`) fails still...

---

> >    I tried using IsEqual and then git stash.
> >    In this case, npm test passed (even though nothing changed).
> 
> Umm...WDYM?

This means I "saved" the file (`source/paths.d.ts`) but nothing changed.

---

### Current env

```terminal
$ npx tsc -v
Version 5.9.3

$ node -v
v24.11.1

$ npm -v
11.8.0
```

### Re-revert
```diff
diff --git a/source/is-equal.d.ts b/source/is-equal.d.ts
index 306b468..2a562c0 100644
--- a/source/is-equal.d.ts
+++ b/source/is-equal.d.ts
@@ -1,3 +1,6 @@
+import type {SimplifyDeep} from './simplify-deep.d.ts';
+import type {UniqueUnionDeep} from './internal/type.d.ts';
+
 /**
 Returns a boolean for whether the two given types are equal.

@@ -25,13 +28,24 @@ type Includes<Value extends readonly any[], Item> =
 @category Utilities
 */
 export type IsEqual<A, B> =
-       [A] extends [B]
-               ? [B] extends [A]
-                       ? _IsEqual<A, B>
-                       : false
+       [A, B] extends [B, A]
+               ? [A, B] extends [object, object]
+                       ? _IsEqual<SimplifyDeep<UniqueUnionDeep<A>>, SimplifyDeep<UniqueUnionDeep<B
>>>
+                       : _IsEqual<A, B>
                : false;

-// This version fails the `equalWrappedTupleIntersectionToBeNeverAndNeverExpanded` test in `test-d
/is-equal.ts`.
+/**
+Note that this doesn't regard the identical union/intersection type `T | T` and/or `T & T` as `T`
recursively.
+e.g., `{a: 0} | {a: 0}` and/or `{a: 0} & {a: 0}` as `{a: 0}`.
+
+@example
+```
+type IDUnionIsTrue = _IsEqual<{a: {b: 0}} | {a: {b: 0}}, {a: {b: 0}}>; // true
+type RecurivelyIDUnionIsFalse = _IsEqual<{a: {b: 0} | {b: 0}}, {a: {b: 0}}>; // => false
+```
+
+This version fails the `equalWrappedTupleIntersectionToBeNeverAndNeverExpanded` test in `test-d/is
-equal.ts`.
+*/
 type _IsEqual<A, B> =
        (<G>() => G extends A & G | G ? 1 : 2) extends
        (<G>() => G extends B & G | G ? 1 : 2)
diff --git a/test-d/is-equal.ts b/test-d/is-equal.ts
index c7856c5..6004284 100644
--- a/test-d/is-equal.ts
+++ b/test-d/is-equal.ts
@@ -89,3 +89,21 @@ expectType<true>(equalTupleIntersectionAndTuple);
 // Test for Issue https://github.com/sindresorhus/type-fest/issues/1305
 type Assignability<T, U, _V extends IsEqual<T, U>> = any;
 type TestAssignability<T> = Assignability<T, T, true>;
+
+// Ensure `{a: t; b: s}` === `{a: t} & {b: s}`, not equal to `{a: u} & {b: v}` if `u` !== `t` or `
v` !== `s`.
+expectType<true>({} as IsEqual<{a: 0} & {b: 0}, {a: 0; b: 0}>);
+expectType<true>({} as IsEqual<{aa: {a: {x: 0} & {y: 0}} & {b: 0}}, {aa: {a: {x: 0; y: 0}; b: 0}}>
);
+expectType<false>({} as IsEqual<{a: 1} & {b: 0}, {a: 0; b: 0}>);
+expectType<false>({} as IsEqual<{aa: {a: {x: 1} & {y: 0}} & {b: 0}}, {aa: {a: {x: 0; y: 0}; b: 0}}
>);
+
+// Ensure `{a: t} | {a: t}` === `{a: t}`
+expectType<true>({} as IsEqual<{a: 0} & ({b: 0} | {b: 0}), {a: 0; b: 0}>); // eslint-disable-line
@typescript-eslint/no-duplicate-type-constituents
+expectType<true>({} as IsEqual<{aa: {a: {x: 0} & ({y: 0} | {y: 0})} & {b: 0}}, {aa: {a: {x: 0; y:
0}; b: 0}}>); // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
+expectType<false>({} as IsEqual<{readonly a: 0} & ({b: 0} | {b: 0}), {a: 0; b: 0}>); // eslint-dis
able-line @typescript-eslint/no-duplicate-type-constituents
+expectType<false>({} as IsEqual<{readonly aa: {a: 0} & ({b: 0} | {b: 0})}, {aa: {a: 0; b: 0}}>); /
/ eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
+
+// `readonly key` should not be equal to `key` whether recursively or not.
+expectType<false>({} as IsEqual<{readonly a: 0} & {b: 0}, {a: 0; b: 0}>);
+expectType<false>({} as IsEqual<{readonly aa: {a: 0} & {b: 0}}, {aa: {a: 0; b: 0}}>);
+expectType<false>({} as IsEqual<{readonly aa: {a: 0} & {b: 0} | {a: 0} & {b: 0}}, {aa: {a: 0; b: 0
}}>); // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
+expectType<false>({} as IsEqual<{aa: {a: 0} & {b: 0} | {a: 0} & {b: 0}}, {aa: {readonly a: 0; b: 0
}}>); // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
```

### Log

```terminal
$ npm test

> type-fest@5.4.3 test
> run-p test:*


> type-fest@5.4.3 test:tsc
> node --max-old-space-size=6144 ./node_modules/.bin/tsc


> type-fest@5.4.3 test:tsd
> node --max-old-space-size=6144 ./node_modules/.bin/tsd


> type-fest@5.4.3 test:xo
> node --max-old-space-size=6144 ./node_modules/.bin/xo --ignores=lint-processors/fixtures/**/*.d.ts


> type-fest@5.4.3 test:linter
> node --test

▶ jsdoc-codeblocks processor
  ✔ valid - No JSDoc (663.411893ms)
  ✔ valid - JSDoc without code block (663.905976ms)
  ✔ valid - Valid code block (665.525556ms)
  ✔ valid - With text before and after (665.733528ms)
  ✔ valid - With line breaks before and after (666.06781ms)
  ✔ valid - With @example tag (666.241051ms)
  ✔ valid - With ts language specifier (666.247042ms)
  ✔ valid - With typescript language specifier (666.626564ms)
  ✔ valid - Multiple code blocks (666.863455ms)
  ✔ valid - Multiple exports and multiple properties (666.908235ms)
  ✔ valid - Indented code blocks (667.145296ms)
  ✔ valid - Ignore codeblocks with inconsistent indentation (667.300927ms)
  ✔ invalid - With text before and after (873.546799ms)
  ✔ invalid - With line breaks before and after (873.752141ms)
  ✔ invalid - With @example tag (873.831382ms)
  ✔ invalid - With language specifiers (873.930272ms)
  ✔ invalid - With typescript language specifiers (874.037693ms)
  ✔ invalid - Multiple code blocks (874.262775ms)
  ✔ invalid - Multiple exports and multiple properties (874.005213ms)
  ✔ invalid - Indented code blocks (874.313675ms)
  ✔ invalid - Error and fix starting at the first character of the codeblock (874.511807ms)
  ✔ invalid - Error and fix in the middle of the codeblock (874.658677ms)
  ✔ invalid - Error and fix ending at the last character of the codeblock (874.887059ms)
  ✔ invalid - Error spanning multiple lines (875.022039ms)
  ✔ invalid - Multiline fix (875.16459ms)
  ✔ invalid - Multiple errors (875.293091ms)
  ✔ invalid - Overlapping errors (875.454692ms)
  ✔ invalid - Error reporting location different from fix location (875.600682ms)
  ✔ invalid - Non fixable error (875.775844ms)
  ✔ invalid - Error outside JSDoc (875.956784ms)
  ✔ invalid - Error ending at first column (876.100646ms)
✔ jsdoc-codeblocks processor (881.621342ms)
✔ lint-rules/import-path.test.js (747.580852ms)
✔ lint-rules/require-export.test.js (763.810857ms)
✔ lint-rules/require-exported-types.test.js (1340.041811ms)
✔ lint-rules/source-files-extension.test.js (702.628004ms)
✔ lint-rules/test-utils.js (644.15943ms)
✔ lint-rules/validate-jsdoc-codeblocks.test.js (9019.142587ms)
ℹ tests 37
ℹ suites 1
ℹ pass 37
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 9075.169707

  test-d/jsonify.ts:2:1
  ⚠    2:1   Unexpected todo comment: TODO: Convert the `interface`'s to....                                                                                                                                                   no-warning-comments
  ⚠   70:1   Unexpected todo comment: TODO: Convert this to a `type`..                                                                                                                                                         no-warning-comments

  test-d/conditional-simplify-deep.ts:27:1
  ⚠   27:1   Unexpected todo comment: TODO: Convert this to a `type`..                                                                                                                                                         no-warning-comments

  test-d/set-parameter-type.ts:43:1
  ⚠   43:1   Unexpected todo comment: TODO: Fix. The `this` is reported as....                                                                                                                                                 no-warning-comments

  test-d/undefined-on-partial-deep.ts:1:1
  ⚠    1:1   Unexpected todo comment: TODO: Test equality.                                                                                                                                                                     no-warning-comments

  source/arrayable.d.ts:28:1
  ⚠   28:1   Unexpected todo comment: TODO: Use `readonly T[]` when this issue....                                                                                                                                             no-warning-comments

  source/merge-deep.d.ts:17:94
  ⚠   17:94  Unexpected todo comment: TODO: Remove this.                                                                                                                                                                       no-warning-comments

  source/subtract.d.ts:37:1
  ⚠   37:1   Unexpected todo comment: TODO: Support big integer..                                                                                                                                                              no-warning-comments

  source/sum.d.ts:34:1
  ⚠   34:1   Unexpected todo comment: TODO: Support big integer..                                                                                                                                                              no-warning-comments

  test-d/internal/has-multiple-call-signatures.ts:11:2
  ⚠   11:2   Unused eslint-disable directive (no problems were reported from @typescript-eslint/unified-signatures).                                                                                  

  test-d/internal/union-max.ts:10:1
  ⚠   10:1   Unexpected todo comment: TODO: push `negative-union-max-min`....                                                                                                                                                  no-warning-comments

  test-d/internal/union-min.ts:10:1
  ⚠   10:1   Unexpected todo comment: TODO: push `negative-union-max-min`....                                                                                                                                                  no-warning-comments

  test-d/internal/unique-union-deep.ts:21:1
  ⚠   21:1   This line has a length of 260. Maximum allowed is 200.                                                                                                                                                            @stylistic/max-len

  test-d/internal/unique-union.ts:18:1
  ⚠   18:1   This line has a length of 256. Maximum allowed is 200.                                                                                                                                                            @stylistic/max-len

  source/internal/array.d.ts:18:1
  ⚠   18:1   Unexpected todo comment: TODO: should unknown-array be updated?.                                                                                                                                                  no-warning-comments

  source/paths.d.ts:80:2
  ✖   80:2   Expected twoslash comment to be: //=> id | author | author.name | author.name.first | author.name.last | author.id, but found: //=> id | author | author.id | author.name | author.name.first | author.name.last  type-fest/validate-jsdoc-codeblocks
  ✖   83:2   Expected twoslash comment to be: //=> id | author.name.first | author.name.last | author.id, but found: //=> id | author.id | author.name.first | author.name.last                                                type-fest/validate-jsdoc-codeblocks
  ✖  130:2   Expected twoslash comment to be: //=> author.name | author.id, but found: //=> author.id | author.name                                                                                                            type-fest/validate-jsdoc-codeblocks

  15 warnings
  3 errors

ERROR: "test:xo" exited with 1.
```

taiyakihitotsu: Even if I fix this now, it might only be a temporary solution.
Any subsequent changes might still break the consistent order of Unions in the arrow docs, due to TypeScript's internal behavior.

I want to implement a safeguard to prevent these minor but problematic diffs.
(Using `IsEqual` could be an option for this.)

som-sm: @taiyakihitotsu I get it now! I have opened [a PR](https://github.com/sindresorhus/type-fest/pull/1347) for the same. Thanks for bringing this up!

This example made things _really_ clear:
```ts
type ArrayElement<T extends readonly unknown[]> = T[number];

declare const foo: 'b'; // Presence of this statement affects the union order in `T1` below

type T1 = ArrayElement<['a', 'b']>;
//   ^? type T1 = "b" | "a"
```
Playground: https://tsplay.dev/weGQ1m

```ts
type ArrayElement<T extends readonly unknown[]> = T[number];

// declare const foo: 'b';

type T1 = ArrayElement<['a', 'b']>;
//   ^? type T1 = "a" | "b" (Notice the order changed)
```

taiyakihitotsu: @som-sm

I see the PR, thank you so much!

So, this `IsEqual` PR will be ready to review after your PR is merged.
(But I might create separate PRs for `ExcludeExactly` and `LastOfUnion`.)

som-sm: > So, this `IsEqual` PR will be ready to review after your PR is merged.

Yeah mostly, but trying rebasing once with `main`, that had solved the issue for me locally.

> (But I might create separate PRs for `ExcludeExactly` and `LastOfUnion`.)

I haven't been able to look at that yet, I'll give it read and let you know.

source/is-equal.d.ts
```diff
@@ -26,13 +28,24 @@ type Includes<Value extends readonly any[], Item> =
 @category Utilities
 */
 export type IsEqual<A, B> =
-	[A] extends [B]
-		? [B] extends [A]
-			? _IsEqual<A, B>
-			: false
+	[A, B] extends [B, A]
+		? [A, B] extends [object, object]
+			? _IsEqual<SimplifyDeep<UniqueUnionDeep<A>>, SimplifyDeep<UniqueUnionDeep<B>>>
```
som-sm: @taiyakihitotsu Not sure if want to make `IsEqual` so complicated just to make `{a: t, b: s}` and `{a: t} & {b: s}` equal. Isn't there a simpler way of achieving this? And maybe in the future, this might just work automatically, refer https://github.com/microsoft/TypeScript/pull/60726.

taiyakihitotsu: @som-sm
I've not found an elegant solution to unify a union type...

> And maybe in the future, this might just work automatically

Thanks for your information!
I hope it to be merged, but I think we should not wait for it because the last comment was posted a month ago, no reviews.
We cannot predict when and (even) whether it will be merged.

I believe we should close any potential bugs which we can.
Even if this PR is merged and results in redundant logic in `IsEqual`, I expect type safety will not be compromised.
(But, if merged, a separate PR for performance optimization would likely be required later.)

---

I've done a refactor for `UniqueUnion` with `UnionToTuple`.

```typescript
type UniqueUnion<U> = UnionToTuple<U>[number]
```

Simple.
But in this naive definition, `npm test` spits a stack-overflow.

```terminal
✔ lint-rules/validate-jsdoc-codeblocks.test.js (11785.496384ms)
ℹ tests 37
ℹ suites 1
ℹ pass 37
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 11849.291703

<--- Last few GCs --->

[25259:0x556c1abd6000]    64324 ms: Scavenge 6096.5 (6140.3) -> 6084.4 (6141.8) MB, pooled: 0 MB, 4.60 / 0.00 ms  (average mu = 0.359, current mu = 0.354) allocation failure;
[25259:0x556c1abd6000]    65764 ms: Mark-Compact (reduce) 6110.4 (6154.5) -> 6063.4 (6113.3) MB, pooled: 0 MB, 47.19 / 0.00 ms  (+ 1233.1 ms in 247 steps since start of marking, biggest step 5.2 ms, walltime since start of marking 1416 ms) (average mu = 0
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
----- Native stack trace -----

 1: 0x556bfb9bddd9 node::OOMErrorHandler(char const*, v8::OOMDetails const&) [node]
 2: 0x556bfbe20da4  [node]
 3: 0x556bfbe20e89  [node]
 4: 0x556bfc0a91fb  [node]
 5: 0x556bfc0a92f2  [node]
 6: 0x556bfc0a93c5  [node]
 7: 0x556bfc0b8cec  [node]
 8: 0x556bfc0bbcc5  [node]
 9: 0x556bfcb0de17  [node]
ERROR: "test:tsc" exited with 134.
```

I found it was caused by the definition of `UniqueUnion` and `UnionToTuple`.

 - In `UniqueUnion`, a result of `UnionToTuple` should be stored with `infer`.
 - In `UnionToTuple`, a result of `ExcludeExactly<T, L>` before expanding it.

```typescript
export type UnionToTuple<T, L = LastOfUnion<T>> =
IsNever<T> extends false
	? ExcludeExactly<T, L> extends infer E // Improve performance.
		? [...UnionToTuple<E>, L]
		: never // Unreachable.
	: [];
```

Implementing either of these is enough to fix the stack overflow though, I’ve applied both just to be safe.

taiyakihitotsu:
See this PR: https://github.com/sindresorhus/type-fest/pull/1349

---

And I found another bug in `UnionToTuple`:

```typescript
type DifferentModifiers = {a: 0} | {readonly a: 0} | {a?: 0} | {readonly a?: 0};
expectType<UnionToTuple<DifferentModifiers>[number]>({} as DifferentModifiers);
```

Expected this to pass, but `npm test` says:
```terminal
  test-d/union-to-tuple.ts:16:0
  ✖  16:0  Parameter type { readonly a?: 0; } is not identical to argument type DifferentModifiers.
```

`UnionToTuple` in this PR uses `ExcludeExactly` instead of the built-in `Exclude` or `ExcludeStrict` in `type-fest` to fix it.

---

To separate this point from this PR, `LastOfUnion` and `ExcludeExactly` should be defined independently.

And we should avoid duplicate definitions.
`LastOfUnion` was defined as a local in `IsNever`.

So I will make a separate PR for it, including:
  - `LastOfUnion`: Useful to define recursion of union types.
  - `ExcludeExactly`: Stricter version of built-in `Exclude` and type-fest's `ExcludeStrict`.
  - `UnionToTuple`: Fix the above case, improve performance.

taiyakihitotsu:
**NOTE**

https://github.com/sindresorhus/type-fest/pull/1338/commits/944a74b01e46e1f51eaaba99aa5eb569d42d3e5e

This change is a temporary fix to trigger the CI.
I’m keeping this PR as a Draft until the PR https://github.com/sindresorhus/type-fest/pull/1347 is merged, even if there are no conflicts here.

---

https://github.com/sindresorhus/type-fest/pull/1347 was merged, but this PR waits for https://github.com/sindresorhus/type-fest/pull/1349 as well.

source/is-equal.d.ts
som-sm: > I believe we should close any potential bugs which we can.

Yeah true, I just linked that PR so that in future when that gets merged we might be able to revert the changes done here.

Let me give this a thorough read sometime next week. I'll also go through all your previous attempts and the issues that existed in those.

In the meanwhile, feel free to open PRs for:
- Exposing `LastOfUnion`.
- Adding `ExcludeExactly`.
- Fixing the `UnionToTuple` bug.

These should be easier to review and get merged.

taiyakihitotsu: @som-sm

I've pushed a PR!
https://github.com/sindresorhus/type-fest/pull/1349

source/exclude-exactly.d.ts
```diff
@@ -0,0 +1,98 @@
...
+This `ExcludeExactly` keeps the union objects element if the keys of the first and the second aren't identical.
+
+@example
+```
+import type {ExcludeExactly} from 'type-fest';
+
+type ExcludeNever = ExcludeExactly<{a: 0} | {a: 0} | {readonly a: 0}, never>; // => {a: 0} | {a: 0} | {readonly a: 0}
+type ExcludeReadonlyKey = ExcludeExactly<{a: 0} | {readonly a: 0}, {readonly a: 0}>; // => {a: 0}
+type ExcludeKey = ExcludeExactly<{readonly a: 0}, {a: 0}>; // => {readonly a: 0}
+type ExcludeReadonly = ExcludeExactly<{readonly a: 0}, {readonly a: 0}>; // => {readonly a: 0}
```
sindresorhus:
```suggestion
type ExcludeReadonly = ExcludeExactly<{readonly a: 0}, {readonly a: 0}>; // => never
```

?

At least that's what the tests show.

source/exclude-exactly.d.ts
```diff
@@ -0,0 +1,98 @@
...
+type NeverReturned_1 = ExcludeStrict<{a: 0} | {readonly a: 0}, {readonly a: 0}>; // => never
+```
+
+This `ExcludeExactly` keeps the union objects element if the keys of the first and the second aren't identical.
+
+@example
+```
+import type {ExcludeExactly} from 'type-fest';
+
+type ExcludeNever = ExcludeExactly<{a: 0} | {a: 0} | {readonly a: 0}, never>; // => {a: 0} | {a: 0} | {readonly a: 0}
```
sindresorhus:
```suggestion
type ExcludeNever = ExcludeExactly<{a: 0} | {a: 0} | {readonly a: 0}, never>; 
//=> {a: 0} | {a: 0} | {readonly a: 0}
```

same with the other ones

test-d/is-equal.ts
```diff
@@ -89,3 +89,21 @@ expectType<true>(equalTupleIntersectionAndTuple);
...
+expectType<true>({} as IsEqual<{aa: {a: {x: 0} & ({y: 0} | {y: 0})} & {b: 0}}, {aa: {a: {x: 0; y: 0}; b: 0}}>); // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
+expectType<false>({} as IsEqual<{readonly a: 0} & ({b: 0} | {b: 0}), {a: 0; b: 0}>); // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
+expectType<false>({} as IsEqual<{readonly aa: {a: 0} & ({b: 0} | {b: 0})}, {aa: {a: 0; b: 0}}>); // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
+
+// `readonly key` should not be equal to `key` whether recursively or not.
+expectType<false>({} as IsEqual<{readonly a: 0} & {b: 0}, {a: 0; b: 0}>);
+expectType<false>({} as IsEqual<{readonly aa: {a: 0} & {b: 0}}, {aa: {a: 0; b: 0}}>);
+expectType<false>({} as IsEqual<{readonly aa: {a: 0} & {b: 0} | {a: 0} & {b: 0}}, {aa: {a: 0; b: 0}}>); // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
+expectType<false>({} as IsEqual<{aa: {a: 0} & {b: 0} | {a: 0} & {b: 0}}, {aa: {readonly a: 0; b: 0}}>); // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
```
sindresorhus, line 225: A couple of more tests. Maybe add around line 34:

```ts
expectType<true>({} as IsEqual<(value: {a: 1}) => {b: 1}, (value: {a: 1}) => {b: 1}>);
expectType<false>({} as IsEqual<(value: {a: 1}) => {b: 1}, (value: {a: 2}) => {b: 1}>);
```

sindresorhus (changes requested):

taiyakihitotsu:
@sindresorhus 

Thank you!

I’ve left `source/exclude-exactly.d.ts` part as is because it’s waiting for #1349 to be merged.
These will be resolved automatically.

While adding the tests you suggested, I discovered a similar issue about lambda arguments and included an update for it as well.
(I'll post about it here.)

taiyakihitotsu:
Found another edge case where lambda arguments and return types are identical unions or intersections.

This isn't a regression; the previous `UniqueUnionDeep` simply didn't support it, so I've included a fix.

---

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbwKYA8xIMYwCoE90C+cAZlBCHAOQwDOAJpQNwCwAUGzPknAHICuIAEZIoAGQCGQuuLgBeOAAoAbuIA2fJAC5E47QDsBwqAQCUcgHyJB+wyIItWndL1tjJg6QEk9METUwwwBB6cooq6lo6NkJ2cABkUXAGMcZmspYI1kmu9hxcLikSUuIAqnpBIfLKahraCLrZKUQAPonJRqYWVtEdDk7c-IXu0gBKSDB8UHrevlD+WBWh1RF1De12aRlZ68bx3Y29ec6DRkUe4mMTU2WLVeG1bTmb+zstLzkOR9wA+p40AKIARz4agAPABBAA0cAAQpZZGwAJAKUEAcXMCmeqLgqF8ejoNDg4L22Na2IA-HAAIxwbQAJjMuKQ+JoSJR6MxXWxTJZsJJcDJcEpNPpJiRiMpMCgGnF2mIan8n1YqHQWDw6FBUo0GIQRHEhN+AOBYJOIjOXh8fgCFWhprcxXMJgcAHpnekcVAyFA2CqAuqkJrpUgdXqDX8gSDVKC7ebSuVgrbXLHHS63ZYRF6fWg-VxA9qFLq4Pq4IaIyak8MLuNJtNLXNrQmCqdKym2K73RnoFnVThc1rgwXQyXw8aozHK5cazdG+OHU622mPV6gA

```typescript

type NumberLambda = (value: {a: number}) => {b: number};
type NumberLambdaIntersection = (value: {a: number} & {a: number}) => {b: number};
type NumberLambdaUnion = (value: {a: number} | {a: number}) => {b: number};
type NumberLambdaReturnIntersection = (value: {a: number}) => {b: number} & {b: number};
type NumberLambdaReturnUnion = (value: {a: number}) => {b: number} | {b: number};

type _IsEqual<A, B> =
		? true
		: false;

expectType<true>({} as _IsEqual<NumberLambdaIntersection, NumberLambda>);
//=> error
expectType<true>({} as _IsEqual<NumberLambdaUnion, NumberLambda>);
//=> error
expectType<true>({} as _IsEqual<NumberLambdaReturnIntersection, NumberLambda>);
//=> error
expectType<true>({} as _IsEqual<NumberLambdaReturnUnion, NumberLambda>);
//=> error
```

taiyakihitotsu:
I've reopen this PR.

I decided to re-introduce my definition `ExcludeExactly` to resolve this following error in `test-d/exact.ts`, despite our previous discussion: https://github.com/sindresorhus/type-fest/pull/1349#discussion_r2774682198

```terminal
  test-d/exact.ts:554:1
  ✖  554:1  Type of property a circularly references itself in mapped type SimplifyUniqueUnionDeep<A>.
```

I'm not entirely sure if this is the ideal solution.

It seems the error is triggered by the `Exact` definition using the new `IsEqual` introduced in this PR.
If `Exact` doesn't require extreme strictness, we can move `_IsEqual` from `source/is-equal.d.ts` to `source/internal/type.d.ts` then export it under a new name for use within `Exact`.

I've verified this approach also resolves the circular reference issue.

```typescript
// source/is-equal.d.ts
type _IsEqual<A, B> =
		? true
		: false;
```

@som-sm
What do you think?

---

As a side note, the current `IsEqual` in this PR now correctly handles identical union/intersection types within tuples, records, and lambda arguments/returns.

sindresorhus:
`IsEqual` now runs function types through the deep union-deduping helper, which rebuilds callables from `Parameters<T>` and `ReturnType<T>`. That drops extra callable structure instead of preserving it.

Minimal repro:

```ts
type A = ((value: number) => void) & {foo?: 1};
type B = (value: number) => void;

type C = {
	(value: 'a'): 1;
	(value: string): 1;
};
type D = (value: string) => 1;
```

On this branch, `IsEqual<A, B>` and `IsEqual<C, D>` both evaluate to `true`.
With the current `main` equality implementation, both evaluate to `false`.

So this change makes distinct function types compare equal. That also means any utility that short-circuits on `IsEqual` can now accept mismatched callables a bit too early.

taiyakihitotsu:
@sindresorhus

Thanks!

Firstly, I added these test cases in `test-d/is-equal.ts`.

I pasted them in 2 separate sections.
The tail comments were removed in the latest commits.
Note that the `(not) type error` comments represent the status as of 8a7056bfbb96f4dba5ea8e0dc4904de30c9b54e2.

```typescript
// Branded Type with Tuple
expectType<false>({} as IsEqual<[0, 1] & {foo?: 1}, [0, 1]>) // circular ref
expectType<true>({} as IsEqual<[0, 1] & {foo?: 1}, [0, 1] & {foo?: 1}>) // circular ref
```

```typescript
// Branded Type with Primitive
expectType<false>({} as IsEqual<1 & {foo?: 1}, 1>); // not type error
expectType<true>({} as IsEqual<1 & {foo?: 1}, 1 & {foo?: 1}>); // not type error
// Branded Type with Lambda
expectType<false>({} as IsEqual<{bar: 'a'} & {foo?: 1}, {bar: 'a'}>); // not type error
expectType<false>({} as IsEqual<((value: number) => void) & {foo?: 1}, (value: number) => void>); // type error

// Overload
expectType<false>({} as IsEqual<{(value: 'a'): 1; (value: string): 1;}, (value: string) => 1>); // type error
expectType<false>({} as IsEqual<{(value: 'a'): 1; (value: string): 1; key: 'value'}, (value: string) => 1>); // not type error
expectType<false>({} as IsEqual<{(value: 'a'): 1; (value: string): 1; key: 'value'}, {(value: string): 1; key: 'value'}>); // type error
```

The first two "Branded Type with Tuple" cases were caused by `SimplifyDeep` (used in `UniqueUnionDeep` in `source/internal/type.d.ts`).
I have fixed this by removing them in commit: eb8828bbebe0fca67255f5e9176dc09db0465266

The second set of cases addresses the points raised in the review (I also added more branded type cases).
I implemented logic similar to [`PartialDeep` logic](https://github.com/sindresorhus/type-fest/blob/0329b2b2dc568df61ed6bfbbfa0f513b060a02bb/source/partial-deep.d.ts#L110) to fix this by avoiding the evaluation of objects that include overloads.
The commit: 4f490d2a8d325b96460e681db1bac6af5a1c2a8d

```typescript
T extends (...arguments_: any[]) => unknown
	? IsNever<keyof T> extends true
		? T // For functions with no properties
		: HasMultipleCallSignatures<T> extends true
			? T
			: ((...arguments_: Parameters<T>) => ReturnType<T>) & PartialObjectDeep<T, Options>
```

---

My next steps will be:

- Instead of completely bypassing the evaluation of objects containing overloads, I'll simplify identical union/intersection types within their non-overload elements.
- Address `Set` and similar structures, at least those handled in `PartialDeep`.

To keep things on track, I'm posting this as an interim report in response to the review.

I'll follow up once the two points above are resolved.

taiyakihitotsu:
Sorry for leaving this PR pending for a while.

The latest commit fixes the case where an object contains overloads with identical union/intersection types.

However, I think TypeScript has a limitation when it comes to separating overloaded object types. We currently need to separate them manually, for example:

```ts
// source/internal/type.d.ts

type UniqueOverload<T extends (...args_: any[]) => unknown> =
	T extends {
		(...args_: infer Arguments0): infer Return0;
		(...args_: infer Arguments1): infer Return1;
		(...args_: infer Arguments2): infer Return2;
		(...args_: infer Arguments3): infer Return3;
	}
		? DedupeTuple<[
				(...args_: Arguments0) => Return0,
				(...args_: Arguments1) => Return1,
				(...args_: Arguments2) => Return2,
				(...args_: Arguments3) => Return3
			]> extends infer TupleOverload extends Lambda[]
			? TupleToOverload<TupleOverload>
			: never
		: never; // Unreachable.
```

Because of this, if the target object type has more than five overloads, the current implementation of `IsEqual` in this PR will break.

https://github.com/sindresorhus/type-fest/pull/1399
At a glance, this approach seems like it may run into the same underlying problem. I haven't had a chance to dig into the implementation or the surrounding code in any detail yet, so I may be missing something, but I also wonder whether some of the overload-related code introduced here could eventually be superseded by the approach proposed in this PR.

That said, aside from overload cases, this PR correctly fixes `IsEqual` for record and tuple types.

For that reason, I'd like to split this PR into two parts:

* Fix `IsEqual` only for record and tuple types.
* Create a separate PR to address overload cases.

In the follow-up PR, we can discuss how overloads should be handled. Rather than giving up on supporting them, I would recommend introducing a `OverloadIsEqual` feature.

@som-sm 
What do you think?

PR #1343, opened 2026-01-29
`Paths`: Fix behavior with generic types

### LINKED ISSUE #1340: `SetRequiredDeep` does not work with generics
### Bug description

```typescript
type Foo<T> = { bar?: { baz?: T } };
type Test<T> = SetRequiredDeep<Foo<T>, "bar.baz">;
```

```
Type '"bar.baz"' does not satisfy the constraint '"bar" | ("baz" | (_Paths<T, { bracketNotation: false; maxRecursionDepth: 3; leavesOnly: false; depth: number; }> extends infer SubPath ? SubPath extends string | number ? `baz.${SubPath}` : never : never) extends infer SubPath ? SubPath extends string | number ? `bar.${SubPath}` : never : never)'.
```

I think it's because `Paths<Foo<T>>` is not including `"bar.baz"` for some reason.

### Repro

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBDAnmApnA3gZRTASigRwFdgoUATAERRTAF84AzKCEOAciVQFpGUBnGOwDcAWABQErmgBiECAB4AKgD44AXgxwARgEMoAfgBcWvQC9jcJXAZ0x46VYExlazdjyESZKjTAK5RVUAGjgAIj0oADpzMJV7IA

som-sm:
Fixes #1340 

<br>

## Reproducing the problem

This is an interesting one.

Let's consider this really simplified implementation of `Paths`:
```ts
type SimplifiedPaths<T> = T extends object
	? {
		[P in keyof T]: P | `${P & string}.${SimplifiedPaths<T[P]> & string}`;
	}[keyof T]
	: never;

type T = SimplifiedPaths<{a: {b: {c: string}; d: number}; e: boolean}>;
//=> 'a' | 'e' | 'a.b' | 'a.d' | 'a.b.c'
```

Now, let's recreate the problem in #1340 using this simplified implementation:
```ts
type Foo<T> = {bar: {baz: T}};
type Test<T> = SomeTypeWithConstraint<Foo<T>, 'bar.baz'>; // ✅ Works

type SomeTypeWithConstraint<T, _U extends SimplifiedPaths<T>> = never;
```
Everything works fine here. `SimplifiedPaths` is able to figure out that `'bar.baz'` is a valid path for `Foo<T>`.

And, if you intentionally instantiate `SomeTypeWithConstraint` with some incorrect path, then you'd see that the complete value for `SimplifiedPaths<Foo<T>>` is `` 'bar' | 'bar.baz' | `bar.baz.${SimplifiedPaths<T> & string}` ``, which is perfect!

image

simplified-paths-gif

## A small change that breaks it

Now let’s make a small change to `SimplifiedPaths`:
```ts
type SimplifiedPaths<T> = T extends object
	? {
		[P in keyof T]: P | (SimplifiedPaths<T[P]> extends infer SubPaths
			? `${P & string}.${SubPaths & string}`
			: never);
	}[keyof T]
	: never;

type T = SimplifiedPaths<{a: {b: {c: string}; d: number}; e: boolean}>;
//=> 'a' | 'e' | 'a.b' | 'a.d' | 'a.b.c'
```

The only difference is that we first capture `SimplifiedPaths<T[P]>` into `SubPaths` and then use that.

Looks harmless, but this now causes an error in our `SomeTypeWithConstraint` instantiation:
```ts
type Foo<T> = {bar: {baz: T}};
type Test<T> = SomeTypeWithConstraint<Foo<T>, 'bar.baz'>; // 💥 Errors

type SomeTypeWithConstraint<T, _U extends SimplifiedPaths<T>> = never;
```

image

If you inspect the error, `SimplifiedPaths<Foo<T>>` now looks like this:
```ts
'bar' | ('baz' | (SimplifiedPaths<T> extends infer SubPaths
    ? `baz.${SubPaths & string}`
    : never) extends infer SubPaths 
        ? `bar.${SubPaths & string}`
        : never)
```
simplified-paths-gif-2

### What changed?
Now, the entire subpath chain is wrapped in conditionals that depend on `SimplifiedPaths<T>`. That makes even `'bar.baz'` indirectly dependent on `T`.

So, `'bar.baz'` is no longer a clearly known union member. It is buried inside generic-dependent conditionals, which is why the constraint fails.

## Why did `Paths` extract the recursive result in the first place?

This was done to handle `bracketNotation` correctly.

If a subpath starts with bracket notation, we must not insert a dot before it. So the caller needs to know what the recursive result looks like before deciding whether to join with `.` or not. And that is what introduced the generic dependency problem.

Refer to the following lines in the existing implementation:
https://github.com/sindresorhus/type-fest/blob/77672ac7ef852022628817fffaf6ade3b8b9f2dc/source/paths.d.ts#L256-L268

## The fix

This PR moves dot handling into the recursive step itself.

Instead of the caller deciding whether to add a dot, each recursive subpath now returns paths that are already correctly prefixed.

Because of that:

- The caller no longer needs to extract the recursive result into an intermediate type.
- The recursive call can be used directly.
- The full path chain is no longer wrapped in generic-dependent conditionals.

As a result, TS can simplify known prefixes like `'bar.baz'`, and the constraint works as expected.

<br>

## Additional improvement

Additionally, this also uncollapses some paths that were previously getting collapsed.

For example, earlier:

```ts
Paths<{a: {[x: string]: {b: string; c: number}}}>
```

returned:

```ts
'a' | `a.${string}`
```

This effectively collapsed `` `a.${string}.b` `` and `` `a.${string}.c` `` into just `` `a.${string}` ``. But those paths do not actually collapse into `` `a.${string}` ``. This was only happening because of a limitation in the previous implementation.

```ts
type T = `a.${string}` | `a.${string}.c` | `a.${string}.b` // TS doesn't collapse this
//=> `a.${string}` | `a.${string}.c` | `a.${string}.b`
```

The issue was that the subpath being returned for `'a'` was `` string | `${string}.b` | `${string}.c` `` which collapses to just `string`. When that gets joined with `'a'`, the final result becomes only `` 'a' | `a.${string}` ``.

So even though the entire result didn't require any collapsing, it happened because the subpath had already collapsed.

With the updated logic, subpaths are prefixed with a dot. In this case, the subpath now becomes:

```ts
`.${string}` | `.${string}.b` | `.${string}.c`
```

This union does not collapse. So when joined with `'a'`, we correctly get:

```ts
'a' | `a.${string}` | `a.${string}.b` | `a.${string}.c`
```

So this PR also removes any unnecessary collapsing.

> [!NOTE]
> This does not eliminate collapsing entirely. Something like:
>
> ```ts
> Paths<{[x: string]: string; a: string}>
> ```
>
> still resolves to just `string`, not `(string & {}) | 'a'`. So #1190 still remains valid.

### DIFF
--- a/test-d/paths.ts
+++ b/test-d/paths.ts
@@ -137,6 +137,20 @@ type Object3 = {
 };
 expectType<Paths<Object3, {bracketNotation: true}>>({} as '[1]' | '[2]');
 
+type Object4 = {
+	1: {
+		a: string;
+	};
+};
+expectType<Paths<Object4, {bracketNotation: true}>>({} as '[1]' | '[1].a');
+
+type Object5 = {
+	1: {
+		2: string;
+	};
+};
+expectType<Paths<Object5, {bracketNotation: true}>>({} as '[1]' | '[1][2]');
+
 type deepArray = {
 	arr: Array<Array<Array<{a: string}>>>;
 };


source/internal/numeric.d.ts
```diff
@@ -44,10 +44,11 @@ type E = IsNumberLike<'a'>;
...
 */
 export type IsNumberLike<N> =
-	IsAnyOrNever<N> extends true ? N
-		: N extends number | `${number}`
+	IfNotAnyOrNever<N,
+		N extends number | `${number}`
 			? true
-			: false;
+			: false,
+		boolean, false>;
```
som-sm, line 51: `IsNumberLike` should always return a boolean, but currently it returns `any` for `any`, which makes it harder to instantiate at places that have a `boolean` constraint, like:
```ts
And<Options['bracketNotation'], IsNumberLike<Key>>
```

sindresorhus:
I added a couple of more tests just to make it comprehensive (not directly related to this change): https://github.com/sindresorhus/type-fest/pull/1343/commits/2f93f4b2f3c2d424acbd0efe3e01c451c0d47240

som-sm:
> I added a couple of more tests just to make it comprehensive (not directly related to this change): [2f93f4b](https://github.com/sindresorhus/type-fest/commit/2f93f4b2f3c2d424acbd0efe3e01c451c0d47240)

Cool, looks good!

Issue #1341, opened 2026-01-28
`ArrayLength` internal type is unused and not exported


jericirenej:
## Issue
The `source/internal/array.d.ts` contains the following type:

```ts
/**
Infer the length of the given array `<T>`.

@link https://itnext.io/implementing-arithmetic-within-typescripts-type-system-a1ef140a6f6f
*/
type ArrayLength<T extends readonly unknown[]> = T extends {readonly length: infer L} ? L : never;
```

Currently, this type is not performing any function within the library, because:
- It is not not used within this file.  
- It is inaccessible to any other helpers or exposed types, since it is not exported. 

## Proposal
The type should either be exported or deleted. Exporting it should be considered in the case that it can be useful to other helper types. 

Initial observation **suggests limited utility**. If we want to check the length of an array we have probably already inferred it as an array at that point via something like `extends UnknownArray` and then the `length` is available automatically. 

**That is why deletion is proposed.**

Alternatively, we can also export it - any of the two options is probably preferable leaving it in its unused state. If we were to export it, then some trivial tests could also be made to safe-guard against possible future regressions.

sindresorhus:
We are happy to expose more internal types.

jericirenej:
OK, I'll se if I can create some default tests for it and then export it 👍

PR #1344, opened 2026-01-30
Expose `ArrayLength` type

jericirenej:
Issue #1341 proposed either deleting or exporting the unused `ArrayLength` internal type. 

The issue was marked as a duplicate of #676 with the comment that [exposing internal types is encouraged](https://github.com/sindresorhus/type-fest/issues/1341#issuecomment-3815336611).

In this PR:
- `ArrayLength` is moved from `ìnternal/array.d.ts` to a separate file in `source`.
- General tests added.
- Type documentation has been expanded.

### DIFF
--- a/test-d/array-length.ts
+++ b/test-d/array-length.ts
@@ -14,7 +14,6 @@ expectType<number>({} as ArrayLength<[1, 2, ...unknown[]]>);
 expectType<number>({} as ArrayLength<[1, 2?, ...unknown[]]>);
 expectType<number>({} as ArrayLength<[...unknown[], 1, 2]>);
 expectType<number>({} as ArrayLength<[0, ...unknown[], 1, 2]>);
-expectType<0 | 2>({} as ArrayLength<[] | [1, 2]>);
 
 // Read-only arrays
 expectType<number>({} as ArrayLength<readonly unknown[]>);
@@ -28,7 +27,13 @@ expectType<number>({} as ArrayLength<readonly [1, 2, ...unknown[]]>);
 expectType<number>({} as ArrayLength<readonly [1, 2?, ...unknown[]]>);
 expectType<number>({} as ArrayLength<readonly [...unknown[], 1, 2]>);
 expectType<number>({} as ArrayLength<readonly [0, ...unknown[], 1, 2]>);
+
+// Unions
+expectType<0 | 2>({} as ArrayLength<[] | [1, 2]>);
 expectType<0 | 2>({} as ArrayLength<readonly [] | readonly [1, 2]>);
+expectType<0 | 2>({} as ArrayLength<[] | readonly [1, 2]>);
+expectType<1 | 2 | 3 | 4>({} as ArrayLength<[1, 2?, 3?] | ['one', 'two', 'three', 'four']>);
+expectType<number>({} as ArrayLength<readonly [1] | [1, ...number[]]>);
 
 // Edge cases and disallowed types
 expectType<never>({} as ArrayLength<never>);


source/array-length.d.ts
```diff
@@ -0,0 +1,38 @@
+/**
+Infers the length of an array. Tuples resolve to numeric literals, while non-tuples resolve to the `number` type.
+
+Useful for enforcing fixed-length arrays and distinguishing between tuple and non-tuple like arrays.
```
som-sm: > Useful for enforcing fixed-length arrays

What do you mean by this? Can you also add an example for this.

> distinguishing between tuple and non-tuple like arrays

I suggest removing this because we already have the `IsTuple` type, which handles this case better. For example, `IsTuple` allows `[number, ...string[]]` to be treated as both a tuple and a non-tuple, whereas `ArrayLength` can only treat it as a non-tuple.

source/array-length.d.ts
```diff
@@ -0,0 +1,38 @@
+/**
+Infers the length of an array. Tuples resolve to numeric literals, while non-tuples resolve to the `number` type.
```
som-sm:
```suggestion
Return the length of an array. Tuples resolve to numeric literals, while non-tuples resolve to the `number` type.
```

source/array-length.d.ts
```diff
@@ -0,0 +1,38 @@
...
+type TupleLength = ArrayLength<[1, 2, 3]>;
+//=> 3
+
+type TupleWithOptionalMembersLength = ArrayLength<[1, 2, number?]>;
+//=> 2 | 3
+
+type NonTupleArray = ArrayLength<string[]>;
+//=> number
+
+type TupleWithSpreadElements = ArrayLength<[1, 2, ...string[]]>;
```
som-sm:
```suggestion
type TupleWithRestElementLength = ArrayLength<[1, 2, ...string[]]>;
```

source/array-length.d.ts
```diff
@@ -0,0 +1,38 @@
...
+```
+import type {ArrayLength} from 'type-fest';
+
+type TupleLength = ArrayLength<[1, 2, 3]>;
+//=> 3
+
+type TupleWithOptionalMembersLength = ArrayLength<[1, 2, number?]>;
+//=> 2 | 3
+
+type NonTupleArray = ArrayLength<string[]>;
```
som-sm:
```suggestion
type NonTupleArrayLength = ArrayLength<string[]>;
```

Consistently suffix all examples with `Length`.

source/array-length.d.ts
```diff
@@ -0,0 +1,38 @@
...
+//=> number
+
+// Distinguish between arrays with determinable and non-determinable lengths
+type IsDeterminableLength<T extends readonly unknown[]> = number extends ArrayLength<T> ? false : true;
+
+type A = IsDeterminableLength<number[]>;
+//=> false
+
+type B = IsDeterminableLength<[1, 2, 3]>;
+//=> true
```
som-sm:
```suggestion
type IsFixedLengthArray<T extends readonly unknown[]> = number extends ArrayLength<T> ? false : true;

type A = IsFixedLengthArray<number[]>;
//=> false

type B = IsFixedLengthArray<[1, 2, 3]>;
//=> true
```

source/array-length.d.ts
```diff
@@ -0,0 +1,38 @@
...
+```
+
+@link https://itnext.io/implementing-arithmetic-within-typescripts-type-system-a1ef140a6f6f
```
som-sm: Remove this, I don't think it's required.

test-d/array-length.ts
```diff
@@ -0,0 +1,46 @@
+import {expectType} from 'tsd';
+import type {ArrayLength} from '../source/array-length.d.ts';
+import type {Primitive} from '../source/primitive.d.ts';
+
+// Non-tuples
+expectType<number>({} as ArrayLength<unknown[]>);
+expectType<0>({} as ArrayLength<[]>);
```
som-sm: Move this one down to "Tuples" section, this is not a "Non-tuple".

test-d/array-length.ts
```diff
@@ -0,0 +1,46 @@
...
+expectType<1>({} as ArrayLength<readonly [never]>);
+expectType<3>({} as ArrayLength<readonly ['one', 2, true]>);
+expectType<2 | 3 | 4>({} as ArrayLength<readonly [number, string, boolean?, boolean?]>);
+
+expectType<number>({} as ArrayLength<readonly [1, 2, ...unknown[]]>);
+expectType<number>({} as ArrayLength<readonly [...unknown[], 1, 2]>);
+expectType<number>({} as ArrayLength<readonly [0, ...unknown[], 1, 2]>);
+
+// Edge cases and disallowed types
+expectType<never>({} as never);
```
som-sm: Add an `any` case as well.

readme.md
```diff
@@ -266,6 +266,8 @@ Click the type names for complete docs.
 - [`ExtractRestElement`](source/extract-rest-element.d.ts) - Extract the [`rest`](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types) element type from an array.
 - [`ExcludeRestElement`](source/exclude-rest-element.d.ts) - Create a tuple with the [`rest`](https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types) element removed.
 - [`ArrayReverse`](source/array-reverse.d.ts) - Reverse the order of elements in a tuple type.
+- [`ArrayLength`](source/array-length.d.ts) - Infers the length of an array. Tuples resolve to numeric literals,
```
som-sm:
```suggestion
- [`ArrayLength`](source/array-length.d.ts) - Return the length of an array. Tuples resolve to numeric literals, 
```

source/array-length.d.ts
jericirenej: Yes, I agree, I didn't include it initially, but added it later, since it was part of the original type documentation. I'll remove it.

source/array-length.d.ts
jericirenej: Yes, it's a bit vague, I agree! 
For enforcing fixed-length array this meant that it could have been used as part of a combined type that would be checking against some other numeric literal condition. For example:
```ts
type IsLength<
  Arr extends readonly unknown[],
  Num extends number,
> = IsEqual<ArrayLength<Arr>, Num>;
```

But this might all be a bit too contrived.  And I agree with your point about the tuple / non-tuple distinction - the `IsTuple` helper does a much better job. 

I'm fine with removing the "usefulness" sentence if it is not needed. I added it to replicate the documentation pattern observed in some other helper types. But the simplicity of the current one might make it hard to do so.

p.s.: I was resolving the conversations /comments as I was applying the changes - I should have left it to you to resolve it, since you made the actual change requests.  Sorry about that, I was on auto-pilot :smile:  Please feel free to unresolve any of them if I made an error somewhere.

jericirenej:
Strange, the IDE inferred `unknown` when specifying `ArrayLength<any>`, but tests failed, because there it was inferred as an `any` type.  Updated.

source/array-length.d.ts
som-sm: I think you can remove this entirely. Adding contrived use cases doesn't really help, and it's a pretty simple type, so it doesn't necessarily need a use case section.

source/array-length.d.ts
```diff
@@ -0,0 +1,36 @@
...
+type TupleWithOptionalMembersLength = ArrayLength<[1, 2, number?]>;
+//=> 2 | 3
+
+type NonTupleArrayLength = ArrayLength<string[]>;
+//=> number
+
+type TupleWithRestElementLength = ArrayLength<[1, 2, ...string[]]>;
+//=> number
+
+// Distinguish between arrays with determinable and non-determinable lengths
```
som-sm:
```suggestion
// Distinguish between arrays with fixed and non-fixed lengths
```

source/array-length.d.ts
jericirenej: Agreed. I'll remove and push

sindresorhus:
https://github.com/sindresorhus/type-fest/pull/1344/changes#diff-a6653dddc77761046fd084a2c26be9d7c4af1eaf6fc2394dc42d8d7f67281e3cR34 fails.

som-sm:
@jericirenej Updated the implementation of `ArrayLength` to the following:
```ts
export type ArrayLength<T extends readonly unknown[]> = T['length'];

type T = ArrayLength<any>
//=> any
```

The previous implementation returned `unknown` when instantiated with `any`. IMO, returning `any` is _better_.
```ts
type ArrayLength<T extends readonly unknown[]> = T extends {readonly length: infer L} ? L : never;

type T = ArrayLength<any>
//=> unknown
```

Also, the updated implementation is _much_ simpler.

jericirenej:
@som-sm Looks great, thanks! :+1: 
I was also curious initially why the original type was inferring the length instead of just accessing `length` directly. But at that point I did not venture further and just focused on providing some tests and documentation which would satisfy the requirements for the type to be exported.

This possibility of accessing the length directly was also one of the reasons that - when faced with the options to expose the type or remove it - I was leaning more into arguing for its omission in my [initial post](https://github.com/sindresorhus/type-fest/issues/1341).

som-sm:
> This possibility of accessing the length directly was also one of the reasons that - when faced with the options to expose the type or remove it - I was leaning more into arguing for its omission in my [initial post](https://github.com/sindresorhus/type-fest/issues/1341).

I agree, this type isn’t _that_ useful. It’s essentially _just_ syntactic sugar, `T["length"]` vs `ArrayLength<T>`. I’m also fine with removing it. @sindresorhus, WDYT?

sindresorhus:
I think it's useful as documentation. It's easy to say that `T["length"]` is better, but most don't even know about it. I also find `ArrayLength` slightly more readable at callsite. So our docs can make it clar that `T["length"]` is also a good alternative.

jericirenej:
Ok, I'll update the base type description and docs later today!

jericirenej:
Rebased to current `main` and updated documentation. 

Hopefully the added documentation is clear and concise enough:
```Return the length of an array. Equivalent to `T['length']` where `T` extends any array.```

jericirenej:
Is there anything else that needs to change here or is the current solution good enough for merging @sindresorhus?

sindresorhus:
Thanks for working on this :)

PR #1345, opened 2026-02-03
fix: `validate-jsdoc-codeblock` handles union symbols regardless of order

taiyakihitotsu:
In the current (before this PR) `validate-jsdoc-codeblocks.js` doesn't sort union types including symbols.
And TypeScript doesn't ensure the order or unions, whether it's placed in `actualComment` or `expectedComment`.

This PR sorts both, keeps the original comments to pass a report.

## Background
The union order error doesn't occur in the current main, https://github.com/sindresorhus/type-fest/commit/051325acc22f044863e52d872eef23a79e170bcb including the commit to fix path, https://github.com/sindresorhus/type-fest/commit/8f0419ca4cbc2be7348f2c7eb1027d04d9828628, but in my PR: https://github.com/sindresorhus/type-fest/pull/1338

## Error log

```terminal
npm test

> type-fest@5.4.3 test
> run-p test:*


> type-fest@5.4.3 test:linter
> node --test


> type-fest@5.4.3 test:tsd
> node --max-old-space-size=6144 ./node_modules/.bin/tsd


> type-fest@5.4.3 test:xo
> node --max-old-space-size=6144 ./node_modules/.bin/xo --ignores=lint-processors/fixtures/**/*.d.ts


> type-fest@5.4.3 test:tsc
> node --max-old-space-size=6144 ./node_modules/.bin/tsc

▶ jsdoc-codeblocks processor
  ✔ valid - No JSDoc (690.561768ms)
  ✔ valid - JSDoc without code block (690.678318ms)
  ✔ valid - Valid code block (690.788088ms)
  ✔ valid - With text before and after (690.823118ms)
  ✔ valid - With line breaks before and after (690.994059ms)
  ✔ valid - With @example tag (691.139441ms)
  ✔ valid - With ts language specifier (691.288462ms)
  ✔ valid - With typescript language specifier (691.587244ms)
  ✔ valid - Multiple code blocks (691.242572ms)
  ✔ valid - Multiple exports and multiple properties (691.630694ms)
  ✔ valid - Indented code blocks (691.762755ms)
  ✔ valid - Ignore codeblocks with inconsistent indentation (691.947576ms)
  ✔ invalid - With text before and after (855.647256ms)
  ✔ invalid - With line breaks before and after (855.894208ms)
  ✔ invalid - With @example tag (856.027278ms)
  ✔ invalid - With language specifiers (856.162729ms)
  ✔ invalid - With typescript language specifiers (856.278529ms)
  ✔ invalid - Multiple code blocks (856.39915ms)
  ✔ invalid - Multiple exports and multiple properties (856.496211ms)
  ✔ invalid - Indented code blocks (856.587461ms)
  ✔ invalid - Error and fix starting at the first character of the codeblock (856.706621ms)
  ✔ invalid - Error and fix in the middle of the codeblock (856.844393ms)
  ✔ invalid - Error and fix ending at the last character of the codeblock (857.064254ms)
  ✔ invalid - Error spanning multiple lines (857.208505ms)
  ✔ invalid - Multiline fix (857.367806ms)
  ✔ invalid - Multiple errors (857.533086ms)
  ✔ invalid - Overlapping errors (857.674898ms)
  ✔ invalid - Error reporting location different from fix location (857.764178ms)
  ✔ invalid - Non fixable error (857.91842ms)
  ✔ invalid - Error outside JSDoc (858.0627ms)
  ✔ invalid - Error ending at first column (858.234671ms)
✔ jsdoc-codeblocks processor (863.804412ms)
✔ lint-rules/import-path.test.js (719.636064ms)
✔ lint-rules/require-export.test.js (641.388524ms)
✔ lint-rules/require-exported-types.test.js (1234.725503ms)
✔ lint-rules/source-files-extension.test.js (643.330826ms)
✔ lint-rules/test-utils.js (557.495563ms)
[baseline-browser-mapping] The data in this module is over two months old.  To ensure accurate Baseline data, please update: `npm i baseline-browser-mapping@latest -D`

  test-d/jsonify.ts:2:1
  ⚠    2:1   Unexpected todo comment: TODO: Convert the `interface`'s to....                                                                                                                                                   no-warning-comments
  ⚠   70:1   Unexpected todo comment: TODO: Convert this to a `type`..                                                                                                                                                         no-warning-comments

  source/arrayable.d.ts:28:1
  ⚠   28:1   Unexpected todo comment: TODO: Use `readonly T[]` when this issue....                                                                                                                                             no-warning-comments

  source/merge-deep.d.ts:17:94
  ⚠   17:94  Unexpected todo comment: TODO: Remove this.                                                                                                                                                                       no-warning-comments

  source/subtract.d.ts:37:1
  ⚠   37:1   Unexpected todo comment: TODO: Support big integer..                                                                                                                                                              no-warning-comments

  source/sum.d.ts:34:1
  ⚠   34:1   Unexpected todo comment: TODO: Support big integer..                                                                                                                                                              no-warning-comments

  test-d/conditional-simplify-deep.ts:27:1
  ⚠   27:1   Unexpected todo comment: TODO: Convert this to a `type`..                                                                                                                                                         no-warning-comments

  test-d/set-parameter-type.ts:43:1
  ⚠   43:1   Unexpected todo comment: TODO: Fix. The `this` is reported as....                                                                                                                                                 no-warning-comments

  test-d/undefined-on-partial-deep.ts:1:1
  ⚠    1:1   Unexpected todo comment: TODO: Test equality.                                                                                                                                                                     no-warning-comments

  source/internal/array.d.ts:18:1
  ⚠   18:1   Unexpected todo comment: TODO: should unknown-array be updated?.                                                                                                                                                  no-warning-comments

  test-d/internal/has-multiple-call-signatures.ts:11:2
  ⚠   11:2   Unused eslint-disable directive (no problems were reported from @typescript-eslint/unified-signatures).                                                                                  

  test-d/internal/union-max.ts:10:1
  ⚠   10:1   Unexpected todo comment: TODO: push `negative-union-max-min`....                                                                                                                                                  no-warning-comments

  test-d/internal/union-min.ts:10:1
  ⚠   10:1   Unexpected todo comment: TODO: push `negative-union-max-min`....                                                                                                                                                  no-warning-comments

  test-d/internal/unique-union-deep.ts:21:1
  ⚠   21:1   This line has a length of 260. Maximum allowed is 200.                                                                                                                                                            @stylistic/max-len

  test-d/internal/unique-union.ts:18:1
  ⚠   18:1   This line has a length of 256. Maximum allowed is 200.                                                                                                                                                            @stylistic/max-len

  source/paths.d.ts:80:2
  ✖   80:2   Expected twoslash comment to be: //=> id | author | author.name | author.name.first | author.name.last | author.id, but found: //=> id | author | author.id | author.name | author.name.first | author.name.last  type-fest/validate-jsdoc-codeblocks
  ✖   83:2   Expected twoslash comment to be: //=> id | author.name.first | author.name.last | author.id, but found: //=> id | author.id | author.name.first | author.name.last                                                type-fest/validate-jsdoc-codeblocks
  ✖  130:2   Expected twoslash comment to be: //=> author.name | author.id, but found: //=> author.id | author.name                                                                                                            type-fest/validate-jsdoc-codeblocks

  15 warnings
  3 errors

✖ lint-rules/validate-jsdoc-codeblocks.test.js (6969.422889ms)
ℹ tests 37
ℹ suites 1
ℹ pass 36
ℹ fail 0
ℹ cancelled 1
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 7026.163875

✖ failing tests:

test at lint-rules/validate-jsdoc-codeblocks.test.js:1:1
✖ lint-rules/validate-jsdoc-codeblocks.test.js (6969.422889ms)
  'Promise resolution is still pending but the event loop has already resolved'
ERROR: "test:xo" exited with 1.
```

## PS
Changed this PR https://github.com/sindresorhus/type-fest/pull/1338 to draft because of this PR.

### DIFF
--- a/lint-rules/validate-jsdoc-codeblocks.test.js
+++ b/lint-rules/validate-jsdoc-codeblocks.test.js
@@ -805,6 +805,26 @@ ruleTester.run('validate-jsdoc-codeblocks', validateJSDocCodeblocksRule, {
 			//=> '🦄' | '🐶' | '🐇'
 		`))),
 
+		// Union symbols are sorted.
+		exportTypeAndOption(jsdoc(fence(dedenter`
+			namespace AAA {
+				export type aaa = {a: string};
+			}
+			namespace BBB {
+				export type aaa = {a: string};
+			}
+			namespace CCC {
+				export type aaa = {a: string};
+				export type bbb = {a: string};
+			}
+
+			type Symbols_0 = AAA.aaa | CCC.bbb | CCC.aaa | BBB.aaa;
+			//=> AAA.aaa | BBB.aaa | CCC.aaa | CCC.bbb
+
+			type Symbols_1 = AAA.aaa | CCC.bbb | CCC.aaa | BBB.aaa;
+			//=> BBB.aaa | AAA.aaa | CCC.aaa | CCC.bbb
+		`))),
+
 		// === Different types of quick info ===
 		// Function
 		exportTypeAndOption(jsdoc(fence(dedenter`
--- a/lint-rules/validate-jsdoc-codeblocks.js
+++ b/lint-rules/validate-jsdoc-codeblocks.js
@@ -239,7 +239,7 @@ function normalizeUnions(type) {
 				.map(t => [print(t), t])
 				.sort(([a], [b]) =>
 					// Numbers are sorted only wrt other numbers
-					isNumeric(a) && isNumeric(b) ? Number(a) - Number(b) : 0,
+					isNumeric(a) && isNumeric(b) ? Number(a) - Number(b) : a.localeCompare(b),
 				)
 				.map(t => t[1]);
 
@@ -274,6 +274,18 @@ function normalizeUnions(type) {
 		const remainingSpaces = indentation.length % spacesPerTab;
 		return '\t'.repeat(tabCount) + ' '.repeat(remainingSpaces);
 	});
+}
+
+function flattenObjectExpression(maybeObjectExpression) {
+	return maybeObjectExpression
+		.replaceAll(/\r?\n\s*/g, ' ') // Collapse into single line
+		.replaceAll(/{\s+/g, '{') // Remove spaces after `{`
+		.replaceAll(/\s+}/g, '}') // Remove spaces before `}`
+		.replaceAll(/;(?=})/g, ''); // Remove semicolons before `}`
+}
+
+function toArrowCommentExpression(typeExpression) {
+	return TWOSLASH_COMMENT + ' ' + typeExpression.replaceAll('\n', '\n// ');
 }
 
 function validateTwoslashTypes(context, env, code, codeStartIndex) {
@@ -311,19 +323,25 @@ function validateTwoslashTypes(context, env, code, codeStartIndex) {
 			const quickInfo = env.languageService.getQuickInfoAtPosition(FILENAME, previousLineOffset + i);
 
 			if (quickInfo?.displayParts) {
-				let expectedType = normalizeUnions(extractTypeFromQuickInfo(quickInfo));
+				const expectedType = normalizeUnions(extractTypeFromQuickInfo(quickInfo));
 
-				if (expectedType.length < 80) {
-					expectedType = expectedType
-						.replaceAll(/\r?\n\s*/g, ' ') // Collapse into single line
-						.replaceAll(/{\s+/g, '{') // Remove spaces after `{`
-						.replaceAll(/\s+}/g, '}') // Remove spaces before `}`
-						.replaceAll(/;(?=})/g, ''); // Remove semicolons before `}`
+				// Save both `expectedComment` and `actualComment` to pass them to report.
+				let saveExpectedType = expectedType;
+				if (saveExpectedType.length < 80) {
+					saveExpectedType = flattenObjectExpression(saveExpectedType);
 				}
 
-				const expectedComment = TWOSLASH_COMMENT + ' ' + expectedType.replaceAll('\n', '\n// ');
+				const saveExpectedComment = toArrowCommentExpression(saveExpectedType);
+				const saveActualComment = actualComment;
+				// To compare with `actualComment`.
+				const expectedComment = toArrowCommentExpression(flattenObjectExpression(expectedType));
 
-				if (actualComment !== expectedComment) {
+				const isPassed = (/\s*\/\/=>\s\S+/g).test(actualComment);
+				const isNotPassed = (/\/\/}/g.test(actualComment)) || (actualComment.length < 80 && (/\[{\n/g.test(actualComment)));
+
+				actualComment = flattenObjectExpression(TWOSLASH_COMMENT + ' ' + normalizeUnions(actualComment.replaceAll(/(\/\/=>)|(\/\/)/g, '')));
+
+				if (!isPassed || isNotPassed || actualComment !== expectedComment) {
 					const actualCommentIndex = line.indexOf(TWOSLASH_COMMENT);
 
 					const actualCommentStartOffset = sourceFile.getPositionOfLineAndCharacter(index, actualCommentIndex);
@@ -339,15 +357,15 @@ function validateTwoslashTypes(context, env, code, codeStartIndex) {
 						},
 						messageId: 'typeMismatch',
 						data: {
-							expectedComment,
-							actualComment,
+							expectedComment: saveExpectedComment,
+							actualComment: saveActualComment,
 						},
 						fix(fixer) {
 							const indent = line.slice(0, actualCommentIndex);
 
 							return fixer.replaceTextRange(
 								[start, end],
-								expectedComment.replaceAll('\n', `\n${indent}`),
+								saveExpectedComment.replaceAll('\n', `\n${indent}`),
 							);
 						},
 					});


taiyakihitotsu:
@som-sm
Hi!

I am not entirely confident in my approach, particularly here:

https://github.com/sindresorhus/type-fest/pull/1345/changes#diff-e8b043e91a0916ef1a0281b7d350f9915ec81cffeac533891e45fd25b03d9844R340

```typescript
const isNotPassed = (/\/\/}/g.test(actualComment)) || (actualComment.length < 80 && (/\[{\n/g.test(actualComment)));
```

My logic assumes that both `actualComment` and `expectedComment` should be sorted, so, they cannot be directly compared with each other.

To simplify the implementation, I used regular expressions to exclude unsupported patterns.

Does this align with your intent, or is there any part of the implementation that seems off?
What do you think?

Thanks for reading.

lint-rules/validate-jsdoc-codeblocks.test.js
```diff
@@ -805,6 +805,26 @@ ruleTester.run('validate-jsdoc-codeblocks', validateJSDocCodeblocksRule, {
...
+			namespace CCC {
+				export type aaa = {a: string};
+				export type bbb = {a: string};
+			}
+
+			type Symbols_0 = AAA.aaa | CCC.bbb | CCC.aaa | BBB.aaa;
+			//=> AAA.aaa | BBB.aaa | CCC.aaa | CCC.bbb
+
+			type Symbols_1 = AAA.aaa | CCC.bbb | CCC.aaa | BBB.aaa;
+			//=> BBB.aaa | AAA.aaa | CCC.aaa | CCC.bbb
```
som-sm, line 825: Umm...didn't get this. This example already works, and the result is already sorted.
```ts
namespace AAA {
	export type aaa = {a: string};
}
namespace BBB {
	export type aaa = {a: string};
}
namespace CCC {
	export type aaa = {a: string};
	export type bbb = {a: string};
}

type Symbols_0 = AAA.aaa | CCC.bbb | CCC.aaa | BBB.aaa;
//   ^? type Symbols_0 = AAA.aaa | BBB.aaa | CCC.aaa | CCC.bbb

type Symbols_1 = AAA.aaa | CCC.bbb | CCC.aaa | BBB.aaa;
//   ^? type Symbols_1 = AAA.aaa | BBB.aaa | CCC.aaa | CCC.bbb
```
Playground: https://tsplay.dev/W4oPAm

som-sm:
@taiyakihitotsu This error seems fine:

image

If you change something that alters the output of `Paths`, then the JSDoc should be updated accordingly.

The intent behind custom sorting was to _only_ sort numbers, because unordered numbers can create confusion. For everything else, we want to stick to what shows up in the Quick Info tooltip. Custom sorting everything, will only make things more confusing.

So if the Quick Info tooltip shows `id | author | author.name | author.name.first | author.name.last | author.id`, we want the twoslash comment to match that. We should not change it to something else.

Also, at times you can tweak the implementation and get the desired order, so it's worth double-checking that as well.

lint-rules/validate-jsdoc-codeblocks.test.js
taiyakihitotsu: Sorry.
The goal was to verify that both `actual` and `expected` are sorted consistently even if they are Union types, allowing them to be compared via `actualComment !== expectedComment` regardless of their order.
However, I'm concerned that my test case might not be reproducing this scenario correctly.

This is what I want to fix:
https://github.com/sindresorhus/type-fest/actions/runs/21571234433/job/62150630102?pr=1338

If you pull this PR https://github.com/sindresorhus/type-fest/pull/1338 and run `npm test`, can you reproduce the same error?

som-sm: > This is what I want to fix:
> https://github.com/sindresorhus/type-fest/actions/runs/21571234433/job/62150630102?pr=1338

Refer https://github.com/sindresorhus/type-fest/pull/1345#issuecomment-3842208650.

taiyakihitotsu:
@som-sm 

Thank for explanation!

> So if the Quick Info tooltip shows id | author | author.name | author.name.first | author.name.last | author.id, we want the twoslash comment to match that. We should not change it to something else.

But in my environment, the main branch passes `npm test` but this PR https://github.com/sindresorhus/type-fest/pull/1338 doesn't.

My concern is that if we don't sort Unions, we might end up having to update (unrelated) documentation every time there's a change, just because the output order shifted.

(Regardless of whether my approach is the right way to go or not)

What do you think?

som-sm:
> But in my environment, the main branch passes `npm test` but this PR #1338 doesn't.

Because your PR most likely introduces some change that affects the order of `Paths`. Can you figure out why the order is changing?

> My concern is that if we don't sort Unions, we might end up having to update (unrelated) documentation every time there's a change, just because the output order shifted.

But the order wouldn't change unless the underlying implementation changes. If the implementation is the same, the order should/would remain the same.

---

And, if you look at the example, the existing order is correct, because it follows the order in the input object.
image

`id` comes first in the object, then `author`, then `author.name` and so on. So, it shouldn't be changed.

taiyakihitotsu:
> > My concern is that if we don't sort Unions, we might end up having to update (unrelated) documentation every time there's a change, just because the output order shifted.
>
> But the order wouldn't change unless the underlying implementation changes. If the implementation is the same, the order should/would remain the same.

Thank you for clarifying!
I see your point now.

I withdraw this PR (and reopen #1338).

Thanks!

PR #1349, opened 2026-02-05
`UnionToTuple`: Fix behavior when a union member is a supertype of another; Add `ExcludeExactly` type

taiyakihitotsu:
The current (7c82a2191042e39841dc1b4a900328b03085b1f5) `UnionToTuple` fails in the following case:

```typescript
type DifferentModifiers = {a: 0} | {readonly a: 0} | {a?: 0} | {readonly a?: 0};
expectType<UnionToTuple<DifferentModifiers>[number]>({} as DifferentModifiers);
```

This is `npm test` error:

```terminal
  test-d/union-to-tuple.ts:16:0
  ✖  16:0  Parameter type { readonly a?: 0; } is not identical to argument type DifferentModifiers.
```

## Add `ExcludeExactly`

This is caused by built-in `Exclude`, used in `UnionToTuple`.
This doesn't distinguish between `readonly` modifiers of objects.

```typescript
type A = Exclude<{readonly a: 0} | {a: 0} | {readonly a?: 0} | {a?: 0}, {a: 0}>;
//=> {readonly a?: 0} | {a?: 0}
type B = Exclude<{readonly a: 0} | {a: 0}, {a: 0}>;
//=> never

type A = ExcludeStrict<{readonly a: 0} | {a: 0} | {readonly a?: 0} | {a?: 0}, {a: 0}>;
//=> {readonly a?: 0} | {a?: 0}
type B = ExcludeStrict<{readonly a: 0} | {a: 0}, {a: 0}>;
//=> never
```

This is a spec: https://www.typescriptlang.org/docs/handbook/utility-types.html#excludeuniontype-excludedmembers
> Constructs a type by excluding from UnionType all union members that are **assignable** to ExcludedMembers.

And I think it's not a bug of `type-fest`'s `ExcludeStrict`, because it also refers to assignability here: https://github.com/sindresorhus/type-fest/blob/main/source/extends-strict.d.ts#L5
> A stricter, non-distributive version of `extends` for checking whether one type is **assignable** to another.

Both are based on assignability, but we need stricter version of `Exclude` and this PR defines it as `ExcludeExactly`.

## Add `LastOfUnion`

And to define it, we need another local type `LastOfUnion`, using a well-known definition, defined already in `IsNever`.
This is useful to recurse union-types, so I think this should be exported as a global.
(And we can evade a duplicated definition potentially).

And the current `LastOfUnion` used in `IsNever`, this test case doesn't pass:

```typescript
expectType<never>({} as LastOfUnion<never>);
```

because `LastOfUnion<never>` returns `unknown`.
Ideally, `LastOfUnion` reduces a type, e.g., $A \cup B$ -> $B$, so $\emptyset$ -> $\emptyset$ intuitively ($\bot$ -> $\bot$).
This PR's `LastOfUnion` handles this.

## Fix `UnionToTuple`
Here is the definition:

```typescript
export type UnionToTuple<T, L = LastOfUnion<T>> =
IsNever<T> extends false
	? ExcludeExactly<T, L> extends infer E // Improve performance.
		? [...UnionToTuple<E>, L]
		: never // Unreachable.
	: [];
```

There're 2 diff.

 - `ExcludeExactly` instead of `Exclude` or `ExcludeStrict`.
 - `ExcludeExactly<T, L>` isn't expanded directly. It stores it to `infer E` to improve memory performance. (See the context: https://github.com/sindresorhus/type-fest/pull/1338#discussion_r2769839547)

---

This PR includes the above 3 changes.

Refers to: https://github.com/sindresorhus/type-fest/blob/main/.github/contributing.md
> One type addition per pull request, unless they are connected.

I think this is a "connected" case, because

  - `UnionToTuple` needs `ExcludeExactly`
  - `ExcludeExactly` needs `LastOfUnion`

(This PR can be split into three parts, though it would require careful consideration of the merge order: `LastOfUnion` -> `ExcludeExactly` -> `UnionToTuple`)

---

NOTE:
This is separate from https://github.com/sindresorhus/type-fest/pull/1338.

### DIFF
--- a/source/union-to-tuple.d.ts
+++ b/source/union-to-tuple.d.ts
@@ -1,48 +1,20 @@
-import type {IsNever} from './is-never.d.ts';
 import type {ExcludeExactly} from './exclude-exactly.d.ts';
+import type {IsNever} from './is-never.d.ts';
 import type {UnionToIntersection} from './union-to-intersection.d.ts';
 
 /**
-Return a member of a union type. Order is not guaranteed.
-Returns `never` when the input is `never`.
-
-@see https://github.com/microsoft/TypeScript/issues/13298#issuecomment-468375328
-
-Use-cases:
-- Implementing recursive type functions that accept a union type.
-- Reducing a union one member at a time, for example when building tuples.
-
-It can detect a termination case using {@link IsNever `IsNever`}.
+Returns the last element of a union type.
 
 @example
 ```
-import type {LastOfUnion, ExcludeExactly, IsNever} from 'type-fest';
-
-export type UnionToTuple<T, L = LastOfUnion<T>> =
-	IsNever<T> extends false
-		? [...UnionToTuple<ExcludeExactly<T, L>>, L]
-		: [];
-```
-
-@example
-```
-import type {LastOfUnion} from 'type-fest';
-
 type Last = LastOfUnion<1 | 2 | 3>;
 //=> 3
-
-type LastNever = LastOfUnion<never>;
-//=> never
 ```
-
-@category Type
 */
 type LastOfUnion<T> =
-	true extends IsNever<T>
-		? never
-		: UnionToIntersection<T extends any ? () => T : never> extends () => (infer R)
-			? R
-			: never;
+UnionToIntersection<T extends any ? () => T : never> extends () => (infer R)
+	? R
+	: never;
 
 /**
 Convert a union type into an unordered tuple type of its elements.
@@ -81,9 +53,7 @@ const petList = Object.keys(pets) as UnionToTuple<Pet>;
 */
 export type UnionToTuple<T, L = LastOfUnion<T>> =
 IsNever<T> extends false
-	? ExcludeExactly<T, L> extends infer E // Improve performance.
-		? [...UnionToTuple<E>, L]
-		: never // Unreachable.
+	? [...UnionToTuple<ExcludeExactly<T, L>>, L]
 	: [];
 
 export {};


readme.md
```diff
@@ -191,6 +191,7 @@ Click the type names for complete docs.
 - [`ConditionalSimplify`](source/conditional-simplify.d.ts) - Simplifies a type while including and/or excluding certain types from being simplified.
 - [`ConditionalSimplifyDeep`](source/conditional-simplify-deep.d.ts) - Recursively simplifies a type while including and/or excluding certain types from being simplified.
 - [`ExclusifyUnion`](source/exclusify-union.d.ts) - Ensure mutual exclusivity in object unions by adding other members’ keys as `?: never`.
+- [`LastOfUnion`](source/last-of-union.d.ts) - Return a type picked in a union type. Order is not guaranteed.
```
```suggestion
- [`LastOfUnion`](source/last-of-union.d.ts) - Return a member of a union type. Order is not guaranteed.
```

source/last-of-union.d.ts
```diff
@@ -0,0 +1,41 @@
+import type {IsNever} from './is-never.d.ts';
+import type {UnionToIntersection} from './union-to-intersection.d.ts';
+
+/**
+Returns the last element of a union type; otherwise returns `never` if `never` is passed.
+Note that this is non-deterministic because the order of union members is not guaranteed.
```
sindresorhus:
```suggestion
Return a member of a union type. Order is not guaranteed.
Returns `never` when the input is `never`.
```

source/last-of-union.d.ts
```diff
@@ -0,0 +1,41 @@
...
+
+@see https://github.com/microsoft/TypeScript/issues/13298#issuecomment-468375328
+
+This can be used to implement a recursive type function that accepts a union type.
```
sindresorhus:
```suggestion
Use-cases:
- Implementing recursive type functions that accept a union type.
- Reducing a union one member at a time, for example when building tuples.

```

source/last-of-union.d.ts
```diff
@@ -0,0 +1,41 @@
...
+```
+import type {LastOfUnion} from 'type-fest';
+
+type Last = LastOfUnion<1 | 2 | 3>;
+//=> 3
+
+type LastNever = LastOfUnion<never>;
+//=> never
+```
+*/
```
sindresorhus:
```suggestion

@category Type
*/
```

source/exclude-exactly.d.ts
```diff
@@ -0,0 +1,120 @@
...
+//=> number
+type ExcludeAllSet = ExcludeExactly<0 | 1 | number, number>;
+//=> never
+type ExcludeFromUnknown = ExcludeExactly<unknown, string>;
+//=> unknown
+type ExcludeFromUnknownArray = ExcludeExactly<number[] | unknown[], number[]>;
+//=> unknown[]
+```
+*/
+export type ExcludeExactly<UnionU, DeleteT> =
```
sindresorhus:
```suggestion

@category Improved Built-in
*/
export type ExcludeExactly<UnionU, DeleteT> =
```

source/exclude-exactly.d.ts
```diff
@@ -0,0 +1,120 @@
...
+	[unknown, B] extends [A, never]
+		? A
+		// This equality code base below doesn't work if `A` is `unknown` and `B` is `never` case.
+		// So this branch should be wrapped to take care of this.
+		: (<G>() => G extends A & G | G ? 1 : 2) extends (<G>() => G extends B & G | G ? 1 : 2)
+			? never
+			: A;
+
+/**
+TypeScript's built-in `Exclude` and `ExcludeStrict` in `type-fest` don't distinguishes key modifiers of objects.
```
sindresorhus:
```suggestion
A stricter version of `Exclude<T, U>` that ensures objects with different key modifiers are not considered identical.

TypeScript's built-in `Exclude` and `ExcludeStrict` in `type-fest` don't distinguish key modifiers of objects.
```

source/exclude-exactly.d.ts
```diff
@@ -0,0 +1,120 @@
...
+
+@example
+```
+type RO_0 = MatchOrNever<{readonly a: 0}, {a: 0}>;
+//=> {readonly a: 0}
+type RO_1 = MatchOrNever<{a: 0}, {readonly a: 0}>;
+//=> {a: 0}
+```
+
+`unknown` and `never` cases, which easily break equality in type level code base.
```
sindresorhus:
```suggestion
`unknown` and `never` cases, which easily break equality in type-level codebase.
```

source/exclude-exactly.d.ts
```diff
@@ -0,0 +1,120 @@
+import type {IsUnknown} from './is-unknown.d.ts';
+import type {IsNever} from './is-never.d.ts';
+import type {IsAny} from './is-any.d.ts';
+import type {LastOfUnion} from './last-of-union.d.ts';
+
+/**
+Return `never` if the 1st and the 2nd arguments are mutually identical.
+Return the 1st if not.
+(But there's a limitation about union/intersection type. See `MatchOrNever` or `_IsEqual` in `source/is-equal.d.ts` doc.)
```
sindresorhus:
```suggestion
Return `never` if the first and second arguments are identical.
Return the first argument if not.
(But there's a limitation about union/intersection type. See `IsEqual` in `source/is-equal.d.ts`.)
```

som-sm:
> This doesn't distinguish between `readonly` modifiers of objects.

BTW, this is _not_ just about property modifiers on objects. This would happen anytime `TupleToUnion` is instantiated with a union that has a member that is a supertype of another member, and `LastOfUnion` picks up the supertype member before the subtype member. Refer to these examples:
```ts
type Union1 = {a: string; b: string} | {a: string};
type Tuple1 = UnionToTuple<Union1>;
//=> [{a: string}]

type Union2 = {a: 1} | {b: 1} | {[x: string]: number};
type Tuple2 = UnionToTuple<Union2>;
//=> [{[x: string]: number}]
```
Note: In both the above examples, for the error to be noticeable, it's important that `LastOfUnion` picks up the supertype before the subtype. Otherwise, the error wouldn't surface. Refer to the following:
```ts
import type {UnionToTuple} from 'type-fest';

type Union1 = {a: string} | {a: string; b: string};
type Tuple1 = UnionToTuple<Union1>;
//=> [{a: string}, {a: string; b: string}]

type Union2 = {[x: string]: number} | {a: 1} | {b: 1};
type Tuple2 = UnionToTuple<Union2>;
//=> [{[x: string]: number}, {a: 1}, {b: 1}]
```

som-sm:
> BTW, this is _not_ just about property modifiers on objects. This would happen anytime `TupleToUnion` is instantiated with a union that has a member that is a supertype of another member, and `LastOfUnion` picks up the supertype member before the subtype member. Refer to these examples:

In order to make the test reliable, we need a union that contains two members `A` and `B` such that both `A extends B` and `B extends A` hold true. One such union is:
```ts
type Union = {a: string} | {readonly a: string}
```

In the above union, both the object types are supertypes of each other because the `readonly` modifier currently doesn't affect assignability:
```ts
type T1 = {a: string} extends {readonly a: string} ? true : false;
//=> true
type T2 = {readonly a: string} extends {a: string} ? true : false;
//=> true
```

Hence, the above union is a test case that is always guaranteed to fail:
```ts
type Union1 = {readonly a: string} | {a: string};
type Tuple1 = UnionToTuple<Union1>;
//=> [{a: string}]

type Union2 = {a: string} | {readonly a: string};
type Tuple2 = UnionToTuple<Union2>;
//=> [{readonly a: string}]
```

taiyakihitotsu:
@sindresorhus 

Thank you for fixing and refining the documentation!

---

@som-sm

> BTW, this is not just about property modifiers on objects.

Thanks for your information!
I've added test cases to pass it.

```typescript
// Super type cases.
type UnionSuperType0 = {a: string; b: string} | {a: string};
expectType<UnionSuperType0>({} as UnionToTuple<UnionSuperType0>[number]);

type UnionSuperType1 = {a: 1} | {b: 1} | {[x: string]: number};
expectType<UnionSuperType1>({} as UnionToTuple<UnionSuperType1>[number]);
```

---
> In the above union, both the object types are supertypes of each other because the readonly modifier currently doesn't affect assignability:

```typescript
type T1 = {a: string} extends {readonly a: string} ? true : false;
//=> true
type T2 = {readonly a: string} extends {a: string} ? true : false;
//=> true
```

I believe this test case aims to prove that the order of the union does not affect the result of `UnionToTuple` (or to be more precise, the result of `UnionToTuple<U>[number]`), correct?
Please let me know if I have overlooked something.
If I’ve misunderstood your point, please feel free to ignore the rest of this comment.

If my understanding is correct, we still cannot predict the order of union member picking as we might hope.
We should probably assume the order is completely arbitrary, as TypeScript does not guarantee it by any means, unfortunately.

I’ve dug into `checker.ts` in the TypeScript source (v5.9.3, https://raw.githubusercontent.com/microsoft/TypeScript/refs/tags/v5.9.3/src/compiler/checker.ts) to observe how this behaves.
It seems we cannot control the order because it relies on internal ids and binarySearch.
Since we cannot expect to control these internal factors, relying on a specific order might be risky.

Taking this into account, I believe your proposed test case can be fully covered by the following cases. 

https://github.com/sindresorhus/type-fest/pull/1349/changes#diff-6ce133a34fd79b9e6b57c337b69241c8a393399e40f7e37e64bba9e421d7c2b0R5
```typescript
// `LastOfUnion` distinguishes between different modifiers.
type UnionType = {a: 0} | {b: 0} | {a?: 0} | {readonly a?: 0} | {readonly a: 0};
expectType<true>({} as LastOfUnion<UnionType> extends UnionType ? true : false);
expectType<false>({} as UnionType extends LastOfUnion<UnionType> ? true : false);
```

What are your thoughts on this?

## `checker.ts` code

```typescript
    function createType(flags: TypeFlags): Type {
        const result = new Type(checker, flags);
        typeCount++;
        result.id = typeCount;
        tracing?.recordType(result);
        return result;
    }
```

```typescript
    function addTypeToUnion(typeSet: Type[], includes: TypeFlags, type: Type) {
        const flags = type.flags;
        // We ignore 'never' types in unions
        if (!(flags & TypeFlags.Never)) {
            includes |= flags & TypeFlags.IncludesMask;
            if (flags & TypeFlags.Instantiable) includes |= TypeFlags.IncludesInstantiable;
            if (flags & TypeFlags.Intersection && getObjectFlags(type) & ObjectFlags.IsConstrainedTypeVariable) includes |= TypeFlags.IncludesConstrainedTypeVariable;
            if (type === wildcardType) includes |= TypeFlags.IncludesWildcard;
            if (isErrorType(type)) includes |= TypeFlags.IncludesError;
            if (!strictNullChecks && flags & TypeFlags.Nullable) {
                if (!(getObjectFlags(type) & ObjectFlags.ContainsWideningType)) includes |= TypeFlags.IncludesNonWideningType;
            }
            else {
                const len = typeSet.length;
                const index = len && type.id > typeSet[len - 1].id ? ~len : binarySearch(typeSet, type, getTypeId, compareValues);
                if (index < 0) {
                    typeSet.splice(~index, 0, type);
                }
            }
        }
        return includes;
    }
```

som-sm:
@taiyakihitotsu 
> ```ts
> // Super type cases.
> type UnionSuperType0 = {a: string; b: string} | {a: string};
> expectType<UnionSuperType0>({} as UnionToTuple<UnionSuperType0>[number]);
> 
> type UnionSuperType1 = {a: 1} | {b: 1} | {[x: string]: number};
> expectType<UnionSuperType1>({} as UnionToTuple<UnionSuperType1>[number]);
> ```

These test cases are _not_ reliable because they _may_ pass even with the existing incorrect implementation. 

For example, in the first example, if the first type picked up from the union is `{a: string; b: string}`, then `Exclude<UnionSuperType0, {a: string; b: string}>` would correctly return `{a: string}`, and there would be "no" error. 

The error _only_ happens if the first type picked up is `{a: string}` because then `Exclude<UnionSuperType0, {a: string}>` would return `never` instead of `{a: string; b: string}`.

<br>


> If my understanding is correct, we still cannot predict the order of union member picking as we might hope. We should probably assume the order is completely arbitrary, as TypeScript does not guarantee it by any means, unfortunately.

You didn’t get me. The order is _not_ predictable. That’s clear, that’s the whole point.

What I meant is that we need a test case that "always" fails with the existing implementation, regardless of the order in which things happen. Therefore, I suggested a test case that's "guaranteed" to fail with the existing implementation.
```ts
type Union = {readonly a: string} | {a: string};
expectType<Union>({} as UnionToTuple<Union>[number]);
```
In this case, with the existing implementation, the output of `UnionToTuple` will _only_ ever have a single element. We "cannot" tell which element it will be, but it will always be just one. Thus, this case always fails.
 
<br>

> Taking this into account, I believe your proposed test case can be fully covered by the following cases.
> 
> https://github.com/sindresorhus/type-fest/pull/1349/changes#diff-6ce133a34fd79b9e6b57c337b69241c8a393399e40f7e37e64bba9e421d7c2b0R5
> 
> ```ts
> // `LastOfUnion` distinguishes between different modifiers.
> type UnionType = {a: 0} | {b: 0} | {a?: 0} | {readonly a?: 0} | {readonly a: 0};
> expectType<true>({} as LastOfUnion<UnionType> extends UnionType ? true : false);
> expectType<false>({} as UnionType extends LastOfUnion<UnionType> ? true : false);
> ```

The proposed test case is "already" fully covered. In the above test case the _only_ necessary members are `{a: 0}` and `{readonly a: 0}`, all the other members are unnecessary.

---

Again, this problem has nothing to do with object property modifiers. The issue can occur even without property modifiers, as illustrated in the earlier examples. But, to ensure our test case is _not_ flaky, we need a case where `A extends B` and `B extends A`, and it just so happens that `{readonly property: type}` extends `{property: type}` and vice versa. Hope it's clear now!

som-sm:
> > This doesn't distinguish between `readonly` modifiers of objects.
> 
> BTW, this is _not_ just about property modifiers on objects. This would happen anytime `TupleToUnion` is instantiated with a union that has a member that is a supertype of another member, and `LastOfUnion` picks up the supertype member before the subtype member. Refer to these examples:
> 
> ```ts
> type Union1 = {a: string; b: string} | {a: string};
> type Tuple1 = UnionToTuple<Union1>;
> //=> [{a: string}]
> 
> type Union2 = {a: 1} | {b: 1} | {[x: string]: number};
> type Tuple2 = UnionToTuple<Union2>;
> //=> [{[x: string]: number}]
> ```
> 
> Note: In both the above examples, for the error to be noticeable, it's important that `LastOfUnion` picks up the supertype before the subtype. Otherwise, the error wouldn't surface. Refer to the following:
> 
> ```ts
> import type {UnionToTuple} from 'type-fest';
> 
> type Union1 = {a: string} | {a: string; b: string};
> type Tuple1 = UnionToTuple<Union1>;
> //=> [{a: string}, {a: string; b: string}]
> 
> type Union2 = {[x: string]: number} | {a: 1} | {b: 1};
> type Tuple2 = UnionToTuple<Union2>;
> //=> [{[x: string]: number}, {a: 1}, {b: 1}]
> ```

Just to clarify, here I was _only_ trying to emphasize that this problem is _not_ limited to object property modifiers. I did "not" mean that this should be used as a test case. **It should not be, since the behavior here is not predictable.**

test-d/exclude-exactly.ts
```diff
@@ -0,0 +1,39 @@
...
+expectType<[unknown]>({} as ExcludeExactly<[unknown], [number]>);
+expectType<unknown[]>({} as ExcludeExactly<unknown[], number[]>);
+expectType<{a: unknown}>({} as ExcludeExactly<{a: unknown}, {a: number}>);
+expectType<unknown[]>({} as ExcludeExactly<number[] | unknown[], number[]>);
+
+// `unknown` and `any` exclude themselves.
+expectType<never>({} as ExcludeExactly<unknown, unknown>);
+expectType<never>({} as ExcludeExactly<unknown, any>);
+expectType<never>({} as ExcludeExactly<any, any>);
+expectType<never>({} as ExcludeExactly<any, unknown>);
```
som-sm: > // `unknown` and `any` exclude themselves.

Why? Is this a requirement _or_ a limitation of the existing implementation?

If this type does an equality check (instead of an extends check), then IMO the ideal behaviour should be:
```ts
expectType<unknown>({} as ExcludeExactly<unknown, any>);
expectType<any>({} as ExcludeExactly<any, unknown>);
```

test-d/exclude-exactly.ts
```diff
@@ -0,0 +1,39 @@
...
+
+// `unknown` and `any` exclude other types.
+expectType<never>({} as ExcludeExactly<string | number, unknown>);
+expectType<never>({} as ExcludeExactly<string | number, any>);
```
som-sm: Similar question here.

Should be this IMO:
```ts
expectType<string | number>({} as ExcludeExactly<string | number, unknown>);
expectType<string | number>({} as ExcludeExactly<string | number, any>);
```

source/exclude-exactly.d.ts
som-sm, line 1: Is this implementation simpler/better? It passes all the existing tests and also works as suggested with `any` and `unknown`.

```ts
export type ExcludeExactly<Union, Delete> =
	IfNotAnyOrNever<
		Union,
		_ExcludeExactly<Union, Delete>,
		// If `Union` is `any`, then if `Delete` is `any`, return `never`, else return `Union`.
		If<IsAny<Delete>, never, Union>,
		// If `Union` is `never`, then if `Delete` is `never`, return `never`, else return `Union`.
		If<IsNever<Delete>, never, Union>
	>;

type _ExcludeExactly<Union, Delete> =
	IfNotAnyOrNever<Delete,
		Union extends unknown // For distributing `Union`
			? [Delete extends unknown // For distributing `Delete`
				? If<SimpleIsEqual<Union, Delete>, true, never>
				: never] extends [never] ? Union : never
			: never,
		// If `Delete` is `any` or `never`, then return `Union`,
		// because `Union` cannot be `any` or `never` here.
		Union, Union
	>;

type SimpleIsEqual<A, B> =
		? true
		: false;
```

taiyakihitotsu:
@som-sm

Thanks for your clarification and review!

What you wanted to say was that we shouldn't focus on the unpredictable union order itself, but on the test cases that must fail with the old `UnionToTuple` but must pass with the new `UnionToTuple`, right?

The case below makes your point clear to me.

```typescript
type OldUnionToTuple<T, L = LastOfUnion<T>> =
IsNever<T> extends false
	? Exclude<T, L> extends infer E // Improve performance.
		? [...UnionToTuple<E>, L]
		: never // Unreachable.
	: [];

type Union1 = {a: string; b: string} | {a: string}
type Tuple1 = OldUnionToTuple<Union1>;
//=> [{a: string}]

type Union2 = {a: string} | {a: string; b: string}
type Tuple2 = OldUnionToTuple<Union2>;
//=> [{a: string}, {a: string, b: string}]
```

(But, extremely, I think, we shouldn't expect that writing `A | B` means "writing `A | B`"; it's possible for it to be interpreted as "writing `B | A`". I don't know how to handle it, and I think it cannot be done theoretically...)

---

Now I've addressed your feedback by adding reversed test cases in `test-d/union-to-tuple.ts`.

And add 2 cases more:
 - **Identical union case**: because `A | B` is not done by eager evaluation directly to `A` even if we know `A` is equal to `B`. So it leaves a possibility that we can doubt if `UnionToTuple` seems to return a 2-length tuple in this case.
 - **`length` test**: Important especially for identical union case.

Do these new test cases cover your concerns?
I hope so.

source/exclude-exactly.d.ts
taiyakihitotsu: Your definition doesn't use `LastOfUnion` and is clearer about the comparison part, since the `SimpleIsEqual` pattern is well-known, and all tests pass.
But `SimpleIsEqual` is already defined in `is-equal.d.ts`, so should we move it to internal, or even export it? Is that ok?
(I vaguely think it is acceptable because this comparison construct is useful to define some type functions.)

And in this situation, `LastOfUnion` is still worth exporting as a standalone utility.

test-d/exclude-exactly.ts
taiyakihitotsu: > > // `unknown` and `any` exclude themselves.
>
> Why? Is this a requirement or a limitation of the existing implementation?

~~`ExcludeExactly<A, B>` would be expected to return $(A \cup B) \setminus B$.
If $B$ is `unknown` or `any`,  $A \subseteq B$ must be true, so it's expected to return $\emptyset$ (`never`).~~

https://github.com/sindresorhus/type-fest/pull/1349#discussion_r2777447219

test-d/exclude-exactly.ts
taiyakihitotsu: https://github.com/sindresorhus/type-fest/pull/1349#discussion_r2775880301

test-d/exclude-exactly.ts
taiyakihitotsu:
```typescript
expectType<any>({} as ExcludeExactly<any, unknown>);
```

In this case, returning `any` might actually be fine, as you suggested.

som-sm:
> (But, extremely, I think, we shouldn't expect that writing `A | B` means "writing `A | B`"; it's possible for it to be interpreted as "writing `B | A`". I don't know how to handle it, and I think it cannot be done theoretically...)

Again, there's some confusion, I "never" meant we should treat `A | B` as `A | B`. Let me clarify again:

1. When I say "existing implementation", I mean the implementation of `UnionToTuple` that is there on `main`, i.e., this one:
   https://github.com/sindresorhus/type-fest/blob/2c34128d2cf6553976a0b08b180d6403ddba4abc/source/union-to-tuple.d.ts#L53-L56

   If that was the source of the confusion, I’d suggest revisiting the comments with this clarified.

2. We need a test case surely fails with the "existing implementation" of `UnionToTuple`. So, we can't use the following test case because it is _not_ guaranteed to fail.
    ```ts
    // Super type cases.
    type UnionSuperType0 = {a: string; b: string} | {a: string};
    expectType<UnionSuperType0>({} as UnionToTuple<UnionSuperType0>[number]);
    ```
    So, in future, someone might remove the fix applied in this PR, and the tests might still "pass", which is something we want to prevent.

test-d/exclude-exactly.ts
som-sm: > `ExcludeExactly<A, B>` would be expected to return $(A \cup B) \setminus B$.

Isn't $(A \cup B) \setminus B$ same as $A \setminus B$?

If `ExcludeExactly<A, B>` is doing _just_ $A \setminus B$, what’s the need for it? The built-in `Exclude` type already provides that behavior.

Also, if `ExcludeExactly<A, B>` is $A \setminus B$, then shouldn't `ExcludeExactly<1, number>` return `never`, instead of `1`?

<br>

Wasn't the whole point of building `ExcludeExactly` that it _shouldn't_ behave like $A \setminus B$?

<br>

Currently, this type is _not_ consistent in its behavior. Sometimes it behaves like $A \setminus B$ (like `ExcludeExactly<1, any>`), but sometimes it does not (like `ExcludeExactly<1, number>`).

taiyakihitotsu: Sorry, completely my bad...

Firstly, not $(A \cup B) \setminus B$, $A \setminus B$ is right (as you pointed it out).

Secondly, I confused some cases: if `A` is `1`, `number`, `1 | number`, and if `B` is `number` or `unknown`.
`1 | number` means `number`, so `ExcludeExactly` properly removes it with `ExcludeExactly<1 | number, number>`.
This shouldn't mean `ExcludeExactly<1, number>` returns `never`.
For the same reason, `ExcludeExactly<A, unknown>` or `, any>` shouldn't return `never`.

You're right, and I think it should use your definition of `ExcludeExactly`.
I've addressed it, moving `SimpleIsEqual` to `internal` and adding test cases for it (4baac7ea71a4d0b55bc5e3aedc30b338d8b5368f and 0699dc068fef495dab7db4e3757ce7989c478383).

taiyakihitotsu:
@som-sm

Do these tests fill the requirements?

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAbwKYA8xIMYwCoE90C+cAZlBCHAOQwDOAJpQNwBQokscM+SiAqgHbAI-bBACS-GEig1MMIfwA0cMTQBySAG7TlqgKIBHAK4BDADZFS5Kl3QBaYkhowmzZqnbxbPADInnAPLEAgoAPNgAfHAAvMwAkDBQRjyoUvx0NCrqWtLhEfFxAPxw-DlQBQBccCHCohJSMnJh2HCpSOmZJvy4cMUAFACUMVEtVaXaUFFtHXCDw7PA-I5QcABKAwVFa5tjZSzuaNBe3NWCtRDYRmBmSOHKPjFwfoHBZ-x5UbGqGhN5rShpDIkcyyeLFPQoDBmIx0W7Ye5TAHtIGLZZwPRwAD0mJU4DI2jg6CgxGgIC6GCQADoCsUANqUhk1EQXK43UJ6CL3AC6lRKZSxOIEUCQJgwAAsTAAjG7UuJVWlc-beOAAEWAxGW7RgAFkIHR1cBpJloogTFUAAxEAA+iGFJjowjMPTNcEtLA8cjw6FCTNEl2utzVGukWt1+uIhpkEVp-CMIEl0i5ET6CCI-lV6s1kjDBqNA3daE93B9bz9rMDmZD2b1uajtMoN34AHMYGLKEm+gAmfNuZWrMqyOhBrM6msRo2PBB2h38J1wF2WuA2hALggF9BYL23X0sgOhfsTQfDquj8ORmjR2PxxPJ1PzzIHo1IIeV4XVs959dF707-1sx+NC+wZvqetYXvWjYtm2HbdiwQA

```typescript
type DifferentModifiers = {a: 0} | {readonly a: 0};
expectType<UnionToTuple<DifferentModifiers>[number]>({} as DifferentModifiers);
expectType<UnionToTuple<DifferentModifiers>['length']>(2);
```

`expectType<UnionToTuple<DifferentModifiers>[number]>({} as DifferentModifiers);` ensures the identity of elements.
`expectType<UnionToTuple<DifferentModifiers>['length']>(2);` ensures the identity of the count of the union type and the length of result.

These tests were added and passed by this PR's `UnionToTuple`:

```typescript
// Different modifiers cases.
type DifferentModifiers = {a: 0} | {readonly a: 0};
expectType<UnionToTuple<DifferentModifiers>[number]>({} as DifferentModifiers);
expectType<UnionToTuple<DifferentModifiers>['length']>(2);

type ReversedDifferentModifiers = {readonly a: 0} | {a: 0};
expectType<UnionToTuple<ReversedDifferentModifiers>[number]>({} as ReversedDifferentModifiers);
expectType<UnionToTuple<ReversedDifferentModifiers>['length']>(2);
```

source/exclude-exactly.d.ts
taiyakihitotsu: https://github.com/sindresorhus/type-fest/pull/1349#discussion_r2777447219

taiyakihitotsu:
@som-sm

Regarding your point about the deficiency in the `UnionToTuple` test cases, I realized the same issue could apply to `LastOfUnion`, so I’ve added tests for it as well, 01cbb49943a4ab4eaf630ae77c60e9acd3946f4b.

The structure of these tests is identical to the length check I added for `UnionToTuple`, but the specific goal here is to verify whether `LastOfUnion` can truly handle all members of a union.
For instance,`LastOfUnion<{readonly a: 0} | {a: 0}>` has a risk that only one of them is returned.
The same concern applies to supertype cases like `LastOfUnion<{a: 0, b: 0} | {a: 0}>`.

In these tests, I needed a `UnionToTuple` definition that uses the `LastOfUnion`.
Therefore, I have redefined my first definition as a local type within `test-d/last-of-union.ts`.

If we go with my definition (of course it should be fixed about `any` and `unknown` cases), this entire local redefinition can be removed.
On the other hand, if we use your definition, it would be possible to split the changes to `last-of-union.d.ts` into a separate PR.

What are your thoughts on this?

---

After I've re-read your comment, I've realized this is what you pointed out now.

https://github.com/sindresorhus/type-fest/pull/1349#issuecomment-3860341776
> The proposed test case is "already" fully covered. In the above test case the only necessary members are {a: 0} and {readonly a: 0}, all the other members are unnecessary.

test-d/union-to-tuple.ts
```diff
@@ -11,3 +11,39 @@ expectType<Options1[number]>({} as (1 | 2 | 3));
 
 type Options2 = UnionToTuple<boolean | 1>;
 expectType<Options2[number]>({} as (1 | false | true));
```
sindresorhus, line 13:
```suggestion
expectType<Options2[number]>({} as (1 | false | true));

// Edge cases.
expectType<UnionToTuple<never>['length']>(0);
expectType<UnionToTuple<any>['length']>(1);
expectType<UnionToTuple<unknown>['length']>(1);
```

test-d/exclude-exactly.ts
```diff
@@ -0,0 +1,44 @@
...
+expectType<never>({} as ExcludeExactly<'0' | '1' | string, string>);
+
+// `{readonly a: t}` should not be equal to `{a: t}` because of assignability.
+expectType<{a: 0}>({} as ExcludeExactly<{a: 0} | {readonly a: 0}, {readonly a: 0}>);
+expectType<{readonly a: 0}>({} as ExcludeExactly<{readonly a: 0}, {a: 0}>);
+expectType<never>({} as ExcludeExactly<{readonly a: 0}, {readonly a: 0}>);
+
+// `never` does nothing.
+expectType<0 | 1 | 2>({} as ExcludeExactly<0 | 1 | 2, never>);
+expectType<never>({} as ExcludeExactly<never, never>);
```
sindresorhus, line 17:
```suggestion
expectType<never>({} as ExcludeExactly<never, never>);
expectType<any>({} as ExcludeExactly<any, never>);
expectType<unknown>({} as ExcludeExactly<unknown, never>);
```

taiyakihitotsu:
@sindresorhus 

Thanks for your review!
Addressed!

source/exclude-exactly.d.ts
```diff
@@ -0,0 +1,69 @@
...
+		If<IsAny<Delete>, never, Union>,
+		// If `Union` is `never`, then if `Delete` is `never`, return `never`, else return `Union`.
+		If<IsNever<Delete>, never, Union>
+	>;
+
+type _ExcludeExactly<Union, Delete> =
+	IfNotAnyOrNever<Delete,
+		Union extends unknown // For distributing `Union`
+			? [Delete extends unknown // For distributing `Delete`
+				? If<SimpleIsEqual<Union, Delete>, true, never>
```
som-sm: @taiyakihitotsu In this PR, we can use `IsEqual` instead of `IsSimpleEqual`, right?

Let's simplify this PR to _only_ fix `UnionToTuple`:
1. Let's use `IsEqual` in `ExcludeExactly`, so _no_ need to create `simple-is-equal.ts`.
2. No need to expose `LastOfUnion` in this PR.

So, just add `ExcludeExactly` and use it in `UnionToTuple`. This will help get this merged quickly.

test-d/union-to-tuple.ts
```diff
@@ -11,3 +11,44 @@ expectType<Options1[number]>({} as (1 | 2 | 3));
...
+expectType<ReversedUnionSuperType0>({} as UnionToTuple<ReversedUnionSuperType0>[number]);
+expectType<UnionToTuple<ReversedUnionSuperType0>['length']>(2);
+
+type UnionSuperType1 = {a: 1} | {[x: string]: number};
+expectType<UnionSuperType1>({} as UnionToTuple<UnionSuperType1>[number]);
+expectType<UnionToTuple<UnionSuperType1>['length']>(2);
+
+type ReversedUnionSuperType1 = {[x: string]: number} | {a: 1};
+expectType<ReversedUnionSuperType1>({} as UnionToTuple<ReversedUnionSuperType1>[number]);
+expectType<UnionToTuple<ReversedUnionSuperType1>['length']>(2);
```
som-sm: As mentioned previously, these cases _are_ flaky, and all of them "can" pass with the existing incorrect implementation of `UnionToTuple` on `main`. Please remove them.

test-d/union-to-tuple.ts
```diff
@@ -11,3 +11,44 @@ expectType<Options1[number]>({} as (1 | 2 | 3));
...
+expectType<UnionToTuple<unknown>['length']>(1);
+
+// Different modifiers cases.
+type DifferentModifiers = {a: 0} | {readonly a: 0};
+expectType<UnionToTuple<DifferentModifiers>[number]>({} as DifferentModifiers);
+expectType<UnionToTuple<DifferentModifiers>['length']>(2);
+
+type ReversedDifferentModifiers = {readonly a: 0} | {a: 0};
+expectType<UnionToTuple<ReversedDifferentModifiers>[number]>({} as ReversedDifferentModifiers);
+expectType<UnionToTuple<ReversedDifferentModifiers>['length']>(2);
```
som-sm: Remove this; just the above case is "sufficient“, it's enough to catch the bug correctly.

test-d/union-to-tuple.ts
```diff
@@ -11,3 +11,44 @@ expectType<Options1[number]>({} as (1 | 2 | 3));
 
 type Options2 = UnionToTuple<boolean | 1>;
 expectType<Options2[number]>({} as (1 | false | true));
+
+// Edge cases.
+expectType<UnionToTuple<never>['length']>(0);
+expectType<UnionToTuple<any>['length']>(1);
+expectType<UnionToTuple<unknown>['length']>(1);
```
som-sm:
```suggestion
expectType<[]>({} as UnionToTuple<never>);
expectType<[any]>({} as UnionToTuple<any>);
expectType<[unknown]>({} as UnionToTuple<unknown>);
```

This is _better_. Just checking `length` doesn't check the element.

test-d/union-to-tuple.ts
```diff
@@ -11,3 +11,44 @@ expectType<Options1[number]>({} as (1 | 2 | 3));
...
+
+// Identical union cases.
+type UnionIdentical = {a: string} | {a: string}; // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
+expectType<UnionToTuple<UnionIdentical>['length']>(1);
+
+type UnionIdenticalIntersection = {a: string} & {a: string} | {a: string}; // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
+expectType<UnionToTuple<UnionIdenticalIntersection>['length']>(1);
+
+type ReversedUnionIdenticalIntersection = {a: string} | {a: string} & {a: string}; // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
+expectType<UnionToTuple<ReversedUnionIdenticalIntersection>['length']>(1);
```
som-sm: Are these cases related to the changes in this PR? If not, please remove them.

source/exclude-exactly.d.ts
```diff
@@ -0,0 +1,69 @@
+import type {IsNever} from './is-never.d.ts';
+import type {IsAny} from './is-any.d.ts';
+import type {If} from './if.d.ts';
+import type {IfNotAnyOrNever, SimpleIsEqual} from './internal/type.d.ts';
+
+/**
+A stricter version of `Exclude<T, U>` that ensures objects with different key modifiers are not considered identical.
```
som-sm:
```suggestion
A stricter version of `Exclude<T, U>` that excludes types only when they are exactly identical.
```

Again this has nothing to do with key modifiers.

source/exclude-exactly.d.ts
```diff
@@ -0,0 +1,69 @@
...
+
+@example
+```
+import type {ExcludeStrict} from 'type-fest';
+
+type NeverReturned_0 = Exclude<{a: 0} | {readonly a: 0}, {readonly a: 0}>;
+//=> never
+type NeverReturned_1 = ExcludeStrict<{a: 0} | {readonly a: 0}, {readonly a: 0}>;
+//=> never
+```
```
som-sm: Remove this, it's better to do the comparison in one single codeblock. And `ExcludeStrict` is a completely different type, it's anyways not meant for this.

source/exclude-exactly.d.ts
```diff
@@ -0,0 +1,69 @@
...
+type ExcludeReadonly = ExcludeExactly<{readonly a: 0}, {readonly a: 0}>;
+//=> never
+type ExcludeSubType = ExcludeExactly<0 | 1 | number, 1>;
+//=> number
+type ExcludeAllSet = ExcludeExactly<0 | 1 | number, number>;
+//=> never
+type ExcludeFromUnknown = ExcludeExactly<unknown, string>;
+//=> unknown
+type ExcludeFromUnknownArray = ExcludeExactly<number[] | unknown[], number[]>;
+//=> unknown[]
```
som-sm: The existing examples are not quite easy to understand. I think this much should be fine:

```suggestion
import type {ExcludeExactly} from 'type-fest';

type TestExclude1 = Exclude<'a' | 'b' | 'c' | 1 | 2 | 3, string>;
//=> 1 | 2 | 3

type TestExcludeExactly1 = ExcludeExactly<'a' | 'b' | 'c' | 1 | 2 | 3, string>;
//=> 'a' | 'b' | 'c' | 1 | 2 | 3

type TestExclude2 = Exclude<'a' | 'b' | 'c' | 1 | 2 | 3, any>;
//=> never

type TestExcludeExactly2 = ExcludeExactly<'a' | 'b' | 'c' | 1 | 2 | 3, any>;
//=> 'a' | 'b' | 'c' | 1 | 2 | 3

type TestExclude3 = Exclude<{a: string} | {a: string; b: string}, {a: string}>;
//=> never

type TestExcludeExactly3 = ExcludeExactly<{a: string} | {a: string; b: string}, {a: string}>;
//=> {a: string; b: string}
```

If you think anything more needs to be added, please add.

test-d/exclude-exactly.ts
```diff
@@ -0,0 +1,49 @@
+import {expectType} from 'tsd';
+import type {ExcludeExactly} from '../index.d.ts';
+
+expectType<number>({} as ExcludeExactly<0 | 1 | number, '1'>);
```
som-sm: Remove tests like these because `0 | 1 | number` gets collapsed to just `number`. So, this test is exactly same as:
```ts
expectType<number>({} as ExcludeExactly<number, '1'>);
```

test-d/union-to-tuple.ts
som-sm: I am _not_ sure where the confusion is, I thought we were clear on this.

Can you tell me _why_ are these cases required when we already have this:
```ts
// Different modifiers cases.
type DifferentModifiers = {a: 0} | {readonly a: 0};
expectType<UnionToTuple<DifferentModifiers>[number]>({} as DifferentModifiers);
```

source/exclude-exactly.d.ts
taiyakihitotsu: @som-sm 

Thanks for your suggestion on this PR!
Addressed!

NOTE: #1338 uses `ExcludeExactly` via `UnionToTuple`, so this will cause a circular reference.
It should be replaced with `SimpleIsEqual` or so in future, but this isn't related in this PR directly.

test-d/union-to-tuple.ts
taiyakihitotsu: ~~Replaced with super type cases, in `exclude-exactly`, `union-to-tuple`, `last-of-union`.
(And add comment about the union test's limitation.)~~

test-d/union-to-tuple.ts
taiyakihitotsu: Addressed.

test-d/union-to-tuple.ts
taiyakihitotsu: Addressed.

test-d/union-to-tuple.ts
taiyakihitotsu: (I've deleted the last 2 cases because of duplication.)

It's the reason that I'm not convinced that `A | A` is treated as `A` correctly, as `IsEqual (which is used internally in the definition) does. So I want to add this case here.

But for the purpose, the test case isn't correct.
Should be written like the below:

```typescript
type UnionIdentical = {a: {b: 0} | {b: 0}} | {a: {b: 0}}; // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
expectType<UnionToTuple<UnionIdentical>['length']>(1);
```

https://github.com/sindresorhus/type-fest/pull/1349/commits/e0d7dac1cba3383d92c86d65ea7e1f9113cf1af8

And this test fails...
(I think this will be solved automatically in #1338 though.)

source/exclude-exactly.d.ts
taiyakihitotsu: Addressed.

source/exclude-exactly.d.ts
taiyakihitotsu: Addressed.

source/exclude-exactly.d.ts
taiyakihitotsu: I think it's enough for `ExcludeExactly` usage!

test-d/exclude-exactly.ts
taiyakihitotsu: Addressed.

test-d/union-to-tuple.ts
taiyakihitotsu:
```typescript
// TypeScript doesn't distinguish the order of a union type, but this is the only way we can write the tests.
type SuperTypeCase = {a: 0} | {a: 0; b: 0};
expectType<UnionToTuple<SuperTypeCase>[number]>({} as SuperTypeCase);
expectType<UnionToTuple<SuperTypeCase>['length']>(2);

type SuperTypeCaseReversed = {a: 0; b: 0} | {a: 0};
expectType<UnionToTuple<SuperTypeCaseReversed>[number]>({} as SuperTypeCaseReversed);
expectType<UnionToTuple<SuperTypeCaseReversed>['length']>(2);
```

~~If `Exclude` happens to produce the same result as `ExcludeExactly`, it's probably impossible to write a test case that only `ExcludeExactly` would pass.
Given that, I think it's better to keep this case rather than remove it.~~

~~What do you think?~~

taiyakihitotsu: > Can you tell me why are these cases required when we already have this:

Sorry, I misread your comment. Please ignore my previous replies.

test-d/union-to-tuple.ts
taiyakihitotsu: https://github.com/sindresorhus/type-fest/pull/1338#issuecomment-3881279606

Given this newly discovered bug case, it seems difficult to handle cases involving identical unions within the scope of this PR.
(This issue has already been fixed in #1338)

This PR was opened with the purpose of adding `ExcludeExactly` to properly handle records, instead of `Exclude`.

If we decide to include a fix for the union case bug as well, and address both issues at once, I think we would likely need to close this PR, carry over its changes, and continue working in #1338.

What approach do you think we should take?

taiyakihitotsu: https://github.com/sindresorhus/type-fest/pull/1349/changes/f7947ff960d0b262a4a8b4b34579e0a2b0033f9d

I've ported `UniqueUnionDeep` from #1338 and fixed these issues.

```typescript
type UnionIdentical = {a: {b: 0} | {b: 0}} | {a: {b: 0}}; // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
expectType<UnionToTuple<UnionIdentical>['length']>(1);

type NumberLambda = (value: {a: number}) => {b: number};
type NumberLambdaUnion = (value: {a: number} | {a: number}) => {b: number}; // eslint-disable-line @typescript-eslint/no-duplicate-type-constituents
type UnionLambdaIdentical = NumberLambda | NumberLambdaUnion;
expectType<UnionToTuple<UnionLambdaIdentical>['length']>(1);
```

Both tests pass now.

taiyakihitotsu: I think this change would be better than nothing, but in `type-fest`, I don't know the policy on how far we should treat `A | A` (and `A & A`) as `A` outside the context of `IsEqual`.

For example, we could consider simply exporting `UniqueUnionDeep` on its own so that end users can use it freely (just as one of the use cases of `Simplify`).

source/exclude-exactly.d.ts
som-sm: > NOTE: #1338 uses `ExcludeExactly` via `UnionToTuple`, so this will cause a circular reference.
> It should be replaced with `SimpleIsEqual` or so in future, but this isn't related in this PR directly.

That’s fine. If it isn’t a problem in this PR, we shouldn’t address it here. The more tightly scoped a PR is to a specific issue, the easier it is to understand the changes.

If someone revisits this later, they may not realize that IsSimpleEqual was used instead of IsEqual due to issues in another PR. It’s better to handle this in #1338, since that’s where the change is actually required.

test-d/union-to-tuple.ts
som-sm: > Sorry, I misread your comment. Please ignore my previous replies.

So, we're clear on this, right? That we don't need all these cases.

taiyakihitotsu: > So, we're clear on this, right? That we don't need all these cases.

Currently, I've left `DifferentModifierUnion` test and removed `UnionSuperType0` ~ `ReversedUnionSuperType1` in `test-d/union-to-tuple.ts`.

If that's what you were referring to, then yes.

test-d/union-to-tuple.ts
som-sm: I’m not a fan of the approach where we identify a failing case and patch _just_ that specific scenario. There will likely be many such cases, and we won’t be able to address them one by one.

`UniqueUnionDeep` seems to currently fails on cases like these, when the union is wrapped in a tuple:
```ts
type T = UniqueUnionDeep<{a: [{b: 0} | {b: 0}]} | {a: [{b: 0}]}>;
//=> {a: [{b: 0} | {b: 0}]} | {a: [{b: 0}]}
```

These are fairly contrived edge cases. If we can’t address them cleanly with a more general solution, I think it’s reasonable not to handle them.

taiyakihitotsu: Yeah, I agree.
At least, this might not be proper approach in this PR.

> If we can’t address them cleanly with a more general solution, I think it’s reasonable not to handle them.

`UniqueUnionDeep` handles record- and lambda-types already, and it should be fixed for tuples (as you pointed out, Thanks!).
If recursion and enumeration of possible types are sufficient to define `UniqueUnionDeep` correctly (and it's enough, maybe), I think it would be acceptable, as similar patterns already exist in `type-fest`, e.g., the current `PartialDeep`: https://github.com/sindresorhus/type-fest/blob/main/source/partial-deep.d.ts

But, implementing that, discussing the details, and adding more tests would likely scope-creep this PR.

---

Based on your comments and the point above, I propose the following next steps:

In this PR, we'll set aside the "ultra-edge cases" and `UniqueUnionDeep`. Those should be discussed in a more relevant PR such as #1338. Once `IsEqual` is fixed, (I hope) `UnionToTuple` will be fixed automatically as it relies on IsEqual internally.

The plan:

 - Revert f7947ff960d0b262a4a8b4b34579e0a2b0033f9d
 - Move the `UniqueUnionDeep` topic to #1338
 - Merge this `ExcludeExactly` PR first.
 - Subsequently, #1338 will be able to utilize the fixed `UnionToTuple`.

What do you think about this flow?

som-sm: > What do you think about this flow?

Yeah, looks good! So, this PR should basically do this:

> Let's simplify this PR to _only_ fix `UnionToTuple`:
> 
> 1. Let's use `IsEqual` in `ExcludeExactly`, so _no_ need to create `simple-is-equal.ts`.
> 2. No need to expose `LastOfUnion` in this PR.
> 
> So, just add `ExcludeExactly` and use it in `UnionToTuple`. This will help get this merged quickly.

taiyakihitotsu:
@som-sm

I've removed any changes about the identical union case tests and `UniqueUnionDeep`.
How does it look now?

source/internal/type.d.ts
som-sm, line 1: No need to move `LastOfUnion` to `internal` in this PR. Please revert this change, it's not related to this PR.

source/is-equal.d.ts
```diff
@@ -32,7 +31,6 @@ export type IsEqual<A, B> =
 			: false
 		: false;
 
-// This version fails the `equalWrappedTupleIntersectionToBeNeverAndNeverExpanded` test in `test-d/is-equal.ts`.
```
som-sm, line 35: Why remove this comment?

source/internal/type.d.ts
som-sm: Are there any changes as well in `LastOfUnion`?

test-d/union-to-tuple.ts
```diff
@@ -11,3 +11,19 @@ expectType<Options1[number]>({} as (1 | 2 | 3));
...
+// Test for https://github.com/sindresorhus/type-fest/issues/1352
+type DifferentModifierUnion = {readonly a: 0} | {a: 0};
+expectType<DifferentModifierUnion>({} as UnionToTuple<DifferentModifierUnion>[number]);
+expectType<2>({} as UnionToTuple<DifferentModifierUnion>['length']);
+
+// Note: Union order is not guaranteed, but this test is still valuable.
+// https://github.com/microsoft/TypeScript/issues/13298#issuecomment-468375328
+type ReversedDifferentModifierUnion = {a: 0} | {readonly a: 0};
+expectType<ReversedDifferentModifierUnion>({} as UnionToTuple<ReversedDifferentModifierUnion>[number]);
+expectType<2>({} as UnionToTuple<ReversedDifferentModifierUnion>['length']);
```
som-sm: Please remove this. Just `{readonly a: 0} | {a: 0}` is "sufficient".

test-d/union-to-tuple.ts
```diff
@@ -11,3 +11,19 @@ expectType<Options1[number]>({} as (1 | 2 | 3));
...
+
+// Edge cases.
+expectType<[]>({} as UnionToTuple<never>);
+expectType<[any]>({} as UnionToTuple<any>);
+expectType<[unknown]>({} as UnionToTuple<unknown>);
+
+// Test for https://github.com/sindresorhus/type-fest/issues/1352
+type DifferentModifierUnion = {readonly a: 0} | {a: 0};
+expectType<DifferentModifierUnion>({} as UnionToTuple<DifferentModifierUnion>[number]);
+expectType<2>({} as UnionToTuple<DifferentModifierUnion>['length']);
```
som-sm: Remove this, checking `length` is not required, we're already checking the exact elements.

test-d/union-to-tuple.ts
```diff
@@ -11,3 +11,19 @@ expectType<Options1[number]>({} as (1 | 2 | 3));
 
 type Options2 = UnionToTuple<boolean | 1>;
 expectType<Options2[number]>({} as (1 | false | true));
+
+// Edge cases.
+expectType<[]>({} as UnionToTuple<never>);
+expectType<[any]>({} as UnionToTuple<any>);
+expectType<[unknown]>({} as UnionToTuple<unknown>);
```
som-sm: Move this section at the very bottom.

source/union-to-tuple.d.ts
```diff
@@ -52,7 +39,9 @@ const petList = Object.keys(pets) as UnionToTuple<Pet>;
 */
 export type UnionToTuple<T, L = LastOfUnion<T>> =
 IsNever<T> extends false
-	? [...UnionToTuple<Exclude<T, L>>, L]
+	? ExcludeExactly<T, L> extends infer E // Improve performance.
```
som-sm: Can there be a test that validates this perf improvement?

source/internal/type.d.ts
taiyakihitotsu: Addressed.

(`ExcludeExactly` doesn't use `LastOfUnion` now)

source/is-equal.d.ts
taiyakihitotsu: Addressed. Unintentionally...

test-d/union-to-tuple.ts
taiyakihitotsu: Addressed.

test-d/union-to-tuple.ts
taiyakihitotsu: Addressed.

test-d/union-to-tuple.ts
taiyakihitotsu: Addressed.

source/union-to-tuple.d.ts
taiyakihitotsu: > Can there be a test that validates this perf improvement?

Currently, no.
I haven’t started working on this yet...

---

(This is the [comment](https://github.com/sindresorhus/type-fest/pull/1338#discussion_r2769839547) about this gc issue I posted firstly.)

I made a [branch](https://github.com/taiyakihitotsu/type-fest/tree/perftest) to reproduce this.

This commit, https://github.com/taiyakihitotsu/type-fest/commit/e8f09672639b2251f56f90cf42fc3f822856191a, fixes the issue.
And `npm test` will crash if this is reverted.

---

The error logs don't explicitly point to `UnionToTuple` as the cause, and it’s not a standard TS2589 error.

We might need to implement granular tests for individual type definitions within the package, rather than just running a full type-check on test files.
Without these, it’s difficult to pinpoint exactly where the performance bottleneck or crash is originating.

Also, if we do implement such tests, we’ll need to separately discuss where to set the thresholds (e.g., memory limits or complexity scores).

taiyakihitotsu:
@som-sm 

> I've also made some improvements in `ExcludeExactly` tests, please review them once.

Thank you!
I definitely missed the `unknown` and `any` scenarios.

source/union-to-tuple.d.ts
som-sm, line 1: Please create another PR for the purpose of exposing `LastOfUnion` and move all these changes there.

som-sm: Keep this PR only for fixing `UnionToTuple` and adding new `ExcludeExactly` type.

som-sm (changes requested):

source/union-to-tuple.d.ts
taiyakihitotsu: @som-sm 

Yes.
I've opened `LastOfUnion` PR, and updated this PR's title.
 - #1368

Note:
`test-d/last-of-union.ts` in #1368  uses the old (my defined) `ExcludeExactly` which is written with `LastOfUnion`, because `LastOfUnion` should also be guaranteed to keep its union members.

taiyakihitotsu: Here is the expected workflow:

 - Merge `LastOfUnion` (PR #1368)
 - Merge `UnionToTuple` (here)
 - Discuss about `UniqueUnionDeep` in (#1338)
 - (Maybe) Merge `IsEqual` (PR #1338)
 - (And This PR #1241 is also related. This has conflicts and I've set it aside currently as a draft.)

How does this look?

som-sm: > This PR uses the old (my defined) `ExcludeExactly` which is written with `LastOfUnion`, because `LastOfUnion` should also be guaranteed to keep its union members.

Didn't get this. Can't see `ExcludeExactly` using `LastOfUnion`.

taiyakihitotsu: Sorry for the confusion, I meant this part here: https://github.com/sindresorhus/type-fest/pull/1368/changes#diff-6ce133a34fd79b9e6b57c337b69241c8a393399e40f7e37e64bba9e421d7c2b0R31

I've edited the comment to clear it.

source/union-to-tuple.d.ts
```diff
@@ -1,19 +1,48 @@
...
 */
 type LastOfUnion<T> =
-UnionToIntersection<T extends any ? () => T : never> extends () => (infer R)
-	? R
-	: never;
+	true extends IsNever<T>
+		? never
+		: UnionToIntersection<T extends any ? () => T : never> extends () => (infer R)
+			? R
+			: never;
```
som-sm: Why are these changes are still here?

taiyakihitotsu: I'll remove this part after #1368 is merged.
Setting this to draft just in case.

som-sm: > I'll remove this part after #1368 is merged.
> Setting this to draft just in case.

But why? This isn't dependent on #1368, right? This is already reviewed, so it can be merged.

taiyakihitotsu: Once #1368 is merged, `union-to-tuple.d.ts` here can import `LastOfUnion` (which might be renamed to `UnionMember`) instead of defining it locally.
This will allow us to finally remove this redundant code.

https://github.com/taiyakihitotsu/type-fest/blob/95798257945e8226e967b270722c92d78103671b/source/union-to-tuple.d.ts#L82

```typescript
export type UnionToTuple<T, L = LastOfUnion<T>> =
```

som-sm: @taiyakihitotsu But that's not necessary, let's remove these changes and finalise this first.

som-sm: > @taiyakihitotsu But that's not necessary, let's remove these changes and finalise this first.

I have removed it in [2d728ee](https://github.com/sindresorhus/type-fest/pull/1349/commits/2d728ee6b7f55a47a4761d185ddf34ec8c25c330), let's see if it breaks anything.

taiyakihitotsu: I see what you mean now after looking at your commit 2d728ee6b7f55a47a4761d185ddf34ec8c25c330.

Thank you!

som-sm: Everything seems to be working fine. Let's remove this PR from draft?

taiyakihitotsu: Reopened.

Issue #1351, opened 2026-02-07
isUnion incorrect behavior


salisbury-espinosa:
### Bug description

let's say there is a type

```typescript
type TestType =
  | { optionalField?: number; test1: number; test2: string }
  | { test1: number; test2: string };
```

so

```typescript
type OptionalFieldType = TestType['optionalField'];
```

=> `TS2339: Property optionalField does not exist on type TestType` 
=> its correct, BUT:

`IsUnion<TestType>` = false

```typescript
const test: IsUnion<TestType> = true;
```
=> `TS2322: Type true is not assignable to type false`

but `TestType` - union....

PR #1353, opened 2026-02-07
`IsUnion`: Fix behavior when the entire union extends all individual members

taiyakihitotsu:
Close https://github.com/sindresorhus/type-fest/issues/1351

### DIFF
--- a/test-d/is-union.ts
+++ b/test-d/is-union.ts
@@ -23,4 +23,5 @@ type TestUnion =
 	// `TestUnion` extends `{a: number; b: string}`.
 	| {opt?: number; a: number; b: string}
 	| {a: number; b: string};
+
 expectType<IsUnion<TestUnion>>(true);


test-d/is-union.ts
```diff
@@ -16,3 +16,8 @@ expectType<IsUnion<'foo' | 'bar'>>(true);
 expectType<IsUnion<'foo' | 'bar' | 1>>(true);
 expectType<IsUnion<'foo' | 1>>(true);
 expectType<IsUnion<[] | {}>>(true);
+
+type SuperTypeUnion =
+	{optionalField?: number; test1: number; test2: string}
+	| {test1: number; test2: string};
+expectType<true>({} as IsUnion<SuperTypeUnion>);
```
som-sm:
```suggestion
type TestUnion =
	// Here, the entire union extends both the individual members:
	// So, `TestUnion` extends `{opt?: number; a: number; b: string}`, and
	// `TestUnion` extends `{a: number; b: string}`.
	| {opt?: number; a: number; b: string}
	| {a: number; b: string};

expectType<IsUnion<TestUnion>>(true);
```

This makes the test case more clear.

source/is-union.d.ts
```diff
@@ -24,7 +25,7 @@ type InternalIsUnion<T, U = T> =
 	IsNever<T> extends true
 		? false
 		: T extends any
-			? [U] extends [T]
+			? true extends IsEqual<[U], [T]>
```
som-sm:
```suggestion
			? IsEqual<U, T> extends true
```

`U` and `T` don't need to wrapped in a tuple anymore.

taiyakihitotsu:
@som-sm 

Thanks for your quick reviews! Addressed!

PR #931, opened 2024-08-08
Add `EmptyArray` and `IsEmptyArray` types

silverwind:
Fixes: https://github.com/sindresorhus/type-fest/issues/929

PR #1355, opened 2026-02-08
Add `EmptyArray` and `IsEmptyArray` types

### LINKED ISSUE #929: `EmptyArray` type
There is already [`EmptyObject`](https://github.com/sindresorhus/type-fest/blob/main/source/empty-object.d.ts), and I think a type for a strictly empty array would be useful.

derodero24:
Closes #929

Adds `EmptyArray` (`never[]`) for representing a strictly empty array, analogous to `EmptyObject`. Also adds `IsEmptyArray<T>` as a type guard.

`IsEmptyArray` handles:
- `[]`, `never[]`, `readonly []`, `readonly never[]` → `true`
- Non-empty tuples, `string[]`, `any`, `never` → `false`

This picks up where #931 left off, with the readonly array tests that were requested in review.

### DIFF
--- a/source/empty-array.d.ts
+++ b/source/empty-array.d.ts
@@ -1,0 +1,50 @@
+/**
+Represents a strictly empty array, the `[]` value.
+
+Use-cases:
+- Declaring a stable default value for an array parameter to avoid re-renders in React.
+- Clearly expressing the intent of an empty array, similar to how {@link EmptyObject} is used for `{}`.
+
+@example
+```
+import type {EmptyArray} from 'type-fest';
+
+const defaultItems: EmptyArray = [];
+
+// Prevents accidental mutation
+// @ts-expect-error
+defaultItems.push(1);
+```
+
+@category Array
+*/
+export type EmptyArray = [];
+
+/**
+Returns a `boolean` for whether the given array type is an empty array (`[]`).
+
+@example
+```
+import type {IsEmptyArray} from 'type-fest';
+
+type A = IsEmptyArray<[]>; //=> true
+type B = IsEmptyArray<readonly []>; //=> true
+type C = IsEmptyArray<[1, 2, 3]>; //=> false
+type D = IsEmptyArray<string[]>; //=> false
+```
+
+@see {@link EmptyArray}
+@category Array
+*/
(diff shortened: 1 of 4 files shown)


test-d/empty-array.ts
```diff
@@ -0,0 +1,42 @@
...
+// IsEmptyArray — Non-empty cases
+expectType<IsEmptyArray<[1]>>(false);
+expectType<IsEmptyArray<[string, number]>>(false);
+expectType<IsEmptyArray<readonly [1]>>(false);
+expectType<IsEmptyArray<string[]>>(false);
+expectType<IsEmptyArray<readonly string[]>>(false);
+
+// IsEmptyArray — Edge cases
+expectType<IsEmptyArray<any>>(false);
+expectType<IsEmptyArray<never>>(false);
```
sindresorhus, line 45: Some more tests:


```suggestion
expectType<IsEmptyArray<never>>(false);

// IsEmptyArray — Union behavior
expectType<IsEmptyArray<[] | [1]>>({} as boolean);
expectType<IsEmptyArray<readonly [] | readonly [1]>>({} as boolean);

// IsEmptyArray — More non-empty/array boundaries
expectType<IsEmptyArray<[never]>>(false);
expectType<IsEmptyArray<unknown[]>>(false);
expectType<IsEmptyArray<any[]>>(false);
```

source/empty-array.d.ts
```diff
@@ -0,0 +1,50 @@
...
+const defaultItems: EmptyArray = [];
+
+// Prevents accidental mutation
+// @ts-expect-error
+defaultItems.push(1);
+```
+
+@category Array
+*/
+export type EmptyArray = never[];
```
sindresorhus: Why not 
```suggestion
export type EmptyArray = [];
```

?

test-d/empty-array.ts
sindresorhus: Maybe also:

```ts
// @ts-expect-error
emptyArray.length = 1;
```

source/empty-array.d.ts
derodero24: All addressed — simplified to `[]`, added union/boundary tests, and the `length` immutability test.

PR #1356, opened 2026-02-08
Add `SetNullable` type

### LINKED ISSUE #697: [FR] SetNullable
Can we get SetNullable as the opposite of [SetNonNullable](https://github.com/sindresorhus/type-fest/blob/main/source/set-non-nullable.d.ts)?

derodero24:
Closes #697

Adds `SetNullable<BaseType, Keys>` — the inverse of `SetNonNullable`. Makes specified keys nullable by adding `| null` to their value types, while preserving `readonly` and optional modifiers.

If no keys are specified, all keys are made nullable (matching `SetNonNullable`'s default behavior).

Implementation mirrors `SetNonNullable` using a homomorphic mapped type.

### DIFF
--- a/test-d/set-nullable.ts
+++ b/test-d/set-nullable.ts
@@ -55,4 +55,3 @@ expectType<{a: string | undefined | null; b: number}>(variation12);
 // Works with unions where only some keys are shared across branches.
 declare const variation13: SetNullable<{a: number; c: boolean} | {a: string; d: boolean}, 'a'>;
 expectType<{a: number | null; c: boolean} | {a: string | null; d: boolean}>(variation13);
-


test-d/set-nullable.ts
som-sm, line 1: Please add more tests, things like unions, index signatures, primitives, `any`/`never` cases, etc. I realise this mirrors `SetNonNullable`, but we can improve there as well.

derodero24: @som-sm Added tests for unions, index signatures, and `any`/`never` keys.

test-d/set-nullable.ts
```diff
@@ -0,0 +1,47 @@
...
+expectType<{[k: string]: unknown; a: number | null; b: string}>(variation9);
+
+// Makes all keys nullable when `Keys` is `any`.
+declare const variation10: SetNullable<{readonly a: number; b?: string; c: boolean}, any>;
+expectType<{readonly a: number | null; b?: string | null; c: boolean | null}>(variation10);
+
+// Does nothing when `Keys` is `never`.
+declare const variation11: SetNullable<{a: number; readonly b?: string; readonly c: boolean}, never>;
+expectType<{a: number; readonly b?: string; readonly c: boolean}>(variation11);
+
```
sindresorhus, line 50:
```suggestion

// Preserves existing `undefined` when adding `null`.
declare const variation12: SetNullable<{a: string | undefined; b: number}, 'a'>;
expectType<{a: string | undefined | null; b: number}>(variation12);

// Works with unions where only some keys are shared across branches.
declare const variation13: SetNullable<{a: number; c: boolean} | {a: string; d: boolean}, 'a'>;
expectType<{a: number | null; c: boolean} | {a: string | null; d: boolean}>(variation13);
```

derodero24: Added both tests.

test-d/set-nullable.ts
```diff
@@ -0,0 +1,55 @@
...
+// Preserve optional modifier.
+declare const variation3: SetNullable<{a: number; b?: string}, 'b'>;
+expectType<{a: number; b?: string | null}>(variation3);
+
+// Key already nullable remains unchanged.
+declare const variation4: SetNullable<{a: number | null; b: string}, 'a'>;
+expectType<{a: number | null; b: string}>(variation4);
+
+// Fail if type changes even if nullable is right.
```
som-sm, line 20: WDYM by this? What's failing here?

som-sm:
Also, *maybe* we should remove the `keyof` constraint from the second argument because, in the case of unions, it prevents updating keys that are *not* common across all members.

Currently, the JSDoc says:

> If no keys are given, all keys will be made nullable.

But this is *not* true for unions:

```ts
type T = SetNullable<{a: number; b: number} | {a: string; c: string}>;
//=> {a: number | null; b: number} | {a: string | null; c: string}
```

And it's not that this behavior cannot be achieved. It *can* be achieved by using `any` as the second argument:

```ts
type T = SetNullable<{a: number; b: number} | {a: string; c: string}, any>;
//=> {a: number | null; b: number | null} | {a: string | null; c: string | null}
```

So even if we keep the `keyof` constraint, I guess a better default would be `any`. But I think the constraint should either be removed completely or at least replaced with `KeysOfUnion`. @sindresorhus WDYT?

sindresorhus:
I think we should use `KeysOfUnion`.

We should probably make the same change to `SetNonNullable` too for consistency.

PR #1357, opened 2026-02-08
Add function parameter constraint examples to numeric comparison types

### LINKED ISSUE #957: Can i use GreaterThan as parameter type?
Newbie here...

Is it possible to use GreaterThan type for the function/method parameters?

when I write the following function:
```
import type {GreaterThan} from 'type-fest'

function printNumberOfBooks <N extends number> (numberOfBooks: GreaterThan<N, 0>): void {
    console.log(numberOfBooks)
}
```

and... :

```
printNumberOfBooks(2)
```

I get the following typescript error:
> Argument of type 'number' is not assignable to parameter of type 'never'

It took me hours... I didn't find a solution

derodero24:
Closes #957

Adds a second `@example` block to `GreaterThan` showing how to use it as a function parameter constraint.

Users attempting to use `GreaterThan<N, 0>` directly as a parameter type get `never` because the generic is unresolved. The added example demonstrates the correct pattern using a conditional type:

```ts
declare function setPositive<N extends number>(value: GreaterThan<N, 0> extends true ? N : never): void;
```

### DIFF
--- a/source/greater-than.d.ts
+++ b/source/greater-than.d.ts
@@ -44,13 +44,14 @@ import type {GreaterThan} from 'type-fest';
 // Use `GreaterThan` to constrain a function parameter to positive numbers.
 declare function setPositive<N extends number>(value: GreaterThan<N, 0> extends true ? N : never): void;
 
-setPositive(1);
-
-// @ts-expect-error
-setPositive(-1);
+setPositive(1); // ✅ Allowed
+setPositive(2); // ✅ Allowed
 
 // @ts-expect-error
 setPositive(0);
+
+// @ts-expect-error
+setPositive(-1);
 ```
 */
 export type GreaterThan<A extends number, B extends number> =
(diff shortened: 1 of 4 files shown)


som-sm (changes requested):
Please resolve the conflict.

source/greater-than.d.ts
```diff
@@ -20,6 +20,22 @@ type B = GreaterThan<1, 1>;
...
+// Use `GreaterThan` to constrain a function parameter to positive numbers.
+declare function setPositive<N extends number>(value: GreaterThan<N, 0> extends true ? N : never): void;
+
+setPositive(1);
+
+// @ts-expect-error
+setPositive(-1);
+
+// @ts-expect-error
+setPositive(0);
```
som-sm: Please also add similar examples to `LessThan`, `LessThanOrEqual` and `GreaterThanOrEqual`.

derodero24: Done. Added similar examples to all three types and resolved the conflict.

Issue #1358, opened 2026-02-08
`ConditionalPickDeep` returns `{}` instead of `never` when no keys match


derodero24:
Same issue as #1197 but for `ConditionalPickDeep`.

When no keys match the given condition, `ConditionalPickDeep` returns `{}` instead of `never`:

```ts
import type {ConditionalPickDeep} from 'type-fest';

type Example = {
  a: string;
  b: string | boolean;
  c: {
    d: string;
    e: {
      f?: string;
      g?: boolean;
    };
    j: boolean;
  };
};

type NumberPick = ConditionalPickDeep<Example, number>;
//   ^? {} — should be `never`

const test: NumberPick = {
  anything: 'is accepted', // No error, but should be
};
```

`ConditionalPick` is being fixed in #1197. `ConditionalPickDeep` has the same behavior and should be addressed separately since it uses a different internal implementation (`ConditionalExcept`-based).

PR #1359, opened 2026-02-08
`ConditionalPick`: Fix returning `{}` instead of `never` when no keys match

### LINKED ISSUE #1197: Bug ConditionalPick
### Bug description

```typescript
import type { ConditionalPick } from 'type-fest';

interface TestInterface {
  a: string;
}

// TS validates that everything is ok - WHY ? 
// ConditionalPick<TestInterface, number> === {} .....

const test123123123: ConditionalPick<TestInterface, number> = {
  sdfsdfsdf: 'asdasd',
};
```

derodero24:
Fixes #1197

## Summary
- When no keys match the given `Condition`, `ConditionalPick` previously returned `Pick<Base, never>` which resolves to `{}` (empty object type). This allowed any properties to be assigned without type errors.
- Now it checks if `ConditionalKeys` returns `never` using `IsNever`, and returns `never` in that case.
- Added test cases to verify the `never` return behavior.

## Note
The same issue exists in `ConditionalPickDeep` — tracked separately in #1358.

### DIFF
--- a/test-d/conditional-pick.ts
+++ b/test-d/conditional-pick.ts
@@ -31,5 +31,5 @@ expectType<{a: string; c?: string}>(exampleConditionalPickWithUndefined);
 declare const noMatchingKeys: ConditionalPick<Example, number>;
 expectType<never>(noMatchingKeys);
 
-declare const noMatchingKeysBoolean: ConditionalPick<{a: string; b: number}, boolean>;
-expectType<never>(noMatchingKeysBoolean);
+declare const noMatchingKeys2: ConditionalPick<{a: string; b: number}, boolean>;
+expectType<never>(noMatchingKeys2);


source/conditional-pick.d.ts
```diff
@@ -38,9 +39,8 @@ type StringKeysOnly = ConditionalPick<Example, string>;
 
 @category Object
 */
-export type ConditionalPick<Base, Condition> = Pick<
-	Base,
-	ConditionalKeys<Base, Condition>
->;
+export type ConditionalPick<Base, Condition> = IsNever<ConditionalKeys<Base, Condition>> extends true
+	? never
+	: Pick<Base, ConditionalKeys<Base, Condition>>;
```
som-sm:
```suggestion
export type ConditionalPick<Base, Condition> = ConditionalKeys<Base, Condition> extends infer Keys
	? IsNever<Keys> extends true
		? never
		: Pick<Base, Keys & keyof Base>
	: never;
```

Better than repeating `ConditionalKeys<Base, Condition>` twice.

test-d/conditional-pick.ts
```diff
@@ -26,3 +26,10 @@ expectType<{name: string; successes: number; failures: bigint}>(awesomeCondition
 
 declare const exampleConditionalPickWithUndefined: ConditionalPick<Example, string | undefined>;
 expectType<{a: string; c?: string}>(exampleConditionalPickWithUndefined);
+
+// Returns `never` when no keys match the condition
+declare const noMatchingKeys: ConditionalPick<Example, number>;
+expectNever(noMatchingKeys);
+
+declare const noMatchingKeysBoolean: ConditionalPick<{a: string; b: number}, boolean>;
+expectNever(noMatchingKeysBoolean);
```
som-sm:
```suggestion
declare const noMatchingKeys: ConditionalPick<Example, number>;
expectType<never>(noMatchingKeys);

declare const noMatchingKeysBoolean: ConditionalPick<{a: string; b: number}, boolean>;
expectType<never>(noMatchingKeysBoolean);
```

Use `expectType` instead of `expectNever` because the editor marks all the code after `expectNever` as unreachable (not good DX).

Image

source/conditional-pick.d.ts
derodero24: @som-sm Thanks, updated both. The `infer Keys` approach is much cleaner.

PR #1360, opened 2026-02-08
`ConditionalPickDeep`: Fix returning `{}` instead of `never` when no keys match

### LINKED ISSUE #1358: `ConditionalPickDeep` returns `{}` instead of `never` when no keys match

derodero24:
Fixes #1358

## Summary
- When no keys match the given `Condition` at any depth, `ConditionalPickDeep` previously returned `{}` (empty object type). This allowed any properties to be assigned without type errors.
- Added a `_NeverIfEmpty` helper that checks if the result has no keys and returns `never` in that case. This is applied only at the top level to avoid interfering with the internal recursive `EmptyObject` exclusion logic.
- Updated JSDoc twoslash comment and test cases accordingly.

## Note
This is the deep variant of the same issue fixed in #1359 for `ConditionalPick`.

### DIFF
--- a/source/conditional-pick-deep.d.ts
+++ b/source/conditional-pick-deep.d.ts
@@ -105,7 +105,7 @@ export type ConditionalPickDeep<
 	ApplyDefaultOptions<ConditionalPickDeepOptions, DefaultConditionalPickDeepOptions, Options>
 >>;
 
-type _NeverIfEmpty<Type> = [keyof Type] extends [never] ? never : Type;
+type _NeverIfEmpty<Type> = Type extends EmptyObject ? never : Type;
 
 type _ConditionalPickDeep<
 	Type,
--- a/test-d/conditional-pick-deep.ts
+++ b/test-d/conditional-pick-deep.ts
@@ -108,6 +108,10 @@ expectType<never>(emptyEqualityPick);
 declare const noMatchingKeys: ConditionalPickDeep<{a: string; b: number}, boolean>;
 expectType<never>(noMatchingKeys);
 
+// Union with no common properties
+declare const unionNoCommon: ConditionalPickDeep<{a: string} | {b: string}, string>;
+expectType<{a: string} | {b: string}>(unionNoCommon);
+
 declare const stringOrBooleanPick: ConditionalPickDeep<Example, string | boolean>;
 expectType<{
 	literal: 'foo';


source/conditional-pick-deep.d.ts
```diff
@@ -99,11 +99,13 @@ export type ConditionalPickDeep<
...
 	Options extends ConditionalPickDeepOptions = {},
-> = _ConditionalPickDeep<
+> = _NeverIfEmpty<_ConditionalPickDeep<
 	Type,
 	Condition,
 	ApplyDefaultOptions<ConditionalPickDeepOptions, DefaultConditionalPickDeepOptions, Options>
->;
+>>;
+
+type _NeverIfEmpty<Type> = [keyof Type] extends [never] ? never : Type;
```
som-sm: This is incorrect, because `keyof Type` would be `never` even for unions with _no_ common properties. For example, `ConditionalPickDeep<{a: string} | {b: string}>, string>` should return `{a: string} | {b: string}`, but with this change it returns `never`.


```suggestion
type _NeverIfEmpty<Type> = Type extends EmptyObject ? never : Type;
```
The above is better because it lets `Type` distribute, and `EmptyObject` is preferable to `keyof Type` since `EmptyObject` is already used to achieve the same behavior in nested cases.
https://github.com/sindresorhus/type-fest/blob/3995003ac0ddab36ebd37e8afbaac1652ea68d4b/source/conditional-pick-deep.d.ts#L118

derodero24: Good catch, fixed. Also added a test for the union case with no common properties.

som-sm (approved):

PR #1362, opened 2026-02-09
Fix syntax for specifying types in certain JSDoc codeblocks

som-sm:
Noticed at certain places we don't use the twoslash (`//=>`) syntax for specifying types. This PR fixes that.

### DIFF
--- a/source/set-non-nullable.d.ts
+++ b/source/set-non-nullable.d.ts
@@ -15,6 +15,7 @@ type Foo = {
 	c?: boolean | null;
 };
 
+// Note: In the following example, `c` can no longer be `null`, but it's still optional.
 type SomeNonNullable = SetNonNullable<Foo, 'b' | 'c'>;
 //=> {a: null | number; b: string; c?: boolean}
 


source/pick-deep.d.ts
```diff
@@ -86,7 +71,7 @@ export type PickDeep<T, PathUnion extends Paths<T>> =
 			}[PathUnion]
 			>
 			: T extends object
-				? Simplify<UnionToIntersection<{
+				? SimplifyDeep<UnionToIntersection<{
```
som-sm, line 74: Had to do this because _just_ `Simplify` wasn't enough:

Image

Notice the unsimplified `&` in the output.

source/set-optional.d.ts
```diff
@@ -18,11 +18,7 @@ type Foo = {
 };
 
 type SomeOptional = SetOptional<Foo, 'b' | 'c'>;
-// type SomeOptional = {
-// 	a: number;
-// 	b?: string; // Was already optional and still is.
-// 	c?: boolean; // Is now optional.
-// }
```
som-sm, line 25: These comments are not required, the type/behavior here is pretty self explanatory.

Same with all other comments in this PR.

source/set-non-nullable.d.ts
```diff
@@ -15,26 +15,19 @@ type Foo = {
...
-// }
+//=> {a: number; b: string; c?: boolean}
 ```
 
 @category Object
 */
 export type SetNonNullable<BaseType, Keys extends keyof BaseType = keyof BaseType> = {
 	[Key in keyof BaseType]: Key extends Keys
-		? NonNullable<BaseType[Key]>
+		? BaseType[Key] & {} // `& {}` is used instead of `NonNullable<BaseType[Key]>` because `NonNullable` doesn't get simplified.
```
som-sm, line 30: Had to change this because `NonNullable` doesn't get simplified.

Image

This doesn't affect behavior because `NonNullable` internally is:
```ts
type NonNullable<T> = T & {};
```
---

Image

Note: Once we have the [expandable hover previews](https://devblogs.microsoft.com/typescript/announcing-typescript-5-9/#expandable-hovers-(preview)) feature available via API, we'd not need changes like these.

test-d/pick-deep.ts
```diff
@@ -71,7 +71,7 @@ declare const genericTest: PickDeep<GenericType<number>, 'genericKey'>;
 expectType<{genericKey: number}>(genericTest);
 
 declare const union: PickDeep<Testing, 'object.number' | 'object.string'>;
-expectType<{object: {number: number} & {string: string}}>(union);
+expectType<{object: {number: number; string: string}}>(union);
```
som-sm, line 74: This is required because `expectType` doesn't treat `{a: number; b: number}` same as `{a: number} & {b: number}`.

source/set-non-nullable.d.ts
som-sm: > Note: Once we have the [expandable hover previews](https://devblogs.microsoft.com/typescript/announcing-typescript-5-9/#expandable-hovers-(preview)) feature available via API, we'd not need changes like these.

This is _already_ available, it's just marked internal.
https://github.com/microsoft/TypeScript/blob/dd830711041b7b0cfd3da7937755996b1e1b1c7e/src/services/types.ts#L588

I’ve created a separate PR (#1364) to make use of this. In that PR, I’ll revert this back to `NonNullable` and also check if previously any other changes were made _just_ to satisfy the twoslash validation.

PR #1364, opened 2026-02-09
Validate twoslash types against different verbosity levels

som-sm:
# Problem

Currently, the twoslash validator _only_ allows types that match the default quickinfo output. This is somewhat restrictive, because in some cases we want the twoslash type to be simpler than what the default quickinfo displays.

For example, in the following code block we want the twoslash type to be `string | undefined`, and not `Optional<string>`.

Screenshot 2026-03-07 at 11 33 03 AM<br>

**This behavior can be improved using the newly added [`verbosityLevel` parameter](https://github.com/microsoft/TypeScript/blob/dd830711041b7b0cfd3da7937755996b1e1b1c7e/src/services/types.ts#L588) in `getQuickInfoAtPosition`.**

# Attempt 1
My initial idea was to always enforce twoslash types to be at maximum verbosity (i.e., `verbosityLevel` set to `Infinity`).

This turned out to be problematic because in some cases we _don't_ want the type to be fully simplified. For example, in the following code block we want the twoslash type to reference `Config`, instead of expanding it to `{featureFlags: {[x: string]: boolean}}`.

Screenshot 2026-03-07 at 10 48 17 AM Screenshot 2026-03-07 at 10 33 17 AM

# Attempt 2
Next, I thought of exposing `verbosityLevel` as a rule option, so that we can tweak it wherever required. 

But the problem with this was that we may require different verbosity levels across different twoslash types in the _same_ file, which would not have been possible with this approach.

# Solution
Twoslash types are now validated against different verbosity levels. By default, levels `0` and `Infinity` are allowed.

Screenshot 2026-03-07 at 10 55 06 AM<br>

If levels other than `0` and `Infinity` are required, the lint rule accepts a `verbosityLevels` option that takes in an array of allowed verbosity levels. This allows for multiple different verbosity levels in the same file, which addresses the issue described in "Attempt 2".

Screenshot 2026-03-05 at 11: 46: 24 PM<br>

Note that `0` and `Infinity` are always supported, even if `verbosityLevels` is explicitly configured, since these levels are required in most cases.

<br>

For auto-fix type at `Infinity` verbosity level is used:

Untitled (1920 x 661 px)

### DIFF
--- a/lint-rules/validate-jsdoc-codeblocks.js
+++ b/lint-rules/validate-jsdoc-codeblocks.js
@@ -100,8 +100,11 @@ export const validateJSDocCodeblocksRule = /** @type {const} */ ({
 				},
 			},
 		}],
+		/** @type {unknown[]} */
+		defaultOptions: [{
+			verbosityLevels: [],
+		}],
 	},
-	defaultOptions: [],
 	create(context) {
 		const filename = context.filename.replaceAll('\\', '/');
 
@@ -336,7 +339,7 @@ function validateTwoslashTypes(context, env, code, codeStartIndex) {
 	const sourceFile = env.languageService.getProgram().getSourceFile(FILENAME);
 	const lines = code.split('\n');
 
-	const specifiedVerbosityLevels = context.options[0]?.verbosityLevels ?? [];
+	const specifiedVerbosityLevels = context.options[0].verbosityLevels;
 	const verbosityLevels = [0, ...specifiedVerbosityLevels, Infinity]; // Keep `Infinity` last since suggestion logic relies on the order
 
 	for (const [index, line] of lines.entries()) {


som-sm:
I've also reverted certain changes that were made only to satisfy twoslash validation.

image

lint-rules/validate-jsdoc-codeblocks.js
```diff
@@ -87,9 +87,24 @@ export const validateJSDocCodeblocksRule = /** @type {const} */ ({
...
+					type: 'array',
+					uniqueItems: true,
+					items: {
+						minimum: 0,
+						type: 'number',
+					},
+				},
+			},
+		}],
+		/** @type {unknown[]} */
```
som-sm, line 103: Without this, we get this error:

Image

PR #1365, opened 2026-02-09
docs(readme): introduce Used by section

sarahdayan:
## Summary

This introduces a **Used by** section in the README to showcase the top users of `type-fest` and how many repositories rely on this package.

The image is generated with [usedby.dev](https://usedby.dev/) and refreshed every day.

I added the image low in the README not to disrupt the flow, but happy to move it higher if this makes sense!


sindresorhus:
Thanks for the suggestion, but this is not something I want.

PR #1368, opened 2026-02-14
Add `UnionMember` type

taiyakihitotsu:
## Add `UnionMember`

 - This is useful to recurse union-types, so I think this should be exported as a global for end-users.
 - We can evade a duplicated definition potentially.

And the current `UnionMember`, which locally defined in `UnionToTuple`, returns `unknown`.

```typescript
expectType<never>({} as UnionMember<never>);
//=> unknown
```

This would cause an error if end-users use `IsNever` for termination check.

Ideally, `UnionMember` reduces a type, e.g., assuming $A$ and $B$ are type arguments of `UnionMember<A | B>`, $A \cup B$ -> $B$, so $\emptyset$ -> $\emptyset$ intuitively ($\bot$ -> $\bot$).
This PR's `UnionMember` handles this.

---

### Note
 - This PR is separated from https://github.com/sindresorhus/type-fest/pull/1349


readme.md
```diff
@@ -191,6 +191,7 @@ Click the type names for complete docs.
 - [`ConditionalSimplify`](source/conditional-simplify.d.ts) - Simplifies a type while including and/or excluding certain types from being simplified.
 - [`ConditionalSimplifyDeep`](source/conditional-simplify-deep.d.ts) - Recursively simplifies a type while including and/or excluding certain types from being simplified.
 - [`ExclusifyUnion`](source/exclusify-union.d.ts) - Ensure mutual exclusivity in object unions by adding other members’ keys as `?: never`.
+- [`LastOfUnion`](source/last-of-union.d.ts) - Return a member of a union type. Order is not guaranteed.
```
```suggestion
- [`LastOfUnion`](source/last-of-union.d.ts) - Returns a member of a union type. Order is not guaranteed.
```

test-d/last-of-union.ts
```diff
@@ -0,0 +1,43 @@
+import {expectType} from 'tsd';
+import type {LastOfUnion, IsAny, IsUnknown, IsNever} from '../index.d.ts';
+
+// `LastOfUnion` distinguishes between different modifiers.
+type UnionType = {a: 0} | {b: 0} | {a?: 0} | {readonly a?: 0} | {readonly a: 0};
+expectType<true>({} as LastOfUnion<UnionType> extends UnionType ? true : false);
+expectType<false>({} as UnionType extends LastOfUnion<UnionType> ? true : false);
+
+// `never` acts as a termination condition with `IsNever`.
+expectType<never>({} as LastOfUnion<never>);
```
sindresorhus:
```suggestion
expectType<never>({} as LastOfUnion<never>);
expectType<1>({} as LastOfUnion<1 | never>);
```

test-d/last-of-union.ts
```diff
@@ -0,0 +1,43 @@
...
+
+expectType<true>({} as IsUnknown<LastOfUnion<unknown>>);
+expectType<true>({} as IsAny<LastOfUnion<any>>);
+
```
sindresorhus:
```suggestion
expectType<true>({} as IsAny<LastOfUnion<any>>);

type UnionToTupleWithExclude<T, L = LastOfUnion<T>> =
	IsNever<T> extends false
		? [...UnionToTupleWithExclude<Exclude<T, L>>, L]
		: [];

expectType<1 | 2 | 3>({} as UnionToTupleWithExclude<1 | 2 | 3>[number]);
```

test-d/last-of-union.ts
```diff
@@ -0,0 +1,43 @@
...
+type UnionToTuple<T, L = LastOfUnion<T>> =
+IsNever<T> extends false
+	? [...UnionToTuple<ExcludeExactly<T, L>>, L]
+	: [];
+
+type MatchOrNever<A, B> =
+	[unknown, B] extends [A, never]
+		? A
+		// This equality code base below doesn't work if `A` is `unknown` and `B` is `never` case.
+		// So this branch should be wrapped to take care of this.
```
sindresorhus:
```suggestion
		// The equality comparison below does not work when `A` is `unknown` and `B` is `never`.
		// This branch handles that case.
```

sindresorhus (changes requested):

test-d/last-of-union.ts
taiyakihitotsu: Adding `never` here is redundant because TypeScript's union types automatically exclude `never`. Should we skip this change?

taiyakihitotsu:
@sindresorhus

Thanks for your reviews!

And I believe this suggestion implies that `test-d/last-of-union.ts` should have a test for recursion without directly importing `UnionToTuple` to clarify it. Right?

If so, `UnionToTuple` in `test-d/last-of-union.ts` fulfills the requirement. Would it be okay to integrate both in `test-d/last-of-union.ts`?

---

https://github.com/sindresorhus/type-fest/pull/1368#discussion_r2807867095

```typescript
type UnionToTupleWithExclude<T, L = LastOfUnion<T>> =
	IsNever<T> extends false
		? [...UnionToTupleWithExclude<Exclude<T, L>>, L]
		: [];

expectType<1 | 2 | 3>({} as UnionToTupleWithExclude<1 | 2 | 3>[number]);
```

source/last-of-union.d.ts
```diff
@@ -0,0 +1,46 @@
...
+type Last = LastOfUnion<1 | 2 | 3>;
+//=> 3
+
+type LastNever = LastOfUnion<never>;
+//=> never
+```
+
+@category Type
+*/
+export type LastOfUnion<T> =
```
som-sm: This type shouldn't be called `LastOfUnion` IMO, because it's not "really" last.

source/last-of-union.d.ts
```diff
@@ -0,0 +1,46 @@
...
+	[T] extends [never]
+		? never
```
som-sm: Use `IsNever`?

source/last-of-union.d.ts
```diff
@@ -0,0 +1,46 @@
...
+@example
+```
+import type {LastOfUnion} from 'type-fest';
+
+type Last = LastOfUnion<1 | 2 | 3>;
+//=> 3
+
+type LastNever = LastOfUnion<never>;
+//=> never
+```
```
som-sm: Move this example before the other one.

test-d/last-of-union.ts
```diff
@@ -0,0 +1,50 @@
...
+// `LastOfUnion` distinguishes between different modifiers.
+type UnionType = {a: 0} | {b: 0} | {a?: 0} | {readonly a?: 0} | {readonly a: 0};
+expectType<true>({} as LastOfUnion<UnionType> extends UnionType ? true : false);
+expectType<false>({} as UnionType extends LastOfUnion<UnionType> ? true : false);
+
+// `never` acts as a termination condition with `IsNever`.
+expectType<never>({} as LastOfUnion<never>);
+
+expectType<true>({} as IsUnknown<LastOfUnion<unknown>>);
+expectType<true>({} as IsAny<LastOfUnion<any>>);
```
som-sm:
```suggestion
expectType<unknown>({} as LastOfUnion<unknown>);
expectType<any>({} as LastOfUnion<any>);
```

test-d/last-of-union.ts
```diff
@@ -0,0 +1,50 @@
...
+			: ExcludeExactly<_ExcludeExactly<UnionU, D>, _ExcludeExactly<DeleteT, D>>
+		: never;
+
+type _ExcludeExactly<UnionU, DeleteT> =
+	UnionU extends unknown // Only for union distribution.
+		? MatchOrNever<UnionU, DeleteT>
+		: never;
+
+type DifferentModifierUnion = {readonly a: 0} | {a: 0};
+expectType<DifferentModifierUnion>({} as UnionToTuple<DifferentModifierUnion>[number]);
```
som-sm: Umm...I don't think this approach for testing is ideal where we implement something with `LastOfUnion` and then test the implementation. We'll have to find a better way.

som-sm (changes requested):

test-d/last-of-union.ts
som-sm: Let's just use something similar to `UnionToTuple`, like:

```ts
type Test<T, L = LastOfUnion<T>> =
IsNever<T> extends false
	? Test<Exclude<T, L>> | L
	: never;

expectType<1 | 2 | 3>({} as Test<1 | 2 | 3>);
```

Also, remove everything else. Here, we are not trying to test quirks of a specific implementation, such as `Exclude` behaving unexpectedly in certain cases. The goal here is to test `LastOfUnion` only, so the surrounding implementation should be as minimal as possible.

source/last-of-union.d.ts
som-sm: Maybe, `MemberOfUnion`? And then add names like `LastOfUnion` / `FirstOfUnion` in the "Alternative type names" section of the README.

sindresorhus: I agree. How about `UnionMember`. Shorter. Or if we want to really make it clear how it works: `ArbitraryUnionMember` ?

som-sm: I think `UnionMember` is fine.

taiyakihitotsu: I renamed this to `UnionMember`.

test-d/last-of-union.ts
som-sm: > ```ts
> type Test<T, L = LastOfUnion<T>> =
> IsNever<T> extends false
> 	? Test<Exclude<T, L>> | L
> 	: never;
> ```

Even this is not ideal, because it does _not_ guarantee that `LastOfUnion` picks exactly one member at a time. For example, in the `1 | 2 | 3` case, if `LastOfUnion<1 | 2 | 3>` returns `1 | 2` first, and then `LastOfUnion<Exclude<1 | 2 | 3, 1 | 2>>` returns `3`, the test would still pass.

This is a better approach I guess:
```ts
type Test<T, L = LastOfUnion<T>> =
IsNever<T> extends false
	? Test<Exclude<T, L>> | [L]
	: never;

expectType<[1] | [2] | [3]>({} as Test<1 | 2 | 3>);
```

I hope this makes sense?

taiyakihitotsu: I kept only this definition.
https://github.com/sindresorhus/type-fest/pull/1368#discussion_r2807867095

---

After both are merged, #1368 and #1349, then open a PR to fix `test-d/union-member.ts` to add the ~~`readonly`~~`ExcludeExactly` case.
Or include the test case into #1349 (and merge this after #1368).

Does this sound acceptable?
If so, which approach do you prefer?

taiyakihitotsu: Sorry, our comments crossed!

> Even this is not ideal, because it does not guarantee that LastOfUnion picks exactly one member at a time.

I’ve added your test case to `test-d/union-member.ts` and included a comment explaining the scenario it covers.

taiyakihitotsu: > After both are merged, https://github.com/sindresorhus/type-fest/pull/1368 and https://github.com/sindresorhus/type-fest/pull/1349, then open a PR to fix test-d/union-member.ts to add the ~~`readonly`~~`ExcludeExactly` case.

And I've updated this my comment, because this is NOT `readonly` but `ExcludeExacly` case.

som-sm: > I kept only this definition.
> [#1368 (comment)](https://github.com/sindresorhus/type-fest/pull/1368#discussion_r2807867095)

This test case is _not_ great because it would **pass** even for this:
```ts
import type {IsNever} from 'type-fest';

type LastOfUnion<T> = T; // Incorrect implementation

type UnionToTupleWithExclude<T, L = LastOfUnion<T>> =
	IsNever<T> extends false
		? [...UnionToTupleWithExclude<Exclude<T, L>>, L]
		: [];

type Test = UnionToTupleWithExclude<1 | 2 | 3>[number] // Passes
//=> 1 | 2 | 3
```

source/union-member.d.ts
```diff
@@ -0,0 +1,47 @@
+import type {UnionToIntersection} from './union-to-intersection.d.ts';
+import type {IsNever} from './is-never.d.ts';
+
+/**
+Returns a member of a union type. Order is not guaranteed.
```
som-sm:
```suggestion
Returns an arbitrary member of a union type.
```

I think this is better because `UnionMember` returns _just_ a single item, so saying "order" could be confusing.

source/union-member.d.ts
```diff
@@ -0,0 +1,47 @@
...
+import type {UnionMember} from 'type-fest';
+
+type Last = UnionMember<1 | 2 | 3>;
+//=> 3
+
+type LastNever = UnionMember<never>;
+//=> never
+```
+
+@see https://github.com/microsoft/TypeScript/issues/13298#issuecomment-468375328
```
som-sm: Remove this.

Without any context, it could be confusing why this is linked.

source/union-member.d.ts
```diff
@@ -0,0 +1,47 @@
...
+
+Use-cases:
+- Implementing recursive type functions that accept a union type.
```
som-sm:
```suggestion
- Implementing recursive types that accept a union type.
```

source/union-member.d.ts
```diff
@@ -0,0 +1,47 @@
...
+- Reducing a union one member at a time, for example when building tuples.
```
som-sm: Remove this point, this covers the same idea as the previous point.

source/union-member.d.ts
```diff
@@ -0,0 +1,47 @@
...
+
+It can detect a termination case using {@link IsNever `IsNever`}.
```
som-sm: This can be removed I guess.

test-d/union-member.ts
```diff
@@ -0,0 +1,20 @@
...
+
+expectType<unknown>({} as UnionMember<unknown>);
+expectType<any>({} as UnionMember<any>);
+
+// Ensure exactly one member is selected at a time, while covering all members in the union.
+type UnionToTupleWithExclude<T, L = UnionMember<T>> =
+	IsNever<T> extends false
+		? UnionToTupleWithExclude<Exclude<T, L>> | [L]
+		: never;
+expectType<[1] | [2] | [3]>({} as UnionToTupleWithExclude<1 | 2 | 3>);
```
som-sm:
```suggestion
// `WrapMemberInTuple` ensures `UnionMember` selects exactly one member at a time.
type WrapMemberInTuple<T, L = UnionMember<T>> =
	IsNever<T> extends false
		? WrapMemberInTuple<Exclude<T, L>> | [L]
		: never;

expectType<[1] | [2] | [3]>({} as WrapMemberInTuple<1 | 2 | 3>);
expectType<['foo'] | ['bar'] | ['baz']>({} as WrapMemberInTuple<'foo' | 'bar' | 'baz'>);
expectType<[1] | ['foo'] | [true] | [100n] | [null] | [undefined]>(
	{} as WrapMemberInTuple<1 | 'foo' | true | 100n | null | undefined>,
);
expectType<[{a: string}] | [{b: number}]>({} as WrapMemberInTuple<{a: string} | {b: number}>);
```

som-sm: > while covering all members in the union.

This depends on how we're excluding the members from the union, it is _not_ related to `UnionMember`.

test-d/union-member.ts
```diff
@@ -0,0 +1,20 @@
+import {expectType} from 'tsd';
+import type {UnionMember, IsNever} from '../index.d.ts';
+
+// `UnionMember` distinguishes between different modifiers.
+type UnionType = {a: 0} | {b: 0} | {a?: 0} | {readonly a?: 0} | {readonly a: 0};
+expectType<true>({} as UnionMember<UnionType> extends UnionType ? true : false);
+expectType<false>({} as UnionType extends UnionMember<UnionType> ? true : false);
```
som-sm: > // `UnionMember` distinguishes between different modifiers.

With these tests you can't _really_ tell if `UnionMember` distinguishes b/w `{readonly a: 0}` and `{a: 0}`.

taiyakihitotsu: @som-sm

Thanks for your reviews! Addressed!

---

https://github.com/sindresorhus/type-fest/pull/1368/changes#diff-a6c3dfecf92726bb7087525c14ee131d87b59973417d501dd1e78577ed090fdfR4

```typescript
type UnionType = {a: 0} | {readonly a: 0};
type PickedUnionMember = UnionMember<UnionType>;
expectType<true>({} as PickedUnionMember extends UnionType ? true : false);
expectType<false>({} as IsEqual<UnionType, PickedUnionMember>);
```

This ensures that `PickedUnionMember` is a constituent of `UnionType` and that they are not identical.

The subsequent tests using `WrapMemberInTuple` further ensure that the result is exactly one member of a union type.

(While I think these tests could be rewritten using `ExcludeExactly` for integration, I'm not sure if it’s worth opening a separate PR just for that.)

Does this look good to you?

som-sm: > ```ts
> expectType<false>({} as IsEqual<UnionType, PickedUnionMember>);
> ```

This one is fine. It correctly validates that `PickedUnionMember` is either `{a: 0}` or `{readonly a: 0}` and not both `{a: 0} | {readonly a: 0}`.



> ```ts
> expectType<true>({} as PickedUnionMember extends UnionType ? true : false);
> ```

Remove this because this would pass even if `PickedUnionMember` were `{a: 0} | {readonly a: 0}`.

source/union-member.d.ts
```diff
@@ -0,0 +1,39 @@
...
+Returns an arbitrary member of a union type.
+
+Returns `never` when the input is `never`.
+
+@example
+```
+import type {UnionMember} from 'type-fest';
+
+type Last = UnionMember<1 | 2 | 3>;
+//=> 3
```
som-sm: This is not a correct example because there's no guarantee that `UnionMember` will return the last member.  And, our twoslash lint rule will also start complaining if it's not `3`.

Umm...let me think how we can structure the docs properly.

source/union-member.d.ts
```diff
@@ -0,0 +1,39 @@
...
+```
+
+@example
+```
+import type {UnionMember, IsNever} from 'type-fest';
+
+type UnionToTuple<T, L = UnionMember<T>> =
+	IsNever<T> extends false
+		? [...UnionToTuple<Exclude<T, L>>, L]
+		: [];
```
som-sm: Here also, it'd be nice if we could give an example like:
```ts
type T = UnionToTuple<1 | 2 | 3>
//=> [1, 2, 3]
```
But we can't do this because the order is not guaranteed. I guess that's why you chose not to add an example.

source/union-member.d.ts
```diff
@@ -17,14 +17,6 @@ type LastNever = UnionMember<never>;
 //=> never
 ```
 
-@see https://github.com/microsoft/TypeScript/issues/13298#issuecomment-468375328
-
-Use-cases:
-- Implementing recursive type functions that accept a union type.
-- Reducing a union one member at a time, for example when building tuples.
```
som-sm: I didn't mean that remove the "Use-cases" section completely, I meant just keep the first one and remove the second point.

source/union-member.d.ts
som-sm: How about this? Let me know your thoughts.

````ts
/**
Returns an arbitrary member of a union type.

Use-cases:
- Implementing recursive type functions that accept a union type.

@example
```
import type {UnionMember, IsNever} from 'type-fest';

type UnionLength<T, Acc extends any[] = []> =
	UnionMember<T> extends infer Member
		? IsNever<Member> extends false
			? UnionLength<Exclude<T, Member>, [...Acc, Member]>
			: Acc['length']
		: never;

type T1 = UnionLength<'foo' | 'bar' | 'baz'>;
//=> 3

type T2 = UnionLength<{a: string}>;
//=> 1
```

- Picking an arbitrary member from a union

@example
```
import type {UnionMember, Primitive, LiteralToPrimitive} from 'type-fest';

type IsHomogenous<T extends Primitive> = [T] extends [LiteralToPrimitive<UnionMember<T>>] ? true : false;

type T1 = IsHomogenous<1 | 2 | 3 | 4>;
//=> true

type T2 = IsHomogenous<'foo' | 'bar'>;
//=> true

type T3 = IsHomogenous<'foo' | 'bar' | 1>;
//=> false
```

Returns `never` when the input is `never`.

@example
```
import type {UnionMember} from 'type-fest';

type LastNever = UnionMember<never>;
//=> never
```

@category Type
*/
````

taiyakihitotsu:
@som-sm

Thank you!
Picked your comment: https://github.com/sindresorhus/type-fest/pull/1368#discussion_r2873476486

> I guess that's why you chose not to add an example.

Yes.
Now I think `UnionLength` defined in the doc is enough to describe how to write a recursion of union type.

som-sm (approved):

som-sm:
@sindresorhus If this LGTY, this can be merged. Also, #1349.

PR #1374, opened 2026-03-03
Add `Optional` type

### LINKED ISSUE #31: Add Maybe for generic or undefined

sindresorhus:
Fixes #31

### DIFF
--- a/source/optional.d.ts
+++ b/source/optional.d.ts
@@ -21,11 +21,13 @@ type Config = {
 	name: string;
 	description: Optional<string>;
 };
+
+type Description = Config['description'];
+//=> string | undefined
 ```
 
 @category Utilities
 */
-// Uses `Exclude` instead of a conditional type so that `Optional<null>` resolves to `undefined` rather than `never`.
 export type Optional<Value> = Exclude<Value, null> | undefined;
 
 export {};


test-d/optional.ts
```diff
@@ -0,0 +1,46 @@
...
+expectType<undefined>(nullOrUndefined);
+
+// Pure undefined stays undefined
+declare const pureUndefined: Optional<undefined>;
+expectType<undefined>(pureUndefined);
+
+// Nested Optional is idempotent
+expectType<string | undefined>({} as Optional<Optional<string>>);
+
+// Void
```
som-sm:
```suggestion
// Strips `null`
expectType<string | undefined>({} as Optional<string | null>);
expectType<string | undefined>({} as Optional<string | null | undefined>);
expectType<number | boolean | undefined>({} as Optional<number | null | boolean>);
expectType<true | undefined>({} as Optional<true | null>);

// Already `undefined` (idempotent)
expectType<string | undefined>({} as Optional<string | undefined>);

// Pure `null` becomes `undefined` (`null` is stripped, `undefined` remains)
declare const pureNull: Optional<null>;
expectType<undefined>(pureNull);

// `null | undefined` becomes `undefined`
declare const nullOrUndefined: Optional<null | undefined>;
expectType<undefined>(nullOrUndefined);

// Pure `undefined` stays `undefined`
declare const pureUndefined: Optional<undefined>;
expectType<undefined>(pureUndefined);

// Nested optional is idempotent
expectType<string | undefined>({} as Optional<Optional<string>>);

// `void`
```

source/optional.d.ts
```diff
@@ -0,0 +1,31 @@
...
+
+type Config = {
+	name: string;
+	description: Optional<string>;
+};
+```
+
+@category Utilities
+*/
+// Uses `Exclude` instead of a conditional type so that `Optional<null>` resolves to `undefined` rather than `never`.
```
som-sm: Nit: I guess this comment can be removed.

I think it tries to illustrate the problem if the implementation were the following:

```ts
export type Optional<Value> = Value extends null ? never : Value | undefined;

type Test = Optional<never>; //=> never
```

But that would have _not_ been an issue if the `| undefined` were added to the entire conditional rather than the else branch.
```ts
export type Optional<Value> = (Value extends null ? never : Value) | undefined;

type Test = Optional<never>; //=> undefined
```

So, this comment could be a bit confusing and anyways we have a test covering this behavior.

source/optional.d.ts
```diff
@@ -0,0 +1,31 @@
...
+//=> number | undefined
+
+// Strips `null` from the type
+type NullableString = Optional<string | null>;
+//=> string | undefined
+
+type Config = {
+	name: string;
+	description: Optional<string>;
+};
```
som-sm, line 23: I guess here you _didn't_ add a twoslash comment like:

```ts
type Config = {
	name: string;
	description: Optional<string>;
	//=> string | undefined
};
```
because the linter wants the type to be `Optional<string>` instead of `string | undefined`. This would be fixed by #1364.

som-sm: For now, I guess you can add this:

```ts
type Description = Config['description'];
//=> string | undefined
```

som-sm (approved):

test-d/optional.ts
```diff
@@ -0,0 +1,46 @@
...
+
+// `null | undefined` becomes `undefined`
+declare const nullOrUndefined: Optional<null | undefined>;
+expectType<undefined>(nullOrUndefined);
+
+// Pure `undefined` stays `undefined`
+declare const pureUndefined: Optional<undefined>;
+expectType<undefined>(pureUndefined);
+
+// Nested optional is idempotent
```
som-sm: Apologies, I missed this!

```suggestion
// Nested `Optional` is idempotent
```

PR #1375, opened 2026-03-03
`Merge`: Improve documentation

### LINKED ISSUE #101: Documentation improvement request: Merge versus intersection
Recently someone pushed [this commit](https://github.com/sindresorhus/got/commit/d3b972efaca28d8033468c7232a469e8cb034677) to Got with the following message:

> Use `Merge` as it's stricter than the intersection operator

I came here looking for an explanation on what exactly this means, and found none. I noticed that Merge uses the intersection operator under the hood, but I'd like to understand this better. Thanks.

Relates to #8

sindresorhus:
Fixes #101

### DIFF
--- a/source/merge.d.ts
+++ b/source/merge.d.ts
@@ -29,12 +29,12 @@ type Bar = {
 };
 
 // With `&`, `a` becomes `string & number` which is `never`. Not what you want.
-type FooAndBar = Foo & Bar;
-//=> Foo & Bar
+type WithIntersection = (Foo & Bar)['a'];
+//=> never
 
 // With `Merge`, `a` is cleanly overridden to `number`.
-type FooMergedWithBar = Merge<Foo, Bar>;
-//=> {b: number; a: number; c: boolean}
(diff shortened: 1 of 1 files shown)


source/merge.d.ts
```diff
@@ -12,6 +12,31 @@ type SimpleMerge<Destination, Source> = Simplify<{
...
+	c: boolean;
+};
+
+// With `&`, `a` becomes `string & number` which is `never`. Not what you want.
+type FooAndBar = Foo & Bar;
+//=> Foo & Bar
+
+// With `Merge`, `a` is cleanly overridden to `number`.
+type FooMergedWithBar = Merge<Foo, Bar>;
+//=> {b: number; a: number; c: boolean}
```
som-sm: Is this _slightly_ better?

```suggestion
type WithIntersection = (Foo & Bar)['a'];
//=> never

// With `Merge`, `a` is cleanly overridden to `number`.
type WithMerge = Merge<Foo, Bar>['a'];
//=> number
```

PR #1376, opened 2026-03-10
`MergeDeep`: Fix optionality preservation for shared keys

### LINKED ISSUE #941: MergeDeep takes type of property of source instead of using the destination type in case of undefined value
I am trying out remeda and came around the MergeDeep type when using the mergeDeep function from remeda.

My problem is that the MergeDeep type always uses the property type of the source value instead of using the value of the destination incase the source value is undefined.

Does this work as intended or is this a bug?

[stackblitz](https://stackblitz.com/edit/stackblitz-starters-nmz1ug?file=index.ts)

```typescript
import { MergeDeep } from 'type-fest';

type Person = { name: string; age: number };

type PartialPerson = Partial<Person>;

// type of MergeDeepPersonPartialFirst makes sense, since all Person values will be used
type MergeDeepPersonPartialFirst = MergeDeep<PartialPerson, Person>;

// I assumed this would be inferred as the same type as MergeDeepPersonPartialLast
// but it gets infered as
// {
//   name?: string | undefined;
//   age?: number | undefined;
// }
type MergeDeepPersonPartialLast = MergeDeep<Person, PartialPerson>;

const person: Person = { name: 'a', age: 25 };
const partialPerson: PartialPerson = { name: 'b' };

// both values get inferred as
// {
//    name: string;
//    age: number;
// }
// equivalent to Person
const spreadPartialLast = { ...person, ...partialPerson };
const spreadPartiaFirst = { ...partialPerson, ...person };

// on a sideonde
// these get infered as string
type test1 = string & (string | undefined);
type test2 = (string | undefined) & string;

// these get infered as something equivalent to Person
type test3 = Partial<Person> & Person;
type test4 = Person & Partial<Person>;

```

Thanks for the your time

imfozilbek:
Fixes #941.

## Problem

`MergeDeep` incorrectly makes required destination keys optional when source has the same key as optional:

```ts
type Dest = {name: string; age: number};
type Src = {name?: string};

type Result = MergeDeep<Dest, Src>;
// ❌ Before: {name?: string; age: number}  — name became optional
// ✅ After:  {name: string; age: number}   — name stays required
```

## Root cause

In `DoMergeDeepRecord`, Case 3 (`[Key in keyof Source as Key extends keyof Destination ? Key : never]`) iterates over `keyof Source`, which preserves Source's optional modifier (`?`). When Source has `name?`, the result key inherits that optionality regardless of Destination.

## Fix

Split Case 3 into two sub-cases using `{} extends Pick<Type, Key>` to detect optionality:

- **Case 3a**: Key is optional in **both** Source and Destination → result is optional (`?`)
- **Case 3b**: Key is required in **at least one** of Source or Destination → result is required (`-?`)

Both sub-cases wrap values in `Required<>` before passing to `MergeDeepRecordProperty` to strip the optional `undefined` from value types.

## Tests

- Fixed 2 existing tests that encoded the buggy behavior
- Added regression test for #941 (required dest + optional source)
- Added tests for both-optional, source-required + dest-optional
- Added edge case tests: `any`, `never`, deep nesting, empty objects, same-type

All tests pass (`npm test`).

### DIFF
--- a/source/merge-deep.d.ts
+++ b/source/merge-deep.d.ts
@@ -57,8 +57,26 @@ type DoMergeDeepRecord<
 		[Key in keyof Source as Key extends keyof Destination ? never : Key]: Source[Key];
 	}
 // Case in rule 3: Both the source and the destination contain the key.
+// Split into two sub-cases to correctly handle optionality:
+// 3a: Key is optional in both Source and Destination → result is optional.
 	& {
-		[Key in keyof Source as Key extends keyof Destination ? Key : never]: MergeDeepRecordProperty<Required<Destination>[Key], Required<Source>[Key], Options>;
+		[Key in keyof Source as Key extends keyof Destination
+			? {} extends Pick<Source, Key>
+				? {} extends Pick<Destination, Key>
+					? Key
+					: never
+				: never
+			: never]?: MergeDeepRecordProperty<Required<Destination>[Key], Required<Source>[Key], Options>;
+	}
+// 3b: Key is required in at least one of Source or Destination → result is required.
+	& {
+		[Key in keyof Source as Key extends keyof Destination
+			? {} extends Pick<Source, Key>
+				? {} extends Pick<Destination, Key>
+					? never
+					: Key
+				: Key
+			: never]-?: MergeDeepRecordProperty<Required<Destination>[Key], Required<Source>[Key], Options>;
 	};
 
 /**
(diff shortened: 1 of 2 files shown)


som-sm:
> `MergeDeep` incorrectly makes required destination keys optional when source has the same key as optional:

This is intentional, `MergeDeep` is _only_ meant to be used at the type-level. To type runtime values use the newly added [`ObjectMerge` type](https://github.com/sindresorhus/type-fest/pull/1324).

https://github.com/sindresorhus/type-fest/blob/d3302ce03a9d42edcb79332421dad942a95c67c9/source/merge.d.ts#L69

```ts

import type {ObjectMerge} from 'type-fest';

type Dest = {name: string; age: number};
type Src = {name?: string};

type Result = ObjectMerge<Dest, Src>;
//=> { age: number; name: string; }
```
Playground: https://tsplay.dev/N5raVm

PR #1380, opened 2026-03-14
Add `SomeExtend` type

### LINKED ISSUE #1379: AtLeastOneExtend
related PR : https://github.com/sindresorhus/type-fest/pull/1378

we can just modify
https://github.com/sindresorhus/type-fest/blob/main/source/all-extend.d.ts#L114

```typescript
First extends Type
				? _AllExtend<Rest, Type, Options>
				: false
```

=>

```typescript
First extends Type
          ? true
          : _AllExtend<Rest, Type, Options>
```

but most likely a base type is needed, where will the option be `mode = 'all' | 'atLeaseOne'` which will already be used by these types.

unit tests example:

```typescript
// Basic matches
expectType<AtLeastOneExtend<[1, 2, 3], number>>(true);
expectType<AtLeastOneExtend<[1, '2', 3], string>>(true);
expectType<AtLeastOneExtend<[1, 2, 3], string>>(false);
expectType<AtLeastOneExtend<[1, 2, 3], number>>(true);
expectType<AtLeastOneExtend<[1, 2, '3'], number>>(true);
expectType<AtLeastOneExtend<[1, 2, '3'], string>>(true);
expectType<AtLeastOneExtend<['1', '2', '3'], number>>(false);
expectType<AtLeastOneExtend<['a', 'b', 'c'], string>>(true);

// Optional Elements
expectType<AtLeastOneExtend<[1?, 2?], number>>(true);
expectType<AtLeastOneExtend<[number?, string?], string>>(true);
expectType<AtLeastOneExtend<[number?], string>>(false);

/ Edge case: Checking against any[] specifically
expectType<AtLeastOneExtend<[Array<number>, string], Array<any>>>(true);
expectType<AtLeastOneExtend<[number, string], Array<any>>>(false);

/* Edge case: Checking against never[] specifically
   Only never[] (or any) will extend never[] */
expectType<AtLeastOneExtend<[Array<never>, number], Array<never>>>(true);
expectType<AtLeastOneExtend<[Array<number>, string], Array<never>>>(false);

// Basic string array check
expectType<AtLeastOneExtend<[Array<string>, number], Array<string>>>(true);

// Multiple matches
expectType<
  AtLeastOneExtend<[Array<string>, Array<string>, boolean], Array<string>>
>(true);

// Partial union match (Union containing string[])
expectType<AtLeastOneExtend<[number, Array<string> | number], Array<string>>>(
  true,
);

// Negative case (Array of different type)
expectType<AtLeastOneExtend<[Array<number>, boolean], Array<string>>>(false);

// Negative case (Individual string does not extend string[])
expectType<AtLeastOneExtend<['hello', number], Array<string>>>(false);

// Readonly string array check
expectType<AtLeastOneExtend<[ReadonlyArray<string>, number], Array<string>>>(
  true,
);

// Tuple that extends string[]
expectType<AtLeastOneExtend<[['a', 'b'], number], Array<string>>>(true);

// Intersections
expectType<AtLeastOneExtend<[{ a: 1 } & { b: 2 }, { c: 3 }], { a: 1 }>>(true);
expectType<AtLeastOneExtend<[{ a: 1 }, { b: 2 }], { a: 1 } & { b: 2 }>>(false);
```

### Type source

_No response_

som-sm:
Adds `SomeExtend` type, similar to the `AllExtend` type.

Closes #1379.

### DIFF
--- a/source/some-extend.d.ts
+++ b/source/some-extend.d.ts
@@ -34,10 +34,10 @@ export type SomeExtendOptions = {
 	type D = SomeExtend<[1, never], never, {strictNever: false}>;
 	//=> true
 
-	type E = SomeExtend<['a', 'b', never], any, {strictNever: true}>;
+	type E = SomeExtend<[never], any, {strictNever: true}>;
 	//=> true
 
-	type F = SomeExtend<['a', 'b', never], any, {strictNever: false}>;
+	type F = SomeExtend<[never], any, {strictNever: false}>;
 	//=> true
 	```
 	*/


source/all-extend.d.ts
som-sm, line 1: For `SomeExtend`, we can't directly return `true` if the target is `any` because `SomeExtend<[], any>` should return `false`. So, adjusted `AllExtend`'s implementation as well so that it stays aligned with the implementation of `SomeExtend`.

test-d/all-extend.ts
```diff
@@ -40,7 +40,7 @@ expectType<AllExtend<readonly [...ReadonlyArray<string | undefined>, string, str
 
 // Optional elements
 // If `exactOptionalPropertyTypes` were disabled, the target type would need an additional `| undefined` for a successful match.
-// For example, `AllExtend<[1?, 2?], number>` would return `false`. To make it return `true`, the target type must be `number | undefined`.
+// For example, `AllExtend<[1?, 2?], number>` would return `boolean`. To make it return `true`, the target type must be `number | undefined`.
```
som-sm, line 43: Fixed typo.

sindresorhus:
Looks great 👍

PR #1383, opened 2026-03-18
Add `AndAll` type

som-sm:
Adds `AndAll` type.

### DIFF
--- a/source/and-all.d.ts
+++ b/source/and-all.d.ts
@@ -66,7 +66,7 @@ type B = AndAll<[true, any]>;
 //=> boolean
 ```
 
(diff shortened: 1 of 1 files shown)


test-d/and-all.ts
```diff
@@ -0,0 +1,63 @@
...
+// Readonly arrays
+expectType<AndAll<readonly [true, true, true]>>(true);
+expectType<AndAll<readonly [true, true, false]>>(false);
+expectType<AndAll<readonly [true, true, boolean]>>(boolean);
+expectType<AndAll<ReadonlyArray<true>>>(true);
+expectType<AndAll<ReadonlyArray<false>>>(false);
+expectType<AndAll<readonly boolean[]>>(boolean);
+
+// Boundary cases
+expectType<AndAll<[]>>(true);
```
som-sm, line 49: `AndAll`, when instantiated with an empty array (`[]`), returns `true`, unlike `OrAll`, which returns `false`. This behavior is intentional/correct; see:
- https://en.wikipedia.org/wiki/Logical_conjunction#:~:text=the%20empty%20conjunction%20(AND%2Ding%20over%20an%20empty%20set%20of%20operands)%20is%20often%20defined%20as%20having%20the%20result%20true.
- https://en.wikipedia.org/wiki/Clause_%28logic%29?utm_source=chatgpt.com#:~:text=The%20truth%20evaluation,vacuous%20truth.

This is the same reason why `Array.prototype.every` returns `true` for empty arrays, while `Array.prototype.some` returns `false`.

PR #1390, opened 2026-03-24
Disable `no-unnecessary-type-arguments` lint rule

som-sm:
Because of [certain recent changes](https://github.com/typescript-eslint/typescript-eslint/pull/11918) in the `@typescript-eslint/no-unnecessary-type-arguments` rule, our test files now have multiple errors (see [run](https://github.com/sindresorhus/type-fest/actions/runs/22629084671/job/68310393578?pr=1373)).

image<br>

This PR disables this rule. 

- We don’t want this rule in our test files, as we need explicit type arguments there.
- It also feels unnecessary in source files. And, there appears to be some bug as well, for example:

  ```ts
  type Test<T, U = unknown> = [T, U];
  type T1<T extends unknown[]> = Test<string, T[number]>;
  //                                          ~~~~~~~~~
  // This is the default value for this type parameter, so it can be omitted. (@typescript-eslint/no-unnecessary-type-arguments)
  ```
  In this case, `Test<string>` is not equivalent to `Test<string, T[number]>`, but the rule reports an error, which it shouldn't.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,5 +41,5 @@ jobs:
         with:
           node-version: 24
       - run: npm install
-      - run: npm install typescript@${{ matrix.typescript-version }}
+      - run: npm install typescript@${{ matrix.typescript-version }} --legacy-peer-deps
       - run: NODE_OPTIONS="--max-old-space-size=6144" npx tsc
--- a/.github/workflows/ts-canary.yml
+++ b/.github/workflows/ts-canary.yml
@@ -20,7 +20,7 @@ jobs:
         with:
           node-version: lts/*
       - run: npm install
-      - run: npm install typescript@${{ matrix.typescript-version }}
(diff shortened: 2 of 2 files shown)


som-sm:
TS `v6` also [got released yesterday](https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/), because of which `npm install typescript@latest` now has peer dependency issues (see [run](https://github.com/sindresorhus/type-fest/actions/runs/23479120768/job/68318313250?pr=1390)):
```
npm error Could not resolve dependency:
npm error peer typescript@">=4.8.4 <6.0.0" from @typescript-eslint/parser@8.57.2
npm error node_modules/@typescript-eslint/parser
```

Should we temporarily add the `--legacy-peer-deps` flag while installing (refer [2969233](https://github.com/sindresorhus/type-fest/pull/1390/commits/2969233ef104a8424e26586eed6105a84b41c388))?

PR #1393, opened 2026-03-29
Upgrade to XO v2

som-sm:
Update to XO v2 & ESLint v10.

Majority of the changes in this PR are styling-related. The ones which are not are also pretty simple, like:
- Replacing `.sort` with `.toSorted`.
- Replacing `BigInt` constructor calls to bigint literals.
- Adding `v` flag to regular expression.

### DIFF
--- a/xo.config.js
+++ b/xo.config.js
@@ -26,9 +26,12 @@ const xoConfig = [
 			'@typescript-eslint/consistent-indexed-object-style': 'off',
 			'@typescript-eslint/unified-signatures': 'off', // Temp
 			'@typescript-eslint/no-unnecessary-type-arguments': 'off',
+			'@typescript-eslint/no-unsafe-type-assertion': 'off',
 			'@stylistic/quote-props': 'off',
 			'@stylistic/function-paren-newline': 'off',
 			'@stylistic/object-curly-newline': 'off',
+			'@stylistic/curly-newline': 'off',
+			'@stylistic/operator-linebreak': ['error', 'before', {overrides: {'=': 'after'}}],
 			'n/file-extension-in-import': 'off',
 			'object-curly-newline': [
 				'error',
@@ -44,7 +47,7 @@ const xoConfig = [
 		},
 	},
 	{
-		files: 'source/**/*.d.ts',
+		files: ['source/**/*.d.ts', 'index.d.ts'],
 		rules: {
 			'no-restricted-imports': [
 				'error',
@@ -52,6 +55,14 @@ const xoConfig = [
 					paths: ['tsd', 'expect-type'],
 				},
 			],
+			'unicorn/require-module-specifiers': 'off',
+		},
+	},
+	{
+		files: 'test-d/**/*.ts',
+		rules: {
+			'unicorn/no-immediate-mutation': 'off',
+			'require-unicode-regexp': 'off',
 		},
 	},
 	{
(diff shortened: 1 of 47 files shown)


xo.config.js
```diff
@@ -26,17 +26,23 @@ const xoConfig = [
 			'@typescript-eslint/consistent-indexed-object-style': 'off',
 			'@typescript-eslint/unified-signatures': 'off', // Temp
 			'@typescript-eslint/no-unnecessary-type-arguments': 'off',
+			'@typescript-eslint/no-unsafe-type-assertion': 'off',
 			'@stylistic/quote-props': 'off',
 			'@stylistic/function-paren-newline': 'off',
 			'@stylistic/object-curly-newline': 'off',
+			'@stylistic/curly-newline': 'off',
+			'@stylistic/operator-linebreak': ['error', 'before', {overrides: {'=': 'after'}}],
```
som-sm, line 34: `'@stylistic/operator-linebreak'` is a useful rule, so instead of turning it off, fixed all the errors reported by this, like:

Image 

Image <br>

`{overrides: {'=': 'after'}}` is because I feel like it's better to have the line break after `=` rather than before it.

Image 

Image

xo.config.js
```diff
@@ -26,17 +26,23 @@ const xoConfig = [
 			'@typescript-eslint/consistent-indexed-object-style': 'off',
 			'@typescript-eslint/unified-signatures': 'off', // Temp
 			'@typescript-eslint/no-unnecessary-type-arguments': 'off',
+			'@typescript-eslint/no-unsafe-type-assertion': 'off',
```
som-sm, line 29: Can't have this because we have assertions like `{} as ...` everywhere in tests. And in source files, there's no scope for assertions.

xo.config.js
```diff
@@ -26,17 +26,23 @@ const xoConfig = [
...
 			'@stylistic/quote-props': 'off',
 			'@stylistic/function-paren-newline': 'off',
 			'@stylistic/object-curly-newline': 'off',
+			'@stylistic/curly-newline': 'off',
```
som-sm, line 33: Turned this off because it seems to have conflicts with other rules. And also there's no good default for this rule for our codebase.

xo.config.js
```diff
@@ -26,17 +26,23 @@ const xoConfig = [
...
+			'@stylistic/operator-linebreak': ['error', 'before', {overrides: {'=': 'after'}}],
 			'n/file-extension-in-import': 'off',
+			'unicorn/require-module-specifiers': 'off',
```
som-sm: Turned this off because source files have `export {}`.

xo.config.js
```diff
@@ -44,14 +47,22 @@ const xoConfig = [
...
-		files: 'source/**/*.d.ts',
+		files: ['source/**/*.d.ts', 'index.d.ts'],
 		rules: {
 			'no-restricted-imports': [
 				'error',
 				{
 					paths: ['tsd', 'expect-type'],
 				},
 			],
+			'unicorn/require-module-specifiers': 'off',
```
som-sm, line 58: Can't have this in `.d.ts` files because there we have `export {}`.

xo.config.js
```diff
@@ -110,6 +121,7 @@ const xoConfig = [
 			'@stylistic/eol-last': 'off',
 			'capitalized-comments': 'off',
 			'unicorn/prefer-structured-clone': 'off',
+			'unicorn/no-immediate-mutation': 'off',
 		},
```
som-sm, line 125: Can't have `unicorn/no-immediate-mutation` enabled in tests and JSDoc codeblocks because we need this at times.

PR #1394, opened 2026-03-29
Align behavior of [Camel|Pascal|Kebab|Snake]-Cased types with lodash

### LINKED ISSUE #1229: `KebabCasedProperties`: Align behavior with Lodash
See: https://github.com/sindresorhus/type-fest/issues/1074#issuecomment-2698729943

jericirenej:
- `WordsOptions`: `splitOnPunctuation` option added. Defaults to `false` for now, so no breaking changes are introduced. Updated JSDoc for the `Words` type.
- `CamelCase, PascalCase, KebabCase, SnakeCase`: 
  - added tests for the new punctuation option. 
  - Updated JSDocs to indicate how the new `splitOnPunctuation` option can be used. 
  - `CamelCase` uses `_DefaultWordsOptions` instead of specifying defaults directly (this would require updates every time a new default option would be added).
- Updated JSDocs follow the `type NamedType1 = NamedType<Value>; //=> result`  pattern.
 
 
Closes #1229

### DIFF
--- a/source/kebab-case.d.ts
+++ b/source/kebab-case.d.ts
@@ -12,12 +12,19 @@ This can be useful when, for example, converting a camel-cased object property t
 import type {KebabCase} from 'type-fest';
 
 // Simple
+type KebabCase1 = KebabCase<'fooBar'>;
+//=> 'foo-bar'
 
-const someVariable: KebabCase<'fooBar'> = 'foo-bar';
-const someVariableNoSplitOnNumbers: KebabCase<'p2pNetwork', {splitOnNumbers: false}> = 'p2p-network';
+type KebabCase2 = KebabCase<'p2pNetwork', {splitOnNumbers: false}>;
+//=> 'p2p-network'
+
+type KebabCase3 = KebabCase<'div.card::after', {splitOnPunctuation: true}>;
+//=> 'div-card-after'
+
+type KebabCase4 = KebabCase<'foo-bar::01', {splitOnPunctuation: true}>
+//=> 'foo-bar-01'
 
 // Advanced
-
 type KebabCasedProperties<T> = {
 	[K in keyof T as KebabCase<K>]: T[K]
 };
--- a/source/kebab-case.d.ts
+++ b/source/kebab-case.d.ts
@@ -12,14 +12,10 @@ This can be useful when, for example, converting a camel-cased object property t
 import type {KebabCase} from 'type-fest';
 
 // Simple
-type KebabCase1 = KebabCase<'fooBar'>;
-//=> 'foo-bar'
 
-type KebabCase2 = KebabCase<'p2pNetwork', {splitOnNumbers: false}>;
-//=> 'p2p-network'
-
-type KebabCase3 = KebabCase<'div.card::after', {splitOnPunctuation: true}>;
-//=> 'div-card-after'
+const someVariable: KebabCase<'fooBar'> = 'foo-bar';
+const someVariableNoSplitOnNumbers: KebabCase<'p2pNetwork', {splitOnNumbers: false}> = 'p2p-network';
+const someVariableWithPunctuation: KebabCase<'div.card::after', {splitOnPunctuation: true}> = 'div-card-after';
 
 // Advanced
 type KebabCasedProperties<T> = {
--- a/source/kebab-case.d.ts
+++ b/source/kebab-case.d.ts
@@ -21,9 +21,6 @@ type KebabCase2 = KebabCase<'p2pNetwork', {splitOnNumbers: false}>;
 type KebabCase3 = KebabCase<'div.card::after', {splitOnPunctuation: true}>;
 //=> 'div-card-after'
 
-type KebabCase4 = KebabCase<'foo-bar::01', {splitOnPunctuation: true}>;
-//=> 'foo-bar-01'
-
 // Advanced
 type KebabCasedProperties<T> = {
 	[K in keyof T as KebabCase<K>]: T[K]
--- a/source/kebab-case.d.ts
+++ b/source/kebab-case.d.ts
@@ -18,6 +18,7 @@ const someVariableNoSplitOnNumbers: KebabCase<'p2pNetwork', {splitOnNumbers: fal
 const someVariableWithPunctuation: KebabCase<'div.card::after', {splitOnPunctuation: true}> = 'div-card-after';
 
 // Advanced
+
 type KebabCasedProperties<T> = {
 	[K in keyof T as KebabCase<K>]: T[K]
 };
--- a/source/kebab-case.d.ts
+++ b/source/kebab-case.d.ts
@@ -21,7 +21,7 @@ type KebabCase2 = KebabCase<'p2pNetwork', {splitOnNumbers: false}>;
 type KebabCase3 = KebabCase<'div.card::after', {splitOnPunctuation: true}>;
 //=> 'div-card-after'
 
-type KebabCase4 = KebabCase<'foo-bar::01', {splitOnPunctuation: true}>
+type KebabCase4 = KebabCase<'foo-bar::01', {splitOnPunctuation: true}>;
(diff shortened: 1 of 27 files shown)


test-d/words.ts
```diff
@@ -93,3 +93,18 @@ expectType<Words<'01item01'>>(['01', 'item', '01']);
...
+expectType<Words<'hello:world'>>(['hello', ':world']);
+expectType<Words<'hello::world'>>(['hello', ':', ':world']);
+expectType<Words<'hello-braveNew:world'>>(['hello', 'brave', 'New', ':world']);
+
+expectType<Words<':', {splitOnPunctuation: true}>>([]);
+expectType<Words<'::', {splitOnPunctuation: true}>>([]);
+expectType<Words<'hello::', {splitOnPunctuation: true}>>(['hello']);
+expectType<Words<'hello:world', {splitOnPunctuation: true}>>(['hello', 'world']);
+expectType<Words<'hello::world', {splitOnPunctuation: true}>>(['hello', 'world']);
+expectType<Words<'hello-braveNew:world', {splitOnPunctuation: true}>>(['hello', 'brave', 'New', 'world']);
```
som-sm, line 110: Have you cross-checked all these cases with Lodash? If not, can you please do that?

test-d/words.ts
som-sm, line 1: Also add tests with both `splitOnNumbers` and `splitOnPunctuation`.

test-d/snake-case.ts
```diff
@@ -82,20 +82,23 @@ expectType<'fo_o2bar'>(snakeFromNumberInTheMiddleNoSplitOnNumbersEdgeCase2);
...
-expectType<'card::after'>(withPunctuation3);
-
-declare const withPunctuation4: SnakeCase<'div.card::after'>;
-expectType<'div.card::after'>(withPunctuation4);
-
-declare const withPunctuationAndNumber: SnakeCase<'foo-bar::01'>;
-expectType<'foo_bar::01'>(withPunctuationAndNumber);
-
-declare const withPunctuationAndNumber2: SnakeCase<'foo-bar::01', {splitOnNumbers: true}>;
-expectType<'foo_bar::_01'>(withPunctuationAndNumber2);
```
som-sm, line 101: Please don't update existing tests in the PR. Add tests using the existing format in that particular file.

som-sm: Same for all other test files.

som-sm (changes requested):
Please don't change the way docs and tests are written as part of this PR. Keep this PR focused on `splitOnPunctuation` changes. If you want to streamline the docs and tests, consider opening a separate PR for that.

som-sm:
Also, please add tests for `*CasedProperties` and `*CasedPropertiesDeep`.

test-d/words.ts
jericirenej:
```ts
import words from "lodash/fp/words";

[
  ":",
  "::",
  "hello::",
  "hello:world",
  "hello::world",
  "hello-braveNew:world",
].forEach((word) => {
  console.log("Original:", word, "\nWords:", words(word), "\n\n");
});
```
Returns
```
Original: : 
Words: [] 


Original: :: 
Words: [] 


Original: hello:: 
Words: [ "hello" ] 


Original: hello:world 
Words: [ "hello", "world" ] 


Original: hello::world 
Words: [ "hello", "world" ] 


Original: hello-braveNew:world 
Words: [ "hello", "brave", "New", "world" ] 
```

test-d/snake-case.ts
jericirenej: All right, I'll do that. 
The reason to why this was done in tests was because:
- although declarations had the benefit of being a condensed "describe" placeholder, they increased the test suite length, so I removed them for that specific testing part, as this change did not affect the test scope. Also,
-  I did not see an explicit policy that would forbid this kind of optimization.

jericirenej:
> Please don't change the way docs and tests are written as part of this PR. Keep this PR focused on `splitOnPunctuation` changes. If you want to streamline the docs and tests, consider opening a separate PR for that.

Ok, I'll change it so that it's in the similar pattern as the originals. The changes were very minor and aimed to increase readability of the JSDoc. I understand that here they are seen as problematic, because they are outside the ticket scope

som-sm:
> The changes were very minor and aimed to increase readability of the JSDoc.

Yeah, the changes themselves were fine, just not in the scope of this PR.

test-d/words.ts
jericirenej: Added.

test-d/snake-case.ts
jericirenej: I adjusted tests so they follow the old form patterns. I needed to change the names of one or two old existing constants to keep everything consistent within the same test file. 

I also added a couple of "failsafe" tests where `splitOnNumbers` is set to true and false explicitly within the punctuation tests, even though its setting should not play any role when the `splitOnPunctuation` is set to true.

jericirenej:
> Also, please add tests for `*CasedProperties` and `*CasedPropertiesDeep`.

Added.

I also updated the JSDocs for the x-cased-properties and x-cased-properties-deep types

source/pascal-case.d.ts
```diff
@@ -48,4 +49,5 @@ type _PascalCase<Value, Options extends Required<CamelCaseOptions>> = CamelCase<
 	? Capitalize<CamelCase<Value, Options>>
 	: CamelCase<Value, Options>;
 
-export {};
+export { };
+
```
som-sm: Unnecessary change.

source/words.d.ts
```diff
@@ -72,21 +90,35 @@ type Words4 = Words<'lifeIs42'>;
...
 */
-export type Words<Sentence extends string, Options extends WordsOptions = {}> =
-	WordsImplementation<Sentence, ApplyDefaultOptions<WordsOptions, _DefaultWordsOptions, Options>>;
+export type Words<
+	Sentence extends string,
+	Options extends WordsOptions = {},
+> = WordsImplementation<
+	Sentence,
+	ApplyDefaultOptions<WordsOptions, _DefaultWordsOptions, Options>
+>;
```
som-sm: Unnecessary change.

test-d/internal/has-multiple-call-signatures.ts
```diff
@@ -8,7 +8,7 @@ type Overloaded = {
 
 type Overloaded2 = {
 	(foo: number | undefined): string;
-	// eslint-disable-next-line @typescript-eslint/unified-signatures
+
```
som-sm: Unnecessary change. 

I realise this is an unused disable comment and gets removed if you run `--fix`. This will get fixed with #1393, so it won’t be an issue going forward.

source/kebab-case.d.ts
```diff
@@ -15,9 +15,9 @@ import type {KebabCase} from 'type-fest';
 
 const someVariable: KebabCase<'fooBar'> = 'foo-bar';
 const someVariableNoSplitOnNumbers: KebabCase<'p2pNetwork', {splitOnNumbers: false}> = 'p2p-network';
+const someVariableWithPunctuation: KebabCase<'div.card::after', {splitOnPunctuation: true}> = 'div-card-after';
 
 // Advanced
-
```
som-sm, line 20: Unnecessary change.

som-sm (commented):
PR LGTM, just a few minor comments. Nice work, thanks for picking this up!

test-d/snake-case.ts
som-sm: Feel free to address any formatting/structuring improvements or JSDoc improvements in a follow-up PR. Always welcome!

test-d/internal/has-multiple-call-signatures.ts
jericirenej: I probably accidentally included this while running an `npx xo --fix` script without a specified file list. I'll revert this.

PR #1396, opened 2026-03-31
Add lint rule to validate type descriptions in README

som-sm:
This PR adds a lint rule that enforces that type descriptions in the README exactly match the first line of their source JSDoc.

https://github.com/user-attachments/assets/2b6918c4-0c1f-42e1-9355-9bf322425a0f

<br>

The lint caught around 60 mismatches in the existing README file.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,13 +59,13 @@
 		"@typescript-eslint/parser": "^8.44.0",
 		"@typescript/vfs": "^1.6.1",
 		"dedent": "^1.7.0",
-		"eslint": "^9.35.0",
+		"eslint": "^10.1.0",
 		"expect-type": "^1.2.2",
 		"npm-run-all2": "^8.0.4",
 		"tsd": "^0.33.0",
 		"typescript": "^5.9.2",
 		"typescript-eslint": "^8.47.0",
-		"xo": "^1.2.2"
+		"xo": "^2.0.2"
 	},
 	"tsd": {
 		"compilerOptions": {
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
 	"scripts": {
 		"test:tsc": "node --max-old-space-size=6144 ./node_modules/.bin/tsc",
 		"test:tsd": "node --max-old-space-size=6144 ./node_modules/.bin/tsd",
-		"test:xo": "node --max-old-space-size=6144 ./node_modules/.bin/xo --ignores=lint-processors/fixtures/**/*.d.ts '**/*.{js,ts}' readme.md",
+		"test:xo": "node --max-old-space-size=6144 ./node_modules/.bin/xo --ignores=lint-processors/fixtures/**/*.d.ts '**/*.{js,ts,md}'",
 		"test:linter": "node --test",
 		"test": "run-p test:*"
 	},
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
 		"tagged-tag": "^1.0.0"
 	},
 	"devDependencies": {
-		"@eslint/markdown": "^7.5.1",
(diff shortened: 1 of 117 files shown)


som-sm:
`xo.config.js` uses the `FlatXoConfig` type from `xo`, and `FlatXoConfig` itself relies on `eslint` types. However, those `eslint` types are being incorrectly resolved from `@types/eslint`, which is causing errors in the config.

I’ve opened a discussion in the ESLint repository about this: https://github.com/eslint/eslint/discussions/20704.

source/literal-to-primitive.d.ts
```diff
@@ -1,5 +1,5 @@
 /**
-Given a [literal type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types) return the {@link Primitive | primitive type} it belongs to, or `never` if it's not a primitive.
+Given a [literal type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types) return the [primitive type](https://developer.mozilla.org/en-US/docs/Glossary/Primitive) it belongs to, or `never` if it's not a primitive.
```
som-sm, line 2: Updated the link to an external link, because otherwise the README description would need manual adjusting because it'd be really complicated to automatically figure out where a particular symbol lives.

If a symbol link is absolutely required, then we'd have to disable the lint-rule for that specific list item and adjust the description manually, as mentioned here:
https://github.com/sindresorhus/type-fest/blob/8fb539c806fe80f5e7769747db0d48b074eb03b2/lint-rules/readme-jsdoc-sync.js#L84-L89

xo.config.js
```diff
@@ -1,15 +1,18 @@
...
 import {importPathRule} from './lint-rules/import-path.js';
 import {sourceFilesExtensionRule} from './lint-rules/source-files-extension.js';
 import {requireExportedTypesRule} from './lint-rules/require-exported-types.js';
 import {requireExportRule} from './lint-rules/require-export.js';
 import {validateJSDocCodeblocksRule} from './lint-rules/validate-jsdoc-codeblocks.js';
+import {readmeJSDocSyncRule} from './lint-rules/readme-jsdoc-sync.js';
 import {jsdocCodeblocksProcessor} from './lint-processors/jsdoc-codeblocks.js';
 
-/** @type {import('xo').FlatXoConfig} */
+/** @type {Array<import('eslint').Linter.Config>} */
```
som-sm, line 12: Had to switch to `eslint` type because of [this issue](https://github.com/sindresorhus/type-fest/pull/1396#issuecomment-4162426104).

package.json
```diff
@@ -27,7 +27,7 @@
 	"scripts": {
 		"test:tsc": "node --max-old-space-size=6144 ./node_modules/.bin/tsc",
 		"test:tsd": "node --max-old-space-size=6144 ./node_modules/.bin/tsd",
-		"test:xo": "node --max-old-space-size=6144 ./node_modules/.bin/xo --ignores=lint-processors/fixtures/**/*.d.ts",
+		"test:xo": "node --max-old-space-size=6144 ./node_modules/.bin/xo --ignores=lint-processors/fixtures/**/*.d.ts '**/*.{js,ts}' readme.md",
```
som-sm: This is because XO doesn't seem to lint `.md` files even if there's an entry for `.md` files in the config.

som-sm:
I initially expected only minor discrepancies between the README and JSDoc descriptions, such as typos. However, there were several instances where the two were entirely different. So, I had to review them manually and select the better version. And at certain places, I had to make changes beyond just picking between the two.

As a result, for some types, the JSDoc description has changed, while for some others, the README description has changed. And for some, both have changed.

PR #1399, opened 2026-04-08
Add `Overloads`, `OverloadParameters`, `OverloadReturnType`

### LINKED ISSUE #585: `OverloadParameters` and `OverloadReturnType`

### LINKED ISSUE #868: A type util to produce an alternation of function overloads.

ikeyan:
Fixes #868
Fixes #585
Closes #1264

## Summary

Adds three types for working with overloaded function signatures:

- `Overloads<F>` — extracts all overloads as a tuple, preserving declaration order
- `OverloadParameters<F>` — overload-aware `Parameters` (returns a union of all parameter tuples)
- `OverloadReturnType<F>` — overload-aware `ReturnType` (returns a union of all return types)

TypeScript's built-in `Parameters` and `ReturnType` only work with the last overload; these types extract all of them. To get a union of overloads instead of a tuple, use `Overloads<F>[number]`.

## Improvements over #1264

- Renamed from `FunctionOverloads` to `Overloads` and changed the return type from a union to a tuple — overloads are ordered, so a tuple is the natural representation. Analogous to `Parameters` / `ReturnType` in naming style
- Added `OverloadParameters` and `OverloadReturnType` as direct replacements for `Parameters` and `ReturnType` on overloaded functions
- `HasExplicitThis` correctly distinguishes implicit `this` from explicit `this: unknown` by exploiting TypeScript's overload deduplication rules
- Fixed a bug in #1264 where encountering an overload with the same parameter list as a previous one would terminate iteration early, missing remaining overloads
- `CollectOverloads` uses an effect-observation loop with N-advancement to extract overloads preceding generic overloads, which the previous convergence-based approach could not reach
- Internal documentation explains TypeScript's overload enumeration behavior (deduplication rules, implicit `this` semantics)

## Known limitations

- Generic overloads:
  - Type parameters are lost -- they are replaced by their upper bound (e.g. `<T>` becomes `unknown`, `<T extends string>` becomes `string`)
  - When there are 4 or more generic overloads, extraction stops at the 4th-from-last generic overload and any overloads before it are omitted
  - Implicit `this` (no annotation) is indistinguishable from explicit `this: unknown`, so the output always includes `this: unknown`
- TypeScript deduplicates overloads that it considers identical. When both overloads have explicit `this`, all three of `(This, Parameters, Return)` must match to be deduplicated. However, when one or both have implicit `this` (no annotation), only `(Parameters, Return)` is compared — meaning an implicit-`this` overload and an explicit-`this` overload with the same params/return are treated as duplicates, and whichever appears first suppresses the other


sindresorhus (approved):

ikeyan:
## Naming suggestion

~I'd like to propose renaming `FunctionOverloads` before merging. Two concerns with the current name:~

1. ~The `Function` prefix is redundant — "overloads" already implies functions~
2. ~The name doesn't convey that the result is a **union** of the overload signatures~

~Suggested alternatives:~

- ~**`OverloadsToUnion`** — follows the existing `UnionToIntersection`, `UnionToTuple`, `TupleToUnion` naming pattern~
- ~**`OverloadsUnion`** — shorter, similar to `LiteralUnion`, `TaggedUnion`~

~I'd lean toward `OverloadsToUnion` for consistency with the `XToY` conversion types already in the library. What do you think @sindresorhus?~

source/function-overloads.d.ts
```diff
@@ -0,0 +1,117 @@
...
+type RequestReturnType = ReturnType<RequestOverloads>;
+//=> Promise<string> | Promise<unknown>
+```
+
+@see https://github.com/microsoft/TypeScript/issues/14107
+@see https://github.com/microsoft/TypeScript/issues/32164
+
+@category Function
+*/
+export type FunctionOverloads<FunctionType> = FunctionType extends unknown
```
som-sm: I think the type argument should have a constraint:

```suggestion
export type FunctionOverloads<FunctionType extends (...args: any) => any> = FunctionType extends unknown
```

source/function-overloads.d.ts
```diff
@@ -0,0 +1,117 @@
...
+	? IsAny<FunctionType> extends true
+		? (...arguments_: readonly unknown[]) => unknown
```
som-sm: This could _just_ be `any`. If input is `any`, we could _simply_ return back `any`.

source/overloads.d.ts
som-sm, line 1: I think the implementation can be simplified if we extract out the logic for distinguishing between explicit and implicit `this` parameter using something like the following:

```ts
declare const unique: unique symbol;

export type HasExplicitThis<T extends (...args: any[]) => any> =
	IsUnknown<ThisParameterType<T>> extends true
		? ThisParameterType<((this: typeof unique, ...args: Parameters<T>) => ReturnType<T>) & T> extends typeof unique
			? false
			: true
		: true;

type T1 = HasExplicitThis<(this: unknown, x: string) => void>; //=> true
type T2 = HasExplicitThis<(x: string) => void>; //=> false
type T3 = HasExplicitThis<(this: any, x: string) => void>; //=> true
```

som-sm (changes requested):

source/overloads.d.ts
ikeyan: Implemented `HasExplicitThis` in source/internal/.

source/function-overloads.d.ts
ikeyan: done

source/function-overloads.d.ts
ikeyan: done

sindresorhus:
`Overloads` does not just lose the generic parameter, it can drop earlier non-generic overloads entirely. A minimal repro:

```ts
declare function genericLast(input: string): 1;
declare function genericLast<Type>(input: Type): Type;

type Result = Overloads<typeof genericLast>;
// actual:   [(this: unknown, input: unknown) => unknown]
// expected: [(input: string) => 1, (input: unknown) => unknown]
```

Same thing happens when the generic overload is in the middle of the list:

```ts
declare function genericMiddle(input: string): 1;
declare function genericMiddle<Type>(input: Type): Type;
declare function genericMiddle(input: number): 2;

type Result = Overloads<typeof genericMiddle>;
// actual:   [(this: unknown, input: unknown) => unknown, (input: number) => 2]
// expected: [(input: string) => 1, (input: unknown) => unknown, (input: number) => 2]
```

I reran this locally with `tsd` to make sure I wasn't misreading it. So the current limitation is a bit bigger then the docs say: generic overloads can suppress earlier overloads altogether, which also makes `OverloadParameters` and `OverloadReturnType` wrong for those cases.

source/internal/function.d.ts
```diff
@@ -0,0 +1,110 @@
...
+import type {IsUnknown} from '../is-unknown.d.ts';
+import type {UnknownArray} from '../unknown-array.d.ts';
+
+/**
+ * Obtain the parameters of a function type in a tuple
+ * This works even when the parameters type is a readonly array
+ */
+export type Parameters<T extends (...args: any) => any> = T extends (...args: infer P extends UnknownArray) => any ? P : never;
+
+// ========================================================================
```
sindresorhus: Use `/* */` multiline comment for this whole block.

source/internal/function.d.ts
```diff
@@ -0,0 +1,110 @@
...
+// the constraint on the right-hand side.
+// ========================================================================
+
+declare const unique: unique symbol;
+type Unique = typeof unique;
+
+/**
+Detect whether the last overload of a function type has an explicit `this` annotation.
+
+Intersects `(this: Unique, ...)` from the right. Per TypeScript's deduplication rules
```
sindresorhus: Don't hard-wrap.

sindresorhus (changes requested):

source/function-overloads.d.ts
som-sm: @ikeyan This still seems pending.

ikeyan: Already fixed.
https://github.com/sindresorhus/type-fest/pull/1399/changes#diff-30f68cc8b7b8951a704612d6b92566aa63dad4b1110124e296262946d9c0d30aR42-R46

som-sm: I mean `Overloads<any>` should yield `any`.

source/internal/function.d.ts
```diff
@@ -0,0 +1,112 @@
...
+import type {IsUnknown} from '../is-unknown.d.ts';
+import type {Sum} from '../sum.d.ts';
+import type {UnknownArray} from '../unknown-array.d.ts';
+
+/**
+Obtain the parameters of a function type in a tuple.
+
+This works even when the parameters type is a readonly array.
+*/
+export type Parameters<T extends (...args: any) => any> = T extends (...args: infer P extends UnknownArray) => any ? P : never;
```
som-sm: Give it a different name.

source/internal/function.d.ts
```diff
@@ -0,0 +1,112 @@
...
+type Unique = typeof unique;
+
+/**
+Detect whether an overload with the given (This, Parameters, Return) has an explicit `this` annotation in the original function type.
+
+Intersects `(this: Unique, ...args: Parameters) => Return` onto the function type from the right. Per TypeScript's deduplication rules (see "Overload enumeration" above), implicit `this` absorbs the sentinel (same (P,R), one implicit -- first-wins), so `ThisParameterType` stays `unknown`. Explicit `this: unknown` does not absorb it (both explicit, different `This` -- not duplicate), so the sentinel becomes the last overload and `ThisParameterType` returns `Unique`.
+*/
+type HasExplicitThis<
+	FunctionType extends (...args: any) => any,
+	This, Parameters_ extends UnknownArray, Return,
```
som-sm: Should be on separate lines.

source/internal/function.d.ts
```diff
@@ -0,0 +1,112 @@
...
+> = IsUnknown<This> extends true
+	? IsEqual<ThisParameterType<FunctionType & ((this: Unique, ...args: Parameters_) => Return)>, Unique>
+	: true;
+
+type MaxOverloadPatterns = 4;
+type OverloadIndex = IntRange<0, MaxOverloadPatterns>;
+/**
```
som-sm: Line break.

source/internal/function.d.ts
```diff
@@ -0,0 +1,112 @@
...
+
+- **Effect observed** (the two differ): the intersection advanced the view. Output `ExtractedN`, intersect it onto `AllOverloads`, and continue with the same N.
+- **No effect** (they are equal): the intersection would not advance the view (e.g. aliasing generic overloads that infer to the same concrete signature). Output `ExtractedN` without intersecting, and advance N to the next position.
+- **Sentinel hit** (`NthLastOverload` returns `undefined`): no original overload at this depth. Return the accumulated result.
+
+The loop terminates when either N exceeds `OverloadIndex` or the sentinel is reached.
+
+@see https://github.com/microsoft/TypeScript/issues/32164#issuecomment-1146737709
+*/
+export type CollectOverloads<
```
som-sm: `CollectOverloads` and `NthLastOverload` overload should be in `function-overloads.d.ts` since they're closely tied together.

source/internal/function.d.ts
som-sm, line 1: Internal types also need proper tests.

source/function-overloads.d.ts
```diff
@@ -0,0 +1,92 @@
...
+
+type AllParameters = OverloadParameters<typeof request>;
+//=> [url: string] | [url: string, options: {json: true}]
+```
+
+Known limitations are the same as {@link Overloads}.
+
+@category Function
+*/
+export type OverloadParameters<FunctionType extends (...args: any) => any> = Parameters<Overloads<FunctionType>[number]>;
```
som-sm: Should be in a separate file.

source/function-overloads.d.ts
```diff
@@ -0,0 +1,92 @@
...
+
+type AllReturnTypes = OverloadReturnType<typeof request>;
+//=> Promise<string> | Promise<unknown>
+```
+
+Known limitations are the same as {@link Overloads}.
+
+@category Function
+*/
+export type OverloadReturnType<FunctionType extends (...args: any) => any> = ReturnType<Overloads<FunctionType>[number]>;
```
som-sm: Should be in a separate file.

source/overloads.d.ts
som-sm, line 1: File name should match the name of the exported type.

som-sm (changes requested):

som-sm:
Also, looks like the explicit `this` logic is not quite working with generic functions, refer:
```ts
type T = Overloads<{<T>(x: T): T}>;
//=> [(this: unknown, x: unknown) => unknown]
```

source/internal/function.d.ts
```diff
@@ -0,0 +1,112 @@
...
+	(this: infer T0, ...args: infer P0 extends UnknownArray): infer R0;
+}
+	? ({
+		3: [T3, P3, R3];
+		2: [T2, P2, R2];
+		1: [T1, P1, R1];
+		0: [T0, P0, R0];
+	}[N] extends [infer T, infer P extends UnknownArray, infer R]
+		? IsEqual<R, Unique> extends true
+			? undefined // No overload at this position
```
som-sm: I think this bit should be inside `CollectOverloads`, `NthLastOverload` should return the `Unique` function, and then checking should happen in `CollectOverloads`. Because `CollectOverloads` prepends the `Unique` function, it'd be nicer if it handled it explicitly. And then `NthLastOverload` should simply return the `N`th last overload.

source/function-overloads.d.ts
ikeyan: done

ikeyan:
> Also, looks like the explicit `this` logic is not quite working with generic functions, refer:
> 
> ```ts
> type T = Overloads<{<T>(x: T): T}>;
> //=> [(this: unknown, x: unknown) => unknown]
> ```

Added to known limitations

ikeyan:
@sindresorhus 

Right. Same underlying limitation, just bounded by a larger constant now.

Quick context first: those repros were against the version of the code at the time the review landed, which used a 1-slot `NthLastOverload`. The code now uses a 4-slot template, so applying those exact examples to current `Overloads` no longer drops anything: `Overloads<typeof genericLast>` returns the `[(input: string) => 1, (input: unknown) => unknown]` you expected. But the structural problem is unchanged; only the constant moved from 1 to 4. To reproduce the same cascade against the current code you need ≥ 4 generic overloads in the source, which is what the existing limitations note calls out.

Why it remains structural: the technique matches the function against a target shape whose slots are non-generic call signatures introduced by `infer` (`(this: infer Ti, ...args: infer Pi): infer Ri`). `infer` cannot bind a type parameter *inside* a captured signature (only variables bound to the enclosing conditional), so a generic call signature cannot be captured as generic in a slot; every captured slot is monomorphized to the signature's constraint form (so `<Type>(input: Type) => Type` instantiates to `(input: unknown) => unknown`). `NthLastOverload` itself does not halt on a generic overload as long as the source has fewer generic overloads than the slot count, but once the count exceeds the slot count, the monomorphized slot collides with the original generic when intersected back onto `AllOverloads` in `CollectOverloads`'s advancement step. The dedup-merge there is what causes overloads on the lossy side of the collision to disappear. That is exactly why in the original 1-slot repros the `(input: string) => 1` *before* the generic vanished while the `(input: number) => 2` *after* survived.

So the limitation is bounded by a constant rather than absent. Closing it (rather than just pushing the constant up further) would need TypeScript itself to grow a way to capture a single, possibly-generic call signature through conditional inference, such as `extends infer F extends AnyNonOverloadedFunction`, where the constraint pins `F` to one non-overloaded signature so each overload could be matched one at a time. That does not exist today.

som-sm:
@ikeyan I've slightly cleaned up `overload.d.ts` in [6282d78](https://github.com/sindresorhus/type-fest/pull/1399/commits/6282d78dfc68e275389f4332405326d067bc31c7), please review once.

No logic changes, just removed the `AllOverloads` type argument, and inlined comments.

PR #1401, opened 2026-04-13
Add `NonNullableDeep` type

### LINKED ISSUE #1400: NonNullableDeep

`NonNullableDeep` has already been requested in #795. However, the https://github.com/sindresorhus/type-fest/pull/1117 that was supposed to add it added only `SetNonNullableDeep`, and `NonNullableDeep` has never been added.

The proposal is to add `NonNullableDeep` which recursively removes `null`, `undefined`, and optionality from all properties.

### Type source

_No response_

som-sm:
Closes #1400 

Also, handles the following from #1117:
> `SetNonNullableDeep` currently doesn't deeply remove all nullables if `KeyPaths` is `any`.


sindresorhus:
Looks great 👍

PR #1404, opened 2026-04-14
`CamelCase`: Add `preserveLeadingUnderscores` option

### LINKED ISSUE #1299: Make `CamelCasedPropertiesDeep` preserve leading underscores
Version [9.0.0 of camelcase](https://github.com/sindresorhus/camelcase/releases/tag/v9.0.0) changed the behavior to preserve leading underscores. This means that now trying to deeply camel case an object by mapping the keys through `camelcase` results in a different runtime output than the result of `CamelCasedPropertiesDeep`.

For example:

```ts
type DeeplyCamelCase<T> = CamelCasedPropertiesDeep<
  T,
  { preserveConsecutiveUppercase: true }
>;

/**
 * Deeply camelCases an object or array
 */
function deeplyCamelCase<T>(input: T): DeeplyCamelCase<T>;
function deeplyCamelCase(obj: unknown): unknown {
  if (Array.isArray(obj)) {
    return obj.map(deeplyCamelCase);
  }

  if (obj !== null && typeof obj === 'object') {
    return Object.entries(obj).reduce<Record<string, unknown>>(
      (result, [key, value]) => {
        const camelCaseKey = camelCase(key, {
          preserveConsecutiveUppercase: true,
        });
        result[camelCaseKey] = deeplyCamelCase(value);
        return result;
      },
      {},
    );
  }

  return obj;
}

const dcc = deeplyCamelCase({ _foo_bar: 'baz' });
// At this point dcc has type `{ fooBar }`, but the runtime output is `{ _fooBar }`
```

I think that these types should be kept in sync (or at least have an option, like `preserveConsecutiveUppercase`, to keep leading underscores).

obarcelonap:
Fixes #1299

`camelcase` v9 changed its behavior to preserve leading underscores (e.g. `_foo_bar` → `_fooBar`). This adds a matching opt-in `preserveLeadingUnderscores` option to `CamelCase`, `CamelCasedProperties`, and `CamelCasedPropertiesDeep` so the type-level and runtime behaviors can be kept in sync.

The option is opt-in (default `false`) to preserve backwards compatibility.

### DIFF
--- a/source/camel-case.d.ts
+++ b/source/camel-case.d.ts
@@ -46,7 +46,6 @@ type LeadingUnderscores<Type extends string, Underscores extends string = ''> =
 		? LeadingUnderscores<Rest, `_${Underscores}`>
 		: Underscores;
 
-
 /**
 Convert an array of words to camel-case.
 */
@@ -117,9 +116,9 @@ export type CamelCase<Type, Options extends CamelCaseOptions = {}> = Type extend
 		: `${Options['preserveLeadingUnderscores'] extends true
 			? LeadingUnderscores<Type>
 			: ''}${Uncapitalize<CamelCaseFromArray<
-			Words<Type extends Uppercase<Type> ? Lowercase<Type> : Type, Options>,
-			ApplyDefaultOptions<CamelCaseOptions, _DefaultCamelCaseOptions, Options>
-		>>}`
+				Words<Type extends Uppercase<Type> ? Lowercase<Type> : Type, Options>,
+				ApplyDefaultOptions<CamelCaseOptions, _DefaultCamelCaseOptions, Options>
+			>>}`
 	: Type;
 
 export {};
--- a/source/camel-case.d.ts
+++ b/source/camel-case.d.ts
@@ -46,7 +46,6 @@ type LeadingUnderscores<Type extends string, Underscores extends string = ''> =
 		? LeadingUnderscores<Rest, `_${Underscores}`>
 		: Underscores;
 
-type StripLeadingUnderscores<S extends string> = S extends `_${infer Rest}` ? StripLeadingUnderscores<Rest> : S;
 
 /**
 Convert an array of words to camel-case.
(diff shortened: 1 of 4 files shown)


source/camel-case.d.ts
```diff
@@ -83,10 +100,17 @@ const dbResult: CamelCasedProperties<RawOptions> = {
...
+			? StripLeadingUnderscores<Type> extends infer Stripped extends string
+				? `${LeadingUnderscores<Type>}${Uncapitalize<CamelCaseFromArray<
+					Words<Stripped extends Uppercase<Stripped> ? Lowercase<Stripped> : Stripped, Options>,
+					ApplyDefaultOptions<CamelCaseOptions, _DefaultCamelCaseOptions, Options>
+				>>}`
+				: never
+			: Uncapitalize<CamelCaseFromArray<
+				Words<Type extends Uppercase<Type> ? Lowercase<Type> : Type, Options>,
+				ApplyDefaultOptions<CamelCaseOptions, _DefaultCamelCaseOptions, Options>
+			>>
```
som-sm: I don't think you need `StripLeadingUnderscores`, you can simply do:

```ts
export type CamelCase<Type, Options extends CamelCaseOptions = {}> = Type extends string
	? string extends Type
		? Type
		: `${Options['preserveLeadingUnderscores'] extends true
			? LeadingUnderscores<Type>
			: ''
		}${Uncapitalize<CamelCaseFromArray<
			Words<Type extends Uppercase<Type> ? Lowercase<Type> : Type, Options>,
			ApplyDefaultOptions<CamelCaseOptions, _DefaultCamelCaseOptions, Options>
		>>}`
	: Type;
```

source/camel-case.d.ts
```diff
@@ -13,12 +13,26 @@ export type CamelCaseOptions = WordsOptions & {
...
+	*/
+	preserveLeadingUnderscores?: boolean;
 };
 
 export type _DefaultCamelCaseOptions = _DefaultWordsOptions & {
 	preserveConsecutiveUppercase: false;
+	preserveLeadingUnderscores: false;
 };
 
+type LeadingUnderscores<S extends string> = S extends `_${infer Rest}` ? `_${LeadingUnderscores<Rest>}` : '';
```
som-sm: Better to use a tail-recursive implementation:
```ts
type LeadingUnderscores<Type, Underscores extends string = ''> =
	Type extends `_${infer Rest}`
		? LeadingUnderscores<Rest, `_${Underscores}`>
		: Underscores;
```

<br>

Also, add a minimal doc string:

````ts
/**
Extract leading underscores from a string.

@example
```
type A = LeadingUnderscores<'__foo_bar'>;
//=> '__'

type B = LeadingUnderscores<'foo_bar'>;
//=> ''
```
*/
````

File: test-d/camel-case.ts [hunk 3]
```diff
@@ -99,3 +99,12 @@ expectType<CamelCase<'fooBAR:biz', {splitOnPunctuation: true; preserveConsecutiv
 expectType<CamelCase<'foo-bar::01', {splitOnPunctuation: true}>>('fooBar01');
 expectType<CamelCase<'foo-bar::01', {splitOnPunctuation: true; splitOnNumbers: false}>>('fooBar01');
 expectType<CamelCase<'foo-bar::01', {splitOnPunctuation: true; splitOnNumbers: true}>>('fooBar01');
+
+expectType<CamelCase<'_foo_bar', {preserveLeadingUnderscores: true}>>('_fooBar');
```
som-sm, line 104: Also add case where this option is enabled and the string has no leading underscores.

som-sm (changes requested):

source/camel-case.d.ts
obarcelonap: fixed

source/camel-case.d.ts
obarcelonap: done

som-sm (approved):

obarcelonap:
@som-sm Thx for the CI fix ❤️ 
I was wondering what are the next steps here and the ETA to get these changes into a published version.

PR #1405, opened 2026-04-14
`UnionToTuple`: Fix behavior with large unions

som-sm:
Currently, `UnionToTuple` hits recursion limits when the input union contains around 45 members.

This PR refactors `UnionToTuple` to leverage tail recursion, allowing it to generate longer tuples.

### DIFF
--- a/test-d/union-to-tuple.ts
+++ b/test-d/union-to-tuple.ts
@@ -1,5 +1,5 @@
 import {expectAssignable, expectType} from 'tsd';
-import type {UnionToTuple} from '../index.d.ts';
+import type {IntClosedRange, UnionToTuple} from '../index.d.ts';
 
 type Options = UnionToTuple<'a' | 'b' | 'c'>;
 // Results unordered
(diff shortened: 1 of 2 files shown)


test-d/union-to-tuple.ts
```diff
@@ -19,6 +19,13 @@ expectType<Options2[number]>({} as (1 | false | true));
 type DifferentModifierUnion = {readonly a: 0} | {a: 0};
 expectType<DifferentModifierUnion>({} as UnionToTuple<DifferentModifierUnion>[number]);
 
+// Long unions
+type OneToFifty = IntClosedRange<1, 50>;
+type OneToTwoHundred = IntClosedRange<1, 200>;
```
som-sm: NOTE: With `UnionToTuple`, we can't really go up to `1000`, like some of our other test cases, because before hitting the recursion limit of `1000`, it hits the instantiation limit of `5M`. 

It could go up to around 600, but the test case doesn't use `600` because that makes everything super slow. And also, the in-editor experience is just not usable at that point.

som-sm: Here are some diagnostics for `UnionToTuple`:

image

Notice there are more than 5M instantiations.

<br>

From initial testing, it looks like the use of `ExcludeExactly` is bumping up the number of instantiations. Here's another diagnostic run but using `Exclude` instead of `ExcludeExactly`:

image
  
It doesn't error out this time, but it's still extremely slow, and practically unusable.

test-d/union-to-tuple.ts
```diff
@@ -19,6 +19,13 @@ expectType<Options2[number]>({} as (1 | false | true));
...
+
+expectType<UnionToTuple<OneToFifty>[number]>({} as OneToFifty);
```
sindresorhus:
```suggestion
expectType<UnionToTuple<OneToFifty>[number]>({} as OneToFifty);
expectType<200>({} as UnionToTuple<OneToTwoHundred>['length']);
```

som-sm: Much better, also inlined everything, doesn't need extracting into a type alias:
```ts
expectType<50>({} as UnionToTuple<IntClosedRange<1, 50>>['length']);
```

PR #1406, opened 2026-04-15
`TsConfigJson`: Add TypeScript 6.0 fields

ntnyq:
This PR add TypeScript v6 support to `tsconfig-json.d.ts`

Refs:

- [TypeScript 6.0](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html)


sindresorhus:
A couple of missing things:

`esnext.temporal` is part of TS 6, and `esnext.float16` is also in the upstream lib list.

PR #1407, opened 2026-04-21
`ApplyDefaultOptions`: Fix behavior with explicit `undefined`

som-sm:
Minor change, this PR solves the caveat mentioned below:
https://github.com/sindresorhus/type-fest/blob/a5491644b32160f804dd10d0b44dad461037f4c1/test-d/internal/apply-default-options.ts#L39-L41

### DIFF
--- a/xo.config.js
+++ b/xo.config.js
@@ -66,6 +66,7 @@ const xoConfig = [
 		rules: {
 			'unicorn/no-immediate-mutation': 'off',
 			'require-unicode-regexp': 'off',
+			'@typescript-eslint/no-unnecessary-type-assertion': 'off',
 		},
 	},
 	{
--- a/source/internal/object.d.ts
+++ b/source/internal/object.d.ts
@@ -3,7 +3,6 @@ import type {IsEqual} from '../is-equal.d.ts';
 import type {KeysOfUnion} from '../keys-of-union.d.ts';
 import type {RequiredKeysOf} from '../required-keys-of.d.ts';
 import type {Merge} from '../merge.d.ts';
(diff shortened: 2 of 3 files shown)


xo.config.js
```diff
@@ -66,6 +66,7 @@ const xoConfig = [
 		rules: {
 			'unicorn/no-immediate-mutation': 'off',
 			'require-unicode-regexp': 'off',
+			'@typescript-eslint/no-unnecessary-type-assertion': 'off',
```
som-sm, line 69: This rule is causing several errors in our test files. It looks like there are some changes around this.

som-sm:
@sindresorhus Should we add `continue-on-error: true` to our `Claude Code Review` workflow?

sindresorhus:
`continue-on-error` would hide other failures, and the Claude thing doesn't actually fail the PR, so is it really needed?

som-sm:
> `continue-on-error` would hide other failures, and the Claude thing doesn't actually fail the PR, so is it really needed?

Not really needed actually. But, I did some testing, and looks like we can very easily skip _just_ the "rate limit" errors once https://github.com/anthropics/claude-code-action/issues/1040 is resolved.

```yml
- name: Ignore Rate Limit Error
  if: steps.claude-review.outcome == 'failure'
  run: |
    FILE='${{ steps.claude-review.outputs.execution_file }}'

    if [ ! -f "$FILE" ]; then
      exit 1
    fi

    if grep -qi "You've hit your limit" "$FILE"; then
      exit 0
    fi

    exit 1
```

PR #1408, opened 2026-04-22
`ExtendsStrict`: Add `distributiveUnions`, `strictNever` & `strictAny` options

som-sm:
In this current form, the `ExtendsStrict` type is not quite useful as its default behaviour is quite restrictive and makes it difficult to use the type in different scenarios.

So, this PR adds three new options to `ExtendsStrict`, which should make the type usable in a lot of different contexts.

### DIFF
--- a/source/extends-strict.d.ts
+++ b/source/extends-strict.d.ts
@@ -47,6 +47,8 @@ export type ExtendsStrictOptions = {
 	type T4 = ExtendsStrict<never, any, {strictNever: true}>;
 	//=> true
 	```
+
+	Note: This option only has an effect when checking assignability from `never` (`ExtendsStrict<never, ...>`), and not when checking assignability to `never` (`ExtendsStrict<..., never>`).
 	*/
 	strictNever?: boolean;
 
@@ -89,6 +91,8 @@ export type ExtendsStrictOptions = {
 	type T3 = ExtendsStrict<any, unknown, {strictAny: false; distributiveUnions: true}>;
 	//=> true
 	```
+
+	Note: This option only has an effect when checking assignability from `any` (`ExtendsStrict<any, ...>`), and not when checking assignability to `any` (`ExtendsStrict<..., any>`).
 	*/
 	strictAny?: boolean;
 };
--- a/source/extends-strict.d.ts
+++ b/source/extends-strict.d.ts
@@ -4,6 +4,7 @@ import type {ApplyDefaultOptions} from './internal/object.d.ts';
 import type {And} from './and.d.ts';
 import type {Or} from './or.d.ts';
 import type {IsUnknown} from './is-unknown.d.ts';
+import type {OrAll} from './or-all.d.ts';
 
 export type ExtendsStrictOptions = {
 	/**
@@ -27,7 +28,7 @@ export type ExtendsStrictOptions = {
 	/**
 	Whether `never` extends every other type.
 
-	When enabled, `never` is not treated as a bottom type and only extends itself (or `any`).
+	When enabled, `never` is not treated as a bottom type and only extends itself (or `any` / `unknown`).
 
 	@default true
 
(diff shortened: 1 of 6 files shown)


test-d/paths.ts
```diff
@@ -208,6 +208,9 @@ expectType<`a.${string}`>(unionLeaves2); // Collapsed union
 declare const unionLeaves3: Paths<{a: string | {toLowerCase: number}}, {leavesOnly: true}>;
 expectType<'a' | 'a.toLowerCase'>(unionLeaves3);
 
+declare const unionLeaves4: Paths<{a: {b: string} | {c: string}}, {leavesOnly: true}>; // No common keys b/w `{b: string}` and `{c: string}`, but this shouldn't make `a` a leaf.
+expectType<'a.b' | 'a.c'>(unionLeaves4);
```
som-sm, line 212: This unrelated test case is part of this PR because I tried using `ExtendsStrict` in `Paths` at the following place:
https://github.com/sindresorhus/type-fest/blob/a5491644b32160f804dd10d0b44dad461037f4c1/source/paths.d.ts#L213-L226

If you replace the above with the following:
```ts
			? ((Options['leavesOnly'] extends true
				? Options['maxRecursionDepth'] extends CurrentDepth
					? TransformedKey
					: ExtendsStrict<
						T[Key],
						readonly [] | NonRecursiveType | Exclude<MapsSetsOrArrays, UnknownArray>,
						{distributiveUnions: true; neverExtendsAll: true; anyExtendsAll: true}
					> extends false
						? IsNever<keyof T[Key]> extends true // Check for empty object & `unknown`, because `keyof unknown` is `never`.
							? TransformedKey
							: never
						: TransformedKey
				: TransformedKey
			) extends infer _TransformedKey
```
All the existing tests would pass _except_ this newly added one.

Note: The problem is _not_ with the usage of `ExtendsStrict`, it's just that in this case, in the `else` branch, we assume `T[Key]` is distributed before using `keyof` on it.

source/extends-strict.d.ts
```diff
@@ -1,44 +1,143 @@
...
+	```
+	*/
+	strictAny?: boolean;
+};
+
+type DefaultExtendsStrictOptions = {
+	distributiveUnions: false;
+	strictNever: true;
+	strictAny: false;
+};
```
som-sm, line 108: The default values currently are configured such that there are no breaking changes.

test-d/extends-strict.ts
```diff
@@ -23,7 +23,7 @@ expectType<ExtendsStrict<string, never>>(false);
 
 // Any and unknown
 expectType<ExtendsStrict<any, any>>(true);
-expectType<ExtendsStrict<any, never>>(true);
+expectType<ExtendsStrict<any, never>>(false);
```
som-sm, line 26: This isn't necessarily a change in behavior, it's more of a bug fix. When `distributiveUnions` is disabled (default), then the default TS behavior is that `any` extends everything except `never`, which is how the updated implementation works. I guess it's fine!

It's more of like, when `strictAny` is `false`, we fallback to the default behavior.

```ts
type Test<T, U> = [T] extends [U] ? true : false;

type T1 = Test<any, never>; //=> false
```

test-d/conditional-keys.ts
```diff
@@ -62,7 +62,7 @@ expectType<never>({} as ConditionalKeys<{length: number} | [number], number>);
 expectType<'a' | 'b' | 'c' | 'd' | 'e'>(
 	{} as ConditionalKeys<{a?: string; b: string | number; readonly c: boolean; readonly d?: bigint; e: never}, any>,
 );
-expectType<'a' | 'b' | 'c' | 'd'>(
+expectType<'a' | 'b' | 'c' | 'd' | 'e'>(
 	{} as ConditionalKeys<{a?: string; b: string | number; readonly c: boolean; readonly d?: bigint; e: never}, unknown>,
 );
```
som-sm, line 67: This is a bug. I don't see why `never` shouldn't be selected when the condition is `unknown`.

source/extends-strict.d.ts
som-sm: Ideally, `strictAny` should be `true` by default.

Issue #1389, opened 2026-03-21
possible bug in `ConditionalPickDeep` (enum case)


salisbury-espinosa:
### Bug description

1. unit test:

```typescript
import { expectType, expectNotType } from 'tsd';
import { ConditionalPickDeep, Paths } from 'type-fest';

enum TestEnum {
  TEST1 = 'TEST1',
  TEST2 = 'TEST2',
}

type ExampleForEnum = {
  enum: TestEnum;
  string: string | null;
};

expectType<'enum' | 'string'>(
  {} as Paths<ConditionalPickDeep<ExampleForEnum, TestEnum>>,
);
```

=> pass, but i expected instead of `'enum' | 'string'` => `'enum' `

2. and and we can see how AssertCondition works from  https://github.com/sindresorhus/type-fest/blob/main/source/conditional-pick-deep.d.ts

```typescript
import { expectType, expectNotType } from 'tsd';
import { ConditionalPickDeep, IsEqual, Paths } from 'type-fest';

enum TestEnum {
  TEST1 = 'TEST1',
  TEST2 = 'TEST2',
}

type ExampleForEnum = {
  enum: TestEnum;
  string: string | null;
};


type ConditionalPickDeepOptions = {
  condition?: 'equality' | 'extends';
};
type AssertCondition<
  Type,
  Condition,
  Options extends ConditionalPickDeepOptions,
> = Options['condition'] extends 'equality'
  ? IsEqual<Type, Condition>
  : Type extends Condition
    ? true
    : false;



expectType<
  AssertCondition<ExampleForEnum['string'], TestEnum, { condition: 'equality' }>
>(false);

expectType<
  AssertCondition<ExampleForEnum['string'], TestEnum, { condition: 'extends' }>
>(false);
```

=> pass (ExampleForEnum['string'] not extends TestEnum nor equal)

som-sm:
Can't reproduce it on the playground:
```ts
import { ConditionalPickDeep, Paths } from 'type-fest';

enum TestEnum {
  TEST1 = 'TEST1',
  TEST2 = 'TEST2',
}

type ExampleForEnum = {
  enum: TestEnum;
  string: string | null;
};

type T = Paths<ConditionalPickDeep<ExampleForEnum, TestEnum>>;
//=> "enum"
```
[...]

PR #1409, opened 2026-04-22
Resolve Bug : #1389 , Fix ConditionalPickDeep incorrectly recursing into unions with null/primitive members 

hksinha01:
## Bug

ConditionalPickDeep was incorrectly including keys whose value types are non-object unions (for example string | null) when the Condition is an enum.

This PR fixes the issue in two ways:

1) Makes the condition check non-distributive using tuple wrapping: [Type] extends [Condition]. 
2) Prevents recursion into non-object unions by using: IsPlainObject<NonNullable<Type[Key]>> extends true

Additional tests were added to cover:

- Enum selection edge case
- Union types that must not recurse
- Nested enum cases
- Verification using Paths<>


som-sm:
Refer https://github.com/sindresorhus/type-fest/issues/1389#issuecomment-4295793113

PR #1410, opened 2026-04-24
Add `UnwrapRequired` type

porada:
Following up on #1294 and #1296, this PR introduces `UnwrapRequired`, a type that reverts the `Required` modifier on object types.

`UnwrapRequired` mirrors `UnwrapPartial`’s implementation:

* No-op for non-`Required` types.
* Simplicity over special-casing unions and arrays.
* Includes similar test cases.

### DIFF
--- a/source/unwrap-required.d.ts
+++ b/source/unwrap-required.d.ts
@@ -22,11 +22,7 @@ Note: If the provided type isn’t of `Required<T>`, `UnwrapRequired` has no eff
 */
 export type UnwrapRequired<RequiredObjectType> =
 	RequiredObjectType extends Required<infer ObjectType>
-		? (
-			Required<ObjectType> extends RequiredObjectType
-				? ObjectType
-				: RequiredObjectType
-		)
+		? ObjectType
 		: RequiredObjectType;
 
 export {};
--- a/source/unwrap-required.d.ts
+++ b/source/unwrap-required.d.ts
@@ -18,9 +18,9 @@ type DraftContactFormData = UnwrapRequired<ContactFormData>;
 //=> {email: string; message?: string}
 ```
 
-Note 1: If the provided type isn’t of `Required<T>`, `UnwrapRequired` has no effect on the original type.
-
-Note 2: `Required<T>` is lossy for arrays and tuples with optional elements, so `UnwrapRequired` has no effect on array and tuple types. Object types with optional properties are unaffected by this limitation (and work as expected with `UnwrapRequired`).
+Note:
+	- If the provided type isn’t of the form `Required<T>`, `UnwrapRequired` simply returns the input type.
+	- `UnwrapRequired` doesn't work with arrays, if instantiated with arrays, it simply returns the input type.
 
 @category Object
 */
--- a/source/unwrap-required.d.ts
+++ b/source/unwrap-required.d.ts
@@ -24,7 +24,11 @@ Note 2: `Required<T>` is lossy for tuples with optional elements. `UnwrapRequire
 */
 export type UnwrapRequired<RequiredObjectType> =
 	RequiredObjectType extends Required<infer ObjectType>
-		? ObjectType
+		? (
+			Required<ObjectType> extends RequiredObjectType
+				? ObjectType
+				: RequiredObjectType
+		)
 		: RequiredObjectType;
 
 export {};
--- a/source/unwrap-required.d.ts
+++ b/source/unwrap-required.d.ts
@@ -16,7 +16,9 @@ type DraftContactFormData = UnwrapRequired<ContactFormData>;
 //=> {email: string; message?: string}
 ```
 
-Note: If the provided type isn’t of `Required<T>`, `UnwrapRequired` has no effect on the original type.
+Note 1: If the provided type isn’t of `Required<T>`, `UnwrapRequired` has no effect on the original type.
+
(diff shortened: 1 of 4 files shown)


test-d/unwrap-required.ts
```diff
@@ -0,0 +1,65 @@
...
+// `UnwrapRequired` works with union types
+type RequiredUnionType = Required<TestType> | Required<AnotherTestType>;
+
+expectType<TestType | AnotherTestType>({} as UnwrapRequired<RequiredUnionType>);
+expectType<TestType | AnotherTestType>({} as UnwrapRequired<Required<TestType> | AnotherTestType>);
+
+// `UnwrapRequired` works with arrays and tuples
+type TestTuple = [string?, number?, boolean?];
+
+expectType<[string, number, boolean]>({} as UnwrapRequired<Required<TestTuple>>);
```
sindresorhus:
```suggestion
expectType<TestTuple>({} as UnwrapRequired<Required<TestTuple>>);
```

porada: You’d think that, right? Unfortunately:

```
Parameter type TestTuple is declared too wide for argument type [string, number, boolean].
```

som-sm: But that's an issue then, the result should have been `TestTuple`.

porada: Not exactly. Even though it’s stricter, `[string, number, boolean]` still passes as `TestTuple`. The test stack just expects the tightest version of the type.

I’ll see what can be done about this.

som-sm: No, if `[string, number, boolean]` were the desired outcome, what's the point of unwrapping? `Required<TestTuple>` is already `[string, number, boolean]`. The whole point of this type is that it returns the same type that was used to instantiate `Required`.

porada: Looks like `Required<T>` is not reversible for tuples—the very transformation is lossy. I removed these tests.

som-sm: I looked around this a bit, and this seems like a TS bug/limitation. Can you add a note that this type has known issues with arrays. Also, please add a similar to `UnwrapPartial`.

source/unwrap-required.d.ts
```diff
@@ -0,0 +1,32 @@
...
+Note: If the provided type isn’t of `Required<T>`, `UnwrapRequired` has no effect on the original type.
+
+@category Object
+*/
+export type UnwrapRequired<RequiredObjectType> =
+	RequiredObjectType extends Required<infer ObjectType>
+		? (
+			Required<ObjectType> extends RequiredObjectType
+				? ObjectType
+				: RequiredObjectType
```
som-sm: Also, why did we add this conditional? Currently, no tests seem to fail even if the conditional is replaced with simply `ObjectType`.

```ts
export type UnwrapRequired<RequiredObjectType> =
	RequiredObjectType extends Required<infer ObjectType>
		? ObjectType
		: RequiredObjectType;
```

porada: Good catch. It was a remnant of the initial `UnwrapPartial` definition that got carried over here.

`UnwrapRequired` is now fixed and the fix for `UnwrapPartial` is waiting at #1412. Sorry about that.

test-d/unwrap-required.ts
som-sm: @porada Can you also add the note 👆

source/unwrap-required.d.ts
```diff
@@ -0,0 +1,30 @@
...
+//=> {email: string; message?: string}
+```
+
+Note 1: If the provided type isn’t of `Required<T>`, `UnwrapRequired` has no effect on the original type.
+
+Note 2: `Required<T>` is lossy for tuples with optional elements. `UnwrapRequired` cannot recover the original tuple type in these cases. Object types with optional properties are unaffected by this limitation (and work as expected with `UnwrapRequired`).
+
+@category Object
+*/
+export type UnwrapRequired<RequiredObjectType> =
```
som-sm, line 27: Found few issues:

- Returns `unknown` for functions:
  ```ts
  type T = UnwrapRequired<() => string>;
  //=> unknown
  ```

- Doesn't preserve primitives:
  ```ts
  type T = UnwrapRequired<string>;
  //=> { toString: () => string; charAt: ... }
  ```
  Same for `Map`s, `Set`s, etc.

som-sm (changes requested):

source/unwrap-required.d.ts
porada: Turns out the original definition was correct.

som-sm: > Turns out the original definition was correct.

Not really. If you look closely, `Map`s, `Set`s, etc. _still_ move to the if branch of the nested conditional:
```ts
type UnwrapRequired<RequiredObjectType> =
	RequiredObjectType extends Required<infer ObjectType>
		? (
			Required<ObjectType> extends RequiredObjectType
				? 'If'
				: RequiredObjectType
		)
		: RequiredObjectType;

type T2 = UnwrapRequired<Map<string, number>>;
//=> 'If'
```

Because of this, `Map`s, `Set`s, etc. get unwrapped, and we don't want that.

image

---

You probably need an outer check like we have in many of our other types to skip primitives, `Map`s, `Set`s, etc.

porada:
@som-sm — Let me know if there’s anything else to address.

som-sm (approved):

porada:
🎉

PR #1412, opened 2026-04-27
`UnwrapPartial`: Add test case for non-`Partial` instantiation

porada:
[Context](https://github.com/sindresorhus/type-fest/pull/1410#discussion_r3148406072)

### DIFF
--- a/source/unwrap-partial.d.ts
+++ b/source/unwrap-partial.d.ts
@@ -23,11 +23,7 @@ Note: If the provided type isn’t of `Partial<T>`, `UnwrapPartial` has no effec
 */
 export type UnwrapPartial<PartialObjectType> =
 	PartialObjectType extends Partial<infer ObjectType>
-		? (
-			Partial<ObjectType> extends PartialObjectType
-				? ObjectType
-				: PartialObjectType
-		)
+		? ObjectType
 		: PartialObjectType;
 
 export {};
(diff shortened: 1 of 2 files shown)


sindresorhus:
I think this change is wrong because the removed guard is what preserves the documented no-op behavior for non-`Partial<T>` inputs. Without it, a normal object type with both required and optional properties is inferred through `Partial<infer ObjectType>` and the optional property becomes required.

Please add this regression test:

```ts
type MixedRequiredAndOptional = {
	required: string;
	optional?: number;
};

expectType<MixedRequiredAndOptional>({} as UnwrapPartial<MixedRequiredAndOptional>);
```

This fails with the simplified implementation, but passes with the previous guarded version.

porada:
Whoa, good catch. Thanks for staying vigilant.

I think this is worth merging for the guard alone.

som-sm:
> This fails with the simplified implementation, but passes with the previous guarded version.

Yeah, the check _is_ necessary in `UnwrapPartial`, but looks like it's _not_ necessary in `UnwrapRequired`. Because in `UnwrapRequired`, non-`Required` objects won't match the outer conditional, so the result would come directly from the else branch, refer:

```ts
type UnwrapRequired<RequiredObjectType> =
	RequiredObjectType extends Required<infer ObjectType>
		? ObjectType
		: 'No Match';

type T1 = UnwrapRequired<{a?: number; b: string}>;
//=> 'No Match'

type T2 = UnwrapRequired<{a: number; b?: string}>;
//=> 'No Match'

type T3 = UnwrapRequired<{a?: number; b?: string}>;
//=> 'No Match'
```

porada:
I wonder: can we merge this?

PR #1417, opened 2026-05-08
Add `Unbind` type

razhayat:
Collapse a union type into a single object type by merging all members.

Each property becomes:
- A union of all possible values across the union members
- Optional if it does not exist in every member

This removes the relationships between union members, producing
a loose object shape where properties are independent.

```ts
import type {Unbind} from 'type-fest';

type Success = {
	kind: 'success';
	data: string;
};

type Failure = {
	kind: 'error';
	error: Error;
};

type Result = Unbind<Success | Failure>;
//=> {
// 	kind: 'success' | 'error';
// 	data?: string;
// 	error?: Error;
// }
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -172,6 +172,7 @@ export type {IsOptional} from './source/is-optional.d.ts';
 export type {IsNullable} from './source/is-nullable.d.ts';
 export type {TupleOf} from './source/tuple-of.d.ts';
 export type {ExclusifyUnion} from './source/exclusify-union.d.ts';
+export type {Unbind} from './source/unbind.d.ts';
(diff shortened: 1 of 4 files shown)


sindresorhus:
I think the behavior is already covered by `AllUnionFields`, and `Unbind` is less descriptive.

The implementation is also a bit too clever and has boundary differences for `never` and `any` compared to the existing type. If anything, we could improve the `AllUnionFields` docs with this loose object shape example.

razhayat:
Right, I missed that. When I read about `AllUnionFields` I thought it only takes the shared keys into account.

Closing the PR

Issue #821, opened 2024-02-27
`NonEmptyObject` fails for objects with dynamic properties


DenizUgur:
The following code is considered as valid by Typescript, although I was expecting it to complain that the filter `foo` in `commonArguments` was empty.

```ts
import type { NonEmptyObject } from "type-fest"

interface CommonArguments {
  [filter: string]: NonEmptyObject<{ [argument: string]: string | number | undefined }>
}

export const commonArguments: CommonArguments = {
  foo: {}
}
```

sindresorhus:
// @kkmuffme

Mouvedia:
This makes `NonEmptyObject` and `RequireAtLeastOne` useless when using [template string pattern index signatures](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#symbol-and-template-string-pattern-index-signatures).

PR #1419, opened 2026-05-11
`NonEmptyObject`: Handle index signatures conservatively

KevinBigham:
Addresses #821.

`NonEmptyObject` currently accepts `{}` for types whose only members are index signatures:

```ts
import type {NonEmptyObject} from 'type-fest';

interface CommonArguments {
	[filter: string]: NonEmptyObject<{[argument: string]: string | number | undefined}>;
}

const args: CommonArguments = {foo: {}}; // ← accepted on main; should fail
```

## Why

A type whose only members are index signatures cannot statically express "at least one dynamic key" in TypeScript. The previous definition fell through to `RequireAtLeastOne<T, keyof T>`, which for `keyof T = string` mapped back to an index signature and trivially accepted `{}`.

## What changed

Re-derive the key set via `OmitIndexSignature` so:

- Pure-index-signature types fail closed and resolve to `never`.
- For mixed types, `HasRequiredKeys` / `RequireAtLeastOne` operate on the explicit-key subset, so index signatures no longer mask non-emptiness enforcement.

## Tests

Added to `test-d/non-empty-object.ts`:

1. Pure string index signature → `never`.
2. Pure number index signature → `never`.
3. Exact #821 repro (`{foo: {}}`) is rejected.
4. Mixed (index signature + required explicit key) still returns the original type.
5. Mixed (index signature + only optional explicit keys) requires at least one of the explicit optional keys.

`npm test` passes locally.

### DIFF
--- a/source/non-empty-object.d.ts
+++ b/source/non-empty-object.d.ts
@@ -1,10 +1,13 @@
 import type {HasRequiredKeys} from './has-required-keys.d.ts';
+import type {OmitIndexSignature} from './omit-index-signature.d.ts';
 import type {RequireAtLeastOne} from './require-at-least-one.d.ts';
 
 /**
 Represents an object with at least 1 non-optional key.
 
 This is useful when you need an object where all keys are optional, but there must be at least 1 key.
+
+Note: A type whose only members are index signatures (e.g. `{[key: string]: unknown}`) cannot statically express "at least one dynamic key" in TypeScript. For such types, `NonEmptyObject` fails closed and resolves to `never` rather than silently accepting `{}`.
 
 @example
 ```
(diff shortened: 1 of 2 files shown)


sindresorhus:
I think this accidentally changes `NonEmptyObject<any>` to `never`.

The new explicit-key guard treats `any` like an index-signature-only object: after omitting index signatures there are no keys left, so it falls into the `never` branch. That’s a regresion from `main`, where `NonEmptyObject<any>` stays `any`, and it also differs from `RequireAtLeastOne<any>`, which explicitly preserves `any`.

I manually checked this with `tsd`: this assertion passes on `main` and fails on this branch with “argument type never”.

```ts
expectType<any>({} as NonEmptyObject<any>);
```

Can you special-case `any` before the index-signature-only guard?

PR #1420, opened 2026-05-11
feat: add lowercase types and properties

aminya:
This introduces lowercase types. 
  - `LowerSpaced`, where everything will be lowercase and separated by spaces -> `lodash` behaviour
  - `Lowerdespaced`, where everything will be lowercase joined without any spaces (current implementation of this PR) -> `despaced` behaviour
  - `Lowercase`, where everything becomes lowercase without changing the delimiter.  -> `toLowercase` behaviour

It also fixed a bug in he delimiter-case logic to handle an empty delimiter, which cut the first letter.


sindresorhus:
Missing docs. The existing case-conversion utilities all have JSDoc with examples, categories, and matching README entries, but these new lowercase variants are exported without any public docs and don’t show up in the README’s Change case list.

aminya:
> Missing docs. The existing case-conversion utilities all have JSDoc with examples, categories, and matching README entries, but these new lowercase variants are exported without any public docs and don’t show up in the README’s Change case list.

Fixed in d753f09de15455f45fa5457315d4456e5a21b17d

som-sm (commented):
1. Lodash’s `lowerCase` utility replaces delimiters with spaces. For example, `_.lowerCase("aB")` returns `"a b"`. I think we should do the same because otherwise this type is _just_ similar to the built-in `Lowercase` type (except for the `splitOnPunctuation` option).

2. With `LowerCasedProperties`, users might expect that it simply lowercases the keys without changing anything around the delimiter. Maybe we don't need the `LowerCase` type at all, we could just have `LowerCasedProperties` that simply lowercases the key using the built-in `Lowercase` type.

aminya:
> 1. Lodash’s `lowerCase` utility replaces delimiters with spaces. For example, `_.lowerCase("aB")` returns `"a b"`. I think we should do the same because otherwise this type is _just_ similar to the built-in `Lowercase` type (except for the `splitOnPunctuation` option).
> 
> 2. With `LowerCasedProperties`, users might expect that it simply lowercases the keys without changing anything around the delimiter. Maybe we don't need the `LowerCase` type at all, we could just have `LowerCasedProperties` that simply lowercases the key using the built-in `Lowercase` type.

This PR is designed for the built-in toLowerCase. 
`fileName` becomes `filename`.

som-sm:
> This PR is designed for the built-in toLowerCase.
> `fileName` becomes `filename`.

@aminya Didn't get it, there's no `toLowerCase` built-in TS type utility. I guess you meant `Lowercase`?

We can already convert `'fileName'` to `'filename'` using the [built-in `Lowercase` type](https://www.typescriptlang.org/docs/handbook/utility-types.html#lowercasestringtype). Why do we need a new type for it?

The only difference currently from the built-in `Lowercase` is when the input contains delimiters. Like `Lowercase<'Foo-Bar'>` would give `'foo-bar'`, but this PR's `LowerCased<'Foo-Bar'>` would give `'foobar'`. This is where I mentioned would it be better to return `'foo bar'`?

som-sm:
> @aminya Didn't get it, there's no `toLowerCase` built-in. I guess you meant `Lowercase`?

If you meant [`String.prototype.toLowerCase`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLowerCase), then that behavior is already covered by TS's [built-in `Lowercase` type](https://www.typescriptlang.org/docs/handbook/utility-types.html#lowercasestringtype).

I guess we don't need the `LowerCased` type then. And we can update `LowerCasedProperties` to use the built-in `Lowercase` type as mentioned here:

> 2\. With `LowerCasedProperties`, users might expect that it simply lowercases the keys without changing anything around the delimiter. Maybe we don't need the `LowerCase` type at all, we could just have `LowerCasedProperties` that simply lowercases the key using the built-in `Lowercase` type.

aminya:
I can switch for the built-in Lowercase type, however, it will change the expected behaviour and break the options that the current types have. 

It will not be able to convert kebab-case, snake_case, etc to lowercase. It ignores the delimiter. 

```ts
const lowerFromKebab: LowerCase<'foo-bar'> = 'foobar';

const lowerFromSnake: LowerCase<'foo_bar'> = 'foobar';
```

Also, it will not support `splitOnNumbers` and `splitOnPunctuation`, which makes them inconsistent with other types from `type-fest`. 

```ts
const example: LowerCase<'onDialog:close', {splitOnPunctuation: true}> = 'ondialogclose';
```

The current implementation is extension of what type-fest already does not limited to a particular usecase. 


> The only difference currently from the built-in Lowercase is when the input contains delimiters. Like Lowercase<'Foo-Bar'> would give 'foo-bar', but this PR's LowerCased<'Foo-Bar'> would give 'foobar'. This is where I mentioned would it be better to return 'foo bar'?

I can add the space in if it's what you mean by lowercase. It depends on the definition. It might work for snake_case and kabab-case.

```ts
const lowerFromKebab: LowerCase<'foo-bar'> = 'foo bar';

const lowerFromSnake: LowerCase<'foo_bar'> = 'foo bar';
```

However, for camelCase and PascalCase, if this means `fooBar` and `FooBar` become `foo bar`, this is totally against what I am expecting.

aminya:
It might makes sense to introduce new types to cover all three options.
  - `LowerSpaced`, where everything will be lowercase and separated by spaces -> `lodash` behaviour
  - `Lowerdespaced`, where everything will be lowercase joined without any spaces (current implementation of this PR) -> `despaced` behaviour
  - `Lowercase`, where everything becomes lowercase without changing the delimiter.  -> `toLowercase` behaviour

som-sm:
> It might makes sense to introduce new types to cover all three options.

No, doesn't make sense to have multiple types for the same thing.

We usually try to align with `Lodash`'s behavior whenever possible, so it's better if we use a single space as delimiter:
- `LowerCase<"fooBar">` --> `"foo bar"`
- `LowerCase<"foo_bar">` --> `"foo bar"` 

<br>

Also, replacing the delimiter with empty string (`LowerCase<"foo_bar">` --> `"foobar"`) doesn't _really_ feel like the standard behavior. And, if this use case is required, we _already_ have `DelimiterCase` for that:

```ts
import type {DelimiterCase} from "type-fest";

type T1 = DelimiterCase<"fooBar", "">;
//=> "foobar"

type T2 = DelimiterCase<"foo_bar", "">;
//=> "foobar"
```
Note: There's currently a bug in `DelimiterCase` because of which it breaks with empty string, this should be fixed with #1424.

---

I'll reiterate the two approaches that I suggested:

1. We don't ship any custom `LowerCase` type. We _just_ ship a `LowercasedProperties` type that is built using the built-in `Lowercase` type:

    `LowercasedProperties<{"fooBar": 1, "bar_baz": 2}>` --> `{"foobar": 1, "bar_baz": 2}`

2. We ship a `LowerCase` type that uses space as delimiter similar to Lodash's `lowerCase`:
    - `LowerCase<"fooBar">` --> `"foo bar"`
    - `LowerCase<"foo_bar">` --> `"foo bar"` 
    
    And, we ship a `LowerCasedProperties` type that's built using our custom `LowerCase` type:
    `LowerCasedProperties<{"fooBar": 1, "bar_baz": 2}>` --> `{"foo bar": 1, "bar baz": 2}`

Approach 1 makes more sense to me because we can achieve the behavior in approach 2 using `DelimiterCasedProperties`. @sindresorhus WDYT?

aminya:
> > It might makes sense to introduce new types to cover all three options.
> 
> No, doesn't make sense to have multiple types for the same thing.
> 
> We usually try to align with `Lodash`'s behavior whenever possible, so it's better if we use a single space as delimiter:
> - `LowerCase<"fooBar">` --> `"foo bar"`
> - `LowerCase<"foo_bar">` --> `"foo bar"` 
>

My last commit covers all cases, they are not the same thing. They are different behaviours. What I need is converting fileName to filename. That's not what you're suggesting. 

Your package description doesn't mention that you enforce the Losash behavior. I thought this package is a generic Typescript type utility.

som-sm:
> What I need is converting fileName to filename. That's not what you're suggesting.

You can already achieve this using `DelimiterCase` and `DelimiterCasedProperties`. And if you *only* want to convert `'fileName'` to `'filename'`, you can use the built-in `Lowercase` type.

Did you go through my previous response?

> Also, replacing the delimiter with an empty string (`LowerCase<"foo_bar">` --> `"foobar"`) doesn’t really feel like standard behavior. And if this use case is needed, we already have `DelimiterCase` for that:
>
> ```ts
> import type {DelimiterCase} from "type-fest";
>
> type T1 = DelimiterCase<"fooBar", "">;
> //=> "foobar"
>
> type T2 = DelimiterCase<"foo_bar", "">;
> //=> "foobar"
> ```

<br>

Also, going forward, please provide playground examples. Otherwise, this discussion doesn’t seem to be converging to a conclusion. I think I’ve already explained quite clearly how the different use cases can be handled, so I’m not sure where the confusion still is.

<br>

> I thought this package is a generic TypeScript type utility.

It is, but the behavior you’re suggesting is _not_ generic. For non-standard behavior, we already have the `DelimiterCase*` types.

aminya:
Given that the suggested changes break my usecase I am not sure there's a point in keeping this open unless there's a breakthrough in your point of view

PR #1422, opened 2026-05-12
`DelimiterCase`: Fix behavior with union delimiters

som-sm:
Currently, `DelimiterCase` doesn't distribute properly over the specified `Delimiter`, resulting in strings that contain multiple delimiters.

Current behavior:
```ts
type T = DelimiterCase<'fooBarBaz', '#' | '$'>
//=> "foo#bar#baz" | "foo#bar$baz" | "foo$bar#baz" | "foo$bar$baz"
```

Fixed behavior:
```ts
type T = DelimiterCase<'fooBarBaz', '#' | '$'>
//=> "foo#bar#baz" | "foo$bar$baz"
```

### DIFF
--- a/test-d/delimiter-case.ts
+++ b/test-d/delimiter-case.ts
@@ -170,7 +170,7 @@ declare const anyDelimiter: DelimiterCase<'fooBar', any>;
 expectType<`foo${Lowercase<any>}bar`>(anyDelimiter);
 
 declare const neverDelimiter: DelimiterCase<'fooBar', never>;
-expectType<'fooBar'>(neverDelimiter);
+expectType<never>(neverDelimiter);
 
 // Verifying example
 type OddCasedProperties<T> = {
--- a/test-d/delimiter-case.ts
+++ b/test-d/delimiter-case.ts
@@ -103,11 +103,11 @@ expectType<'foo22#bar'>(
 declare const unionValue: DelimiterCase<'fooBar' | 'barBaz', '#'>;
 expectType<'foo#bar' | 'bar#baz'>(unionValue);
 
-declare const unionDelimiter: DelimiterCase<'fooBar', '#' | '$'>;
-expectType<'foo#bar' | 'foo$bar'>(unionDelimiter);
+declare const unionDelimiter: DelimiterCase<'fooBarBaz', '#' | '$'>;
+expectType<'foo#bar#baz' | 'foo$bar$baz'>(unionDelimiter);
 
-declare const unionValueAndDelimiter: DelimiterCase<'fooBar' | 'barBaz', '#' | '$'>;
-expectType<'foo#bar' | 'bar#baz' | 'foo$bar' | 'bar$baz'>(unionValueAndDelimiter);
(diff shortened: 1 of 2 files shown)


test-d/delimiter-case.ts
```diff
@@ -103,11 +103,11 @@ expectType<'foo22#bar'>(
...
 
-declare const unionDelimiter: DelimiterCase<'fooBar', '#' | '$'>;
-expectType<'foo#bar' | 'foo$bar'>(unionDelimiter);
+declare const unionDelimiter: DelimiterCase<'fooBarBaz', '#' | '$'>;
+expectType<'foo#bar#baz' | 'foo$bar$baz'>(unionDelimiter);
 
-declare const unionValueAndDelimiter: DelimiterCase<'fooBar' | 'barBaz', '#' | '$'>;
-expectType<'foo#bar' | 'bar#baz' | 'foo$bar' | 'bar$baz'>(unionValueAndDelimiter);
+declare const unionValueAndDelimiter: DelimiterCase<'fooBarBaz' | 'barBazFoo', '#' | '$'>;
+expectType<'foo#bar#baz' | 'bar#baz#foo' | 'foo$bar$baz' | 'bar$baz$foo'>(unionValueAndDelimiter);
```
sindresorhus, line 110: I sugest adding this test too, to cover the `any` case. Without it, the unreachable branch can still leak into the result because `any` evaluates both sides of the conditional type.

```suggestion
expectType<'foo#bar#baz' | 'bar#baz#foo' | 'foo$bar$baz' | 'bar$baz$foo'>(unionValueAndDelimiter);

declare const anyDelimiter: DelimiterCase<'fooBar', any>;
expectType<`foo${Lowercase<any>}bar`>(anyDelimiter);
expectNotAssignable<DelimiterCase<'fooBar', any>>('fooBar');
```

som-sm: Addressed in [710d048](https://github.com/sindresorhus/type-fest/pull/1422/commits/710d04852a1a7ff7c973fb3080c53359d78d9ea9).

Also, the `expectNotAssignable` test case shouldn't be required because the `expectType` would catch if the unreachable branch leaks into the result.

som-sm: Also, not _really_ sure what the behavior should be when `Delimiter` is `never`. Currently, I have kept it such that it simply returns the input type (`DelimiterCase<'fooBar', never>` -> `fooBar`).

Maybe we should return `never` (`DelimiterCase<'fooBar', never>` -> `never`)? Or maybe we should treat `never` as an empty string (`DelimiterCase<'fooBar', never>` -> `foobar`)?

som-sm: > Maybe we should return `never` (`DelimiterCase<'fooBar', never>` -> `never`)?

I guess this makes more sense because `` `foo${never}bar` `` is `never`.  Updated behavior in [1ea1f79](https://github.com/sindresorhus/type-fest/pull/1422/commits/1ea1f796a8bfcbfe20814ce449eb37315f0aae6e).

sindresorhus: > I guess this makes more sense because `foo${never}bar` is never. Updated behavior in 1ea1f79.

👍

PR #1425, opened 2026-05-15
`Except`: Enhancing parameter‌ types

### LINKED ISSUE #1421: `Except` not playing well conditional types and/or `infer`
### Bug description

I've been using `type-fest` to replace some of the type utilities we have in our project, but when trying to use `Except` to replace our alternative version of `Omit` (which we needed because `Omit` loses known keys) I was getting errors due to the `keyof ObjectType` requirement in the keys. This doesn't seem to play well with conditional types and/or `infer`

For conditional types:
```ts
type Cond<T> =
  T extends string ? {
    name: string;
    remove: string
  }
  : T extends number ? { prime: boolean }
  : {}


type Removed<T> = Except<Cond<T>, "remove">
//                                  ^ ---> Type 'string' does not satisfy the constraint 'keyof A<T>'.
//                                          Type '"remove"' is not assignable to type '("remove" | "name") & keyof (T extends number ? { prime: boolean; } : {})'.
```

For `infer`, I've been working with the `CompnentProps` type from Vue, which is defined as:
```ts
type ComponentProps<T> =
  T extends new (...args: any) => {
    $props: infer P
  } ? NonNullable<P>
  : T extends (props: infer P, ...args: any) => any ? P
  : {}
```

Trying to use `Except` with it runs into
```ts
type ComponentPropsLessKey<T> =
  Except<ComponentProps<T>, "key">
//                          ^ ---> Type 'string' does not satisfy the constraint 'keyof ComponentProps<T>'.
//                                  Type '"key"' is not assignable to type 'never'.
```
This seems to imply that `keyof ComponentProps<T>` resolves to `never`

I have a type that attempts to get the user defined props from a Vue component, and I get a similar type error as the above when attempting to use `Except` to define it:
```ts
type ComponentPropsUser<T> = Except<
  ComponentProps<T>,
  keyof VNodeProps
  | keyof AllowedComponentProps
  | keyof ComponentCustomProps // This is an empty interface by default as it is meant for extension by users
>
// ^ ---> Type 'string' does not satisfy the constraint 'keyof ComponentProps<T>'.
//  Type 'string' is not assignable to type 'keyof ComponentProps<T>'.
//    Type '"key"' is not assignable to type 'never'.
```

This might be better resolved by a version of `Except` that would be less strict in the types of the keys it expects, but I will leave it up to the project to decide if this is an issue worth fixing.

For reference, I currently use an alternative version of `Omit` defined as:
```ts
// Taken from https://stackoverflow.com/a/65673414
type AlternativeOmit<T, K extends PropertyKey> = {
  [P in keyof T as Exclude<P, K>]: T[P]
};
```

### Repro

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBDAnmApnA3nAogDwMYpjwC+cAZlBCHAORKoC0ZKAzjDQNwCwAUKJLATI0mAMJVIAOxSSYABUpgWcUhSpwARADcArigZ4JEabIb19ACxQAbVFBYbuPXubjjJAEwA8AFQB8cAC8vHBwPnAoODAyHspsUMCSAOZwAPwYIaFwkgCGICgAXHDxiUlOWXBQKCAQWoXFMAnJmcSZReGR0Z7KkjogAEYoUGkYcGAJ+UX9EBDWKDmSKm0Yrc5rrgBK1bUo3v5B2PiEMF7ue34ANJpVNXUafrwuwm5GJvKKLAAyrCwA0iiIXwBYI8UK4AhEU6vGTvCBKIFXDQAawB90ePFc4nAxhhCjhLAAqiwhkCDuDjl5MlipLiPgjMijEBAyHAAGoAOQgHhQeKUmQAPnBGcy4ABBazWCAAd121Jxsl5LAFQoBIrlb1EOjYVEVvD8HCAA

Emiyaaaaa:
- fixed #1421

When the 2nd input param `KeysType` is clearly defined, `keyof ObjectType` can correctly constrain and suggest parameter options. However, when the input type is ambiguous due to generics, `keyof ObjectType` prevents the parameter from passing. `LiteralUnion` can relax the strict input type, and still maintain the parameter suggestions.

If there are ok, I can fix similar problems in other PRs.

### DIFF
--- a/source/except.d.ts
+++ b/source/except.d.ts
@@ -1,5 +1,6 @@
 import type {ApplyDefaultOptions} from './internal/index.d.ts';
 import type {IsEqual} from './is-equal.d.ts';
+import type {LiteralUnion} from './literal-union.d.ts';
 
 /**
 Filter out keys from an object.
@@ -100,10 +101,10 @@ type PostPayloadFixed = Except<UserData, 'email'>;
 
 @category Object
 */
-export type Except<ObjectType, KeysType extends keyof ObjectType, Options extends ExceptOptions = {}> =
+export type Except<ObjectType, KeysType extends LiteralUnion<keyof ObjectType, PropertyKey>, Options extends ExceptOptions = {}> =
 	_Except<ObjectType, KeysType, ApplyDefaultOptions<ExceptOptions, DefaultExceptOptions, Options>>;
 
-type _Except<ObjectType, KeysType extends keyof ObjectType, Options extends Required<ExceptOptions>> = {
+type _Except<ObjectType, KeysType extends LiteralUnion<keyof ObjectType, PropertyKey>, Options extends Required<ExceptOptions>> = {
 	[KeyType in keyof ObjectType as Filter<KeyType, KeysType>]: ObjectType[KeyType];
 } & (Options['requireExactProps'] extends true
 	? Partial<Record<KeysType, never>>
--- a/source/except.d.ts
+++ b/source/except.d.ts
@@ -104,7 +104,7 @@ type PostPayloadFixed = Except<UserData, 'email'>;
 export type Except<ObjectType, KeysType extends LiteralUnion<keyof ObjectType, PropertyKey>, Options extends ExceptOptions = {}> =
 	_Except<ObjectType, KeysType, ApplyDefaultOptions<ExceptOptions, DefaultExceptOptions, Options>>;
 
-type _Except<ObjectType, KeysType extends LiteralUnion<keyof ObjectType, PropertyKey>, Options extends Required<ExceptOptions>> = {
+type _Except<ObjectType, KeysType extends PropertyKey, Options extends Required<ExceptOptions>> = {
 	[KeyType in keyof ObjectType as Filter<KeyType, KeysType>]: ObjectType[KeyType];
 } & (Options['requireExactProps'] extends true
 	? Partial<Record<KeysType, never>>
--- a/test-d/except.ts
+++ b/test-d/except.ts
@@ -28,3 +28,14 @@ type Example = {
 const test: Except<Example, 'bar', {requireExactProps: false}> = {foo: 123, bar: 'asdf'};
 expectType<number>(test.foo);
 expectType<unknown>(test['bar']);
+
+// Test for generic type
+type GenericType<T> = T extends string ? {
+	foo: T;
+	bar: string;
+} : {
+	baz: T;
+};
+type GenericTypeExcept<T> = Except<GenericType<T>, 'foo'>;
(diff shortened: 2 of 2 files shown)


sindresorhus:
I think this changes the contract too much.

`Except` is documented as stricter than `Omit`: it should only allow keys that actually exist, so typos and rename mistakes are caught. With this PR, arbitrary property keys are accepted because the key constraint includes `PropertyKey`.

I agree with fixing the generic conditional type case, but I don't think the fix should drop the strict key guarantee for normal object types. That guarantee is one of the main reasons this type exists, becuase otherwise it gets much closer to `Omit`.

som-sm:
@Emiyaaaaa In cases like these, _simply_ taking an intersection would silence the error:

```ts
import type {Except} from "type-fest";

type GenericType<T> = T extends string ? {
	foo: T;
	bar: string;
} : {
	baz: T;
};

type GenericTypeExcept<T> = Except<GenericType<T>, 'foo' & keyof GenericType<T>>; // ✅ Works

type G1 = GenericTypeExcept<string>;
//=> { bar: string; }

type G2 = GenericTypeExcept<number>;
//=> { baz: number; }
```

This is better IMO because you're acknowledging that the second argument needs to be `keyof T`, and if it's not, the intersection would make it `never` and `Except<T, never>` is basically a no-op.

---

Moreover, in the above example, the compiler rightly complains because `'foo'` is _not_ a key of `{baz: T}`. If both branches had `'foo'`, the compiler would recognise this and stop complaining. Refer:

```ts
import type {Except} from "type-fest";

type GenericType<T> = T extends string ? {
	foo: T;
	bar: string;
} : {
	foo: string; // <--- Added `foo` here too
	bar: T;
};

type GenericTypeExcept<T> = Except<GenericType<T>, 'foo'>; // ✅ Works

type G1 = GenericTypeExcept<string>;
//=> { bar: string; }

type G2 = GenericTypeExcept<number>;
//=> { bar: number; }
```

Uygniqoar:
@sindresorhus @som-sm  Let's see the simpler case:
```ts
type A = { foo: string } | { bar: number };

type B = Except<A, 'foo'> // should throw error or not?
```
Is the union type included in the design of `Except`?

som-sm:
> Is the union type included in the design of `Except`?

@Uygniqoar `Except` _is_ meant to work with union types, but it _only_ allows keys that are present across all union members, like `Except<{a: 1; b: 2} | {b: 3; c: 4}, 'b'>` -> `{a: 1} | {c: 4}`.

You can use the `DistributedOmit` type in case you want to omit keys that are not present across all union members.

som-sm:
@Emiyaaaaa Shall we close this then?

PR #1429, opened 2026-05-20
Export internal types: FirstArrayElement, UnknownArrayOrTuple

### LINKED ISSUE #676: `ArrayElement` only importable from 'type-fest/source/internal'

Malaybhai11:
## Summary
Exports internal types from `type-fest/source/internal` to make them publicly accessible. This fixes issue #676.

## Changes
- Export `FirstArrayElement` from `./source/internal/array.d.ts`
- Export `UnknownArrayOrTuple` from `./source/internal/array.d.ts`

These types are useful for users who need to work with array/tuple types, similar to the existing `ArrayElement` export.

## Related
Closes #676

## Testing
- `npm run test:tsc` passes

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -146,6 +146,8 @@ export type {ArraySlice} from './source/array-slice.d.ts';
 export type {ArraySplice} from './source/array-splice.d.ts';
 export type {ArrayTail} from './source/array-tail.d.ts';
 export type {ArrayElement} from './source/array-element.d.ts';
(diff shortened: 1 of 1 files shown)


sindresorhus:
Fully AI generated PRs are not accepted. Many mistakes here too.

PR #1431, opened 2026-05-21
Add NoInfer to IsNullable and IsOptional types

### LINKED ISSUE #848: Use `NoInfer`?
https://devblogs.microsoft.com/typescript/announcing-typescript-5-4-beta/#the-noinfer-utility-type

We should review our types and see if there are any that could take advantage of it.

algojogacor:
This is an independent contribution and is not part of any organized competition or hackathon.

Summary: Adds the built-in TypeScript NoInfer<T> utility type (TS 5.4+) to IsNullable and IsOptional types, following the pattern already established in the IsAny type.

Root Cause: The IsAny type already uses NoInfer<T> to prevent unwanted type widening in the intersection check. The IsNullable and IsOptional types use Extract<T, null> and Extract<T, undefined> respectively, which can benefit from the same NoInfer<T> pattern.

Fix: Wrapped T with NoInfer<T> in the Extract calls for both IsNullable and IsOptional.

Testing: No behavioral change is expected for standard usage. The existing test suites cover both types and verify the same type-level behavior.

Closes #848

### DIFF
--- a/source/is-nullable.d.ts
+++ b/source/is-nullable.d.ts
@@ -25,6 +25,6 @@ type D = IsNullable<string | null | undefined>;
 @category Type Guard
 @category Utilities
 */
-export type IsNullable<T> = IsAny<T> extends true ? true : Extract<T, null> extends never ? false : true;
+export type IsNullable<T> = IsAny<T> extends true ? true : Extract<NoInfer<T>, null> extends never ? false : true;
 
 export {};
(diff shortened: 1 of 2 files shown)


som-sm:
Looks like a fully AI-generated PR. Also, it doesn't have any tests whatsoever and just adds `NoInfer` to two random types.

PR #1434, opened 2026-05-28
Add `Whitespace` and `WordSeparators` public types

me-saurabhkohli:
## Summary

Exposes two previously internal types from `source/internal/characters.d.ts` as part of the public API.

### `Whitespace`
A union of all 26 Unicode whitespace characters (`\t`, `\n`, `\r`, `' '`, and more). Useful for building custom `TrimStart`/`TrimEnd` template literal types.

### `WordSeparators`
A union of `'-' | '_' | Whitespace`. Mirrors the word separators used internally by `CamelCase`, `Words`, etc., letting users build their own casing/splitting utilities.

## Changes
- `source/whitespace.d.ts` — new public type with full JSDoc + `@example`
- `source/word-separators.d.ts` — new public type with full JSDoc + `@example`
- `test-d/whitespace.ts` — tsd tests (assignability, non-assignability, template literal usage)
- `test-d/word-separators.ts` — tsd tests
- `source/internal/characters.d.ts` — re-exports from new files to avoid duplication
- `index.d.ts` — exports both new types
- `readme.md` — adds both entries under the String section

All 38 tests pass.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -198,6 +198,8 @@ export type {Join} from './source/join.d.ts';
 export type {Split, SplitOptions} from './source/split.d.ts';
 export type {Words, WordsOptions} from './source/words.d.ts';
 export type {Trim} from './source/trim.d.ts';
+export type {Whitespace} from './source/whitespace.d.ts';
+export type {WordSeparators} from './source/word-separators.d.ts';
 export type {Replace, ReplaceOptions} from './source/replace.d.ts';
 export type {StringRepeat} from './source/string-repeat.d.ts';
 export type {Includes} from './source/includes.d.ts';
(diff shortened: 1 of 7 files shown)


sindresorhus:
I think the docs should be a little more precise before this becomes public API. The union is not literally “all valid Unicode whitespace characters”: it is Unicode `White_Space` plus `U+FEFF`. It also does not exactly match ECMAScript whitespace, since JS trims `U+FEFF` but not `U+0085`. Maybe describe it as the whitespace characters Type-Fest string utilities treat as whitespace. That makes the contract clearer.

The tests for the template-literal examples should use exact type assertions instead of assignability. Right now they would still pass if the result widened more than intended. It would also be good to include at least one non-ASCII whitespace case, especially `U+0085` and `U+FEFF`, since those are the interesting parts of the public contract.

sindresorhus:
Actually, this PR looks fully AI-generated. Not interested in that.

PR #1435, opened 2026-05-28
Add `RenameKeys` type

materwelonDhruv:
Renames one or more keys in an object type while preserving each value type and the `?` / `readonly` modifiers. Distributes over union object types.

`RenameKey` is the single-key form; `RenameKeys` takes a map of old-to-new names for batch renames. Useful for deriving a JSON-shape from an internal type (camelCase to snake_case), relabeling a discriminator across a tagged union, or any spot where `Merge` / `OverrideProperties` would let you change a value type but not the key itself, as examples.

I use this in my framework, [seedcord](https://github.com/seedcord/seedcord)'s, docs site, where I have a `WithDeprecationStatus` mixin that attaches a `deprecationStatus` field to entity types (classes, methods, properties). When rendering a child member of a deprecated parent, the section component needs to know the parent's status without claiming the child itself is deprecated, so I derive `WithParentDeprecationStatus = RenameKey<WithDeprecationStatus, 'deprecationStatus', 'parentDeprecationStatus'>`. Because of this, it's just one line and has no risk of the two types drifting. And renaming the source key automatically flags every downstream rename if I ever touch the mixin.

Some other examples below:

```ts
import type { RenameKey, RenameKeys } from 'type-fest';

type User = {
    readonly id: string;
    firstName: string;
    createdAt: Date;
};

type SingleRenamed = RenameKey<User, 'createdAt', 'created_at'>;
//=> {readonly id: string; firstName: string; created_at: Date}

type BatchRenamed = RenameKeys<User, { firstName: 'first_name'; createdAt: 'created_at' }>;
//=> {readonly id: string; first_name: string; created_at: Date}

type Event = { kind: 'click'; target: string } | { kind: 'submit'; target: HTMLFormElement };

type Renamed = RenameKey<Event, 'target', 'element'>;
//=> | {kind: 'click'; element: string}
//   | {kind: 'submit'; element: HTMLFormElement}
```


sindresorhus:
I think this is close, but I would not merge it with collision support as-is.

The collision behavior is unsafe. The tests currently say that renaming `a` to an existing `b` should collapse the value types into `b: number | string`, but modifiers collapse too. For example:

```ts
type Result = RenameKey<{a?: number; b: string}, 'a', 'b'>;
```

This becomes `{b?: number | string}`, so this is accepted:

```ts
const value: Result = {};
```

That means renaming an optional key over a required key accidentally makes the required key optional. I think the simpler and better contract is to not support target-key collisions, except maybe renaming a key to itself. Supporting collisions correctly would need a lot more subtle modifier rules, and I don't think the use-case is worth it.

There is also a small issue with `RenameKeys`: optional entries in the rename map are accepted but silently ignored.

```ts
type Result = RenameKeys<{a: number; b: string}, {a?: 'alpha'}>;
```

This ends up as `{a: number; b: string}`, not `{alpha: number; b: string}`. Either the map entries should be required, or optional map entries should be rejected. Silently doing nothing is suprising.

Minor: there is a new unused `eslint-disable-next-line no-lone-blocks` warning in the tests.

materwelonDhruv:
Thanks! That makes sense. I reworked it to drop collision support entirely and expanded tests with positive + negative coverage for a bunch of edge cases. The comments explain them.

Also, a consequence of this is that since the check is `KeysOfUnion`-based, a target present in another union member (or implied by a string index signature) is also rejected. I think that is good, actually.

I was also wondering about a `RenameKeyDeep` type which would be modeled on `SetRequiredDeep` + `Paths`. But that's for a later PR if you think it would be a good addition.

Re the lint: the lone blocks are intentional as each case block-scopes a reused local `Source`/`Bad` alias to keep the cases uniform. I replaced the per-line directives with a single paired file-level `eslint-disable` for `no-lone-blocks`, with a justification comment.

test-d/rename-keys.ts
```diff
@@ -0,0 +1,203 @@
+import {expectType} from 'tsd';
+import {expectTypeOf} from 'expect-type';
+import type {RenameKeys} from '../source/rename-keys.d.ts';
+
+/* eslint-disable no-lone-blocks -- Each case is wrapped in a block to scope its reused local aliases (`Source`, `Bad`). */
+
+// Basic batch rename.
+{
```
som-sm: Refer other test files to see how we do tests. Usually this is preferred:
```ts
// Basic batch rename.
expectType<{id: string; first_name: string; created_at: Date}>(
	{} as RenameKeys<{id: string; firstName: string; createdAt: Date}, {firstName: 'first_name'; createdAt: 'created_at'}>,
);

// `readonly` and optional combine and travel together to the new name.
expectType<{readonly b?: number; x: string}>({} as RenameKeys<{readonly a?: number; x: string}, {a: 'b'}>);
```

Inlining everything is fine, just provide proper line breaks.

source/rename-keys.d.ts
```diff
@@ -0,0 +1,59 @@
...
+		[Key in keyof RenameMap]: Key extends KeysOfUnion<BaseType>
+			? RenameMap[Key] extends PropertyKey
+				? RenameMap[Key] extends Exclude<KeysOfUnion<BaseType>, keyof RenameMap>
+					? never
+					: RenameMap[Key] extends RenameMap[Exclude<keyof RenameMap, Key>]
+						? never
+						: RenameMap[Key]
+				: never
+			: never;
+	},
```
som-sm: Having complex constraints makes instantiations with generic arguments difficult, like:
```ts
type Test<T, Keys extends string> = 
    RenameKeys<T, {[P in Keys]: `new_${P}`}>; // Errors, but shouldn't ideally
```

Would it be better to remove the conditional and move that logic inside the type, and treat error cases (like collisions) as no-ops? And then list all the no-op scenarios in the JSDoc, because the user anyways has to refer to the documentation to understand why there's an error, because the error message isn't useful.

Image

The error message doesn't tell the user that `'a'` cannot be renamed `'b'` because that would cause a collision with the existing `'b'`.

source/rename-key.d.ts
som-sm, line 1: I don't think we need `RenameKey` when we already have `RenameKeys`. If only a single key needs renaming, the latter can still be used. And moreover, providing two ways of doing the same thing with different APIs can cause confusion, so it's better to just have one.

som-sm (commented):
These test cases fail currently:

```ts
type T1 = RenameKeys<{a: 1; b: 2}, {a: 'x'} | {b: 'y'}>;
//=> {a: 1; b: 2}

type T2 = RenameKeys<{a: 1; b: 2}, {a: 'b' | 'c'}>; // Doesn't error, but should
//=> {b: 1 | 2; c: 1}

type T3 = RenameKeys<{[x: string]: number; a: 1}, {a: 'b'}>; // Errors, but shouldn't
```

som-sm:
A more nuanced issue:

```ts
type T = RenameKeys<{a: number; b: string}, {a: string}>;
//=> {[x: string]: number; b: string}
```

The output `{[x: string]: number; b: string}` is broken, refer:

image

This one would probably be quite hard to tackle properly.

source/rename-keys.d.ts
materwelonDhruv: I agree with the documentation. I'll add examples there. But I think the conditional, and hence the error, should stay, as no-ops are too weak a signal for something that is intentionally not supported.

test-d/rename-keys.ts
materwelonDhruv: Will do in the next commit 👍🏻

source/rename-key.d.ts
materwelonDhruv: That makes sense. Would you rather I remove `RenameKey` then?

source/rename-keys.d.ts
materwelonDhruv: Also, hmm. I see how that generic example you mentioned would be useful. Rn it errors because it can't make sure that there aren't any collisions happening here. I will look into this and try out a few things.

materwelonDhruv:
I'll work on all this feedback and set some stuff up in a TypeScript playground link so we can iterate more quickly before I push.

Ended up testing things in vscode itself...

source/rename-key.d.ts
som-sm: Yeah you can remove `RenameKey`.

materwelonDhruv: On a side note, I had a case for keeping `RenameKey`. Will write back here soon.

materwelonDhruv: The main reason I'd keep `RenameKey` is ergonomics for the common case. `RenameKey<User, 'createdAt', 'created_at'>` takes three positional arguments: source, from-key, to-key. `RenameKeys<User, {createdAt: 'created_at'}>` wraps the same rename in an object literal. For the single-key case, the positional form maps to how you'd phrase the operation in English ("rename `createdAt` to `created_at`").
  
This pair-of-surfaces pattern already exists in the library. `And<A, B>` is a one-line wrapper over `AndAll<[A, B]>` (source/and.d.ts:80), and `Or<A, B>` is the same over `OrAll<[A, B]>`. Both pairs ship together in the README and cross-reference each other with `@see`. The binary positional forms exist for the two-argument case; the variadic forms exist for the general case.

`RenameKey` can follow the same shape (this might change, I'm still testing):

```ts
export type RenameKey<BaseType, FromKey extends KeysOfUnion<BaseType>, ToKey extends PropertyKey> = 
	RenameKeys<BaseType, {[Key in FromKey]: ToKey}>;
```

So no duplicated logic as well, and a fix to RenameKeys applies to RenameKey automatically.

materwelonDhruv:
I pushed a redesign that addresses everything in the thread.

V imp. The constraint is now structural-only (`Record<PropertyKey, PropertyKey>`). All semantic validation moved into the body. This is the only constraint shape that defers cleanly under generic instantiation, which fixes the generic-test case @som-sm wrote above. The tradeoff is that a mistake when you **use** the type returns `never` from the body instead of erroring at the call site and the error surfaces when the result is consumed (assignment, indexing) rather than at the declaration. Probably better too I think per type-fest's role as a composable utility.


- **sindre's collision concern (modifier collapse via optional → required)**: dropped collision support from the constraint. A body-side check returns `never` when any target in the rename map matches a kept literal key of the source (kept = literal keys minus the ones being renamed away). Swapping keys still works because both are simultaneously renamed away.

- **sindre's optional-entry slip-through (`{a?: 'b'}` silently ignored)**: now returns `never` via `HasOptionalKeys<RenameMap>` in the body.

- **sindre's unused `no-lone-blocks` lint**: tests are rewritten in the inline `expectType<X>({} as Y)` style like the rest of the lib now.

- **som-sm's T1 (`{a: 'x'} | {b: 'y'}` doesn't distribute)**: added explicit distribution via `RenameMap extends RenameMap ? body : never`. Each member of the rename-map union runs through the body independently.

- **som-sm's T2 (`{a: 'b' | 'c'}` doesn't error)** and the related `{a: string}` case: rejected via a body-side `_AllTargetsAreSingleLiterals` check that requires each target to be a single literal `PropertyKey` (no unions, no primitives like `string`).

- **som-sm's T3 (index signature false-positive collision)**: kept-key set now excludes index signatures via `OmitIndexSignature`. A literal target onto a string-indexed object is no longer flagged; the index signature doesn't preclude adding a named property. `RenameKey<{[x: string]: number; a: 1}, 'a', 'b'>` now produces `{[x: string]: number; b: 1}` instead of erroring.

- **som-sm's generic instantiation case (`{[P in K]: \`new\_${P}\`}` errors under generic T, K)**: works now too.

- **som-sm on dropping `RenameKey`**: I'd like to keep it. It's now a one-line wrapper:

```ts
export type RenameKey<BaseType, FromKey extends KeysOfUnion<BaseType>, ToKey extends PropertyKey> = RenameKeys<
    BaseType,
    { [Key in FromKey]: ToKey }
>;
```

The pair-of-surfaces pattern already exists in type-fest: `And<A, B>` wraps `AndAll<[A, B]>` ([source/and.d.ts:80](https://github.com/sindresorhus/type-fest/blob/main/source/and.d.ts#L80)), and `Or<A, B>` wraps `OrAll<[A, B]>`. Both pairs ship together and cross-reference each other. `RenameKey` over `RenameKeys` follows the same shape. There's no duplicated logic since the wrapper delegates fully, so a fix to `RenameKeys` applies to `RenameKey` automatically. The positional form (`RenameKey<User, 'createdAt', 'created_at'>`) also reads more directly at the call site for the single-key case than the object-literal form. I also mentioned this above in the comments.

Two cases the constraint can't catch and the type doesn't reject cleanly so I short-circuit directly:

- `IsAny<BaseType>` returns `never`: matches the existing test's "can't prove safe" policy.
- `IsNever<BaseType>` returns `never`: needed because `KeysOfUnion<never>` evaluates to `string | number | symbol` (not `never`), so the natural constraint doesn't reject `never` BaseType.

materwelonDhruv:
If I were to open a PR that sets up husky to run tests before push and lint-staged before commit, would that be merged? I feel that should definitely be there in a library like this.

som-sm:
> If I were to open a PR that sets up husky to run tests before push and lint-staged before commit, would that be merged? I feel that should definitely be there in a library like this.

No, I don’t think we want to add that. These tools add unnecessary friction, and blocking commits tends to make contributing more annoying than helpful. Contributors should be able to run checks locally when it makes sense, with CI checks acting as the final gate.

@sindresorhus

source/rename-key.d.ts
som-sm: I still don't think it makes sense to keep `RenameKey`. `RenameKey`/`RenameKeys` is not the same as `Or`/`OrAll`. There we cannot remove `Or` and just have `OrAll` because `Or` is a pretty standard type, so we need to keep it. But that's not the case here, `RenameKey` doesn't need to be present.

And, in terms of ergonomics, I think `RenameKeys<User, {createdAt: 'created_at'}>` is just about as good as `RenameKey<User, 'createdAt', 'created_at'>`.

sindresorhus:
Yeah, I don't want that. I had those in the past in other projects and caused more problems than they saved.

som-sm:
Few thoughts:

- I think returning input object as-is is better than returning `never` in case of collisions.
- And, maybe it's better to just ignore keys that are not present in the input object. This pattern is kinda common, like `Omit<{a: 1; b: 2}, 'a' | 'c'>` returns back `{b: 2}`, ignoring `'c'`. 

IMO, returning `never` doesn't feel great!

materwelonDhruv:
Hmmmmm. I do agree that `never` doesn't feel great as the result of a type. But, I think missing keys and collisions are different cases as well.

For missing keys (`{nme: 'fullName'}` on `{name, id}`) `Omit` style ignore is fine, ye. Like what I was doing before. Error when using the type itself via the type params.

For collisions tho (`{a: 'b'}` when `b` is kept), it should return something that the user HAS to pay attention to. The map said "overwrite `b` with the type of `a` under the name `b`", and returning the input as-is silently doesn't do that. That's a false-success I feel. Same goes for duplicate targets (`{a: 'x'; b: 'x'}`) where a pass through would have to pick one arbitrarily?

Maybe we go lenient on missing keys like Omit, and keep the never, so strict, on collisions, duplicate targets, basically anything that is "no this should not be allowed" vs "there isn't really a problem with not doing anything about this"

som-sm:
@materwelonDhruv How hard do you think it would be to handle collisions?

I believe these are the only cases we need to consider when multiple properties `p1, p2, …, pn` are all being renamed to the same key `new_p`:

1. **Optionality is uniform (all required or all optional)**

    The value of `new_p` is a union of the values of `p1, p2, …, pn`, and the optional modifier is preserved.
    
    ```ts
    // All required → required
    type T = RenameKeys<{p1: string; p2: number; p3: bigint}, {p1: 'new_p'; p2: 'new_p'; p3: 'new_p'}>;
    //=> {new_p: string | number | bigint}
    ```
    
    ```ts
    // All optional → optional
    type T = RenameKeys<{p1?: string; p2?: number; p3?: bigint}, {p1: 'new_p'; p2: 'new_p'; p3: 'new_p'}>;
    //=> {new_p?: string | number | bigint}
    ```

2. **Mixed optionality (at least one required, at least one optional)**

    If at least one property is required, then `new_p` should be **required**. And it's value is a union of the values of `p1, p2, …, pn`.
    
    Note: In this case, the `exactOptionalPropertyTypes` compiler flag affects the result. When disabled, `new_p`'s value should include an additional `undefined`, refer:
    
    ```ts
    // exactOptionalPropertyTypes: true
    type T = RenameKeys<{p1: string; p2?: number; p3: bigint}, {p1: 'new_p'; p2: 'new_p'; p3: 'new_p'}>;
    //=> {new_p: string | number | bigint}
    ```

    ```ts
    // exactOptionalPropertyTypes: false
    type T = RenameKeys<{p1: string; p2?: number; p3: bigint}, {p1: 'new_p'; p2: 'new_p'; p3: 'new_p'}>;
    //=> {new_p: string | number | bigint | undefined}
    ```

3. **Mixed `readonly` modifiers**

    If at least one source property is `readonly`, we could go either way on the result. I'd lean towards making `new_p` **`readonly`**; it's the safer, more conservative choice.
    
    ```ts
    type T = RenameKeys<{p1: string; readonly p2: number; p3: bigint}, {p1: 'new_p'; p2: 'new_p'; p3: 'new_p'}>;
    //=> {readonly new_p: string | number | bigint}
    ```

materwelonDhruv:
@som-sm just to make sure I'm reading this right. You want these to be valid and produce a merged result, not return `never`? The current impl rejects duplicate targets like `{p1: 'new_p'; p2: 'new_p'}` outright, so I want to confirm you're asking me to drop that check and produce the merge instead. I'll try to figure that out then.

Do you think that has valid use cases? This is basically imitating a merge, right? Also, are you sure this being allowed is a good idea? We should guard against this kind of rename right? Why would someone want to do that? Why **should** someone want to rename keys this way?

som-sm:
> so I want to confirm you're asking me to drop that check and produce the merge instead.

Yup, correct!

> This is basically imitating a merge, right?

Yeah kinda, just that merge is limited to only two items, this can happen with any no. of items. 

<br>

There could be several valid use cases for merging. Here are some AI-suggested ones that look pretty reasonable:
```ts
type SearchInput = {
  textQuery: string;
  voiceQuery: Blob;
  imageQuery: File;
};

type Normalized = RenameKeys<SearchInput, {textQuery: 'query'; voiceQuery: 'query'; imageQuery: 'query'}>;
//=> {query: string | Blob | File}

type LegacyUser = {
  oldId?: string;
  numericId: number; 
  name: string;
};

type Migrated = RenameKeys<LegacyUser, {oldId: 'id'; numericId: 'id'}>;
//=> {id: string | number; name: string}
```

materwelonDhruv:
oki imma try.

materwelonDhruv:
We just came full circle lmao. Because I think @sindresorhus explicitly asked to reject such collisions where we end up in a union 😅

Some tests that currently reject this behavior will change now.

som-sm:
> The collision behavior is unsafe. The tests currently say that renaming `a` to an existing `b` should collapse the value types into `b: number | string`, but modifiers collapse too. For example:
> 
> ```ts
> type Result = RenameKey<{a?: number; b: string}, 'a', 'b'>;
> ```
> 
> This becomes `{b?: number | string}`, so this is accepted:
> 
> ```ts
> const value: Result = {};
> ```
> 
> That means renaming an optional key over a required key accidentally makes the required key optional. I think the simpler and better contract is to not support target-key collisions, except maybe renaming a key to itself. Supporting collisions correctly would need a lot more subtle modifier rules, and I don't think the use-case is worth it.

This was an issue because the merged result was incorrect; it would have been fine if the result were `{b: number | string}`. 

And also, there aren't many cases here, so I guess it's worth solving it!

<br>

> Some tests that currently reject this behavior will change now.

That's fine!

materwelonDhruv:
okay that makes sense. I'm nearly done with it. testing with the cases you gave and some others of my own. and also simplifying the implementation a bit.

test-d/rename-keys.ts
```diff
@@ -0,0 +1,118 @@
...
+expectType<{[x: string]: number; b: 1}>(
+	{} as RenameKeys<{[x: string]: number; a: 1}, {a: 'b'}>,
+);
+
+// Swapping two keys is fine: both keys rename away simultaneously, so no
+// kept-key collision occurs.
+expectType<{b: 1; a: 2}>({} as RenameKeys<{a: 1; b: 2}, {a: 'b'; b: 'a'}>);
+
+// Renaming to itself is a no-op (the only "collision" allowed: the key is
+// renaming away, not into a kept property).
```
som-sm: Update the comment text.

source/rename-key.d.ts
som-sm, line 1: Remove this, refer https://github.com/sindresorhus/type-fest/pull/1435#discussion_r3330668419.

test-d/rename-keys.ts
```diff
@@ -0,0 +1,118 @@
...
+expectType<{a: 1; b: 2}>({} as RenameKeys<{a: 1; b: 2}, {a: 'a'}>);
+
+// Typo'd source key is ignored, matching `Omit`'s behavior on missing keys.
+expectType<{a: 1; b: 2}>({} as RenameKeys<{a: 1; b: 2}, {nme: 'fullName'}>);
+
+// Optional rename-map entry returns `never`. Without this, an optional source
+// key could silently make a required target key optional.
+expectType<never>({} as RenameKeys<{a: number; b: string}, {a?: 'alpha'}>);
```
som-sm: Returning `never` is not ideal, just wrap the map in `Required` and ignore the optional modifier.
```ts
expectType<{alpha: number; b: string}>({} as RenameKeys<{a: number; b: string}, {a?: 'alpha'}>);
```

source/rename-keys.d.ts
```diff
@@ -0,0 +1,131 @@
...
+	{
+		[Key in keyof RenameMap]: _IsSingleLiteral<RenameMap[Key]>;
+	}[keyof RenameMap],
+] extends [true]
+	? true
+	: false;
+
+// With EOPT off, the optional modifier implicitly admits `undefined`, so
+// `{a: undefined}` is assignable to `{a?: string}`. With EOPT on it isn't.
+type _IsExactOptionalDisabled = {a: undefined} extends {a?: string} ? true : false;
```
som-sm: We already have a type for this, reuse that!

som-sm (changes requested):
This currently fails:
```ts
type T = RenameKeys<{a: string}, {a: 'b' | 'c'}>;
//=> Current: never
//=> Expected: {b: string; c: string}
```

materwelonDhruv:
> type T = RenameKeys<{a: string}, {a: 'b' | 'c'}>;

still working on this

som-sm (changes requested):
The current implementation fails in this case:
```ts

type T1 = RenameKeys<{a?: number; b: string}, {a: 'x'; b: 'x'}>;
//=> Current: {x?: string | number}
//=> Expected: {x: string | number}
```

<br>

The current implementation doesn't explicitly handle optional properties. Instead, it relies on TS's default behaviour, which isn't sufficient in this case because by default, if multiple keys are being aliased to the same key, TS picks up the modifier of the first key:

```ts
type Test<BaseType extends object> =
	{
		[Key in keyof BaseType as 'z']: Required<BaseType>[Key];
	};

type T1 = Test<{a: number; b: string; c?: bigint}>;
//=> { z: string | number | bigint; }

type T2 = Test<{a: number; b?: string; c: bigint}>;
//=> { z: string | number | bigint; }

type T3 = Test<{a?: number; b: string; c: bigint}>;
//=> { z?: string | number | bigint; }
```

materwelonDhruv:
I did a fix for this for readonly because it was only picking up the first key's modifier by default. hmmmm. need to do something similar for optional as well. I might have to strip optional while renaming and reapply or something. i'll try some things out and add a few more tests

materwelonDhruv:
found another bug. gonna fix it

```ts
RenameKeys<{[x: string]: number; a?: 1}, {a: 'b'}>
// Expected. {[x: string]: number; b?: 1}
// Actual bugged: {[x: string]: number; b: 1}
```

som-sm (changes requested):
Doesn't seem to add `| undefined` when `exactOptionalPropertyTypes` is disabled, refer:

```ts
type T1 = RenameKeys<{a?: 1; b: 2}, {b: 'x'; a: 'x'}>;
// Current: {x: 1 | 2}
// Expected: {x: 1 | 2 | undefined}

type T2 = RenameKeys<{a?: 1; b: 2}, {b: 'a'}>;
// Current: {a: 1 | 2}
// Expected: {a: 1 | 2 | undefined}

type T3 = RenameKeys<{a?: 1; b: 2}, {a: 'b'}>;
// Current: {b: 1 | 2}
// Expected: {b: 1 | 2 | undefined}
```

materwelonDhruv:
Fixed. The `-?` that forces a renamed key to required was also stripping the `| undefined` an option key carries. So when an optional key and a required key merged into the same name, the merged value lost its undefined. It only shows when `exactOptionalPropertyTypes` is off (with it on, there's no `undefined` to keep).

All three of your cases should now return the correct `| undefined` when EOPT is off. But, I wasn't able to add tests for this because the repo has it on. I tried adding a folder inside the tsd directory with its own tsconfig extending the base and setting EOPT off, but tsd doesn't care about nested folders with their own configs :( what to do? How do I add tests for this? @som-sm 

I'm gonna check if some of my new impl can be replaced by any other type-fest types first before I commit.

And one other thing. Optional jeys show an explicit `| undefined` under EOPT off, e.g. `b?: 1 | undefined` where just `b?: 1` would be enough. It's the same type, but not as clean, but it's explicit. It comes from `_NewValue`'s off-branch. Since `_RestoreMergedUndefined` now adds `undefined` where it's actually needed, `_NewValue` could prolly be simplified to always strip, which would get the `b?` thingy. Not really needed tho. Wdyt?

materwelonDhruv:
I can add a folder outside test-d/ for EOPT off tests. But that's outside the files this PR is currently about. So better to ask first

som-sm:
> How do I add tests for this?

It's fine, you don't need to add tests for `exactOptionalPropertyTypes`, you can just verify it manually!

> And one other thing. Optional jeys show an explicit `| undefined` under EOPT off, e.g. `b?: 1 | undefined` where just `b?: 1` would be enough.

That's fine too, with `exactOptionalPropertyTypes` disabled, `b?: 1` and `b?: 1 | undefined` are the exact same thing.

materwelonDhruv:
I already verified it manually 👍🏻 so it should be good on that front.

test-d/rename-keys.ts
```diff
@@ -0,0 +1,202 @@
...
+
+// Cross-union collision merges per union member.
+expectType<{b: number} | {b: string}>({} as RenameKeys<{a: number} | {b: string}, {a: 'b'}>);
+
+// Duplicate targets across the same rename map merge into a union.
+expectType<{x: 1 | 2; c: 3}>({} as RenameKeys<{a: 1; b: 2; c: 3}, {a: 'x'; b: 'x'}>);
+
+// Non-literal targets return `never`. A `string`-typed target would widen to
+// an index signature.
+expectType<never>({} as RenameKeys<{a: number; b: string}, {a: string}>);
```
som-sm: This shouldn't be `never`, just ignore entries in `RenameMap` that map to a non-literal. So, something like `RenameKeys<{a: number; b: string}, {a: string; b: 'c'}>` should return `{a: number; c: string}`. Also, add a note regarding this in the JSDoc.

Also, add tests for few more non-literals like `Uppercase<string>`, `` `foo${string}` ``, etc.

test-d/rename-keys.ts
```diff
@@ -0,0 +1,202 @@
...
+// Index signature is preserved when the only renamed source is optional. The
+// index sig key must not affect the target's modifier routing.
+expectType<{[x: string]: number; b?: 1}>(
+	{} as RenameKeys<{[x: string]: number; a?: 1}, {a: 'b'}>,
+);
+
+// Three-way swap with the optional source in the middle.
+expectType<{b: 1; c?: 2; a: 3}>(
+	{} as RenameKeys<{a: 1; b?: 2; c: 3}, {a: 'b'; b: 'c'; c: 'a'}>,
+);
```
som-sm, line 234: Umm... this is an interesting case. This kinda suffers from the problem that was mentioned in https://github.com/sindresorhus/type-fest/pull/1435#issuecomment-4576023954, `c` was required in the source object, but is optional in the target object.

A more simpler version of this is:
```ts
type T = RenameKeys<{a: 1; b?: 2}, {a: 'b'; b: 'a'}>;
```

The answer for this is currently `{a?: 2; b: 1}`, but should it be `{a: 2; b: 1}` instead?

som-sm (changes requested):
I have not reviewed the implementation yet, I'll review that once everything else is sorted.

test-d/rename-keys.ts
materwelonDhruv: For this the current implementation makes sense in my opinion.

Because you are renaming a key, in this case `b`, which is optional. And you are doing it deliberately. Since you are only **renaming**, the modifiers that were on the key you rename move to the new key name you pass. 

`RenameKeys` should not affect the modifiers that were on the original key that you are renaming as that is a side effect. And type-fest already has other types that do that kind of modification for you. This should be a pure rename with no side effects.

What you're saying would tie the modifier to the name rather than the property, which has no basis in a rename and would contradict the merge rule. Your original safety worry (from the earlier comment) was an optional key clobbering a required one and silently making it optional. That can't happen here, no required property is being overwritten, `a` is a fresh property from renaming the optional b.

test-d/rename-keys.ts
materwelonDhruv: done

materwelonDhruv:
bump @som-sm :)

source/rename-keys.d.ts
som-sm, line 1: The current implementation seems quite complicated and there’s definitely room for simplification.

Gave it a quick try:
```ts
import type {IsLiteral} from './is-literal.d.ts';
import type {ReadonlyKeysOf} from './readonly-keys-of.d.ts';
import type {RequiredKeysOf} from './required-keys-of.d.ts';
import type {IfNotAnyOrNever, IsExactOptionalPropertyTypesEnabled} from './internal/type.d.ts';
import type {SetRequired} from './set-required.d.ts';
import type {SetReadonly} from './set-readonly.d.ts';
import type {OmitIndexSignature} from './omit-index-signature.d.ts';

export type RenameKeys<
	Type extends object,
	RenameMap extends Record<PropertyKey, PropertyKey>,
> =
	IfNotAnyOrNever<Type,
		Type extends unknown // For distributing `Type`
			? RenameMap extends unknown // For distributing `RenameMap`
				? _RenameKeys<Type, NormalizeRenameMap<RenameMap>>
				: never
			: never>;

type NormalizeRenameMap<RenameMap extends Record<PropertyKey, PropertyKey>> =
	{
		-readonly [Key in keyof RenameMap as true extends IsLiteral<RenameMap[Key]> ? Key : never]-?: RenameMap[Key]
	};

type _RenameKeys<
	Type extends object,
	RenameMap extends Record<PropertyKey, PropertyKey>,
> =
	HandleReadonliness<Type, RenameMap,
		HandleOptionality<Type, RenameMap, {
			[Key in keyof Type as Key extends keyof RenameMap
				? RenameMap[Key] extends PropertyKey
					? RenameMap[Key]
					: Key
				: Key]: (IsExactOptionalPropertyTypesEnabled extends true ? Required<Type> : Type)[Key]
		}>>;

type HandleOptionality<Type extends object, RenameMap extends Record<PropertyKey, PropertyKey>, RenamedType> =
	SetRequired<RenamedType, TargetKey<RequiredKeysOf<OmitIndexSignature<Type>>, RenameMap> & keyof RenamedType>;

type HandleReadonliness<Type extends object, RenameMap extends Record<PropertyKey, PropertyKey>, RenamedType> =
	SetReadonly<RenamedType, TargetKey<ReadonlyKeysOf<OmitIndexSignature<Type>>, RenameMap> & keyof RenamedType>;

type TargetKey<SourceKey, RenameMap extends Record<PropertyKey, PropertyKey>> =
	SourceKey extends keyof RenameMap
		? RenameMap[SourceKey]
		: SourceKey;

export {};
```
This passes all the existing test cases and is clearly much simpler.

Review the solution and see if there are any issues here. Add test cases against those issues and see if those issues can be fixed while keeping the implementation simpler.

test-d/rename-keys.ts
```diff
@@ -0,0 +1,258 @@
...
+
+// Three-member union of rename maps distributes correctly.
+expectType<{x: 1; b: 2; c: 3} | {a: 1; y: 2; c: 3} | {a: 1; b: 2; z: 3}>(
+	{} as RenameKeys<{a: 1; b: 2; c: 3}, {a: 'x'} | {b: 'y'} | {c: 'z'}>,
+);
+
+// `any` source returns `never`. The target key cannot be proven absent from
+// an `any` shape.
+expectType<never>({} as RenameKeys<any, {a: 'x'}>);
```
som-sm:
```suggestion
expectType<any>({} as RenameKeys<any, {a: 'x'}>);
```

`any` should yield `any`.

test-d/rename-keys.ts
```diff
@@ -0,0 +1,258 @@
...
+
+// Generic instantiation. The constraint is structural-only, so the type can
+// be used inside other generics without erroring at the definition site.
+type WrappedGeneric<T, K extends keyof T & string> = RenameKeys<T, {[P in K]: `new_${P}`}>;
```
som-sm:
```suggestion
type WrappedGeneric<T extends object, K extends keyof T & string> = RenameKeys<T, {[P in K]: `new_${P}`}>;
```

som-sm (changes requested):

test-d/rename-keys.ts
som-sm: A fully AI-generated response doesn’t add anything meaningful to the discussion.

A stronger argument for keeping the current behavior is that a key changing from required to optional is only a problem when the change is unintentional. If the change is intentional, that should be fine.

For example, even in a simple case like `RenameKeys<{a: string}, {a: 'b'}>`, the key `'a'` was required before the rename but is no longer required afterward because it is gone entirely. That is acceptable because the change is intentional.

materwelonDhruv: That's not AI generated lmao. I genuinely believe it should not change the optionality and that was me justifying it. What makes it sound like AI anyway? I thought you're asking me to either justify that current behavior, or change it, which is why I wrote that much. 

Either way. Both you and I agree to keep it. Which is great! And the `{a: 'b'}` already works.  And the current behavior which I think is correct: `RenameKeys<{a?: string}, {a: 'b'}>`. For this, it would end up as `{b?: string}`.

materwelonDhruv:
I'm testing that simplified version right now btw. Will reply to your review comments in a bit.

test-d/rename-keys.ts
materwelonDhruv: This is nice. Rejects primitives and unions without an object at the call site.

source/rename-keys.d.ts
materwelonDhruv: Wow. Just one problem with this though. It regressed the EOPT-off `| undefined` fixed earlier. Now, with this, even with EOPT-off you don't get the correct type. Everything else works! Very cool.

```ts
RenameKeys<{a?: 1; x: 2}, {a: 'x'}>
// EOPT off -> should be {x: 2 | 1 | undefined}, this gives {x: 2 | 1}
```

The `Required<Type>` in the value plus `SetRequired`'s `-?` strip the undefined.

I can keep the `_RestoreMergedUndefined` from earlier for the fix (with some tweaks).

Edit: I'll need to add two....
Edit 2: Never mind. I can refactor it a bit more.

materwelonDhruv:
Posted a comment above @som-sm 
I can push with that additional helper for the EOPT fix if you're good with it.

som-sm:
> I can push with that additional helper for the EOPT fix if you're good with it.

Yeah, sure!

materwelonDhruv:
OKAY. I think I'm done. Gonna push now. Also added like 3 more tests.

materwelonDhruv:
I'm trying one other thing. If it's straightforward to get autocomplete on the second arg's map for the target keys. If it's too complex, it won't be worth it.

Edit: Not worth it.

materwelonDhruv:
I like how this started with this 😭:
```ts
export type RenameKey<
	BaseType,
	FromKey extends KeysOfUnion<BaseType>,
	ToKey extends PropertyKey,
> = {
	[Key in keyof BaseType as Key extends FromKey ? ToKey : Key]: BaseType[Key];
};
```
 Nice PR review. V good learning experience. Thanks som.

materwelonDhruv:
Maybe worth it to add CodeRabbit and Greptile code reviews to the repo? Both are free for OSS.

source/rename-keys.d.ts
```diff
@@ -0,0 +1,111 @@
...
+type _RenameOnce<BaseType extends object, RenameMap extends Record<PropertyKey, PropertyKey>> =
+	_RestoreMergedUndefined<BaseType, RenameMap,
+		_ApplyReadonly<BaseType, RenameMap,
+			_ApplyRequired<BaseType, RenameMap,
+				_RenameNaive<BaseType, RenameMap>>>>;
+
+type _RenameNaive<BaseType extends object, RenameMap extends Record<PropertyKey, PropertyKey>> = {
+	// Two keys mapping to one target produce a union value and keep only the first key's modifiers
+	// Like for example, {a?: 1; b: 2} with {a: 'x'; b: 'x'} produces {x?: 1 | 2}, taking a's optional
+	[Key in keyof BaseType as _TargetOf<Key, RenameMap>]: _SourceValue<BaseType, Key>;
```
som-sm: Can this not directly be the following given that we now separately handle `IsExactOptionalPropertyTypesEnabled`?

```suggestion
	[Key in keyof BaseType as _TargetOf<Key, RenameMap>]: Required<BaseType>[Key];
```

materwelonDhruv: Ye. Doing that rn.

materwelonDhruv:
bump @som-sm :)

materwelonDhruv:
bump @som-sm 🙏🏻

som-sm (approved):
PR LGTM, nice work!!

Apologies for delay, was occupied with something else.

materwelonDhruv:
🥳 Thanks a lot Som!

PR #1436, opened 2026-05-29
feat: add RemoveSuffix type

### LINKED ISSUE #1433: proposal: remove suffix

similar to this type: https://github.com/sindresorhus/type-fest/blob/main/source/remove-prefix.d.ts
but for suffix

### Type source

_No response_

xxiaoxiong:
Add `RemoveSuffix` type that removes a specified suffix from the end of a string type, mirroring the existing `RemovePrefix` type.

Closes #1433

### DIFF
--- a/source/remove-suffix.d.ts
+++ b/source/remove-suffix.d.ts
@@ -1,0 +1,133 @@
+import type {ApplyDefaultOptions} from './internal/object.d.ts';
+import type {IfNotAnyOrNever, Not} from './internal/type.d.ts';
+import type {IsStringLiteral} from './is-literal.d.ts';
+import type {IsNever} from './is-never.d.ts';
+import type {Or} from './or.d.ts';
+import type {If} from './if.d.ts';
+
+/**
+@see {@link RemoveSuffix}
+*/
+export type RemoveSuffixOptions = {
+	/**
+	When enabled, instantiations with non-literal suffixes (e.g., `string`, `Uppercase<string>`, `` `${string}` ``) simply return `string`, since their precise structure cannot be statically determined.
+
+	Note: Disabling this option can produce misleading results that might not reflect the actual runtime behavior.
+	For example, ``RemoveSuffix<'on-change', `-${string}`, {strict: false}>`` returns `'on-'`, but at runtime, suffix could be `'-error'` (which satisfies `` `-${string}` ``) and removing `'-error'` from `'on-change'` would not result in `'on-'`.
+
+	So, it is recommended to not disable this option unless you are aware of the implications.
+
+	@default true
+
+	@example
+	```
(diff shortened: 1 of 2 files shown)


source/remove-suffix.d.ts
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+@see {@link RemoveSuffix}
+*/
+export type RemoveSuffixOptions = {
+	/**
+	When enabled, instantiations with non-literal suffixes (e.g., `string`, `Uppercase<string>`, `` `${string}` ``) simply return `string`, since their precise structure cannot be statically determined.
+
+	Note: Disabling this option can produce misleading results that might not reflect the actual runtime behavior.
+	For example, ``RemoveSuffix<'on-change', `-${string}`, {strict: false}>`` returns `'on-'`, but at runtime, suffix could be `'-error'` (which satisfies `` `-${string}` ``) and removing `'-error'` from `'on-change'` would not result in `'on-'`.
```
som-sm, line 16: Should be `'on'` instead of `'on-'`.

```suggestion
	Note: Disabling this option can produce misleading results that might not reflect the actual runtime behavior.
	For example, ``RemoveSuffix<'on-change', `-${string}`, {strict: false}>`` returns `'on'`, but at runtime, suffix could be `'-error'` (which satisfies `` `-${string}` ``) and removing `'-error'` from `'on-change'` would not result in `'on'`.
```

source/remove-suffix.d.ts
```diff
@@ -0,0 +1,133 @@
...
+
+	@example
+	```
+	import type {RemoveSuffix} from 'type-fest';
+
+	type A = RemoveSuffix<'on-change', `-${string}`, {strict: true}>;
+	//=> string
+
+	type B = RemoveSuffix<'on-change', `-${string}`, {strict: false}>;
+	//=> 'on-'
```
som-sm, line 30:
```suggestion
	//=> 'on'
```

source/remove-suffix.d.ts
```diff
@@ -0,0 +1,133 @@
...
+	//=> string
+
+	type B = RemoveSuffix<`on-${string}`, '-change', {strict: false}>;
+	//=> string
+
+	type C = RemoveSuffix<`id-${number}`, '-id', {strict: true}>;
+	//=> `${number}`
+
+	type D = RemoveSuffix<`id-${number}`, '-id', {strict: false}>;
+	//=> `${number}`
```
som-sm, line 61: These examples are incorrect and also don't make sense.

som-sm (commented):
This looks like a fully AI generated PR, so many mistakes in the JSDoc examples itself.

sindresorhus:
Closing for being fully AI-generated.

PR #1438, opened 2026-05-30
`RemovePrefix`: Remove incorrect optimisation and always fallback to `string` for non-literal prefixes

som-sm:
> Note: If it can be statically determined that the input string can never start with the specified non-literal prefix, then the input string is returned as-is, regardless of the value of this option.
For example, ``RemovePrefix<`${string}/${number}`, `${string}:`>`` returns `` `${string}/${number}` ``, since a string of type `` `${string}/${number}` `` can never start with a prefix of type `` `${string}:` ``.

The above note in `RemovePrefix` is actually incorrect, a string of type `` `${string}/${number}` `` _can actually_ start with a prefix of type `` `${string}:` ``.

For example, currently ``RemovePrefix<`on${string}`, `${string}B`>`` returns `` `on${string}` ``, but this wouldn't always align with the runtime behavior, like removing `'onB'` from `'onBounce'` would result in `'ounce'`, which isn't of type `` `on${string}` ``.

Hence, this PR simply removes this optimisation and updates the behavior to always fallback to `string` in `strict` mode if the prefix is a non-literal string type.

### DIFF
--- a/source/remove-prefix.d.ts
+++ b/source/remove-prefix.d.ts
@@ -109,7 +109,7 @@ type _RemovePrefix<S extends string, Prefix extends string, Options extends Requ
 			? S extends `${Prefix}${infer Rest}`
 				? Rest
 				: S // Return back `S` when `Prefix` is not present at the start of `S`
-			: string // Fallback to `string` when `Prefix` is non-literal and `strict` is disabled
+			: string // Fallback to `string` when `Prefix` is non-literal and `strict` is enabled
 		: never;
 
 export {};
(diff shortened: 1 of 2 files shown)


som-sm:
AI review made a suggestion that we should return the input type as-is (instead of `string` fallback) if:

1. The input is a non-literal, and 
2. The prefix isn't found in the input.

For example, `RemovePrefix<'onClick', Uppercase<string>>` should return `'onClick'`. 

While this would be correct, I think it's better to simply fallback to `string` and keep things consistent/simple. And, we already have non-strict mode if anyone really wants that behaviour.

PR #1443, opened 2026-06-02
Add `OptionalPaths` type

### LINKED ISSUE #1426: `OmitOptional` and `OmitOptionalDeep`: Removes non-required keys from an object.

```ts
type Config = { required: string, optional?: string }

// This one already exists
type FullConfig = Required<Config> // -> { required: string, optional: string }

// So, I'm suggesting to make a complementary one
type MinimalConfig = OmitOptional<Config> // -> { required: string }
```

Our usecase is that we're writing tests for our configuration parsing, so we want to test out cases where we give the function the minimum possible configuration, and assert what's computed from it. We'd love to have type enforcement that we're not adding extra keys in this "MinimalConfig" testing.

- Similar to #462 

### Type source

We're currently using this

https://github.com/wasp-lang/wasp/blob/5398d01fca789bf6c42979155d4f38bcd09d92f1/waspc/data/packages/wasp-config/__tests__/spec/testFixtures.ts#L536-L562

```ts
/**
 * Recursively strips optional properties from T.
 * - Branded types are passed through (don't recurse into the brand).
 * - Arrays recurse into element type.
 * - Objects keep only required keys (collapse to EmptyObject when none remain).
 * - Primitives pass through.
 */
export type MinimalConfig<T> =
  T extends Branded<unknown, unknown>
    ? T
    : T extends Array<infer Item>
      ? Array<MinimalConfig<Item>>
      : T extends object
        ? keyof T extends never
          ? EmptyObject
          : MinimalConfigObject<T>
        : T;

type MinimalConfigObject<T> = {
  [K in keyof T as EmptyObject extends Pick<T, K> ? never : K]: MinimalConfig<
    T[K]
  >;
} extends infer Result
  ? Result extends EmptyObject
    ? EmptyObject
    : Result
  : never;
```

som-sm:
Closes #1426

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -177,6 +177,7 @@ export type {ArrayReverse} from './source/array-reverse.d.ts';
 export type {UnionMember} from './source/union-member.d.ts';
 export type {Absolute} from './source/absolute.d.ts';
 export type {UnionLength} from './source/union-length.d.ts';
+export type {OptionalPaths, OptionalPathsOptions} from './source/optional-paths.d.ts';
 
 // Template literal types
 export type {CamelCase, CamelCaseOptions} from './source/camel-case.d.ts';
--- a/readme.md
+++ b/readme.md
@@ -188,6 +188,7 @@ Click the type names for complete docs.
 - [`ArrayTail`](source/array-tail.d.ts) - Extract the type of an array or tuple minus the first element.
 - [`SetFieldType`](source/set-field-type.d.ts) - Create a type that changes the type of the given keys.
 - [`Paths`](source/paths.d.ts) - Generate a union of all possible paths to properties in the given object.
+- [`OptionalPaths`](source/optional-paths.d.ts) - Get a union of all paths to optional properties in the given object.
 - [`SharedUnionFields`](source/shared-union-fields.d.ts) - Create a type with shared fields from a union of object types.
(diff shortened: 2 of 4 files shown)


sindresorhus:
the new type says it can be used with `PickDeep`/`OmitDeep`, but that does not hold for all of the documented output. For example:

```ts
type Tuple = [string, number?];
type Result = OmitDeep<Tuple, OptionalPaths<Tuple>>;
```

`OptionalPaths<Tuple>` includes `1 | '1'`, while the deep path helpers only accept the string tuple path here. So the numeric member makes the composition fail. I think the simplest fix is to make the public result only produce paths that are accepted by the existing deep helpers, instead of returning extra numeric tuple variants.

There is also a generic version of the same problem, which is close to the original use case from the issue:

```ts
type MinimalObject<T extends object> = OmitDeep<T, OptionalPaths<T>>;
```

That currently fails the path constraint. `Extract<OptionalPaths<T>, Paths<T>>` makes the object case work, but users should not need that if the documented use case is direct composition with `OmitDeep`.

and I would add tests for:

```ts
type RequiredUserFields = OmitDeep<UserFields, OptionalPaths<UserFields>>;
type OptionalUserFields = PickDeep<UserFields, OptionalPaths<UserFields>>;
type MinimalObject<T extends object> = OmitDeep<T, OptionalPaths<T>>;
```

And if array support stays in scope, also add a tuple composition test. Right now the standalone `OptionalPaths` tests are pretty thorough, but they miss whether the type actually compsoes with the helpers it is meant to feed into.

PR #1446, opened 2026-06-03
Add `StringToNumber` type

som-sm:
Exposes the internal `StringToNumber` type ~~with some improvements around edge cases~~.

I initially had added support for handling two out of the many cases mentioned in [https://github.com/microsoft/TypeScript/issues/57404](https://github.com/microsoft/TypeScript/issues/57404).

* Fractionals ending in `0` were handled, so `StringToNumber<'1.50'>` would have returned `1.5`.
* Literals with separators were handled, so `StringToNumber<'123_45'>` would have returned `12345`.

But I think that's not correct. There's actually _no_ number equivalent for strings like `'1.50'`, or `'123_45'` as mentioned in https://github.com/microsoft/TypeScript/issues/57404#issuecomment-2138494542.

Here's a quick breakdown of why I think those should _not_ be supported.

# Context

```ts
type ToNumber<T extends string> = T extends `${infer N extends number}`
  ? N
  : never;

```

When you instantiate a type like `ToNumber` with strings such as `'1.50'`, `'0b10'`, or `'1e2'`, you'd get back the generic `number` type rather than a specific number literal.

This happens because TS enforces a **round-trip constraint** when inferring number literals from string literals. Meaning, the inferred number literal, when converted back to a string, must exactly match the original string.

For example, `ToNumber<'123'>` returns `123`. If you convert `123` back to a string (`` `${123}` ``), it produces `'123'`, perfectly matching the original input.

However, for a string like `'1.50'`, this round-trip fails. There is *no* number literal whose string form is exactly `'1.50'`. TS cannot return `1.50` because its string representation is `'1.5'`, which doesn't match `'1.50'`. Similarly, it cannot return `1.5`, because `1.5` also stringifies to `'1.5'`, failing the round-trip constraint.

In other words, there are no number literals whose string representation can be `'1.50'`, `'0b10'`, or `'1e2'`. For reference, the string representation of `1.50` is `'1.5'`, `0b10` is `'2'`, and `1e2` is `'100'`:

```ts
type A = `${1.50}`; //=> "1.5"
type B = `${0b10}`; //=> "2"
type C = `${1e2}`; //=> "100"

```

# Why is this behavior correct?

Let's consider the following example:

```ts
const doubles =  {'1.5': 2, '2.50': 5};

doubles[1.5]; // ✅ Works, yields `2` at runtime
doubles[2.5]; // ❌ Errors, yields `undefined` at runtime
//      ~~~
// Property '2.5' does not exist on type '{ '1.5': number; '2.50': number; }'.

```

When you access an object using a number index, JS implicitly converts that number to a string before performing the lookup. And since `(2.5).toString()` is `'2.5'`, which doesn't match `'2.50'`, the result is `undefined` and not `5`.

In other words, there is *no* number index that can be used to access the key `'2.50'`.

So, if we use `StringToNumber` to normalize keys (like we already do in `NormalizeKeys`), we'd end up getting incorrect keys. For example, if `Paths<{'2.50': 5}>` returned `'2.50' | 2.5`, it would be incorrect because `2.5` isn't a valid path.

Note: `'Infinity'` handling in `StringToNumber` is fine because `toString` of `Infinity` is indeed `'Infinity'`.

# Caveat

There is, however, one caveat in TS's default inference behavior that feels a bit off.

If you try to infer a number from a string like `'1.50'`, TS doesn't simply deny the inference and return `never`, it instead returns `number`. I am _not_ really sure why it behaves like this, but this feels incorrect. Since there is _no_ number equivalent for `'1.50'`, it should behave _just_ like an invalid string (e.g., `'foo'`). Interestingly, for strings with numeric separators like `'123_45'`, TS denies the inference and returns `never`.

Also, if `Paths<{'2.50': 5}>` were to return `'2.50' | number`, that would be _even_ worse.

And, because of this, `ObjectMerge<{1.5: number}, {'1.50': string}>` currently returns `{'1.50': string}`, instead it should have returned `{1.5: number; '1.50': string}`.

# Conclusion

So, these are the _only_ changes to `StringToNumber` that this PR introduces:

| Input | Current | Updated |
|-------|---------|---------|
| `'1.50' \| '0b10' \| '12_345'` | `number` | `never` |
| `string` | `never` | `number` |
| `any` | `Infinity \| -Infinity` | `number` |

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -177,6 +177,7 @@ export type {ArrayReverse} from './source/array-reverse.d.ts';
 export type {UnionMember} from './source/union-member.d.ts';
 export type {Absolute} from './source/absolute.d.ts';
 export type {UnionLength} from './source/union-length.d.ts';
+export type {StringToNumber} from './source/string-to-number.d.ts';
 
 // Template literal types
 export type {CamelCase, CamelCaseOptions} from './source/camel-case.d.ts';
--- a/readme.md
+++ b/readme.md
@@ -310,6 +310,7 @@ Click the type names for complete docs.
 - [`Sum`](source/sum.d.ts) - Returns the sum of two numbers.
 - [`Subtract`](source/subtract.d.ts) - Returns the difference between two numbers.
 - [`Absolute`](source/absolute.d.ts) - Returns the absolute value of the specified number or bigint.
+- [`StringToNumber`](source/string-to-number.d.ts) - Converts a numeric string to a number.
 
 ### Change case
 
--- a/source/absolute.d.ts
+++ b/source/absolute.d.ts
@@ -1,4 +1,4 @@
-import type {StringToNumber} from './internal/string.d.ts';
+import type {StringToNumber} from './string-to-number.d.ts';
 
 /**
 Returns the absolute value of the specified number or bigint.
--- a/source/internal/object.d.ts
+++ b/source/internal/object.d.ts
@@ -6,9 +6,10 @@ import type {Merge} from '../merge.d.ts';
 import type {IsAny} from '../is-any.d.ts';
 import type {If} from '../if.d.ts';
 import type {IsNever} from '../is-never.d.ts';
+import type {StringToNumber} from '../string-to-number.d.ts';
 import type {FilterDefinedKeys, FilterOptionalKeys} from './keys.d.ts';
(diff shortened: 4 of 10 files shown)


som-sm:
@sindresorhus Let me know your thoughts on this.

The tests are failing because of #1453.

PR #1441, opened 2026-05-31
Improve `TupleOf` implementation

airismadeupofair:
Closes: #1440.

### example:
```ts

import type {TupleOf} from 'type-fest';

type TensorSized2000 = TupleOf<2000, number>
//      ^? type TensorSized2000 = [number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, number, ... 1936 more ..., number]

```

PR #1447, opened 2026-06-04
`StringRepeat`: Add support for generating really long strings

som-sm:
This PR refactors the implementation of `StringRepeat` similar to the refactor of `TupleOf` (#1441).

image


source/string-repeat.d.ts
```diff
@@ -9,9 +10,9 @@ Returns a new string which contains the specified number of copies of a given st
 import type {StringRepeat} from 'type-fest';
 
 declare function stringRepeat<
-	Input extends string,
+	S extends string,
 	Count extends number,
->(input: Input, count: Count): StringRepeat<Input, Count>;
+>(input: S, count: Count): StringRepeat<S, Count>;
```
som-sm, line 15: IMO, `S` is better in this context, also more conventional.

PR #1449, opened 2026-06-05
Enhance comparison types to support bigint comparison

### LINKED ISSUE #1413: Numeric comparison types: support for bigints?
I was experimenting with the bigint type a few weeks ago, where I wanted to use types such as `LessThanOrEqual`.  Alas, I couldn't figure it out because this type, like several other numeric types, is restricted to `number` for its type parameters.

I'm not asking for bigint-to-number comparisons, but I'm wondering if these types could be built for bigint-to-bigint operations.

Emiyaaaaa:
- fixed #1413


sindresorhus:
1. The nonliteral `bigint` fallback runs before the infinity handling, so comparisons where one side is broad `bigint` and the other side is `PositiveInfinity`/`NegativeInfinity` widen to `boolean`, even though the answer is fixed. For example, `GreaterThan<PositiveInfinity, bigint>` should be `true`, `GreaterThan<bigint, PositiveInfinity>` should be `false`, and `LessThan<bigint, PositiveInfinity>` should be `true`, but they all currently come out as `boolean`. Moving the broad `bigint` fallback until after the infinity cases should keep the existing precision.

2. The tests add mixed `bigint`/`number` support, but that path is not correct for decimal number literals. For example, `GreaterThan<2n, 1.5>` currently resolves to `false`, and `GreaterThan<1.5, 2n>` resolves to `true`, because the positive numeric string comparison treats `"1.5"` as longer than `"2"`. The issue only asked for bigint-to-bigint support, so the simplest fix is probaly to drop mixed number/bigint support from this PR, or make the contract explicitly integer-only if that is intended.

Also the public docs still only talk aboutt `number` and non-literal `number`, so they should be updated to mention `bigint` and broad `bigint` behavior too

Emiyaaaaa:
DONE

som-sm:
Umm... `GreaterThan<2n, 1.5>` still returns `false`, and `GreaterThan<1.5, 2n>` still returns `true`?

Emiyaaaaa:
> Umm... `GreaterThan<2n, 1.5>` still returns `false`, and `GreaterThan<1.5, 2n>` still returns `true`?

`GreaterThan` previously only supported integers; I have updated the type description.

test-d/greater-than-or-equal.ts
```diff
@@ -12,32 +12,63 @@ expectType<GreaterThanOrEqual<2, 2>>(true);
...
+// Non-literal `bigint`
+expectType<GreaterThanOrEqual<bigint, bigint>>({} as boolean);
+expectType<GreaterThanOrEqual<bigint, number>>({} as boolean);
+expectType<GreaterThanOrEqual<bigint, 1>>({} as boolean);
+expectType<GreaterThanOrEqual<1, bigint>>({} as boolean);
+
+expectType<GreaterThanOrEqual<PositiveInfinity, bigint>>({} as true);
+expectType<GreaterThanOrEqual<bigint, PositiveInfinity>>({} as false);
+expectType<GreaterThanOrEqual<NegativeInfinity, bigint>>({} as false);
+expectType<GreaterThanOrEqual<bigint, NegativeInfinity>>({} as true);
```
som-sm, line 74: Add tests for `PositiveInfinity`/`NegativeInfinity` and `number`.

test-d/greater-than.ts
```diff
@@ -12,30 +12,61 @@ expectType<GreaterThan<2, 2>>(false);
...
+// Non-literal `bigint`
+expectType<GreaterThan<bigint, bigint>>({} as boolean);
+expectType<GreaterThan<bigint, number>>({} as boolean);
+expectType<GreaterThan<bigint, 1>>({} as boolean);
+expectType<GreaterThan<1, bigint>>({} as boolean);
+
+expectType<GreaterThan<PositiveInfinity, bigint>>({} as true);
+expectType<GreaterThan<bigint, PositiveInfinity>>({} as false);
+expectType<GreaterThan<NegativeInfinity, bigint>>({} as false);
+expectType<GreaterThan<bigint, NegativeInfinity>>({} as true);
```
som-sm, line 71: I guess it's fine to return `boolean` in all these cases. `PositiveInfinity` and `NegativeInfinity` are anyway not that common, and returning `boolean` isn't inaccurate, it's just imprecise.

Also, if we handle these cases, then `GreaterThan<PositiveInfinity, number>` should also return `false`, which it currently doesn't (returns `boolean` currently).

som-sm: I think it makes sense to keep the implementation simple, and just return `bigint` if `bigint extends A | B`. And remove the `IsFinite` type.

som-sm (changes requested):

test-d/greater-than.ts
Emiyaaaaa: @sindresorhus

PR #1452, opened 2026-06-08
Simplify ApplyDefaultOptions for supported TypeScript versions

### LINKED ISSUE #1142: Remove redundant code for compatibility with older versions of typescript
In some previously supported but now unsupported versions of typescript, there would be some incorrect circular check, and in order to bypass the circular check, we have to add some non-essential code, for example:
```ts
type A<T> = 
 SomeToolType<T> extends infer B 
  ? B
  : never // Never happen
```

The above code can be simplified now:
```ts
type A<T> = SomeToolType<T>
```

AkshatKardak:
Closes #1142

Remove the redundant TypeScript compatibility wrapper from `ApplyDefaultOptions` and return `_ApplyDefaultOptions<...>` directly.

This simplifies the type definition now that older TypeScript workaround behavior is no longer needed.

### DIFF
--- a/source/internal/object.d.ts
+++ b/source/internal/object.d.ts
@@ -219,13 +219,13 @@ type Result = ApplyDefaultOptions<PathsOptions, DefaultPathsOptions, SpecifiedOp
 ```
 */
 export type ApplyDefaultOptions<
-    Options extends object,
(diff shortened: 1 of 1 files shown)


som-sm:
Bogus PR, just removes a comment.

PR #1454, opened 2026-06-10
Split lint into a separate job

Emiyaaaaa:
Split lint into a separate job. This makes troubleshooting way easier, and since lint doesn’t depend on Node.js versions, we don’t need to run it on every Node version.

And running lint in parallel can make the workflow faster. (3mins -> 2mins)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,6 +9,18 @@ on:
     branches:
       - main
 jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-node@v6
+        with:
+          node-version: 24
+      - run: npm install
+      - run: npm run test:linter
+      - run: npm run test:xo
+
   test:
     name: Node.js ${{ matrix.node-version }}
     runs-on: ubuntu-latest
@@ -25,7 +37,8 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm install
-      - run: npm test
+      - run: npm run test:tsd
+
   types:
     name: TypeScript ${{ matrix.typescript-version }}
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,34 +9,23 @@ on:
     branches:
       - main
 jobs:
-  lint:
-    name: Lint
+  test:
+    name: Test ${{ matrix.script }}
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        script:
+          - tsd
+          - xo
+          - linter
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-node@v6
         with:
           node-version: 24
       - run: npm install
-      - run: npm run test:linter & npm run test:xo
-
-  test:
-    name: Node.js ${{ matrix.node-version }}
-    runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        node-version:
-          - 24
-          - 22
-          - 20
-    steps:
-      - uses: actions/checkout@v6
-      - uses: actions/setup-node@v6
-        with:
-          node-version: ${{ matrix.node-version }}
-      - run: npm install
-      - run: npm run test:tsd
+      - run: npm run test:${{ matrix.script }}
 
   types:
     name: TypeScript ${{ matrix.typescript-version }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,7 +26,6 @@ jobs:
           node-version: 24
       - run: npm install
       - run: npm run test:${{ matrix.script }}
-
   types:
     name: TypeScript ${{ matrix.typescript-version }}
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,8 +18,7 @@ jobs:
         with:
           node-version: 24
       - run: npm install
-      - run: npm run test:linter
-      - run: npm run test:xo
+      - run: npm run test:linter & npm run test:xo
 
   test:
     name: Node.js ${{ matrix.node-version }}


som-sm (commented):
- Aren't the test scripts already running in parallel?
- Maybe `test:tsd` can also just run on a single Node version?

> This makes troubleshooting way easier

Umm... the scripts can be run independently locally, why do we need to split it in CI?

Emiyaaaaa:
> * Aren't the test scripts already running in parallel?

Actually, a runner in GitHub Actions only has 2 physical cores (each physical core is split into two logical cores, yet they still share resources). Therefore, when more than 2 Node.js scripts run in parallel, the parallel execution is not genuine parallelism, especially for these CPU-intensive scripts. 

And this explains why the total execution time decreases after separating the lint CI task; running tasks in parallel across different runners achieves genuine parallel execution.

> Maybe test:tsd can also just run on a single Node version?

Maybe, and in the very very early stage, test:tsd was tested across multiple Node versions, and this decision rests with @sindresorhus 

> Umm... the scripts can be run independently locally, why do we need to split it in CI?

Merely for low coupling and clarity, enabling contributors, maintainers and anyone else reviewing this PR to clearly see which tests passed and which failed. This approach is adopted by most projects, covering lint checks, code tests, performance tests, UI tests and more, instead of executing all of them within a single job.

som-sm (commented):
I guess this is cleaner:

```yml
jobs:
  test:
    name: Run test:${{ matrix.script }}
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
        script:
          - tsd
          - xo
          - linter
    steps:
      - uses: actions/checkout@v6
      - uses: actions/setup-node@v6
        with:
          node-version: 24
      - run: npm install
      - run: npm run test:${{ matrix.script }}
```

It runs all three as separate jobs, but I guess that's fine.

som-sm:
> And this explains why the total execution time decreases after separating the lint CI task; running tasks in parallel across different runners achieves genuine parallel execution.

Umm... I looked into this some time back and ran some quick benchmarks, but I didn't pursue it further because I didn't notice any significant difference.

I ran it again today, and still it's not as significant:

- With 4 CPUs it takes around 1m 50s.
    ```bash
    rm -rf node_modules/.cache/xo-linter && docker run --memory=14g --cpus=4 -it --rm -v "$(pwd)":/app -w /app node:24 bash -c "time npm run test:xo"
    ```

- With 8 CPUs it takes around 1m 30s. 
    ```bash
    rm -rf node_modules/.cache/xo-linter && docker run --memory=14g --cpus=8 -it --rm -v "$(pwd)":/app -w /app node:24 bash -c "time npm run test:xo"
    ```

So, not much of a difference. And then having them in separate jobs has the `npm install` overhead as well, which would take up around ~15-20s. So, I thought it wouldn't matter much. But if you're noticing improvement of around a min, then maybe my testing is flawed.

I guess it makes sense to split it up!

Emiyaaaaa:
Testing a single Node.js script on a multi-core CPU proves nothing. This is because Node.js operates in a single-threaded manner; a single node command typically spawns only one process, which can run on at most one CPU core. Theoretically, the same Node.js script will take roughly the same amount of time to execute on a 1-core CPU and a 100-core CPU (I have not verified this through testing).

The proper testing approach is to measure and compare the execution time of two CPU-intensive Node scripts running in parallel on a single CPU core versus their parallel execution time on a dual CPU core setup (physical cores, not logical cores).

Running two CPU-intensive Node scripts in parallel generates two separate processes. On a dual-core CPU, the operating system will most likely assign each process to a distinct physical core, resulting in a total execution time equal to that of a single script. On a single-core CPU, however, the two processes have to share the sole physical core and execute alternately via time slicing, leading to a total runtime equivalent to sequential execution.

PS：Although the `nproc` command shows that there are 4 CPUs available in GitHub Actions, seemingly four Node.js scripts can run in true parallel, hyper-threading is actually enabled here — this is a dual-core quad-thread configuration with four logical CPU cores. The two logical cores residing on the same physical CPU core will still contend for computing resources when handling certain computational tasks. Therefore, there is little practical difference between a dual-core quad-thread setup and a dual-core dual-thread setup for CPU-intensive scripts.

run `lscpu` in github action
```
CPU(s):                                  4
On-line CPU(s) list:                     0-3
Vendor ID:                               AuthenticAMD
Model name:                              AMD EPYC 7763 64-Core Processor
CPU family:                              25
Model:                                   1
Thread(s) per core:                      2
Core(s) per socket:                      2
```

som-sm:
> Testing a single Node.js script on a multi-core CPU proves nothing. This is because Node.js operates in a single-threaded manner; a single node command typically spawns only one process, which can run on at most one CPU core. Theoretically, the same Node.js script will take roughly the same amount of time to execute on a 1-core CPU and a 100-core CPU (I have not verified this through testing).

This is not true in our case, and it's quite easy to validate. If you run any of these scripts with the `time` utility, you'll notice that the CPU utilization always exceeds 100%. So, these commands _do_ benefit from having multiple cores.

```
> time npm run test:tsc

33.80s user 0.91s system 160% cpu 21.671 total
```

```
> time npm run test:tsd

26.60s user 0.81s system 148% cpu 18.421 total
```

```
> time npm run test:linter

11.07s user 1.90s system 256% cpu 5.051 total
```

```
> rm -rf node_modules/.cache/xo-linter && time npm run test:xo

62.20s user 4.79s system 164% cpu 40.796 total
```

<br>

You can further verify this by running these tests with a single CPU vs 2 CPUs. You'll notice that with 2 CPUs, the tests run significantly faster, which _again_ confirms that there _is_ benefit in having multiple cores.

```
> docker run --memory=6g --cpus=1 -it --rm -v "$(pwd)":/app -w /app node:24 bash -c "time npm run test:tsc"

real    1m4.430s
user    1m2.268s
sys     0m1.830s
```

```
> docker run --memory=6g --cpus=2 -it --rm -v "$(pwd)":/app -w /app node:24 bash -c "time npm run test:tsc"

real    0m29.803s
user    0m41.516s
sys     0m1.350s
```

<br>

> Umm... I looked into this some time back and ran some quick benchmarks, but I didn't pursue it further because I didn't notice any significant difference.

Probably the reason I didn't notice any significant difference in my testing with 4 vs 8 CPUs is that `npm run test:linter` finishes quite quickly, so there's one spare CPU after a short while. So, it isn't strictly one CPU per test for the entire duration.

som-sm (approved):

PR #1455, opened 2026-06-13
`StringRepeat`: Fix behavior with extremely large/small counts

DucMinhNe:
When `Count` is large or small enough that TypeScript stringifies it in scientific notation, `StringRepeat` silently returns a **wrong** string literal:

```ts
import type {StringRepeat} from 'type-fest';

type A = StringRepeat<'a', 1e-7>; //=> 'a'   ❌ (1e-7 = 0.0000001, should repeat 0 times)
type B = StringRepeat<'a', 2e21>; //=> 'aa'  ❌ (an astronomically large count)
type C = StringRepeat<'a', 1e21>; //=> 'a'   ❌
```

### Root cause

`BuildStringDigitByDigit` walks `` `${Count}` `` digit by digit and stops at the first non-digit character. That is exactly how the documented "decimal part is ignored" behavior works — for `2.5`, `"2.5"` is consumed up to the `.`, yielding the integer part `2`.

But for scientific notation the leading digits are **not** the integer part of the value:
- `` `${1e-7}` `` is `"1e-7"` → the builder consumes `"1"` and stops at `e`, producing one copy — but the actual integer part of `1e-7` is `0`.
- `` `${2e21}` `` is `"2e+21"` → consumes `"2"`, producing two copies — but the real count is 2,000,000,000,000,000,000,000.

So any count whose string form contains `e` produces a confidently-wrong literal.

### Fix

These counts can't be expanded into a precise literal (they're either effectively `0` or far beyond any representable repetition — `String.prototype.repeat(2e21)` throws `RangeError` at runtime). So when `` `${Count}` `` is in scientific notation, resolve to `string` — the same honest "can't compute precisely" fallback the type already uses for non-literal number counts:

```ts
: IsNumericLiteral<Count> extends false
    ? string
    : `${Count}` extends `${string}e${string}`
        ? string
        : BuildStringDigitByDigit<S, `${Count}`>;
```

No regression: every scientific-notation count was already wrong, and plain-integer / decimal counts are untouched (the existing `7.5` / `3.14` fractional tests still pass). Added `test-d` assertions for `1e-7`, `2e21`, `1e21`.

Full `tsd` suite passes.

### DIFF
--- a/source/string-repeat.d.ts
+++ b/source/string-repeat.d.ts
@@ -33,6 +33,8 @@ type DecimalCount = StringRepeat<'foo', 2.5>;
 //=> 'foofoo'
 ```
 
+Note: If the count is large or small enough to be represented in scientific notation (for example, `1e21` or `1e-7`), the result resolves to `string`, since the exact repeated literal cannot be computed.
+
 @category String
 @category Template literal
 */
@@ -43,7 +45,9 @@ export type StringRepeat<S extends string, Count extends number> =
 			? ''
 			: IsNumericLiteral<Count> extends false
 				? string
-				: BuildStringDigitByDigit<S, `${Count}`>;
+				: `${Count}` extends `${string}e${string}`
+					? string
+					: BuildStringDigitByDigit<S, `${Count}`>;
 
 type BuildStringDigitByDigit<S extends string, Count extends string, Accumulator extends string = ''> =
 	Count extends `${infer First extends DigitCharacter}${infer Rest}`
--- a/source/string-repeat.d.ts
+++ b/source/string-repeat.d.ts
@@ -37,7 +37,7 @@ type DecimalCount = StringRepeat<'foo', 2.5>;
 @category Template literal
 */
 export type StringRepeat<S extends string, Count extends number> =
-	Count extends Count // Distribute over `Count` so the checks below apply to each union member individually.
+	Count extends unknown // To distribute `Count`
 		? IsNegative<Count> extends true
 			? never
 			: S extends ''
--- a/source/string-repeat.d.ts
+++ b/source/string-repeat.d.ts
@@ -33,21 +33,23 @@ type DecimalCount = StringRepeat<'foo', 2.5>;
 //=> 'foofoo'
 ```
 
-Note: If the count is large or small enough to be represented in scientific notation (for example, `1e21` or `1e-7`), the result resolves to `string`, since the exact repeated literal cannot be computed.
+Note: If the count is large or small enough that it gets stored in scientific notation internally, then in such cases the result is `string`.
 
 @category String
 @category Template literal
 */
 export type StringRepeat<S extends string, Count extends number> =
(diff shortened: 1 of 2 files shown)


source/string-repeat.d.ts
```diff
@@ -33,6 +33,8 @@ type DecimalCount = StringRepeat<'foo', 2.5>;
 //=> 'foofoo'
 ```
 
+Note: If the count is large or small enough to be represented in scientific notation (for example, `1e21` or `1e-7`), the result resolves to `string`, since the exact repeated literal cannot be computed.
```
som-sm: PR makes sense!

Just that this note text is not quite clear. It should have been something like:
```txt
Note: If the count is large or small enough that it gets stored in scientific notation internally, then in such cases the result is `string`.
```

But, I guess we can remove the note altogether, because it's not a common use case.

som-sm (changes requested):

sindresorhus:
I think this still misses union counts. The new exponent check is applied to `` `${Count}` `` as a whole, so it doesn't distribute over a union.

For example:

```ts
expectType<StringRepeat<'0', 1e-7 | 2>>(unknown as string);
```

DucMinhNe:
Thanks for the careful review!

@sindresorhus You're right that the previous fix only checked `` `${Count}` `` as a whole, so it didn't handle a union `Count`. I've reworked the type to distribute over `Count`:

```ts
export type StringRepeat<S extends string, Count extends number> =
	Count extends Count // Distribute over `Count` so the checks below apply to each union member individually.
		? IsNegative<Count> extends true
			? never
			: S extends ''
				? ''
				: IsNumericLiteral<Count> extends false
					? string
					: `${Count}` extends `${string}e${string}`
						? string
						: BuildStringDigitByDigit<S, `${Count}`>
		: never;
```

The `Count extends Count` wrapper forces a distributive conditional, so every member of a union `Count` is evaluated independently — including the scientific-notation guard. For `1e-7 | 2`, the `1e-7` member falls back to `string` and `2` produces `'00'`, and `'00' | string` collapses to `string`.

The requested assertion now passes:

```ts
expectType<StringRepeat<'0', 1e-7 | 2>>(unknown as string);
```

I also added a couple more union cases (a union of two scientific-notation counts, and a sci-notation member mixed with a normal one), and confirmed all existing assertions still pass. `tsd` and `xo` are both green.

@som-sm Good call on the note — I've updated the wording to your suggestion: "Note: If the count is large or small enough that it gets stored in scientific notation internally, then in such cases the result is `string`."

som-sm:
> @som-sm Good call on the note — I've updated the wording to your suggestion: "Note: If the count is large or small enough that it gets stored in scientific notation internally, then in such cases the result is `string`."

Removed the note because it's _not_ adding any value, users wouldn't know what "stored in scientific notation internally" means without enough explanation. And it doesn't make sense to put a detailed explanation for something that is extremely rare.

som-sm (approved):

PR #1462, opened 2026-06-19
`IfNotAnyOrNever`: Lazily evaluate different conditional branches

som-sm:
Consider the following example:
```ts
type CrashIfAnyWrapper<T> = IfNotAnyOrNever<T, CrashIfAny<T>, 'any handled', 'never handled'>;

type CrashIfAny<T, Acc extends unknown[] = []> = 0 extends 1 & T // Check if `T` is `any`
	? CrashIfAny<T, [...Acc, unknown]>
	: never;

// @ts-expect-error
type T1 = CrashIfAny<any>;

type T2 = CrashIfAnyWrapper<any>; // Should not error, but errors!
```

`CrashIfAny` is a type that fails when instantiated with `any`. To fix this, we try wrapping it in `IfNotAnyOrNever` to handle the `any` case explicitly, but this currently doesn't prevent the error. This happens because when we write `IfNotAnyOrNever<T, CrashIfAny<T>, 'any handled'>`, `CrashIfAny` is instantiated with `T` even if its value isn't going to be used.

<br>

Couldn't really find a great fix for this, but accepting the values for different cases in an object seems to fix this issue. It seems to make the compiler not do the instantiation until the value is actually accessed.

```ts
type CrashIfAnyWrapper<T> = IfNotAnyOrNever<T, {ifNot: CrashIfAny<T>; ifAny: 'any handled'; ifNever: 'never handled'}>;

type CrashIfAny<T, Acc extends unknown[] = []> = 0 extends 1 & T // Check if `T` is `any`
	? CrashIfAny<T, [...Acc, unknown]>
	: never;

// @ts-expect-error
type T1 = CrashIfAny<any>;

type T2 = CrashIfAnyWrapper<any>; // Works!
```

### DIFF
--- a/source/internal/type.d.ts
+++ b/source/internal/type.d.ts
@@ -1,9 +1,7 @@
-import type {If} from '../if.d.ts';
 import type {IsAny} from '../is-any.d.ts';
 import type {IsNever} from '../is-never.d.ts';
 import type {Primitive} from '../primitive.d.ts';
 import type {UnknownArray} from '../unknown-array.d.ts';
-import type {UnionToIntersection} from '../union-to-intersection.d.ts';
 
 /**
 Matches any primitive, `void`, `Date`, or `RegExp` value.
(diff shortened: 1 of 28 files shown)


source/internal/type.d.ts
```diff
@@ -91,16 +89,15 @@ An if-else-like type that resolves depending on whether the given type is `any`
 
 @example
 ```
-// When `T` is a NOT `any` or `never` (like `string`) => Returns `IfNotAnyOrNever` branch
-type A = IfNotAnyOrNever<string, 'VALID', 'IS_ANY', 'IS_NEVER'>;
+type A = IfNotAnyOrNever<string, {ifNot: 'VALID'; ifAny: 'IS_ANY'; ifNever: 'IS_NEVER'}>;
```
som-sm, line 93: Umm... would this be better?
```suggestion
type A = IfNotAnyOrNever<string, {neither: 'VALID'; any: 'IS_ANY'; never: 'IS_NEVER'}>;
```

abo-L (approved):
0dafc0d83c3e4c48c0ab2efad4b160d1b8c6d252

test-d/internal/if-not-any-or-never.ts
```diff
@@ -1,12 +1,48 @@
...
+type CrashIfNotAnyWrapper<T> = IfNotAnyOrNever<T, {ifNot: never; ifAny: CrashIfNotAny<T>}>;
+
+type CrashIfNotAny<T, Acc extends unknown[] = []> = 0 extends 1 & T // Check if `T` is `any`
+	? never
+	: CrashIfNotAny<T, [...Acc, unknown]>;
+
+// `CrashIfNotAny<string>` errors with a recursion depth error, but `CrashIfNotAnyWrapper<string>` shouldn't.
+// @ts-expect-error
+type T5 = CrashIfNotAny<string>;
+type T6 = CrashIfNotAnyWrapper<string>;
```
sindresorhus, line 48: maybe also ensure an unused explicit `ifNever` branch is not evaluated

```suggestion
type T6 = CrashIfNotAnyWrapper<string>;

type CrashIfNotNeverWrapper<T> = IfNotAnyOrNever<T, {ifNot: never; ifNever: CrashIfNotNever<T>}>;

type CrashIfNotNever<T, Acc extends unknown[] = []> = [T] extends [never]
	? never
	: CrashIfNotNever<T, [...Acc, unknown]>;

// `CrashIfNotNever<string>` errors with a recursion depth error, but `CrashIfNotNeverWrapper<string>` shouldn't.
// @ts-expect-error
type T7 = CrashIfNotNever<string>;
type T8 = CrashIfNotNeverWrapper<string>;
```

PR #1463, opened 2026-06-21
`Get`: Distribute over unions so a key missing on some members yields `undefined`

### LINKED ISSUE #1205: Possible incorrect behavior of 'GET<>'

ardittirana:
Fixes #1205

### What was wrong
For a **union** base type, `Get` accessed the key on each member, but a member that *lacked* the key resolved to `unknown` — and `unknown` absorbs the rest of the union. So:

```ts
Get<{a: number} | {b: string}, 'a'> // was `unknown`, expected `number | undefined`
Get<{data: {type: 'x'; v: number} | {type: 'y'}}, 'data.v'> // was `unknown`
```

### Fix
`PropertyOf` now detects a union first and, when distributing over its members, treats a missing key as `undefined` for that member. A missing key on a **non-union** type still resolves to `unknown` (the existing, documented behaviour — preserved by an explicit `MissingProperty` parameter).

### Tests
Added cases to `test-d/get.ts` for: key present on all members, key present on some members (→ `number | string | undefined`), a direct union base type, and the unchanged non-union `unknown` case. `npm run test:tsd` and `xo` pass; no existing assertions changed.

### DIFF
--- a/source/get.d.ts
+++ b/source/get.d.ts
@@ -125,29 +125,25 @@ type UncheckedIndex<T, U extends string | number> = [T] extends [Record<string |
 Get a property of an object or array. Works when indexing arrays using number-literal-strings, for example, `PropertyOf<number[], '0'> = number`, and when indexing objects with number keys.
 
 Note:
-- Returns `unknown` if `Key` is not a property of `BaseType`, since TypeScript uses structural typing, and it cannot be guaranteed that extra properties unknown to the type system will exist at runtime.
-- Returns `undefined` from nullish values, to match the behaviour of most deep-key libraries like `lodash`, `dot-prop`, etc.
+- Returns `undefined` if `Key` is not a property of `BaseType`, to match the behaviour of most deep-key libraries like `lodash`, `dot-prop`, etc.
 */
 type PropertyOf<BaseType, Key extends string, Options extends Required<GetOptions>> =
 	IsUnion<BaseType> extends true
 		// Distribute over the union so that a key which is missing on *some*
 		// members resolves to `undefined` for those members instead of
-		// collapsing the whole result to `unknown`. For example,
+		// collapsing the whole result. For example,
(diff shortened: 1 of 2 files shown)


som-sm (commented):
> A missing key on a **non-union** type still resolves to `unknown` (the existing, documented behaviour — preserved by an explicit `MissingProperty` parameter).

I think the existing behaviour here is incorrect, this should _also_ be `undefined`. More so, if for unions it's going to be `undefined`. Let's make it consistently `undefined` for missing properties.

ardittirana:
Makes sense — consistency wins here, and you're right that it's odd to special-case non-unions once unions resolve missing keys to `undefined`. I've updated it so a missing key resolves to `undefined` for both union and non-union base types, matching lodash/dot-prop runtime behaviour.

Changes:
- `PropertyOf` no longer distinguishes the two cases; the now-redundant `MissingProperty` parameter on `SinglePropertyOf` is removed.
- The doc note now reads "Returns `undefined` if `Key` is not a property of `BaseType`".
- Updated the out-of-bounds / missing-key tests (tuples, array index, `Get<{a: number}, 'b'>`, etc.) from `unknown` to `undefined`.

Full `tsd` suite passes and lint is clean. Let me know if you'd prefer any wording tweaks in the docs.

PR #1466, opened 2026-06-28
test: add IsEqual test cases for number vs bigint

### LINKED ISSUE #1442: Bug: `IsEqual<100, 100n>` is False
### Bug description

in js:
```js
100 == 100n // true
100 === 100n // false
```

So, should `IsEqual<100, 100n>` be true? 
Or we need a `isNumberEqual` method to make `isNumberEqual<100, 100n>` true then keep `IsEqual<100, 100n>` false?

### Repro

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBDAnmApnA3gSQM4FECOArgIYA2AvnAGZQQhwDkSqAtFStjAwNwCwAUAOZoYHeAF44OAiVIAeAIwAGJQBo4ypQDsAfH35xDcAPTG4APQD8AoA

DakshSinghDhami:
### Problem

Issue #1442 raised a question about whether `IsEqual<100, 100n>` should be `true` or `false`. After investigation, the current behavior (`false`) is correct because `number` and `bigint` are distinct types in TypeScript and use strict equality (`===`) in JavaScript.

### Changes

- Added test cases validating that `IsEqual` correctly distinguishes between `number` and `bigint` types
- Added tests for bigint literal self-equality

Closes #1442

### DIFF
--- a/test-d/is-equal.ts
+++ b/test-d/is-equal.ts
@@ -8,8 +8,7 @@ expectType<true>({} as IsEqual<'foo', 'foo'>);
 expectType<false>({} as IsEqual<true, false>);
 expectType<true>({} as IsEqual<false, false>);
 
-// Test for Issue https://github.com/sindresorhus/type-fest/issues/1442
-// number and bigint are distinct types in TypeScript and JavaScript (===)
+// Number and bigint are distinct types.
 expectType<false>({} as IsEqual<100, 100n>);
 expectType<false>({} as IsEqual<number, bigint>);
 expectType<false>({} as IsEqual<42, 42n>);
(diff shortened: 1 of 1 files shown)


Emiyaaaaa (approved):

sindresorhus:
#1442 is an open design discussion about whether to add a loose/numeric-equality type (`ToBigint` / `IsLooseEqual` / `isNumberEqual`), since math types like `LessThanOrEqual<100, 100n>` route through `IsEqual`. Adding tests for the current behavior documents the status quo but doesn't resolve that.

PR #1468, opened 2026-06-30
Refactor: Remove redundant circular check infer workarounds (#1142)

### LINKED ISSUE #1142: Remove redundant code for compatibility with older versions of typescript

abhijeetnardele24-hash:
### Description

This PR removes the redundant `extends infer Result` workaround from several types. 

Previously, this pattern was used to bypass circular dependency checks and inference limitations in older versions of TypeScript (e.g. `extends infer Result ? ... : never`). With modern TypeScript's improved conditional type evaluation, these workarounds are no longer necessary and can be safely inlined to simplify the type definitions.

Changes were made to:
- `ArrayReverse`
- `ArrayTail`
- `_SplitOnRestElement`
- `_SetRequired`
- `SumPositives`

*(Note: `UnionToTuple` was intentionally left as-is, as the compiler still requires the explicit nudge there to guarantee an array return type during deep recursion).*

### Linked Issues

Fixes #1142

### DIFF
--- a/source/split-on-rest-element.d.ts
+++ b/source/split-on-rest-element.d.ts
@@ -66,13 +66,8 @@ export type SplitOnRestElement<
 > =
 	Array_ extends unknown // For distributing `Array_`
 		? IfNotAnyOrNever<Array_, {
-			ifNot: _SplitOnRestElement<
-				Array_,
-				ApplyDefaultOptions<SplitOnRestElementOptions, DefaultSplitOnRestElementOptions, Options>
-			>;
-		}> extends infer Result extends UnknownArray
-			? If<IsArrayReadonly<Array_>, Readonly<Result>, Result>
-			: never // Should never happen
+			ifNot: If<IsArrayReadonly<Array_>, Readonly<_SplitOnRestElement<Array_, ApplyDefaultOptions<SplitOnRestElementOptions, DefaultSplitOnRestElementOptions, Options>>>, _SplitOnRestElement<Array_, ApplyDefaultOptions<SplitOnRestElementOptions, DefaultSplitOnRestElementOptions, Options>>>;
+		}>
 		: never; // Should never happen
 
 /**
--- a/source/sum.d.ts
+++ b/source/sum.d.ts
@@ -75,8 +75,6 @@ type SumPostChecks<A extends number, B extends number, AreNegative = [IsNegative
 Adds two positive numbers.
 */
 type SumPositives<A extends number, B extends number> =
-	[...TupleOf<A>, ...TupleOf<B>]['length'] extends infer Result extends number
-		? Result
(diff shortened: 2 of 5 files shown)


source/split-on-rest-element.d.ts
```diff
@@ -66,13 +66,8 @@ export type SplitOnRestElement<
...
 		? IfNotAnyOrNever<Array_, {
-			ifNot: _SplitOnRestElement<
-				Array_,
-				ApplyDefaultOptions<SplitOnRestElementOptions, DefaultSplitOnRestElementOptions, Options>
-			>;
-		}> extends infer Result extends UnknownArray
-			? If<IsArrayReadonly<Array_>, Readonly<Result>, Result>
-			: never // Should never happen
+			ifNot: If<IsArrayReadonly<Array_>, Readonly<_SplitOnRestElement<Array_, ApplyDefaultOptions<SplitOnRestElementOptions, DefaultSplitOnRestElementOptions, Options>>>, _SplitOnRestElement<Array_, ApplyDefaultOptions<SplitOnRestElementOptions, DefaultSplitOnRestElementOptions, Options>>>;
+		}>
```
som-sm, line 70: The result is extracted into a type (`Result` here) solely to keep the code DRY. It also makes the intent a bit clearer.

This is more obvious with this change because we definitely don't want to repeat `_SplitOnRestElement<Array_, ApplyDefaultOptions<SplitOnRestElementOptions, DefaultSplitOnRestElementOptions, Options>>` twice. It's less obvious with other changes because they're much smaller, but the idea is the same, keep the code DRY.

source/sum.d.ts
```diff
@@ -75,8 +75,6 @@ type SumPostChecks<A extends number, B extends number, AreNegative = [IsNegative
 Adds two positive numbers.
 */
 type SumPositives<A extends number, B extends number> =
-	[...TupleOf<A>, ...TupleOf<B>]['length'] extends infer Result extends number
-		? Result
-		: never;
+	[...TupleOf<A>, ...TupleOf<B>]['length'] & number;
```
som-sm, line 78: This change is different from the other changes, but not sure if we want to keep this PR _just_ for this change.

som-sm (commented):
I think we can close this PR, doesn't make sense to keep it open [just for a single change](https://github.com/sindresorhus/type-fest/pull/1468#discussion_r3504083611).

PR #1469, opened 2026-07-01
Fix NonEmptyObject returning non-never for types with only index signatures

### LINKED ISSUE #821: `NonEmptyObject` fails for objects with dynamic properties

highoncomputers:
This fixes #821.

When `NonEmptyObject` is used with a type that has only index signatures (e.g. `NonEmptyObject<{ [argument: string]: string | number | undefined }>`), the type previously returned a type that did not reject empty objects.

The fix introduces a `HasIndexSignature` internal type that checks whether the given type has any `string`, `number`, or `symbol` index signature. When `NonEmptyObject` encounters a type with no required keys but with index signatures, it now returns `never`, properly rejecting empty objects.

### Test cases added:
- Type with only a string index signature → expects `never`
- Type with both index signature and optional keys → expects `never`
- Type with both index signature and required keys → expects the type itself

### DIFF
--- a/source/non-empty-object.d.ts
+++ b/source/non-empty-object.d.ts
@@ -1,4 +1,5 @@
 import type {HasRequiredKeys} from './has-required-keys.d.ts';
+import type {OmitIndexSignature} from './omit-index-signature.d.ts';
 import type {RequireAtLeastOne} from './require-at-least-one.d.ts';
 
 /**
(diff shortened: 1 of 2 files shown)


sindresorhus:
Looks like this still misses the mixed case where the object has an optional named key and an index signature:

```ts
type Example = {
	foo?: string;
	[key: string]: unknown;
};

expectNotAssignable<NonEmptyObject<Example>>({});
```

On this branch, that still fails becuase `{}` is assignable. The new first branch only catches types with no explicit keys at all. For mixed types, `HasRequiredKeys<T>` sees the broad index signature as required and returns `T`, so the empty object gets through.

I think the fix should use the index-signature-stripped type consistently:

```ts
export type NonEmptyObject<T extends object> =
	HasRequiredKeys<OmitIndexSignature<T>> extends true
		? T
		: RequireAtLeastOne<T, keyof OmitIndexSignature<T>>;
```

PR #1470, opened 2026-07-01
`Writable`: Fix behavior with index signatures

### LINKED ISSUE #717: Writable doesn't only just strip readonly from the keys
Hi! I encounter a weird bug, and I am not sure if I am using it wrong or it's broken somehow.

I am encountering an issue where the `Writable` doesn't only strip the `readonly` but it also changes the object's type structure. Check out the following images for further details. I expect that Writable simply removes `readonly` from the first level of keys in the object.

Writable1

Writable2

I like the [@utility-types approach to it](https://github.com/piotrwitek/utility-types/blob/411e83ecf70e428b529fc2a09a49519e8f36c8fa/src/mapped-types.ts#L649), it seems simpler with a very similar result. Would you be open to have the same implementation here?

Thanks!

MahinAnowar:
Fixes #717.

## Problem

When `Writable` is used without the optional `Keys` argument, it doesn't only strip `readonly` — it also changes the object's structure. The clearest symptom is an index signature: `Writable<{readonly [key: string]: number}>` currently resolves to `{[x: string]: number; [x: number]: number}` — a **spurious numeric index signature** gets added.

This happens because the default (all-keys) case still goes through the `Except<BaseType, Keys> & {-readonly [K in keyof Pick<BaseType, Keys>]: …}` rebuild, and that `Except`/`Pick` round-trip does not faithfully reproduce the original shape.

## Fix

As suggested in the issue thread ("add an overload when that parameter is not passed and use the simple handling"), the default case now uses a plain homomorphic mapped type:

```ts
{-readonly [KeyType in keyof BaseType]: BaseType[KeyType]}
```

This only removes the `readonly` modifier and otherwise preserves the input structure exactly, including index signatures — matching the simpler community implementations referenced in the issue. The explicit subset-of-`Keys` path (`Writable<T, 'a' | 'b'>`) is left completely unchanged, so all existing behavior and tests are preserved.

The all-keys case is detected with `IsEqual<Keys, keyof BaseType>` (already used internally by `Except`), which is `true` both when `Keys` is omitted and when the full key set is passed explicitly — semantically identical intents.

## Tests

Added two `tsd` assertions to `test-d/writable.ts`:

- `Writable<{readonly [key: string]: number}>` → `{[key: string]: number}`
- `Writable<{readonly [key: string]: number; readonly foo: number}>` → `{[key: string]: number; foo: number}`

Both **fail on the current code** (the result carries the extra `[x: number]: number` index signature) and **pass with the fix**. Verified with `git stash` on the source change. The full `npm test` (`tsc` + `tsd` + `xo` + linter) passes.

### DIFF
--- a/source/writable.d.ts
+++ b/source/writable.d.ts
@@ -1,4 +1,5 @@
 import type {Except} from './except.d.ts';
+import type {IsEqual} from './is-equal.d.ts';
 import type {Simplify} from './simplify.d.ts';
 
 /**
@@ -58,11 +59,14 @@ export type Writable<BaseType, Keys extends keyof BaseType = keyof BaseType> =
 				// Handle array
 				? WritableArray<BaseType>
 				// Handle object
-				: Simplify<
-					// Pick just the keys that are not writable from the base type.
-					Except<BaseType, Keys>
-					// Pick the keys that should be writable from the base type and make them writable by removing the `readonly` modifier from the key.
-					& {-readonly [KeyType in keyof Pick<BaseType, Keys>]: Pick<BaseType, Keys>[KeyType]}
-				>;
+				: IsEqual<Keys, keyof BaseType> extends true
+					// When every key is made writable (the default, when no `Keys` argument is passed), use a plain homomorphic mapped type. It only strips the `readonly` modifier and otherwise preserves the input structure exactly — including index signatures — instead of rebuilding it via `Except`/`Pick`, which alters the shape. See: https://github.com/sindresorhus/type-fest/issues/717
+					? {-readonly [KeyType in keyof BaseType]: BaseType[KeyType]}
+					: Simplify<
+						// Pick just the keys that are not writable from the base type.
+						Except<BaseType, Keys>
+						// Pick the keys that should be writable from the base type and make them writable by removing the `readonly` modifier from the key.
+						& {-readonly [KeyType in keyof Pick<BaseType, Keys>]: Pick<BaseType, Keys>[KeyType]}
+					>;
 
 export {};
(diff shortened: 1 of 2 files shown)


sindresorhus (approved):

source/writable.d.ts
```diff
@@ -58,11 +59,14 @@ export type Writable<BaseType, Keys extends keyof BaseType = keyof BaseType> =
...
-					& {-readonly [KeyType in keyof Pick<BaseType, Keys>]: Pick<BaseType, Keys>[KeyType]}
-				>;
+				: IsEqual<Keys, keyof BaseType> extends true
+					// When every key is made writable (the default, when no `Keys` argument is passed), use a plain homomorphic mapped type. It only strips the `readonly` modifier and otherwise preserves the input structure exactly — including index signatures — instead of rebuilding it via `Except`/`Pick`, which alters the shape. See: https://github.com/sindresorhus/type-fest/issues/717
+					? {-readonly [KeyType in keyof BaseType]: BaseType[KeyType]}
+					: Simplify<
+						// Pick just the keys that are not writable from the base type.
+						Except<BaseType, Keys>
+						// Pick the keys that should be writable from the base type and make them writable by removing the `readonly` modifier from the key.
+						& {-readonly [KeyType in keyof Pick<BaseType, Keys>]: Pick<BaseType, Keys>[KeyType]}
```
som-sm: A better fix here would have been to change the mapped type after `&` to this:
```ts
& {-readonly [KeyType in keyof BaseType as KeyType extends Keys ? KeyType : never]: BaseType[KeyType]}
```
Passes both the newly added test cases.

som-sm (changes requested):

source/writable.d.ts
MahinAnowar: Good call, that's cleaner — applied in 7a49338. Both new test cases and the full suite pass with it.

som-sm (approved):

PR #1482, opened 2026-07-03
`IsLiteral*`: Fix behavior with unions

som-sm:
This PR primarily fixes the behavior of `IsLiteral*` types when instantiated with a union of literals and non-literals. For example, `IsNumericLiteral<1 | bigint>` currently returns `false`, instead it should return `boolean`.

This PR also makes a couple of other things:
- Replaces certain complicated/contrived JSDoc examples with simpler ones.
- Removes unclear use-cases section from JSDoc. IMO, the use-cases mentioned were not adding much value.
- Separates the different literal types into individual files.
- Rewrites both the implementation and tests for all literal types from scratch.
- Fixes behavior with `Tagged`/`LiteralUnion` types.

Also, the input to these literal types is currently unconstrained, but I think that should be constrained. For example, `IsStringLiteral<T extends string> = ...` instead of _just_ `IsStringLiteral<T> = ...`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -127,13 +127,11 @@ export type {UnionToTuple} from './source/union-to-tuple.d.ts';
 export type {IntRange} from './source/int-range.d.ts';
 export type {IntClosedRange} from './source/int-closed-range.d.ts';
 export type {IsEqual} from './source/is-equal.d.ts';
-export type {
-	IsLiteral,
-	IsStringLiteral,
-	IsNumericLiteral,
-	IsBooleanLiteral,
-	IsSymbolLiteral,
-} from './source/is-literal.d.ts';
+export type {IsLiteral} from './source/is-literal.d.ts';
+export type {IsStringLiteral} from './source/is-string-literal.d.ts';
+export type {IsNumericLiteral} from './source/is-numeric-literal.d.ts';
+export type {IsBooleanLiteral} from './source/is-boolean-literal.d.ts';
+export type {IsSymbolLiteral} from './source/is-symbol-literal.d.ts';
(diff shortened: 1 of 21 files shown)


sindresorhus:
The new union behavior makes sense, but this regresses branded broad primitives:

```ts
type Brand = {readonly __brand: unique symbol};

type A = IsNumericLiteral<number & Brand>;
//=> true, should be false

type B = IsLiteral<Opaque<number, 'NumberTag'>>;
//=> true, should be false
```

The old implementation explicitly stripped branding before checking whether a primitive was broad or literal, so I think the fix is to keep teh new distributive union behavior but reuse that branded-primitive check for numeric/boolean/symbol.

som-sm:
> but this regresses branded broad primitives:

Added a new internal `UnwrapBrand` type to handle these cases. `UnwrapBrand` simply extracts the base type from the specified branded type, so it's not tied to `IsLiteral*` types and can be used in other places as well.

```ts
type Brand = {readonly __brand: unique symbol};

type A = UnwrapBrand<string & Brand>;
//=> string

type B = UnwrapBrand<Tagged<number, 'Tag'>>;
//=> number

type C = UnwrapBrand<Opaque<bigint, 'Tag'>>;
//=> bigint

type D = UnwrapBrand<PropertyKey & Brand>;
//=> PropertyKey

type E = UnwrapBrand<(1 | 200n | 'foo' | 'bar' | Uppercase<string>) & Brand>;
//=> 1 | 200n | 'foo' | 'bar' | Uppercase<string>
```

Notice that `UnwrapBrand` also works with branded types created using `Tagged`/`Opaque`.

PR #1483, opened 2026-07-04
Fix Get<T, Paths<T>> for generic T

### LINKED ISSUE #991: `Get<Paths<T extends {}>>` fails
### Bug description

When `T` has no keys that are known at compile time, the Result of `Paths<T>` is not assignable to the second argument of `Get`.

Repro:

```
import type { Get, Paths } from 'type-fest'

interface Bug<T extends {}> {
  props: Get<T, Paths<T>>
}
```

Error:

```
Type 'Paths<T>' does not satisfy the constraint 'readonly string[] | LiteralStringUnion<ToString<_Paths<T, { maxRecursionDepth: 10; bracketNotation: false; }> | _Paths<T, { maxRecursionDepth: 10; bracketNotation: true; }>>>'.
  Type 'number | `${number}`' is not assignable to type 'readonly string[] | LiteralStringUnion<ToString<_Paths<T, { maxRecursionDepth: 10; bracketNotation: false; }> | _Paths<T, { maxRecursionDepth: 10; bracketNotation: true; }>>>'.
    Type 'number' is not assignable to type 'readonly string[] | LiteralStringUnion<ToString<_Paths<T, { maxRecursionDepth: 10; bracketNotation: false; }> | _Paths<T, { maxRecursionDepth: 10; bracketNotation: true; }>>>'.
```

Workaround:

```
interface Workaround<T extends {someArbitraryPropertyKeyWeDoNotCareAbout?: never}> {
  paths: Get<T, Paths<T>>
}
```

### Repro

https://www.typescriptlang.org/play/?ssl=5&ssc=2&pln=3&pc=1#code/JYWwDg9gTgLgBDAnmApnA3nA4imAaOABQEMYALAZzgF84AzKCEOAciVQFo6UKYWBYAFBDgAOxgoodYgGM0AIQCuAcwA8AFTgoAHhNEATKumoA+DELhwwjMBQBc2XBoIlyFDSZNDqQkeMnScnAA6tAA1sSMigYaWrooBkYUTCgAglAARsAwUJGIhDaSSADSKIjBKAAiEAByEDAAwpFpGRCKMAD8DqIoAG6SpuaClmCklA44MM5EY+7qnt5AA

momomuchu:
`Get<T, Paths<T>>` fails to compile for a generic `T` (#991). `Paths<T>` can include bare numeric-literal keys (e.g. `1` alongside `'1'`), which for an unresolved generic `T` TypeScript treats structurally as `string | number`. `Get`'s `Path` parameter was constrained to string only, so it rejected the `number` and the whole expression failed to typecheck.

This widens `Get`'s `Path` constraint to also accept `Paths<BaseType>` and normalizes the path through `ToString` before splitting, so a bare-number path resolves the same as its string form.

Added a `test-d/get.ts` case for `Get<T, Paths<T>>` with a generic `T`. It fails to compile without the fix and passes with it. `test:tsc`, `test:tsd`, `test:xo` and `npm test` all green.

Closes #991

### DIFF
--- a/test-d/get.ts
+++ b/test-d/get.ts
@@ -1,5 +1,5 @@
 import {expectTypeOf} from 'expect-type';
-import type {Get} from '../index.d.ts';
+import type {Get, Paths} from '../index.d.ts';
 
 type NonStrict = {strict: false};
 
@@ -151,4 +151,21 @@ expectTypeOf<WithDictionary>().toEqualTypeOf<Get<WithDictionary, readonly []>>()
 
 	type FooPaths2 = 'array.1';
 	expectTypeOf<Get<Foo, FooPaths2>>().toEqualTypeOf<string | undefined>();
+}
+
+// eslint-disable-next-line no-lone-blocks
+{
+	// `Paths<GenericType>` must always be assignable to `Get`'s `Path` parameter, even when `GenericType` is an unresolved generic.
+	// https://github.com/sindresorhus/type-fest/issues/991
+	type WrapperOfGeneric<GenericType extends {}> = {
+		value: Get<GenericType, Paths<GenericType>>;
+	};
+
+	type Foo = {
+		key: number;
+	};
+
+	expectTypeOf<Get<Foo, 'key'>>().toEqualTypeOf<number>();
+	expectTypeOf<Get<Foo, Paths<Foo>>>().toEqualTypeOf<number>();
+	expectTypeOf<WrapperOfGeneric<Foo>['value']>().toEqualTypeOf<number>();
 }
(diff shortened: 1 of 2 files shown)


test-d/get.ts
```diff
@@ -152,3 +152,20 @@ expectTypeOf<WithDictionary>().toEqualTypeOf<Get<WithDictionary, readonly []>>()
...
+	};
+
+	type Foo = {
+		key: number;
+	};
+
+	expectTypeOf<Get<Foo, 'key'>>().toEqualTypeOf<number>();
+	expectTypeOf<Get<Foo, Paths<Foo>>>().toEqualTypeOf<number>();
+	expectTypeOf<WrapperOfGeneric<Foo>['value']>().toEqualTypeOf<number>();
+}
```
sindresorhus, line 178:
```suggestion
}

type WithNumberKey = {
	1: boolean;
};

expectTypeOf<Get<WithNumberKey, Paths<WithNumberKey>>>().toEqualTypeOf<boolean>();
expectTypeOf<WrapperOfGeneric<WithNumberKey>['value']>().toEqualTypeOf<boolean>();
```

momomuchu:
Thanks. Applying that verbatim trips TS2304: WrapperOfGeneric is declared inside the no-lone-blocks scoped block, so referencing it after the closing brace puts it out of scope. Two ways to land it: keep both new assertions inside the block (matches the WithNumberKey test already there), or lift the WrapperOfGeneric definition out of the block. Which do you prefer?

PR #1484, opened 2026-07-06
`Jsonify`: Fix optional properties leaking `undefined` into the key set

Linell:
`FilterJsonableKeys` maps each property to its own name and reads all values back out with an indexed access. Reading an optional property adds `undefined` to the result, so for `{a: string; b?: string}` the key union became `"a" | "b" | undefined`. Feeding that union to `Pick` silently builds an object whose key set includes `undefined`, and re-applying `Jsonify` to that output walks the corrupted key set, computes the illegal lookup `T[undefined]`, and collapses the object to `{}`.

Excluding `undefined` from the union keeps the key set clean and makes `Jsonify<Jsonify<T>>` equal `Jsonify<T>`.

Fixes discovered downstream in inngest/inngest#4483.

### DIFF
--- a/source/jsonify.d.ts
+++ b/source/jsonify.d.ts
@@ -25,16 +25,11 @@ type JsonifyList<T extends UnknownArray> = T extends readonly []
 			? JsonValue[]
 			: Array<T[number] extends NotJsonable ? null : Jsonify<UndefinedToNull<T[number]>>>;
 
-// Note: `Exclude<..., undefined>` is required because optional properties add `undefined` to the resulting union, which would otherwise corrupt the key set of the resulting object (e.g., `keyof Jsonify<{a?: string}>` would include `undefined`).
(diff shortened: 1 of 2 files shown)


sindresorhus (approved):

source/jsonify.d.ts
som-sm, line 33: `FilterJsonableKeys` is actually _not_ required at all, that's just unnecessary code.

Updating `JsonifyObject` to the following is a better fix:
```suggestion
type JsonifyObject<T extends object> = {
	[Key in keyof T as T[Key] extends NotJsonable ? never : Key]: Jsonify<T[Key]>;
};
```
This passes all the test cases.

som-sm (changes requested):

source/jsonify.d.ts
Linell: Ohhh you're right, updated! 👍  Thank you!

som-sm (approved):

PR #1485, opened 2026-07-07
Add `ConditionalPaths` type

### LINKED ISSUE #1328: Proposal: `ConditionalPaths`

Get paths recursively, but only when an item extends a passed matcher item. Basically, the result would be like doing `Paths<ConditionalPickDeep<T, Matcher>>`, but the problem with this is that it does double the recursion to dive into the object. By using a `ConditionalPaths`, the recursion would only have to be performed once.

### Type source

_No response_

naman-luthra-rubrik:
This PR:
- Adds `ConditionalPaths<Type, Condition, Options>` - the path-aware counterpart of `ConditionalKeys`. Returns a union of all dot-notation paths in an object whose value type satisfies `Condition`.
- Closes #1328

Supports:
- `extends` mode (default): value assignable to `Condition`
- `equality` mode: exact type match via `IsEqual`
- All `Paths` options except `leavesOnly` (which is a no-op for any specific-type condition since primitive-typed fields are always leaves)

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -150,6 +150,7 @@ export type {ArrayElement} from './source/array-element.d.ts';
 export type {ArrayLength} from './source/array-length.d.ts';
 export type {SetFieldType, SetFieldTypeOptions} from './source/set-field-type.d.ts';
 export type {Paths, PathsOptions} from './source/paths.d.ts';
+export type {ConditionalPaths, ConditionalPathsOptions} from './source/conditional-paths.d.ts';
 export type {AllUnionFields} from './source/all-union-fields.d.ts';
 export type {SharedUnionFields} from './source/shared-union-fields.d.ts';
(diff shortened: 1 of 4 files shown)


sindresorhus:
This is described as the path-aware counterpart of `ConditionalKeys`, but the semantics differ. Optional properties are treated as required before checking the condition, so `{name?: string}` filtered by `string` includes `name`. `ConditionalKeys` requires `string | undefined` for that.

`never` properties also match evry condition because of the tuple-wrapped `extends` check, so `{data: never; name: string}` filtered by `string` includes both paths. That is suprising, and `ConditionalKeys` already avoids this with `ExtendsStrict`.

I'm also not convinced the new type is worth duplicating most of the path-generation logic. If we add this, I think it should be a thin composition over `Paths`, `Get`, and `ExtendsStrict`, unless there is a benchmark showing that approach is too slow.

I feel like this is a bit too niche

laadaadi772-png:
Hey can check my github profile idk how to ise but can suggest some yt video or sth i think u reply me

PR #1489, opened 2026-07-09
NonEmptyObject: use OmitIndexSignature for better index signature handling

Mubashir78:
Use `OmitIndexSignature` before `HasRequiredKeys` in `NonEmptyObject` so that types with only index signatures (no explicit keys) are routed through `RequireAtLeastOne` instead of being incorrectly considered as having required keys.

Previously, index-signature-only types like `{ [x: string]: string }` were incorrectly considered as having required keys (`HasRequiredKeys` returned `true` due to the index signature), causing `NonEmptyObject` to pass them through unchanged. This meant `{}` was silently accepted, defeating the purpose of `NonEmptyObject`.

Now `NonEmptyObject<{ [x: string]: string }>` correctly routes to `RequireAtLeastOne`. Note that `RequireAtLeastOne` itself has a known limitation with pure index signatures that prevents it from fully enforcing non-emptiness in that case, but this change is still an improvement over the previous behavior.

### Test plan
- Existing tests continue to pass
- Added test cases for index-signature-only, optional+index, and required+index type combinations

### DIFF
--- a/source/non-empty-object.d.ts
+++ b/source/non-empty-object.d.ts
@@ -1,4 +1,5 @@
 import type {HasRequiredKeys} from './has-required-keys.d.ts';
+import type {OmitIndexSignature} from './omit-index-signature.d.ts';
 import type {RequireAtLeastOne} from './require-at-least-one.d.ts';
 
 /**
@@ -33,6 +34,6 @@ const update2: UpdateRequest<User> = {};
 
 @category Object
 */
-export type NonEmptyObject<T extends object> = HasRequiredKeys<T> extends true ? T : RequireAtLeastOne<T, keyof T>;
+export type NonEmptyObject<T extends object> = HasRequiredKeys<OmitIndexSignature<T>> extends true ? T : RequireAtLeastOne<T, keyof T>;
 
 export {};
--- a/test-d/non-empty-object.ts
+++ b/test-d/non-empty-object.ts
@@ -18,12 +18,38 @@ type TestType3 = {
 
 type TestType4 = {};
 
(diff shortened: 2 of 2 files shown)


sindresorhus:
Thanks, but this doesn't actualy fix the issue. `RequireAtLeastOne<T, keyof T>` still accepts `{}` for both a pure index signature and an optional property plus an index signature. The tests only confirm which branch is selected, not the behaviour.

A pure index signature cannot be statically made non-empty here, so this shouldn't close #821. Also XO fails because the index signatures need to come before the explicit members.

Mubashir78:
Thanks for the review. I have:
- Removed the `Closes #821` claim since `RequireAtLeastOne` still cannot enforce non-emptiness for pure index signatures (a separate issue)
- Fixed the index signature ordering in tests (signature before explicit members)
- Force-pushed the amended commit
- Updated the PR title and description

This change still improves the behavior by routing index-signature-only types through `RequireAtLeastOne` instead of silently passing `{}`, even though the full enforcement is blocked by a fundamental TypeScript limitation with index signatures.

som-sm (commented):
The PR still doesn't make sense. Routing index signatures to `RequireAtLeastOne` doesn't solve anything since `RequireAtLeastOne` doesn't work with index signatures.

And the test cases are also wrong. We don't want to add incorrect / half-baked assertions.

Mubashir78:
Yeah i see now that routing to `RequireAtLeastOne` doesn't solve the issue either, index signatures are fundamentally unenforceable at the type level in TypeScript. Im gonna close this PR. Thanks for the review.

PR #1492, opened 2026-07-28
Use NoInfer utility type in types

### LINKED ISSUE #848: Use `NoInfer`?

marceli1404:
Closes #848.

Adds a JSDoc reference to the built-in NoInfer<Type> utility type in the IsAny type definition, documenting that NoInfer prevents circular constraint errors when IsAny is used in function generic constraints.

The NoInfer type was already being used in the IsAny implementation (since #993) via a custom type alias, which was later replaced with the built-in after upgrading to TypeScript 5.4+. This PR adds documentation linking to the built-in NoInfer utility type and confirms that a review of all types in the library found no other types that would benefit from using NoInfer.

See the discussion in #848 for context.

### DIFF
--- a/source/is-any.d.ts
+++ b/source/is-any.d.ts
@@ -5,6 +5,8 @@ Returns a boolean for whether the given type is `any`.
 
 Useful in type utilities, such as disallowing `any`s to be passed to a function.
 
+Uses the built-in [`NoInfer`](https://www.typescriptlang.org/docs/handbook/utility-types.html#noinfertype) utility type to prevent circular constraint errors.
+
 @example
 ```
 import type {IsAny} from 'type-fest';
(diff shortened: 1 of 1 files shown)


source/is-any.d.ts
```diff
@@ -5,6 +5,8 @@ Returns a boolean for whether the given type is `any`.
 
 Useful in type utilities, such as disallowing `any`s to be passed to a function.
 
+Uses the built-in [`NoInfer`](https://www.typescriptlang.org/docs/handbook/utility-types.html#noinfertype) utility type to prevent circular constraint errors.
```
som-sm, line 8: I don't think so how `IsAny` is built needs to be mentioned in the documentation, that's an internal detail which the user doesn't need to care about.

sindresorhus:
I don't really see how this is useful.

PR #1493, opened 2026-08-01
fix: remove Partial from Dependency type

dword-design:
Currently the Dependency type is a partial of `Record<string, string>`, which allows `{ 'type-fest': undefined }`.

Tests are failing but I think they are not related.

### DIFF
--- a/source/package-json.d.ts
+++ b/source/package-json.d.ts
@@ -206,7 +206,7 @@ export namespace PackageJson {
 	/**
 	Dependencies of the package. The version range is a string which has one or more space-separated descriptors. Dependencies can also be identified with a tarball or Git URL.
 	*/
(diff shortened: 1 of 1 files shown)


som-sm (changes requested):
This PR makes sense because `undefined` should not be allowed as a valid value. `undefined` should only come up while accessing an unknown property from an index signature, which can be enforced using `noUncheckedIndexedAccess` compiler option.

```ts
// @noUncheckedIndexedAccess: true
declare const x: {[x: string]: number};

const y = x.foo;
//=> number | undefined

// @ts-expect-error
x.bar = undefined; // Not allowed
```

But all instances of `Partial<Record<...` need fixing, not just `Dependency`. 

Also, the failing tests are related to this change, so tests need to be updated as well.

PR #1494, opened 2026-08-02
Add `SetOptionalDeep` type

### LINKED ISSUE #1308: SetOptionalDeep

Like `SetOptional` but works on deeply-nested properties

```typescript
type Foo = { bar: { baz: string } }
type Example = SetOptionalDeep<Foo, "bar.baz">
// Example = { bar: { baz?: string } }
```

### Type source

_No response_

dsavy4:
Closes #1308

This adds `SetOptionalDeep`, the optional counterpart to the existing `SetRequiredDeep`. It makes the given keys optional using dot separated paths, so you can reach into nested objects and arrays instead of only the top level.

```ts
import type {SetOptionalDeep} from 'type-fest';

type Foo = {
	a: number;
	b: string;
	c: Array<{
		d: number;
	}>;
};

type Example = SetOptionalDeep<Foo, 'a' | `c.${number}.d`>;
//=> {b: string; c: {d?: number}[]; a?: number}
```

It mirrors `SetRequiredDeep` closely so the two read the same way, and it reuses the shared internals (`Paths`, `SimplifyDeep`, `PartialDeep`, `UnionToTuple`).

One thing worth calling out about tuples. An optional element can only sit at the tail of a tuple, so making an element optional also makes every element after it optional. Asking for a middle index therefore cascades to the rest of the tuple rather than producing an invalid type.

```ts
type A = SetOptionalDeep<{a: [number, number, number]}, 'a.1'>;
//=> {a: [number, number?, number?]}
```

- [x] Added the type to `index.d.ts`
- [x] Added tests in `test-d/set-optional-deep.ts`
- [x] Added an entry to the readme
- [x] `tsd`, `tsc`, and `xo` all pass locally

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -48,6 +48,7 @@ export type {Optional} from './source/optional.d.ts';
 export type {Opaque, UnwrapOpaque, Tagged, GetTagMetadata, UnwrapTagged} from './source/tagged.d.ts';
 export type {InvariantOf} from './source/invariant-of.d.ts';
 export type {SetOptional} from './source/set-optional.d.ts';
+export type {SetOptionalDeep} from './source/set-optional-deep.d.ts';
 export type {SetReadonly} from './source/set-readonly.d.ts';
 export type {SetRequired} from './source/set-required.d.ts';
 export type {SetRequiredDeep} from './source/set-required-deep.d.ts';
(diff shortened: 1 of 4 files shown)


sindresorhus:
Found one issue:

`number extends TArray['length']` also matches tuples with a rest element, so `[string, number, ...boolean[]]` is treated like a regular array. This means `SetOptionalDeep<{a: [string, number, ...boolean[]]}, 'a.1'>` leaves the second element required, even though `a.1` is a valid path and `[string, number?, ...boolean[]]` is a valid tuple varient.

Could you add a regression test for this and process the fixed elements before returning the array unchanged? `SetRequiredDeep` already handles this shape.

dsavy4:
Good catch, thanks. Fixed.

The dispatch was bailing out whenever `length` was `number`, which also caught tuples that end in a rest element. It now checks for fixed numeric indices instead, so `[string, number, ...boolean[]]` still gets its leading elements processed and only a true non-tuple array like `string[]` is left untouched. The tail helper already knew to stop at the rest element, so it was just that guard.

Added regression tests for the rest-element shape, including the readonly and cascade cases:

```ts
SetOptionalDeep<{a: [string, number, ...boolean[]]}, 'a.1'>;
//=> {a: [string, number?, ...boolean[]]}
```

PR #1498, opened 2026-08-14
Add `UndefinableDeep` type

### LINKED ISSUE #612: Would be good to have DeepUndefinable
See https://github.com/ts-essentials/ts-essentials/tree/master/lib/deep-undefinable

is it ok to put a PR to add something like this?

maximilliangrand:
Closes #612.

Adds `UndefinableDeep` for a fully shaped object whose individual values can be cleared without removing its required containers or properties:

```ts
type Draft = UndefinableDeep<{
	textEditor: {fontSize: number};
	tags: string[];
}>;
// {textEditor: {fontSize: number | undefined}; tags: Array<string | undefined>}
```

The transformation widens leaf values consistently in objects, arrays, tuples, maps, sets and promises, while preserving optional/readonly modifiers. Functions and constructors are leaves, so generic/overloaded signatures and attached properties stay intact rather than being rebuilt with `Parameters` and `ReturnType`. Weak-collection keys stay valid by excluding `undefined`; a `never` leaf becomes `undefined`.

This follows the leaf-value contract requested in review. Tests cover required-container rejection, optional and variadic tuples, collections, weak keys, generic/overloaded callables and constructors, and `never`/`unknown`/`any`. README and API documentation describe the same contract.

Validation: the revised contract tests fail against the previous implementation. Full TypeScript compilation and tsd pass, as do XO for the changed files and all 38 custom-linter tests. `npm test` fails on four XO errors in unchanged `test-d/abstract-class.ts` and `test-d/readonly-deep.ts`; all four reproduce in a separate checkout of the previous PR head using the same dependencies. No unrelated lint fixes are included. AI-assisted implementation and independent source review.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -40,6 +40,7 @@ export type {PickDeep} from './source/pick-deep.d.ts';
 export type {OmitDeep} from './source/omit-deep.d.ts';
 export type {PartialOnUndefinedDeep, PartialOnUndefinedDeepOptions} from './source/partial-on-undefined-deep.d.ts';
 export type {UndefinedOnPartialDeep} from './source/undefined-on-partial-deep.d.ts';
+export type {UndefinableDeep} from './source/undefinable-deep.d.ts';
 export type {ReadonlyDeep} from './source/readonly-deep.d.ts';
 export type {LiteralUnion} from './source/literal-union.d.ts';
 export type {Promisable} from './source/promisable.d.ts';
--- a/readme.md
+++ b/readme.md
@@ -135,6 +135,7 @@ Click the type names for complete docs.
 - [`PartialDeep`](source/partial-deep.d.ts) - Create a deeply optional version of another type.
 - [`PartialOnUndefinedDeep`](source/partial-on-undefined-deep.d.ts) - Create a deep version of another type where all keys accepting `undefined` type are set to optional.
 - [`UndefinedOnPartialDeep`](source/undefined-on-partial-deep.d.ts) - Create a deep version of another type where all optional keys are set to also accept `undefined`.
+- [`UndefinableDeep`](source/undefinable-deep.d.ts) - Create a deep version of another type where all keys are set to also accept `undefined`.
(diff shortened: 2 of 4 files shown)


sindresorhus:
I think this needs some more thought

The issue points to ts-essentials' DeepUndefinable, but this behaves differently: primitive values inside arrays, maps, sets, promises, and tuples are not widened, while an object-valued property itself becomes undefined. Neither contract is necessarily wrong, but it conflicts with the documented "every key" and "every level" behavior. For the fully-shaped object use-case, the referenced behavior seems more coherent: keep object containers present and make their leaf values undefinable.

The type itself makes sense because PartialDeep makes keys optional. I would simplify this around that referenced contract and treat functions and constructors as leaves. Rebuilding callable objects with `Parameters<T>` and `ReturnType<T>` also erases generic signatures, turning a generic identity function into effectively `(value: unknown) => unknown`. Passing callables through avoids this bug and removes the fragile function-specific machinery, which is seperate from the main object-data use-case.

maximilliangrand:
Reworked this around the leaf-value contract you described. Required object containers now stay present, while primitive leaves widen consistently inside arrays, tuples, maps, sets and promises. A single homomorphic mapping preserves object and tuple modifiers.

Functions and constructors pass through as leaves with `undefined` added; their generic and overloaded signatures and attached properties remain intact. This removes the `Parameters`/`ReturnType` reconstruction and the function-specific helper machinery. Weak-collection keys exclude `undefined` to satisfy their key constraint.

Added negative container-shape tests and generic/overloaded callable and constructor regressions, alongside collection and tuple coverage. The revised contract fails against the previous implementation. README and API documentation now match the behavior.

Full tsc and tsd pass, along with XO for the changed files and all 38 custom-linter tests. `npm test` reports four XO errors in unchanged `test-d/abstract-class.ts` and `test-d/readonly-deep.ts`; the same four reproduce in a separate checkout of the previous PR head with identical dependencies. AI-assisted implementation and independent source review.

PR #1500, opened 2026-08-15
`SetOptional`: Add support for making tuple indices optional

som-sm:
Adds support for making tuple indices optional (similar to `SetRequired`):
```ts
type ArrayExample = SetOptional<[number, number, number], 1 | 2>;
//=> [number, number?, number?]
```

This will _also_ help simplify #1494.

### DIFF
--- a/source/set-optional.d.ts
+++ b/source/set-optional.d.ts
@@ -59,14 +59,13 @@ type SetArrayOptional<
 		// Enters this branch if `TArray` contains a rest element
 		? TArray extends readonly [...any[], any]
 			? TArray // If there are elements after the rest element, then we can't make any elements optional.
-			: SetArrayOptionalHelper<TArray, Keys, TArray>
-		: SetArrayOptionalHelper<TArray, Keys, TArray>
+			: SetArrayOptionalHelper<TArray, Keys>
+		: SetArrayOptionalHelper<TArray, Keys>
 	: never;
 
 type SetArrayOptionalHelper<
 	TArray extends UnknownArray,
 	Keys,
-	OriginalArray extends UnknownArray,
 	Counter extends any[] = [],
 	Accumulator extends UnknownArray = [],
 	OptionalsAccumulator extends UnknownArray = [],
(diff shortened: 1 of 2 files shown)


sindresorhus:
`SetOptional<[string, number, boolean], 1>` currently returns `[string, number, boolean]`. The selected index is not made optional at all. It also disagrees with #1494, where selecting a middle index makes the remaining fixed tail optional because TypeScript cannot put a required element after an optional one.

I think the simplest rule is: the first requestd fixed index starts the optional tail. Once that index is found, the recursion can stop and return the accumulated prefix followed by `Partial<StaticPartOfArray<Remaining>>` and `VariablePartOfArray<Remaining>`. This preserves the rest element and removes the need for `OptionalsAccumulator`.

There is also a regression for augmented and tagged arrays. Rebuilding every array through tuple spreads removes non-index members. For example:

```ts
type AugmentedArray = string[] & {tag: boolean};

type A = SetOptional<AugmentedArray, never>;
// Currently: string[]

type B = SetOptional<AugmentedArray, 'tag'>;
// Currently: string[]
// Expected: the existing array shape with `tag` optional
```

I would add a `Keys = never` fast path that returns the input exactly, and only use the tuple-specific path when all requested keys are numeric. Non-index keys can continue through the existing object implementation. Mixed numeric and non-numeric keys could be left unsupported instead of adding more machinery.

I would test these cases seperately:

```ts
expectType<[string, number?, boolean?]>({} as SetOptional<[string, number, boolean], 1>);
expectType<[string?, number?, boolean?]>({} as SetOptional<[string, number, boolean], 0 | 1>);
expectType<[string, number?, boolean?, ...bigint[]]>({} as SetOptional<[string, number, boolean, ...bigint[]], 1>);

type AugmentedArray = string[] & {tag: boolean};
expectType<AugmentedArray>({} as SetOptional<AugmentedArray, never>);

declare const augmentedResult: SetOptional<AugmentedArray, 'tag'>;
expectType<boolean | undefined>(augmentedResult.tag);

type TaggedTuple = Tagged<[string, number], 'Tuple'>;
expectType<TaggedTuple>({} as SetOptional<TaggedTuple, never>);
```

som-sm:
> `SetOptional<[string, number, boolean], 1>` currently returns `[string, number, boolean]`. The selected index is not made optional at all. It also disagrees with #1494, where selecting a middle index makes the remaining fixed tail optional because TypeScript cannot put a required element after an optional one.

@sindresorhus This approach will not align with the behavior of `SetRequired`. For example, `SetRequired<[string, number?, boolean?, bigint?], 2>` does not make everything till index `2` required (i.e., `[string, number, boolean, bigint?]`), instead it does nothing.

And also IMO the behaviour in this PR makes more sense because you can always pass a union of keys to get the behaviour mentioned in #1494, like:

```ts
type T1 = SetOptional<[string, number, boolean], 1 | 2>;
//=> [string, number?, boolean?]

type SomeArray = ['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i'];
type T2 = SetOptional<SomeArray, IntRange<2, SomeArray['length']>>;
//=> ["a", "b", "c"?, "d"?, "e"?, "f"?, "g"?, "h"?, "i"?]
```

I think we should keep the behavior consistent.

sindresorhus:
Fair point. I agree that matching `SetRequired` is preferable here.

It would be good to mention this limitation in the documentation, since an isolated middle index silently doing nothing may otherwise be surprising.

The augmented and tagged array regression is separate and still needs addressing. `SetOptional<AugmentedArray, never>` should remain an exact no-op, and non-index keys should retain the existing object behavior.

PR #1503, opened 2026-08-16
`ConditionalKeys`: Fix declared keys being dropped when an index signature is present

### LINKED ISSUE #1501: ConditionalPick and ConditionalKeys break on types with index signatures
### Bug description

When using ConditionalPick / ConditionalKeys with types which include an index signature for an `unknown` typed property, the created type will never include any results, unless `unknown` is passed as a condition, at which point, *only* those index signature keys will be part of the created type.

**Expected behaviour**: the created type includes all keys that *can* be determined at compile time, optionally also including any index signature properties.

### Repro

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBDAnmApnA3nAwhAdgE2BmDwEMAbABWAGMBrAGjkqlCOADc0BfOAMygQQcAORJUAWj4oAzjBEBuALAAoVcFwwUUPqRppq9UgCNyadKrhW4YWnRT4AXHDmtcAc1UBICGGJkqOwcAfmdcAFcQY21La1wIGEN7Jww4UjDI6KgmYwyo7ThuVSK1FXEDOxMzAFEAD1JwMzgAXmw8Qn9cCiSAHiSqlCYWNmIuAD5lFQB6KeZK0xQ6hrAm1sxbegdnVw13Jl9O7qD8ULgI-Kg4AB84cIIUPg0HQtV1TW1dfTgAVVwNugGGFiVniiWO2xgbk8Kh8fhIXXIADkEkkQnkssCzgkAOqkd4pVLpM6ZbQ5dEFErWOAAbXsiAhUIAus47nR4gB3XDFV5lZBoX7-AZLRpoVo4AjsAK9AXzMxDVggdjjVQzH5-WWLeoilpnFBcKA8jRaHR6fm4NkQTnY6B0GRAlRU2koekuSG7Zm3c0crkqEqqcpqi1Wm0yYUrUVtCWHQL0Hq-IO4a1QW3ykacNA3VnesYq2bx71J21h1apam1Bnulley24BQ08vEi4erM1uvcIA

MahinAnowar:
Fixes #1501.

`ConditionalKeys` returns `never` for any type that has a `string` index signature whose value type doesn't match the condition, even when the declared keys do match:

```ts
type Example = {
	[x: string]: unknown;
	a: string;
	b: number;
};

type StringKeys = ConditionalKeys<Example, string>;
//=> never, expected 'a'
```

`ConditionalPick` and `ConditionalExcept` inherit it, so they come back as `never` and unchanged respectively.

The cause is that a `string` index signature widens `keyof` to `string | number`, which absorbs the literal keys — `'a' | 'b' | string` is just `string`. `_ConditionalKeys` iterates `keyof Base & {}`, and the `& {}` (added in #1198 to stop the mapped type distributing over unions) means it iterates that widened union rather than the declared members, so it only ever asks whether `Base[string]` matches. A `number` or pattern index signature like `[x: Uppercase<string>]` doesn't absorb string literals, which is why those already worked.

So I collect the declared keys separately, with a homomorphic mapped type that drops the index signatures, and match them on their own. The two results get unioned rather than the two key sets — unioning the key sets first would just collapse `'a' | string` back to `string`.

The existing behaviour is unchanged, including the case where the index signature *does* match: `ConditionalKeys<{[x: string]: string; a: string}, string>` still gives `string | number`, since that already subsumes `'a'`. The union handling from #1198 is intact too — `Keys & keyof Base` keeps it non-homomorphic and drops keys that aren't common to every union member.

Added tests to `conditional-keys`, `conditional-pick` and `conditional-except`. Five of them fail on `main` and pass with this change; the symbol-index and no-match cases already passed and are there to pin the surrounding behaviour. `npm test` is green.

### DIFF
--- a/source/conditional-keys.d.ts
+++ b/source/conditional-keys.d.ts
@@ -53,10 +53,28 @@ export type ConditionalKeys<Base, Condition> = (Base extends UnknownArray ? Tupl
 	? IfNotAnyOrNever<_Base, {ifNot: _ConditionalKeys<_Base, Condition>; ifAny: keyof _Base}>
 	: never;
 
-type _ConditionalKeys<Base, Condition> = keyof {
+// A string index signature widens `keyof` to `string | number`, which swallows the declared
+// literal keys, so they have to be collected separately and matched on their own.
+type _ConditionalKeys<Base, Condition> =
+	| _MatchingKeys<Base, Condition, keyof Base>
+	| _MatchingKeys<Base, Condition, DeclaredKeys<Base>>;
+
+type _MatchingKeys<Base, Condition, Keys> = keyof {
 	[
-	Key in (keyof Base & {}) as // `& {}` prevents homomorphism
+	Key in (Keys & keyof Base) as // The intersection prevents homomorphism
 	ExtendsStrict<Base[Key], Condition> extends true ? Key : never
+	]: never
+};
+
+// The keys `Base` declares itself, without the ones contributed by a `string`, `number` or
+// `symbol` index signature. Homomorphic on purpose — that is what keeps the declared keys
+// visible one by one instead of collapsed into the index signature.
+type DeclaredKeys<Base> = keyof {
+	[
+	Key in keyof Base as string extends Key ? never
+		: number extends Key ? never
+			: symbol extends Key ? never
+				: Key
 	]: never
 };
 
--- a/source/conditional-keys.d.ts
+++ b/source/conditional-keys.d.ts
@@ -1,5 +1,6 @@
 import type {ExtendsStrict} from './extends-strict.d.ts';
 import type {IfNotAnyOrNever} from './internal/type.d.ts';
+import type {OmitIndexSignature} from './omit-index-signature.d.ts';
 import type {TupleToObject} from './tuple-to-object.d.ts';
 import type {UnknownArray} from './unknown-array.d.ts';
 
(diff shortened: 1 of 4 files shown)


sindresorhus:
The two-pass approach makes sense, but this still misses declared keys that are covered by a template-pattern index signature. For example:

```ts
type Result = ConditionalKeys<{
	[key: `data-${string}`]: unknown;
	'data-known': string;
}, string>;
//=> never, expected 'data-known'
```

`string extends Key` only filters out the full `string` type. It leaves `` `data-${string}` ``, so `keyof DeclaredKeys<Base>` collapses `'data-known'` back into the pattern and the second matching pass has the same problem as the first. The same thing happens with patterns like `Uppercase<string>`.

Could we reuse the existing `OmitIndexSignature` type here instead?

```ts
type _ConditionalKeys<Base, Condition> =
	| _MatchingKeys<Base, Condition, keyof Base>
	| _MatchingKeys<Base, Condition, keyof OmitIndexSignature<Base>>;
```

I tested this against the current tests and additional string, number, symbol, template-pattern, intersection, and union cases. They all pass with this version. The two key sets still need to be matched seperately, since unioning them first would swallow the literal keys again.

MahinAnowar:
Good catch, you're right. I only filtered the bare `string`/`number`/`symbol` keys, so any pattern signature stayed in the set and re-absorbed the literal keys it covers — my hand-rolled version was solving half the problem.

Switched to `OmitIndexSignature`, which is obviously the right tool and one less thing to maintain. Your example now gives `'data-known'`, and I added it as a test along with the `Uppercase<string>` variants and an intersection case.

One thing I kept deliberately: a pattern signature that *does* match still contributes and subsumes the literals under it, so `ConditionalKeys<{[x: \`data-${string}\`]: string; 'data-known': string}, string>` stays `` `data-${string}` ``. That mirrors the existing `{[x: string]: string; a: string}` behaviour, so I assumed it's intended rather than something to change here — say the word if you'd rather it enumerate.

Both new assertions fail on my previous commit and pass now. `npm test` is green.

Issue #662, opened 2023-08-08
`Destructurable` type


sindresorhus:
See: https://github.com/sindresorhus/type-fest/pull/596

PR #1505, opened 2026-08-18
Add `Destructurable` type

ChrisJr404:
Adds a `Destructurable` type that converts a union of object types into a shape whose members can be destructured directly, resolving #662.

When a function returns different object shapes and you let TypeScript infer the return type, the compiler describes each member with the keys of the *other* members added as optional `undefined` properties, and that inferred shape is exactly what makes the result destructurable in a single statement. The moment you annotate the return type with a plain union (`Success | Failure`), that convenience disappears — destructuring a key that only lives on some members no longer compiles, even though the value is there at runtime. `Destructurable` reconstructs the inferred shape from an explicit union so the union stays destructurable.

Importantly, the original members are preserved, so narrowing on a discriminant property still works exactly as before — the type only makes the pre-narrowing destructuring compile, it doesn't collapse the union. That addresses the main concern raised on the earlier attempt (#596): you don't lose discriminated-union safety.

This revives #596, which the maintainer bumped and then closed only for author inactivity, and follows the current contribution guidelines (new `source/destructurable.d.ts`, tests in `test-d/`, export in `index.d.ts`, readme entry under Utilities).

Tests cover direct destructuring, that narrowing still yields the correct per-branch types, that individual members remain assignable while mixed members are rejected, a single non-union object passing through unchanged, and a three-member union. `npm test` passes locally (tsc, tsd, xo, and the jsdoc-codeblock linter).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -19,6 +19,7 @@ export type {UnknownSet} from './source/unknown-set.d.ts';
 export type {UnknownMap} from './source/unknown-map.d.ts';
 export type {Except, ExceptOptions} from './source/except.d.ts';
 export type {TaggedUnion} from './source/tagged-union.d.ts';
+export type {Destructurable} from './source/destructurable.d.ts';
 export type {Writable} from './source/writable.d.ts';
 export type {WritableDeep} from './source/writable-deep.d.ts';
 export type {Merge} from './source/merge.d.ts';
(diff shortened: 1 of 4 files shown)


sindresorhus:
I don't think we should add this as a seperate public type. `ExclusifyUnion` already supports direct destructuring and correlated narrowing.

I checked both types with and without `exactOptionalPropertyTypes`. Without it, they behave equivalently for this use case. With it enabled, the only meaningful difference is that `Destructurable` allows another member's property to be explicitly present as `undefined`, while `ExclusifyUnion` and TypeScript's inferred type reject it. This makes the new type less strict than the compiler inference it claims to reproduce.

When the implementation is available, `satisfies` can validate the declared union while preserving the inferred destructurable return type. For explicit or ambient return types, `ExclusifyUnion` already solves the problem. I suggest adding a destructuring example to its documentation and closing this as a duplicate.

PR #1519, opened 2026-09-14
Add `UnionMax` type

xianjianlf2:
Expose the existing internal `UnionMax` type through the package entry point, so consumers can derive a maximum from a union of supported numeric literals without importing an internal path or copying the implementation. For example, `UnionMax<0 | 2 | 5>` produces `5` for a fixed set of retry counts.

Related to #676, where consumers requested public `UnionMax` and `UnionMin` exports. This PR exposes only `UnionMax`.

- Move the existing implementation to `source/union-max.d.ts`, updating the existing consumer to import it directly without a compatibility re-export.
- Document practical usage, special inputs, and the existing algorithm's limits: small non-negative integers and infinity are supported; negative/fractional finite numbers are unsupported, and large inputs can hit TypeScript's recursion limit. The former internal comment's blanket “0 to 999” claim is not repeated: `UnionMax<0 | 999>` reaches the compiler's recursion limit in local testing.
- Move the existing tests to the public import and add coverage for zero, a larger supported integer, mixed infinities, a readonly tuple-derived union, and invalid inputs/assignments.

Validation: `npm test` passes (TypeScript, tsd, XO including documentation examples, and all 38 linter tests). The export is grouped with the numeric types, IsAnyOrNever is imported from its defining module, and the original commented negative tests are retained with an updated unsupported-input note. The numeric algorithm is preserved; changes inside it only give the type parameters descriptive names.

AI assistance: OpenAI Codex assisted with implementation, review, and testing.


sindresorhus:
A few small changes: Please place the new export beside the other numeric exports. I would also update the existing internal consumer to import the standalone type directly and remove the compatibility re-export. That deep path is not exposed by the package, and the re-export creates a direct circular dependency. Importing IsAnyOrNever directly from its defining module would keep that dependency clear too.

Finally, please preserve the existing commented negative test cases when moving the test. The old TODO text is stale, so it can be updated to simply say that negative numbers are not supported yet.

